Issue #173, opened 2014-06-26
tab completion during interactive prompts


chaoflow:
It would be nice if click would support tab completion when prompting for files and paths.

nir0s:
Um.. anything on this?

chaoflow:
Do we have tab completion for choice option prompts? Does this fit this issues scope after adjusting title and description or shall I open a new issue?

davidism:
The completion system has been totally rewritten in #1622. The prompt system can get completions for a parameter by calling `param.shell_complete(ctx, incomplete)`. #181 explored using `readline`, but that's not guaranteed to be built with Python, and there's an issue around license. It also points out that there are other nice solutions like [prompt-toolkit](https://github.com/prompt-toolkit/python-prompt-toolkit). I'd like to avoid adding optional dependencies, so the first thing that needs to be done is to design a way for `prompt` to be customized more, so extensions could provide implementations.

kdeldycke:
A FAQ entry has been proposed at https://github.com/pallets/click/pull/3826 to close that discussion.

kdeldycke:
FAQ has been merged upstream.

PR #181, opened 2014-07-01
A first draft of tab-completion for prompts

untitaker:
See #173. This is not ready for merge yet.
- Missing tests, changelog entry, and isn't yet used in click.Path/File.
- Is a very myopic implementation. Users will ask for custom completion
  functions, which bash's read of course doesn't support.


untitaker:
Yeah, this is a dead end.

PR #189, opened 2014-07-13
Add support for different categories to get_app_dir

cgtx:
This will improve support of the [XDG Base Directory Specification](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html), including the proposed [STATE directory](https://wiki.debian.org/XDGBaseDirectorySpecification#state).  Since the default for **category** is CONFIG, the default behavior of get_app_dir is still the same.  However, users can override the category to get a different type of directory.

Examples:

```
>>> APP_NAME = 'foo'
>>> get_app_dir(APP_NAME)
'/home/carl/.config/foo'
>>> get_app_dir(APP_NAME, category='CONFIG')
'/home/carl/.config/foo'
>>> get_app_dir(APP_NAME, category='CACHE')
'/home/carl/.cache/foo'
>>> get_app_dir(APP_NAME, category='DATA')
'/home/carl/.local/share/foo'
>>> get_app_dir(APP_NAME, category='STATE')
'/home/carl/.local/state/foo'
```

### DIFF
--- a/click/utils.py
+++ b/click/utils.py
@@ -314,7 +314,7 @@ def format_filename(filename, shorten=False):
     return filename_to_ui(filename)
 
 
-def get_app_dir(app_name, roaming=True, force_posix=False):
+def get_app_dir(app_name, roaming=True, force_posix=False, category='CONFIG'):
     r"""Returns the config folder for the application.  The default behavior
     is to return whatever is most appropriate for the operating system.
 
@@ -348,6 +348,8 @@ def get_app_dir(app_name, roaming=True, force_posix=False):
                         folder will be stored in the home folder with a leading
                         dot instead of the XDG config home or darwin's
                         application support folder.
+    :param category: the type of directory to return.  Valid values are CONFIG,
+                     CACHE, DATA, or STATE.
     """
     if WIN:
         key = roaming and 'APPDATA' or 'LOCALAPPDATA'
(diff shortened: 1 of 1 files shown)


jck:
Lot's of apps use XDG_CACHE(or ~/.cache) now. Most notably: pip, Firefox, Chrome, Unity, and pretty much every QT5 app including all of KDE Plasma 5.

If it is OK with you, I can submit a PR with separate cross-platform functions: get_config_dir, get_cache_dir and get_data_dir(alias of get_app_dir).

untitaker:
@jck I'm not sure how you would implement `get_cache_dir` on Windows. I'd definetly like to see this but I don't think it's possible. Perhaps we should simply point the caller to a subfolder inside the app dir?

jck:
Yes, that is probably the best way for windows: https://github.com/ActiveState/appdirs/blob/master/appdirs.py#L257

nchammas:
[appdirs](https://github.com/ActiveState/appdirs) is an interesting project, @jck. Thanks for mentioning it. I'm surprised it takes [so much code](https://github.com/ActiveState/appdirs/blob/c3f35455ee00e4531469afb6c156533ea29ac235/appdirs.py) just to compute these values, but what do I know.

@mitsuhiko @untitaker - Here's [how appdirs does it](https://github.com/ActiveState/appdirs#some-example-output):

```
>>> user_cache_dir(appname, appauthor)
'C:\\Users\\trentm\\AppData\\Local\\Acme\\SuperApp\\Cache'
```

PR #221, opened 2014-09-16
Ensure bash completion functions have valid posix names

rayl:
A click application 'foo-bar' would generation a completion function
with a hyphen in the name. This is accepted by Bash interactive shells,
but fails during graphical login on a Ubuntu machine:

/usr/sbin/lightdm-session: 11: /home/rayl/.my_aliases: Syntax error: Bad function name

This leads to a login/logout loop until you login on a terminal and
fix the function name.

Signed-off-by: Ray Lehtiniemi rayl@mail.com

### DIFF
--- a/click/_bashcomplete.py
+++ b/click/_bashcomplete.py
@@ -1,4 +1,5 @@
 import os
+from string import letters, digits
 from .utils import echo
 from .parser import split_arg_string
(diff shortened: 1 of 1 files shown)


untitaker:
@rayl Could you add a changelog and tests for this? Thanks!

PR #252, opened 2014-11-17
Fixed exception in Option.get_help_record()

ulope:
This PR fixes a TypeError ("not all arguments converted during string formatting") in `Option.get_help_record()` that occurred while showing the help for options that are defined with `multiple=True`, `show_default=True` and have a list or tuple as default value.


untitaker:
Thanks!

PR #259, opened 2014-11-24
Support nargs=-1 for option (cleaner)

tristanwietsma:
#222

### DIFF
--- a/click/parser.py
+++ b/click/parser.py
@@ -275,7 +275,7 @@ class OptionParser(object):
 
             nargs = option.nargs
             if nargs == -1:
-                func = lambda (_, v): (v.startswith('-') and len(v) == 2) or v.startswith('--')
+                func = lambda p: (p[1].startswith('-') and len(p[1]) == 2) or p[1].startswith('--')
                 dash_locs = filter(func, enumerate(state.rargs))
                 if len(dash_locs) > 0:
                     nvar = min([idx for (idx, _) in dash_locs])
--- a/click/parser.py
+++ b/click/parser.py
@@ -277,7 +277,8 @@ class OptionParser(object):
             if nargs == -1:
                 dash_locs = filter(lambda pair: pair[1].startswith('-'), enumerate(state.rargs))
                 if len(dash_locs) > 0:
-                    nvar = min([x[0] for x in filter(lambda pair: pair[1].startswith('-'), enumerate(state.rargs))])
+                    func = lambda p: (p[1].startswith('-') and len(p[1]) == 2) or p[1].startswith('--')
+                    nvar = min([x[0] for x in filter(func, enumerate(state.rargs))])
                     value = tuple(state.rargs[:nvar])
                     del state.rargs[:nvar]
                 elif len(state.rargs) == 1:
--- a/click/parser.py
+++ b/click/parser.py
@@ -275,10 +275,10 @@ class OptionParser(object):
 
             nargs = option.nargs
             if nargs == -1:
-                dash_locs = filter(lambda pair: pair[1].startswith('-'), enumerate(state.rargs))
+                func = lambda (_, v): (v.startswith('-') and len(v) == 2) or v.startswith('--')
+                dash_locs = filter(func, enumerate(state.rargs))
                 if len(dash_locs) > 0:
-                    func = lambda p: (p[1].startswith('-') and len(p[1]) == 2) or p[1].startswith('--')
-                    nvar = min([x[0] for x in filter(func, enumerate(state.rargs))])
+                    nvar = min([idx for (idx, _) in dash_locs])
                     value = tuple(state.rargs[:nvar])
                     del state.rargs[:nvar]
                 elif len(state.rargs) == 1:
--- a/click/parser.py
+++ b/click/parser.py
@@ -275,14 +275,10 @@ class OptionParser(object):
 
             nargs = option.nargs
             if nargs == -1:
-                is_dash = lambda v: (v.startswith('-') and len(v) == 2) or v.startswith('--')
-                dash_locs = [idx for (idx, v) in enumerate(state.rargs) if is_dash(v)]
-                if len(dash_locs) > 0:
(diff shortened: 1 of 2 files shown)


untitaker:
What is the behavior when the user doesn't provide a ``--``? When does it fail?

click/parser.py
```diff
@@ -290,6 +298,13 @@ def _match_long_opt(self, opt, explicit_value, state):
 
         option.process(value, state)
 
+    def _next_flag(self, state):
+        is_flag = lambda v: (v.startswith('-') and len(v) == 2) or v.startswith('--')
```
tristanwietsma: @untitaker relevant to your question

tristanwietsma:
if the user doesn't supply '--' or another flag, then there isn't any way to determine where the variadic portion ends.

this usage of '--' is demonstrated in the source elsewhere, so i'm assuming this pattern is fair game.

i'll agree that it feels somewhat awkward in that it requires extra knowledge about the cli. the alternative to variadic options is a comma-separated string of values to be parsed in the command, which isn't very pretty either. bit of a conundrum.

untitaker:
I just read up on click's documentation. I wonder why anybody would want to use such a form if ``multiple=True`` is already available? The resulting commands also seem a lot more explicit.

tristanwietsma:
the particular use case i wrote it for involved a chain of commands in a bash. output from one script was a list of files with unknown length. using this approach, i was able to set nargs=-1 and just pass that output with a " --" appended. alternatively, i could have preprocessed the output in bash.

i think this boils down to a question of where to put the abstraction. i'm not interested in complicating a great library, so i see where you're going. at the same time, the behavior is only in play with "nargs=-1", so it isn't **that** unreasonable to expect the author of the script using this feature to pass additional information in the usage string.

untitaker:
Hmm, I think it might get extremely funky if a misunderstanding user *passes both multiple and nargs=-1*. I'll just leave this open for now.

tristanwietsma:
if you're strongly against it, please feel free to close. this PR is from 11/24; the feature was useful for me and while I see some users have inquired about it, i'm more than happy to just work of my fork. regards.

untitaker:
I'm not strongly against it, I honestly have no strong opinion at all.

warsamebashir:
+1 please merge this

mitsuhiko:
This is opening a can of worms.  It was never intended for nargs < 0 to even remotely work for options and I am not going to support that.  This is however something that might change in the future if sufficient demand comes up with proper justifications about why that's a good idea.

From my perspective it's a terrible idea because it's impossible to unambiguously parse.  This is what arguments are for.

MathieuDuponchelle:
Re "proper justifications about why that's a good idea."

Consider the following case:

I have a CLI that needs to handle two arbitrarily-long lists of distinct filenames (Let's say a bunch of C source files and a bunch of markdown files)

afaiu, the only two solutions are to either:
+ let the user do my_cli --source source.c --source source.h --markdown-file my_file.markdown
+ let the user do my_cli --sources "source.c,source.h" --markdown-files "my_file.markdown"

The first one seems pretty clumsy to me, and the second one error-prone (obviously things will break if a filename contains a comma)

I would like to be able to do my_cli --sources *.c *h --markdown_files *.markdown , which seems to me like a pretty common usecase.

I'm not sure I do understand the issue with allowing this, afaict it should only be a matter of stopping consumption once the parsing code encounters a recognized "--" option ?

untitaker:
>I would like to be able to do my_cli --sources *.c *h --markdown_files *.markdown , which seems to me like a pretty common usecase.

Why not make all of those arguments, like `mv` and `cp` do?

untitaker:
Consider the following example:

    @click.command()
    @click.option('--lol')
    @click.argument('foo', nargs=-1)
    def foo(lol, foo):
        print(lol)
        print(foo)

If I were to add `nargs=-1` to the option, and execute this:

    foo --lol 1 2 3 4 5 6 7 8 9

Click would have no idea what belongs to the option and what to the argument.

MathieuDuponchelle:
> Why not make all of those arguments, like mv and cp do?

I don't understand that question, I have two separate types of files, I can only use nargs=-1 for the final argument so this doesn't solve my problem.

untitaker:
I see, I assumed that you can only specify one markdown file. What if you specify a single `nargs=-1` argument and detect the filetype by file extension?

MathieuDuponchelle:
> Click would have no idea what belongs to the option and what to the argument.

I'm just stating that there is a pretty clear use case here, even more so when taking wildcards into account, I don't know anything about click's implementation but I do know that argparse handles that correctly.

This issue prevents me from using click at all, even though I'd like to use it because I need to resort to funny hacks in argparse to get what click calls "chained commands", it's a pretty silly reason for not using such a nice library :)

MathieuDuponchelle:
> What if you specify a single nargs=-1 argument and detect the filetype by file extension?

This isn't a heuristic I'd be satisfied with, and doesn't solve the general case anyway

untitaker:
As shown, the general case would be ambiguous. Options with nargs=-1 would only be possible if no arguments are specified, which *is* a can of worms.

MathieuDuponchelle:
I would argue that 

+ This could very well be detected at runtime ("Oh no you can't do that you silly")
+ Or afaiu the initial patch set, a separator would seem pretty fair game

MathieuDuponchelle:
Preventing the usage of both nargs=-1 arguments and options doesn't strike me as particularly bad, as it's obvious this can't be done (without a separator), can be easily detected, and simply extends the functionality to let use cases such as mine be implemented with click

tristanwietsma:
Been a while since I looked at this PR, but I believe "--" signals the
option args are complete. See test cases.

MathieuDuponchelle:
Ack

MathieuDuponchelle:
Anyway @mitsuhiko  to summarize my reason for wanting this is "please let me wildcard more than one parameter" :)

PR #265, opened 2014-12-01
Fix erroneous region code on python3 doc page

MrTact:
There is no territory with a region code of "EN"; the correct one is "US." At least, it didn't work when I tried it, but "en_US" did!


untitaker:
Thanks!

Issue #277, opened 2015-01-09
Support for logging-setup


dploeger:
I find click wonderful as it simplifies all my cli-related code lines.

What I also do besides argument checking (what click does) is setting up the logging (using Python's logging-facilities) according to certain parameters.

I'd find it great, if click would also do that. Namely...

* Understand -q for quiet mode (call **logging.basicConfig(level=logging.CRITICAL)**)
* Understand -v for verbose mode (call **logging.basicConfig(level=logging.INFO)**)
* Understand -d for debug mode (call **logging.basicConfig(level=logging.DEBUG)**)

Call **logging.basicConfig(level=logging.ERROR)** without any specific parameter.

That could be done using another decorator like @simplelog.

kissgyorgy:
I think it's unnecessary to have three switches for the same feature. How py.test does it, also it appears in click documentation: with count for e.g. -v, so like `$ command -vvv` is more verbose than `$ command -vv`

dploeger:
I'll add that flavor to my pull request.

kimvais:
Please merge this - @click.simplelog(flavor='vvv') is pretty much what I need.

untitaker:
If you're interested in this, there is a basic implementation at https://github.com/click-contrib/click-log. Patches for other option flavors are welcome!

PR #278, opened 2015-01-09
Simplelog-feature

dploeger:
This is what I meant in #277 :)

Now you can add @click.simplelog() to a command to add logging behaviour to it. Afterwards you can simply call logging.debug for debugging messages and they will only show up, if the user specified the --debug/-d-switch.

If you like it, I can additionally add the relevant documentation.

### DIFF
--- a/click/simplelog.py
+++ b/click/simplelog.py
@@ -16,7 +16,7 @@ def simplelog(*args, **kwargs):
     return make_simplelog
 
 
-class SimpleLog:
+class SimpleLog(object):
 
     __name__ = "SimpleLog"
 
--- a/click/simplelog.py
+++ b/click/simplelog.py
@@ -1,0 +1,135 @@
+from core import Option
+import logging
+
+
+def simplelog(*args, **kwargs):
+
+    """ Decorator to add a very simple logging facility to scripts
+
+    For parameters see SimpleLog.__init__
+    """
+
+    def make_simplelog(f):
+
+        return SimpleLog(f, **kwargs)
+
+    return make_simplelog
+
+
+class SimpleLog:
+
+    __name__ = "SimpleLog"
+
+    def __init__(self, callback, verbose_help=None, debug_help=None,
+                 quiet_help=None, remove_args=False, verbose_args=None,
+                 debug_args=None, quiet_args=None, verbose_arg_name=None,
+                 debug_arg_name=None, quiet_arg_name=None,
+                 default_level=None, quiet_level=None, verbose_level=None,
+                 debug_level=None, logger_name=None):
+
+        """ Handler class for SimpleLog
+
+        :param verbose_help: Override the text displayed for the verbose
+                     parameter [Enable verbose logging]
+        :param debug_help: Override the text displayed for the debug parameter
+                           [Enable debugging]
+        :param quiet_help: Override the text displayed for the quiet parameter
+                           [Enable quiet mode]
+        :param remove_args: Remove the logging arguments, so they are not
+                            submitted to the main command [Disabled]
+        :param verbose_args: List of arguments for the "verbose"-setting
+                             [--verbose and -v]
+        :param debug_args: List of arguments for the "debug"-setting
+                           [--debug and -d]
+        :param quiet_args: List of arguments for the "quiet"-setting
+                           [--quiet and -q]
+        :param verbose_arg_name: Name of the "verbose" argument [verbose]
+        :param debug_arg_name: Name of the "debug" argument [debug]
+        :param quiet_arg_name: Name of the "quiet" argument [quiet]
+        :param default_level: Default level [logging.ERROR]
+        :param verbose_level: Level for "verbose"-setting [logging.INFO]
+        :param debug_level: Level for "debug"-setting [logging.DEBUG]
+        :param quiet_level: Level for "quiet"-setting [logging.CRITICAL]
+        :param logger_name: Logger to configure [root-logger]
+        """
+
+        self.callback = callback
+
+        self.verbose_help = verbose_help or "Enable verbose logging"
+        self.debug_help = debug_help or "Enable debugging"
+        self.quiet_help = quiet_help or "Enable quiet mode"
+        self.remove_args = remove_args or False
+        self.verbose_args = verbose_args or ["--verbose", "-v"]
+        self.debug_args = debug_args or ["--debug", "-d"]
+        self.quiet_args = quiet_args or ["--quiet", "-q"]
+        self.verbose_arg_name = verbose_arg_name or "verbose"
+        self.debug_arg_name = debug_arg_name or "debug"
+        self.quiet_arg_name = quiet_arg_name or "quiet"
+        self.default_level = default_level or logging.ERROR
+        self.verbose_level = verbose_level or logging.INFO
+        self.debug_level = debug_level or logging.DEBUG
+        self.quiet_level = quiet_level or logging.CRITICAL
+        self.logger_name = logger_name or None
+
+        self.__click_params__ = [
+            Option(
+                [self.verbose_arg_name] + self.verbose_args,
+                is_flag=True,
+                help=self.verbose_help
+            ),
+            Option(
+                [self.debug_arg_name] + self.debug_args,
+                is_flag=True,
+                help=self.debug_help
+            ),
+            Option(
+                [self.quiet_arg_name] + self.quiet_args,
+                is_flag=True,
+                help=self.quiet_help
+            )
+        ]
+
+        if hasattr(callback, "__click_params__"):
+
+            # Add parameters of the original command as well
+
+            self.__click_params__ = self.__click_params__ + \
+                callback.__click_params__
+
+
+    def __call__(self, *args, **kwargs):
+
+        """ Call the main program, but set up logging first
+        """
+
+        verbose = kwargs[self.verbose_arg_name]
+        debug = kwargs[self.debug_arg_name]
+        quiet = kwargs[self.quiet_arg_name]
+
+        if self.remove_args:
+
+            for key in (self.verbose_arg_name, self.debug_arg_name,
+                        self.quiet_arg_name):
+
+                kwargs.pop(key, None)
+
+        if not self.logger_name:
+
+            # As a precaution, set up the root logger beforehand
+
+            logging.basicConfig()
+
+        logger = logging.getLogger(self.logger_name)
+
+        logger.setLevel(self.default_level)
+
+        if quiet:
+            logger.setLevel(self.quiet_level)
+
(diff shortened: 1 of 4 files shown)


dploeger:
Yay! Travis build works! ;)

dploeger:
I found, that simplelog needs to be put **after** a pass_context-decorator. If that's a problem, I could investigate more.

untitaker:
@dploeger Currently using https://github.com/untitaker/vdirsyncer/blob/master/vdirsyncer/log.py and looking into this:

- It's **very** nice that this integrates with click in the sense that it adds the relevant options, but otherwise it's extremely tightly coupled to it.
- It seems that this is hightly coupled with the rest of click
- It doesn't set up its own logger, but expects the user to use the global logger. Which is probably not a big deal with CLI apps, but the use of ``tool_resetlog`` looks kinda sad.

Features missing:
- Colored error/warning output

click/simplelog.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+
+    def make_simplelog(f):
+
+        return SimpleLog(f, **kwargs)
+
+    return make_simplelog
+
+
+class SimpleLog:
```
untitaker: Use new-style classes.

click/simplelog.py
```diff
@@ -0,0 +1,199 @@
...
+
+    __name__ = "SimpleLog"
+
+    def __init__(self, callback, verbose_help=None, debug_help=None,
```
untitaker: That's massive. No idea how to fix this, but I think this is not really acceptable. Maybe users should override methods, which gives them more flexibility anyway?

click/simplelog.py
```diff
@@ -0,0 +1,199 @@
...
+                           [--quiet and -q]
+        :param verbose_arg_name: Name of the "verbose" argument [verbose]
+        :param debug_arg_name: Name of the "debug" argument [debug]
+        :param quiet_arg_name: Name of the "quiet" argument [quiet]
+        :param default_level: Default level [logging.ERROR]
+        :param verbose_level: Level for "verbose"-setting [logging.INFO]
+        :param debug_level: Level for "debug"-setting [logging.DEBUG]
+        :param quiet_level: Level for "quiet"-setting [logging.CRITICAL]
+        :param logger_name: Logger to configure [root-logger]
+        :param flavor: Parameter-flavor to use:
```
untitaker: Do we need multiple flavors?

tests/test_simplelog.py
```diff
@@ -0,0 +1,192 @@
...
+import re
+import click
+import logging
+
+log_stream = None
+
+
+def tool_resetlog():
+
+    global log_stream
```
untitaker: Here and in other places you have a lot of unnecessary empty lines.

click/simplelog.py
dploeger: Yes. Will fix that.

click/simplelog.py
dploeger: Hmm... I tried to mimic the way, the other decorators work in click. Perhaps there's a more elegant way somehow.

click/simplelog.py
dploeger: I got that request in #277. I use a "-q, -v, -d"-flavor but the user there was suggesting a "multiple of v"-variant, so I added the flavor support.

It doesn't do harm IMHO.

tests/test_simplelog.py
dploeger: I try to comply to PEP-8, but I see, that there are some useless blank lines. I'll clear that up.

dploeger:
@untitaker About the logger: You CAN use simplelog with your custom logger by using the logger_name-parameter. However, I had some problems unit-testing the logging-output so I used the rather crude tool_resetlog.

davidism:
I don't see why this is in Click.  If someone wants fine-grained control over a logger, they should probably do that themselves in a way that works for their project.

Flask gets in a bunch of trouble (there are numerous issues reported) for messing with logging internally; Click should try to avoid the same mistakes.

This implementation seems very heavy-handed.  Rather than a canonical "Click" way to do things, you offer two.  In addition, there's no way to gain fine-grained control of the log level, for example if I want to be able to specify a fifth log leve, or a custom log level.

Perhaps a cleaner solution would be a `-l LEVEL` flag, such as what Celery uses.  Ultimately though, I don't think this should be part of Click.

dploeger:
I think, you misunderstood a thing or two here.

Simplelog doesn't really interfere with logging. It just sets up a logger for you. You can configure your own logger (you don't actually need to use the root logger) and simplelog will configure it for you.

In my opinion, it's just a helper for those, who regularly write cli programs, that need logging behaviour and implement that feature each and every time - not really DRY.

If you want to set up logging your way, simply don't use the simplelog decorator.

tests/test_simplelog.py
untitaker: That's a good thing, but you can freely disobey PEP8 if you feel your code is better off without it. Also I'd recommend using the ``pep8`` package instead of checking manually.

click/simplelog.py
```diff
@@ -0,0 +1,199 @@
...
+        self.debug_level = debug_level or logging.DEBUG
+        self.quiet_level = quiet_level or logging.CRITICAL
+        self.logger_name = logger_name or None
+        self.flavor = flavor or "qvd"
+
+        if self.flavor == "qvd":
+
+            self.verbose_help = "Enable verbose logging"
+
+            self.__click_params__ = [
```
untitaker: That seems like poking at click's internal state... maybe you could simply wrap the callback normally with ``click.option`` etc?

click/simplelog.py
```diff
@@ -0,0 +1,199 @@
+from .core import Option
+import logging
+
+
+def simplelog(*args, **kwargs):
```
untitaker: This throws ``*args`` away.

click/simplelog.py
untitaker: Yeah, this might be bikeshedding... I think @mitsuhiko should decide on this.

tests/test_simplelog.py
```diff
@@ -0,0 +1,192 @@
+import sys
+
+import re
+import click
+import logging
+
+log_stream = None
+
+
+def tool_resetlog():
```
untitaker: Since the logger is supposed to print to stdout/stderr, you can use [pytest's features](http://pytest.org/latest/capture.html) for this.

Also, please don't use any global state.

davidism:
@dploeger Fair enough.  It's good to have an explanation of the intended use for this.  I'm still not a fan of having more than one way to do things in an official part of the code, but I'll leave that up to the maintainers.

tests/test_simplelog.py
dploeger: Actually, I configured my IntelliJ Code Style accordingly. ;-)

click/simplelog.py
dploeger: Hmm... I need to look into this. I personally dislike the way I did this myself, but I found no better way.

tests/test_simplelog.py
dploeger: Yeah, I *tried* that. However, it didn't succeed. I tried logging to stdout/stderr and check the output after the call, but sometimes the log-lines aren't in the output. That's the cause of this rather ugly thing here.

untitaker: I think the problem is that you didn't explicitly configure any logging handlers and formatters... check out the code I linked at the beginning, it even comes with colorful output!

dploeger: Hmm... Could it be, that pytest together with the runner.invoke method, that is used in the unit tests, doesn't handle stderr-results? However, I tried to do a 

    logging.basicConfig(stream=sys.stdout)

before the test, but that still doesn't work. I guess, because pytest interrupts and captures all output. The code you linked isn't inside this pytest/runner.invoke-cage.

untitaker: Click also captures stdout and stderr and stores it in result.output. Sorry for the earlier comment, bow Idon't think pytest's capturing methods will work in this case.

dploeger: I think, that capturing the stdout and stderr into result.output doesn't work reliably. At least not when running the tests via pytest.

Maybe I could rewrite the code to simply check against result.output, commit it and you could test it on your side?

untitaker: That sounds good!

dploeger: Please see commit 86ea179 - that was the way I originally wanted to test, but it doesn't work at my site.

dploeger:
@untitaker Any news on the logging problems in the test suite?

dploeger:
@untitaker? Hello? ;)

untitaker:
Sorry for the late response... I did already comment on some technical details in the code, but some of them were left unaddressed (see inline diff comments). Some, like introducing flavors, have to be decided by mitsuhiko.

Also,

* superfluous newlines nearly everywhere.
* I'd like to make this set up a specific logger on the context instead of using the global one (as it currently stands, there is no reason why this code should be coupled to click at all)

dploeger:
Thanks, @untitaker, but I was more interested in if you got my test suite running. Remember? I wrapped it together, so that it should theoretically work (by parsing the output), but didn't at my site. You said, you would test it on your site. :)

untitaker:
Ah yes, will try later. But I think these other issues have to be solved anyway

untitaker:
@dploeger The testing issue can be fixed by setting up a custom logger which uses click's echo function (see https://github.com/untitaker/vdirsyncer/blob/master/vdirsyncer/log.py)

What basically happens is that Click's runner tries to monkeypatch sys.stdout, but that doesn't really help because the logging uses its own "pointer" (for lack of a better word) to the old stdout -- it has something like this somewhere:

    x = sys.stdout

and click's testrunner can only modify sys.stdout, but x stays the same.

dploeger:
Thanks, @untitaker. I fixed the tests thanks to you. I also implemented the new style classes. Now, the rest is up to you and/or @mitsuhiko.

untitaker:
@dploeger Please fix two issues:

* The one mentioned in the diff: https://github.com/mitsuhiko/click/pull/278/files#r23776750
* Remove unnecessary newlines. Generallly, if a *single line* is surrounded by two blank lines, the blank lines are unnecessary.

untitaker:
Now that we have [click-contrib](https://github.com/click-contrib), it might be a better idea to pilot this as a third-party extension?

untitaker:
@dploeger Are you interested in updating this? Otherwise I'll close this pull request and maybe create a repo on `click-contrib` sometime.

dploeger:
To be honest, I actually moved to using Cement, as they have already implemented that feature there. Let's close this pull request. Thanks, guys.

PR #285, opened 2015-01-29
ability to pass custom implementations of ``Option`` and ``Argument``

slafs:
allow passing a different class to `option` and `argument` decorators via `cls` keyword argument.

### DIFF
--- a/click/decorators.py
+++ b/click/decorators.py
@@ -140,8 +140,8 @@ def argument(*param_decls, **attrs):
     This is equivalent to creating an :class:`Argument` instance manually
     and attaching it to the :attr:`Command.params` list.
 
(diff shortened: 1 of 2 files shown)


untitaker:
Thanks for your contribution!

Please refer to ``command``: You can avoid ``lwargs.pop`` and define the parameter normally. Also, please adjust the wording so it's similar to the docstring of ``command``.

slafs:
Wording adjusted. 

I don't know what do you mean by avoiding ``kwargs.pop``. Since we have positional args we can't do ``def argument(*param_decls, cls=Argument, **attrs): ``

untitaker:
Thanks!

PR #286, opened 2015-01-30
Modify _bashcomplete for zsh use

anonfunc:
Zsh has support for loading bash completion functions, but dies on the syntax here.
Adding 'env' works on both bash and zsh.

zsh instructions:
1. autoload bashcompinit && bashcompinit
2. Follow bash instructions.

### DIFF
--- a/click/_bashcomplete.py
+++ b/click/_bashcomplete.py
@@ -6,7 +6,7 @@ from .core import MultiCommand, Option
 
 COMPLETION_SCRIPT = '''
 %(complete_func)s() {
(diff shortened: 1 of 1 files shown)


rcaloras:
:+1:  Yeah you can usually get bash completion to work in zsh, this just needed the env in front. 
```
autoload bashcompinit
bashcompinit
source click_completion_file_with_env.
```
http://stackoverflow.com/a/8492043/1282124

allancaffee:
+1

untitaker:
Thanks!

PR #301, opened 2015-02-21
Fix bashism in get_editor()

sephii:
This causes the path to the editor to get printed if the shell spawned by `os.system` is not Bash.


untitaker:
Thanks!

PR #305, opened 2015-03-06
Minor typo.  Variable PATH and not PATHS

manuphatak:
The ENV variable that's being referenced is and should be `PATH` singular not `PATHS` plural.

Also, just for fun, I think it would be way more funny if line 376, referencing dangerous operations was a bit more sinister: `click.echo('Dropped all tables, lol!')`

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -479,7 +479,7 @@ Example usage:
 .. click:example::
 
     @click.command()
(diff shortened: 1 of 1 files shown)


untitaker:
That's not a typo, this short example doesn't have anything to do with the `PATH` environment variable. The environment variable might as well be called something else.

PR #307, opened 2015-03-07
Incorrect code example.  open with 'write' flag

manuphatak:
The example doesn't work as is.  You need the 'w' flag.


untitaker:
Thanks!

PR #309, opened 2015-03-17
Add colorama as a dependency on Windows (works with wheels too). 

ionelmc:
See https://www.python.org/dev/peps/pep-0426/#environment-markers for details.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,4 +25,7 @@ setup(
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
     ],
+    extras_require={
+        ':sys_platform == "win32"': ['colorama'],
+    },
 )


untitaker:
I'm in favor of adding this, but up until now, colorama was an optional dependency. If you suggest making it a hard dep., the following would have to happen, probably in that order:

 * Approval by Armin
* Remove all the fallback code

ionelmc:
As I understood this, correct me if I'm wrong, click can output colors on
unix-like OS-es without colorama. Colorama is just for windows where there
are no ansii escape codes.


Thanks,
-- Ionel Cristian Mărieș, http://blog.ionelmc.ro

untitaker:
Exactly, and on Windows, Colorama is an optional dependency.

mitsuhiko:
I have no idea how I feel about this.   My intuition says no because of how we use colorama currently.  First of all not everybody needs it, secondly colorama is of partially questionable code quality and if it causes issues there should be a good way to disable it.  Going to close this for now.

ionelmc:
@mitsuhiko it would be useful to know what you don't like about colorama. Can you point me to where it has flaws?

PR #325, opened 2015-04-08
Document use of a callable for the default value of an option

pfmoore:


untitaker:
This is already mentioned in the API docs though. Wouldn't it make sense to mention this feature in the context of the envvar-default's behavior (i.e. explain that envvars override prompts, and give an example on how to achieve a different behavior)

pfmoore:
Possibly. The API docs don't typically seem to include usage examples, so I wasn't sure how to fit it in there. Also, when I was looking for this functionality, the place I looked was alongside the "Prompting" section, so I thought that putting a mention there would be helpful.

But if someone has a better way of documenting this, that's fine with me.

As a somewhat related thought, maybe having some sort of "Click Cookbook" which gives example code for common scenarios like this would be useful? Either in the docs, or maybe use the github wiki for this, so that users can more easily contribute examples?

untitaker:
Ok, I have an idea for a middle-path: Modify the current example to be specifically your usecase (passing `lambda: os.environ['something']` or whatever you did), and explain it. I think that should both explain the feature of passing a callable and provide a real usecase.

Regarding a cookbook, I think the current state is that e.g. options-related recipes already go into `options.rst`. I see value for a proper cookbook, but it would've to be properly categorized, like the now defunct Flask-Snippets.

pfmoore:
OK, updated. How does this look?

untitaker:
Pretty great, thanks!

PR #327, opened 2015-04-16
fix for a bug in multiple option

frankho:
hi,
I think I found a bug in the multiple option handling.
Or I expect it should work diffrent than currently :-), which would mean I'm using it wrong.
However, I created a test case to show it.
I hope it is helpful.

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -1150,6 +1150,7 @@ class Parameter(object):
         self.name, self.opts, self.secondary_opts = \
             self._parse_decls(param_decls or (), expose_value)
 
+
         self.type = convert_type(type, default)
 
         # Default nargs to what the type tells us if we have that
--- a/click/types.py
+++ b/click/types.py
@@ -432,6 +432,9 @@ class Tuple(CompositeParamType):
         self.types = [convert_type(ty) for ty in types]
 
     @property
+    def name(self):
+        return "<"+" ".join(ty.name for ty in self.types)+">"
+    @property
     def arity(self):
         return len(self.types)
 
@@ -446,14 +449,18 @@ def convert_type(ty, default=None):
     """Converts a callable or python ty into the most appropriate param
     ty.
     """
+    guessed_type = False
+    if ty is None and default is not None:
+        if isinstance(default,tuple):
+            ty=tuple(map(type,default))
+        else:
+            ty = type(default)
+        guessed_type = True
+
     if isinstance(ty, tuple):
         return Tuple(ty)
     if isinstance(ty, ParamType):
         return ty
-    guessed_type = False
-    if ty is None and default is not None:
-        ty = type(default)
-        guessed_type = True
     if ty is text_type or ty is str or ty is None:
         return STRING
     if ty is int:
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -146,9 +146,9 @@ def test_multiple_default_help(runner):
     @click.command()
     @click.option("--arg1", multiple=True, default=('foo', 'bar'),
                   show_default=True)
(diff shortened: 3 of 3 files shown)


frankho:
I fixed the bugs.
please merge it.

untitaker:
I think the current behavior is intended, as

* the multiple-option doesn't specify the *amount* of accepted params
* Click's assumption is that multiple values are homogenous (have the same type).

frankho:
I will provide an exaple:

    @click.command()
    @click.option("--echo",type=(str,int),multiple=True)
    def cmd(echo):
        for s,num in echo:
            print((s+" ")*num)

usage:

    cmd --echo test 4 --echo foo 1 # works with current version
    cmd --help # doesn't work with the currunt click version


the same should work with default parameters:

    @click.option("--echo",default=("test",1),multiple=True):

but doesn't work with current version.

untitaker:
Ok, I see.

click/core.py
```diff
@@ -1150,6 +1150,7 @@ def __init__(self, param_decls=None, type=None, required=False,
         self.name, self.opts, self.secondary_opts = \
             self._parse_decls(param_decls or (), expose_value)
 
+
```
untitaker: Please remove this.

click/types.py
```diff
@@ -432,6 +432,9 @@ def __init__(self, types):
         self.types = [convert_type(ty) for ty in types]
 
     @property
```
untitaker: What's this for?

tests/test_options.py
```diff
@@ -146,9 +146,9 @@ def test_multiple_default_help(runner):
     @click.command()
     @click.option("--arg1", multiple=True, default=('foo', 'bar'),
                   show_default=True)
-    @click.option("--arg2", multiple=True, default=(1, 2), type=int,
+    @click.option("--arg2", multiple=True, default=(1, 2), type=(int,int),
```
untitaker: This shouldn't have to change.

frankho: why should the type be different from default?

untitaker: I mean we probably shouldn't break backwards compat like that. Changing it back still makes the tests pass it seems.

I'm currently playing around with your PR locally, to see if there's a better way of fixing this bug.

untitaker:
Thanks, I've applied this to 4.x-maintenance and master!

tests/test_options.py
frankho: Ok. If I remember correctly I changed this just because I saw it and thought it might be a typo.
I don't know why click even accept this contradictory options.
It might not be the best thing to include it in this pull request, sry.

davidism:
This doesn't appear to be the correct fix, I'm not sure how no one ran into it in the years since then. Admittedly the concepts of `nargs`, `multiple`, and `types.Tuple` overlap quite a bit, which can make things confusing.

`multiple=True, default=(1, 2)` is describing multiple ints, for which the default is two ints, not a tuple of only two ints. If you wanted a multiple tuple, the default would be `[(1, 2)]`, a list of one tuple of two ints. The example of `multiple=True, type=(str, int)` is the same case, the default would be `[("a", 2)]`. I've opened #1649 to address this.

untitaker:
I don't think multiple=True should allow for heterogenous types. Are you saying we document this behavior anywhere already?

davidism:
It doesn't, it just worked by accident in the example given. That example was actually describing multiple heterogeneous tuples.

davidism:
The original version didn't implement detecting a tuple type from `default`. If `type` wasn't provided, it would have ended up returning `FuncParamType(tuple)`, which I don't think would work as expected either.

Issue #316, opened 2015-04-03
multiple argument, wrong order


liujoey:
```
@click.command()
@click.argument('src', type=click.Path(exists=True), nargs=-1)
@click.argument('abstract', type=click.File('w', encoding='utf-8'), nargs=1)
@click.argument('body', type=click.File('w', encoding='utf-8'), nargs=1)
def cli(src, abstract, body):
```
with above code, when I call:
`mycli input.txt abs.txt body.txt` I suppose in my cli function, `abstract` is the file of `abs.txt` and `body` is the file of `body.txt`.
But the fact is what I got `abstract` is `body.txt` and `body` is `abs.txt`

So, I reverse the `abstract` and `body` declaration like this:
```
@click.command()
@click.argument('src', type=click.Path(exists=True), nargs=-1)
@click.argument('body', type=click.File('w', encoding='utf-8'), nargs=1)
@click.argument('abstract', type=click.File('w', encoding='utf-8'), nargs=1)
def cli(src, abstract, body):
```
Now my function getting the right files. But when I run `mycli --help` it shows:
```
Usage: mycli [OPTIONS] [SRC]... BODY ABSTRACT
```
Which the sequence is wrong again.

ckamradt:
this is due to the code in click/utils.py.  when the unpack_args method encounters an nargs value < 0 it causes fetch(args) to pop arguments from the right side of the deque instead of the left.  its only in the case of 2 or more args after the nargs=-1 argument that anyone notices.  

this is easily demonstrated by slightly altering code from the docstring:
>>> from click.utils import unpack_args
>>> unpack_args(range(6), [1, 2, 1, -1])  # from docstring
((0, (1, 2), 3, (4, 5)), [])
>>> unpack_args(range(6), [-1,1,1,1,1])  # error
(((0, 1), 5, 4, 3, 2), [])
>>> unpack_args(range(6), [-1, 1, 1])      # error
(((0, 1, 2, 3), 5, 4), [])

can work on patch + tests this week

untitaker:
Thanks @ckamradt!

PR #336, opened 2015-05-06
Handle multiple args after star arg

ckamradt:
Addresses issue #316.  Previous behavior only worked if there was
a single arg specified after an arg with nargs=-1.


untitaker:
BTW thanks for the patch @ckamradt!

PR #340, opened 2015-05-09
Make click.echo()'s 'nl=True' thread-safe

mktums:


untitaker:
We should use a global lock to do this, it will solve many more problems, see https://github.com/mitsuhiko/click/issues/293

untitaker:
Nevermind, that idea was silly.

untitaker:
@mktums Thanks, I slightly modified the patch to avoid too complex logic. I don't think there are any resulting performance penalties with this, but I didn't test.

mktums:
@untitaker Thank you, sir! Does it works with Python 2.7?

untitaker:
I've reverted that fix yesterday because it doesn't. There's an open PR that includes this fix, any suggestions would be welcome.

On 24 July 2015 15:22:45 CEST, Mike TUMS <notifications@github.com> wrote:
>@untitaker Thank you, sir! Does it works with Python 2.7?
>
>---
>https://github.com/mitsuhiko/click/pull/340#issuecomment-124518942

-- 
Sent from my phone. Please excuse my brevity.

untitaker:
Now it does, see #388

Issue #338, opened 2015-05-08
click.File() produces an iterator but click.open_file() does not


geowurster:
When setting an argument or option's type to `click.File()` the resulting instance of `TextIOWrapper` is an iterable, but when opening `stdin` with `click.open_file()` the resulting instance of `KeepOpenFile` is not.  `KeepOpenFile` allows access to the underlying file-like object's attributes with a smarter `__getattr__()` so I'm guessing omitting `__iter__()` was accidental.

Assuming this wasn't intentional, it seems like the best solution is to add an `__iter__()` method that just does `return self`.  When the object is iterated over the underlying file-like object's `next()` or `__next__()` is automatically called via the smart `__getattr__()`.  The other solution would be to just set `__iter__` equal to the underlying file-like object's `__iter__` but that seems less intuitive and could cause problems if the underlying object lacks an `__iter__` method, although I looked through the source code and that doesn't appear to be possible.

```python
# Produces an iterable file-like object
@click.option('--infile', type=click.File())

# Does not produce a file-like object
infile = click.open_file(path)
```

geowurster:
Sorry about all the name changes - I hit submit a little too early.

untitaker:
Fixed by https://github.com/mitsuhiko/click/pull/342

PR #342, opened 2015-05-19
Let click.utils.KeepOpenFile()/LazyFile() be an iterator.  For #338.

geowurster:
Added a `__iter__()` method to `KeepOpenFile()` and `LazyFile()` that just returns the underlying `self._file.__iter__()` so an exception will still be raised if the underlying file isn't an iterator.  `LazyFile()` calls its `open()` method first.

Included a unittest in `tests/test_utils.py` as well.

### DIFF
--- a/click/utils.py
+++ b/click/utils.py
@@ -191,6 +191,10 @@ class LazyFile(object):
     def __exit__(self, exc_type, exc_value, tb):
         self.close_intelligently()
 
+    def __iter__(self):
+        self.open()
+        return iter(self._f)
+
 
 class KeepOpenFile(object):
 
@@ -208,6 +212,9 @@ class KeepOpenFile(object):
 
     def __repr__(self):
         return repr(self._file)
+
+    def __iter__(self):
+        return iter(self._file)
 
 
 def echo(message=None, file=None, nl=True, err=False, color=None):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,7 +1,8 @@
 import os
 import sys
-import click
 
+import click
+import click.utils
 import click._termui_impl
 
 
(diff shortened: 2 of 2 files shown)


click/utils.py
```diff
@@ -209,6 +213,9 @@ def __exit__(self, exc_type, exc_value, tb):
     def __repr__(self):
         return repr(self._file)
 
+    def __iter__(self):
+        return self._file.__iter__()
```
untitaker: Could you use `iter()` here, analogous to the `__repr__` implementation?

tests/test_utils.py
```diff
@@ -1,7 +1,9 @@
 import os
 import sys
-import click
+import tempfile
```
untitaker: You should use http://pytest.org/latest/tmpdir.html instead.

geowurster:
@untitaker good points - updated.  The unittests for `KeepOpenFile()` and `LazyFile()` are so similar that I just merged them as well.

untitaker:
Actually I preferred the isolation between the two testcases.

geowurster:
@untitaker done.

untitaker:
Could you use `git rebase -i` to squash all commits into one?

geowurster:
@untitaker not that familiar with rebasing but I think I got it right.

untitaker:
Yup that's it. Thanks!

geowurster:
Looks like Travis is [backed up](https://www.traviscistatus.com/incidents/gks98h4lbdyh) so you may have to manually re-start the pypy builds.  If it helps [here's the build](https://travis-ci.org/geowurster/click) from my fork.

untitaker:
Thanks!

Issue #350, opened 2015-05-29
Ctrl-C in a pager causes program to quit


amjith:
I'm using `echo_via_pager()` in my program to display the output through a pager. But while I'm in the pager if Ctrl-C is pressed then system.exit(1) is called, without giving me an opportunity to handle it. 

I'm using this inside a [repl](https://github.com/dbcli/pgcli) and I'd like to have the ability to handle Ctrl-C. 

Here's an example to reproduce: 

```
import click

@click.command()
@click.option('--foo', prompt=True)
def cli(foo):
    try:
        click.echo_via_pager('foo=%s' % foo)
    except click.Abort:
        print('Tada!')
```

If you run this example and press Ctrl-C while in the pager, it will print 'Aborted!' and quit the program, I'd expect to see the message 'Tada!' instead.

untitaker:
You can still catch SystemExit

amjith:
I tried catching SystemExit and it still doens't work as I expect. 

```
import click

@click.command()
@click.option('--foo', prompt=True)
def cli(foo):
    try:
        click.echo_via_pager('foo=%s' % foo)
    except SystemExit:
        print('Tada!')
```

It still prints `Aborted!` and quits the program.

untitaker:
In my terminal it seems that `^C` kills both pager and program **and** leaves the terminal in an unusuable state.

untitaker:
In your case you need to catch KeyboardInterrupt though.

amjith:
You're right if I catch KeyboardInterrupt it is handled. But it still doesn't solve the problem. 

Looks like the ^C is surfaced by click but the pager is never killed. So even if I handle the KeyboardInterrupt the `less` command continues to print the contents on the screen. 

I feel like the ^C in the page need to be handled by click here: https://github.com/mitsuhiko/click/blob/master/click/_termui_impl.py#L306..L311

Wrap that line in `try/except KeyboardInterrupt` and then call `c.terminate()` and then reraise that KeyboardInterrupt. 

Otherwise there is no way for an app to call the echo_via_pager() and keep working when a user presses ^C in the pager.

untitaker:
I've pushed a fix to 4.x-maintenance and master that fixes your issue. It swallows `^C`, `echo_via_pager` doesn't raise an exception.

This behavior is then consistent with the other pager methods I've seen. I'm not sure if that's the behavior we want, but anything else is fairly hard to implement given the behavior of `less`.

PR #351, opened 2015-06-02
Ignore KeyboardInterrupt while writing to pager.

amjith:
This addresses a corner case in #350. 

@untitaker Your solution to ignore the KeyboardInterrupt in `echo_via_pager()` is absolutely right and it works great. But for cases where the amount of text that is being written into the pager is larger than the OS buffer it blocks on writing to the buffer. https://github.com/mitsuhiko/click/blob/4.x-maintenance/click/_termui_impl.py#L309

So it is necessary to ignore the ^C while that is happening as well. 

Test program:

```
import click

@click.command()
@click.option('--foo', prompt=True)
def cli(foo):
    s = open('large_file.txt')
    while True:
        a = click.prompt('>')
        if a == 'exit':
            break
        click.echo_via_pager(s.read())
```

Run the above program with a `large_file.txt` in the working folder. Make sure it has at least 5000 lines in there and press ^C right after the first page. You will trigger the old behavior where the KeyboardInterrupt is bubbled up to the click app and results in 'Aborted!'. With this PR that corner case if fixed.

### DIFF
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -306,7 +306,7 @@ def _pipepager(text, cmd, color):
     try:
         c.stdin.write(text.encode(encoding, 'replace'))
         c.stdin.close()
-    except IOError:
+    except (IOError, KeyboardInterrupt):
         pass
 
     # Less doesn't respect ^C, but catches it for its own UI purposes (aborting


untitaker:
Is the pager already active while click is writing data into it?

amjith:
Yes, the pager becomes active as soon as the pipe receives the data. In fact with the example I've shown above you can try it out. You might need a setup.py to go with it: 

```
from setuptools import setup

setup(
    name='foo',
    version='1.0',
    py_modules=['foo'],
    include_package_data=True,
    install_requires=[
        'click',
    ],
    entry_points='''
        [console_scripts]
        foo=foo:cli
    ''',
)
```

untitaker:
Thanks!

amjith:
Could you merge this to the 4.x maint branch or should I create a separate PR for that?

untitaker:
I already did.

PR #356, opened 2015-06-10
Enable plugins via setup tools entry-points.

geowurster:
This PR allows developers to open their CLI to external plugins via setuptools entry-points.  Plugins must be a `click.Group()` or `click.Command()` and be registered to an entry-point that is loaded by the `click.group()` decorator.  See `examples/plugins/README.rst` for a a walkthrough.

The primary modification is in `click.group()`, which has a new `plugins` keyword argument accepting `pkg_resources.iter_entry_points()`, or any iterator producing a `pkg_resources.EntryPoint()` per iteration.  Each entry-point is loaded and attached to the group being created by the group decorator.

``` python
from pkg_resources import iter_entry_points

import click

@click.group(plugins=iter_entry_points('entry.point'))
def cli():
   """Main CLI group."""
```

Plugins that raise an exception on load are intercepted and converted to a `click.core.BrokenCommand()` and overrides the command's short description with a note stating the plugin is broken and if invoked, the full traceback is printed.

Some things to consider:
1. The core package gets to decide where plugins are registered, not the plugins themselves.  This can create a UI with a lot of commands as it is difficult for some plugins to register to one group while other plugins register under a different group.  The workaround is for the core package to have additional entry-points for each group if the core package developer wants a deeper group-based CLI instead of one that is flat.
2. I haven't done anything with command chaining other than some R&D so I'm not quite sure of the positive or negative affects for this feature.  It seems like plugins that cannot be chained can can be registered alongside commands that can be chained, which may be confusing for end users.
3. Where else should this feature be documented?

My initial though is that the above two points, especially the second, are really documentation issues and are minor when compared to the added functionality.

/ cc @sgillies

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -4,6 +4,7 @@ import codecs
 from contextlib import contextmanager
 from itertools import repeat
 from functools import update_wrapper
+import traceback
 
 from .types import convert_type, IntRange, BOOL
 from .utils import make_str, make_default_short_help, echo
@@ -1617,6 +1618,50 @@ class Argument(Parameter):
     def add_to_parser(self, parser, ctx):
         parser.add_argument(dest=self.name, nargs=self.nargs,
                             obj=self)
+
+
+class BrokenCommand(Command):
+
+    """
+    Rather than completely crash the CLI when a broken plugin is loaded, this
+    class provides a modified help message informing the user that the plugin is
+    broken and they should contact the owner.  If the user executes the plugin
+    or specifies `--help` a traceback is reported showing the exception the
+    plugin loader encountered.
+    """
+
+    def __init__(self, name):
+
+        """
+        Define the special help messages after instantiating
+        :class:`click.Command()`.
+        """
+
+        Command.__init__(self, name)
+
+        util_name = os.path.basename(sys.argv and sys.argv[0] or __file__)
+
+        if os.environ.get('CLICK_HONESTLY'):  # pragma no cover
+            icon = u'\U0001F4A9'
+        else:
+            icon = u'\u2020'
+
+        self.help = (
+            "\nWarning: entry point could not be loaded. Contact "
+            "its author for help.\n\n\b\n"
+            + traceback.format_exc())
+        self.short_help = (
+            icon + " Warning: could not load plugin. See `%s %s --help`."
+            % (util_name, self.name))
+
+    def invoke(self, ctx):
+
+        """
+        Print the traceback instead of doing nothing.
+        """
+
+        echo(self.help, color=ctx.color)
+        ctx.exit(1)
 
 
 # Circular dependency between decorators and core
--- a/click/core.py
+++ b/click/core.py
@@ -1027,10 +1027,21 @@ class Group(MultiCommand):
     :param commands: a dictionary of commands.
     """
 
-    def __init__(self, name=None, commands=None, **attrs):
+    def __init__(self, name=None, commands=None, plugins=None, **attrs):
         MultiCommand.__init__(self, name, **attrs)
         #: the registered subcommands by their exported names.
         self.commands = commands or {}
+
+        # Register external commands from setuptools entry-points
+        if plugins:
+            for entry_point in plugins:
+                try:
+                    self.add_command(entry_point.load())
+                except Exception:
+                    # Catch this so a busted plugin doesn't take down the CLI.
+                    # Handled by registering a dummy command that does nothing
+                    # other than explain the error.
+                    self.add_command(BrokenCommand(entry_point.name))
 
     def add_command(self, cmd, name=None):
         """Registers another :class:`Command` with this group.  If the name
--- a/click/decorators.py
+++ b/click/decorators.py
@@ -4,6 +4,7 @@ import inspect
 from functools import update_wrapper
 
 from ._compat import iteritems
+from .core import BrokenCommand
 from .utils import echo
 
 
@@ -115,13 +116,43 @@ def command(name=None, cls=None, **attrs):
     return decorator
 
 
-def group(name=None, **attrs):
+def group(name=None, plugins=None, **attrs):
     """Creates a new :class:`Group` with a function as callback.  This
     works otherwise the same as :func:`command` just that the `cls`
     parameter is set to :class:`Group`.
+
+    .. versionadded:: 5.0
+       Added `plugins`.
+
+    :param plugins: An iterable like `pkg_resources.iter_entry_points()` that
+                    procuces one instance of  :class:`pkg_resources.EntryPoint()`
+                    per iteration.  Used to load click commands or groups from
+                    setuptools entry-points and attach them to the group being
+                    created by this decorator.
     """
+
     attrs.setdefault('cls', Group)
-    return command(name, **attrs)
+    attrs.update(name=name)  # Some people may be using name as a positional arg
+
+    def decorator(f):
+
+        attrs.setdefault('cls', Group)
+        obj = command(**attrs)(f)
+
+        if plugins is not None:
+            for entry_point in plugins:
+                try:
+                    obj.add_command(entry_point.load())
+
+                except Exception:
+                    # Catch this so a busted plugin doesn't take down the CLI.
+                    # Handled by registering a dummy command that does nothing
+                    # other than explain the error.
+                    obj.add_command(BrokenCommand(entry_point.name))
+
+        return obj
+
+    return decorator
 
 
 def _param_memo(f, param):
--- a/click/decorators.py
+++ b/click/decorators.py
@@ -116,43 +116,13 @@ def command(name=None, cls=None, **attrs):
     return decorator
 
 
-def group(name=None, plugins=None, **attrs):
+def group(name=None, **attrs):
     """Creates a new :class:`Group` with a function as callback.  This
     works otherwise the same as :func:`command` just that the `cls`
     parameter is set to :class:`Group`.
-
-    .. versionadded:: 5.0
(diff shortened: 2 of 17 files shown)


click/decorators.py
```diff
@@ -115,13 +116,43 @@ def decorator(f):
     return decorator
 
 
-def group(name=None, **attrs):
+def group(name=None, plugins=None, **attrs):
```
untitaker: I don't think the "convenience API" (the decorators) should contain that logic.

untitaker:
I haven't fully looked at it because this PR is massive, looks like very interesting functionality, but I am not sure if it's not a better idea to provide a separate package for that.

What I'd like to see is a `click` organization that holds code extending the core functionality of click with new command types etc.

click/decorators.py
geowurster: Good point - it would probably make sense to put this in `click.core.Group()`.  I wasn't really thinking about how the decorators are technically just provided for convenience.

l0kix2:
Am i correctly understand, that it helps to build core cli tool with some name and add subcommands for this tool by installing python packages with described entrypoints in setup.py? If so, this is very interesting, i was looking for something like that for click.

geowurster:
@l0kix2 Yep.  This PR includes some [examples](https://github.com/geowurster/click/tree/master/examples/plugins) if you didn't already find them.

@untitaker As per your suggestion, I moved the plugin registration from the `group()` decorator to the `Group()` class.  This PR includes several examples so it seems much larger than it actually is, although the implications for adding a new feature like this are still potentially large.  The code that actually implement the feature is the plugin registration in [`click.core.Group()`](https://github.com/geowurster/click/blob/master/click/core.py#L1035-L1044) and the new [`click.core.BrokenCommand()`](https://github.com/geowurster/click/blob/master/click/core.py#L1634-L1675) for when registration fails.

If some real-world examples would help, we needed CLI plugins for [Fiona](https://github.com/toblerity/Fiona) and [rasterio](https://github.com/mapbox/rasterio), two geospatial data IO/processing libraries, and after some [back and forth](https://github.com/Toblerity/Fiona/issues/224) we 
determined that modifying the group decorator offered the most familiar syntax so we implemented plugin support in a common helper module called [cligj](https://github.com/mapbox/cligj) (stands for CLI GeoJSON) that both depend on.  The two projects are for processing two different types of data but they share some idioms so the primary goal of cligj is to unify the CLI between the two projects so the same flag means the same thing in both projects, but it doesn't really make sense to ask the click community to use an extension package specific to the geospatial world.  I am also using plugins for a (non-production ready) [project at work](https://github.com/SkyTruth/gpsdio) that heavily relies on plugins to facilitate adding analytical CLI commands that may have deep and complex dependencies.  Some of these commands are for internal use only but implementing them as plugins lets us flip the repo to public when we're ready.

I agree that a group hosting click extension packages would be good, but supporting plugins within click itself fits with the "batteries included" ethos of Python.  I think extension packages makes sense for [edge cases](https://github.com/geowurster/str2type), domain specific [argument types](https://github.com/mapbox/cligj), or functionality requiring additional dependencies, but this feature cleanly extends something click [already advocates](http://click.pocoo.org/4/setuptools/) via some non-invasive internal changes and a single additional keyword argument in the API.

click/core.py
```diff
@@ -1026,11 +1027,22 @@ class Group(MultiCommand):
...
 
-    def __init__(self, name=None, commands=None, **attrs):
+    def __init__(self, name=None, commands=None, plugins=None, **attrs):
         MultiCommand.__init__(self, name, **attrs)
         #: the registered subcommands by their exported names.
         self.commands = commands or {}
 
+        # Register external commands from setuptools entry-points
+        if plugins:
+            for entry_point in plugins:
```
untitaker: You can use `for entry_point in plugins or ()` here.

examples/plugins/PrinterStyle/printer_style/core.py
```diff
@@ -0,0 +1,46 @@
...
+@click.argument('outfile', type=click.File('w'), default='-')
+@click.option('-c', '--color', type=click.Choice(COLORS), required=True)
+def background(infile, outfile, color):
+
+    """
+    Add a background color.
+    """
+
+    for line in infile:
+        click.echo(line, file=outfile, color=color)
```
untitaker: The example is buggy, `color` has a different meaning. You want to use `secho`.

untitaker:
>I agree that a group hosting click extension packages would be good, but supporting plugins within click itself fits with the "batteries included" ethos of Python.

I personally think that ethos has had mostly negative consequences. Many stdlib modules are broken, and adoption rate of third-party replacements is slow because many projects want to avoid dependencies due to Python's bad packaging story.

I've seen the results of this in another project of Armin too. Werkzeug contains many things in its `contrib` module that are only there because *they are useful for webdev*, not because they are related to Werkzeug's goal of being a WSGI library. At that time it was useful to just have those modules rather than having to deal with setuptools, distutils and whatever you need to install more packages. Click, on the other hand, has a *dependency* on setuptools.

untitaker:
Note that I'm not trying to dispute your need for a plugin architecture (using setuptools entry points), but rather the need for it to be part of Click.

untitaker:
And If you think a third-party package wouldn't result in an API as clean:

    @with_plugins(plugins=iter_entry_points('entry.point'))
    @click.group()
    def cli():
       """Main CLI group."""

geowurster:
> Note that I'm not trying to dispute your need for a plugin architecture (using setuptools entry points), but rather the need for it to be part of Click.

I never thought you were, and my primary concern is making sure this is implemented properly, so if you feel strongly that implementing this externally is the right choice then I will whole-heartedly support that decision and continue to help get it set up.  Ultimately you know more than I do about click's history, source code, and intentions so I defer to your final judgement  :smile:.

If this ends up being in a separate package, how do you envision that being handled?  One concern I have is that if this is just a package that I own at something like `github.com/geowurster/click-plugins` then people may not find it, which means they're missing out on some pretty cool functionality.  You mentioned wanting to see more extension packages so maybe a `ClickExtension` GitHub group (that is advertised in this repo) where features like this are implemented and other lobby to have their own extension packages included with the "official" set?

Now for some additional comments:

> I personally think that ethos has had mostly negative consequences. 

> [...] due to Python's bad packaging story

> [..] Click, on the other hand, has a dependency on setuptools. 

The first two are good points but I'm a little confused about the third.  Could you explain with a little more detail?

The click documentation already does a good job explaining how to use setuptools so to some extent this is an opportunity to encourage good packaging in the Python ecosystem.  Reading that documentation has cleaned up my packaging style and is a big part of why I use setuptools instead of distutils or some custom script.  To be fair, this also means that if this is implemented externally it would be as trivial to install as a `pip install click-plugins` or `install_requires=['click-plugins']`.

> And If you think a third-party package wouldn't result in an API as clean:

My point about the API is mainly that using `click.goup(plugins=...)` doesn't necessarily complicate the API enough to warrant an additional decorator.

untitaker:
>The first two are good points but I'm a little confused about the third. Could you explain with a little more detail?

The third point is more of an example how a few years ago, bundling functionality tangentially related to the core package had clear advantages (see Werkzeug). Today, python-requests still goes as far as bundling all dependencies, so that they can be used just by downloading the tarball and importing it. All of this just to avoid setuptools. But Click doesn't try to avoid setuptools or Python's packaging system at all, so the philosophy from Werkzeug, Requests and Python's stdlib is not appropriate here.

(Explaining this in detail made me realize how poor/contrived this point is)

untitaker:
Regarding a Click org, that's what I was suggesting further above in this thread.

geowurster:
I'm happy to join such an org and help out if that's where this ends up.

I would be interested to see what others have to say about this.  Is anyone else following along with an opinion?

untitaker:
I'm currently writing an "extension" for click. I'd like to set this organization up ASAP but this is something that should be decided by @mitsuhiko who is hard to reach.

geowurster:
Sure.  I can already implement this feature with cligj so I'm not in a huge rush to get this resolved.  I'll let this sit for a while so everyone can think about it.  Thanks for your feedback!

untitaker:
I've created https://github.com/click-contrib and invited you. I'm not sure if
you're able to create repositories, if not, let me know which ones you want to
create.

geowurster:
@untitaker thanks - looks like I cannot create repos.  Are we waiting for Armin to weigh in on this or is this definitely heading towards `click-contrib`?  If so, I'll need a `click-plugins` repo and maybe we want a repo where people can open tickets to submit their own external repos?  Otherwise it will be difficult for people to get involved.

untitaker:
I've created a `click-plugins` repo.

Management of click-contrib repos should happen on this bugtracker -- Armin didn't respond yet, but I don't think he has anything against it.

geowurster:
Hmmm looks like I can't really do anything in `click-plugins`.

What do you mean by "on this bugtracker"?

untitaker:
Sorry, you should have access now. I mean the bugtracker we're currently on -- Click's.

geowurster:
Now I have access - thanks!  Using click tickets to direct contributions makes sense.

untitaker:
@geowurster Can I close this?

BTW I had issues with pushing to the new organization, it might be that your SSH key has to be approved just for the organization.

geowurster:
Yeah go ahead.  I opened an issue in the plugins repo and assigned it to me
so I wouldn't lose track. It will probably be several days until I get
around to migrating.

Thanks for the heads up about the potential key issue. I'll find you if I
run into any problems.

PR #359, opened 2015-06-13
Improve help text formatting.

markbt:
I promised an updated pull request, so here it is.

In fact it turns out that the bug I was seeing had actually been fixed on master, I was just using an out-of-date version of click.

However, I do have a couple of improvements that you might be interested in:
- Clean up the code.  There's actually no need for a separate function to add subsequent indents, we can do what we want with the normal wrap function as long as we put the command in the initial_indent, rather than printing it out first and starting the textwrapper when the output cursor is not in column zero.
- When the command is too long for the arguments to fit on the right of the command help text, wrap them to the next line at a reduced indent.

For comparison, given a long command and a narrow terminal:

With current master:

```
Usage: format.py long_subgroup_name long_command_name [OPTIO
NS]
                                                      ARGUME
NT1
                                                      ARGUME
NT2

  This is a long command.
```

With this change:

```
Usage: format.py long_subgroup_name long_command_name
            [OPTIONS] ARGUMENT1 ARGUMENT2

  This is a long command.
```

Of course, if it gets so narrow that the command name itself doesn't fit, things still start to look wrong.

### DIFF
--- a/click/formatting.py
+++ b/click/formatting.py
@@ -123,20 +123,20 @@ class HelpFormatter(object):
         :param args: whitespace separated list of arguments.
         :param prefix: the prefix for the first line.
         """
-        prefix = '%*s%s ' % (self.current_indent, prefix, prog)
+        usage_prefix = '%*s%s ' % (self.current_indent, prefix, prog)
         text_width = self.width - self.current_indent
 
-        if text_width >= (term_len(prefix) + 20):
+        if text_width >= (term_len(usage_prefix) + 20):
             # The arguments will fit to the right of the prefix.
-            indent = ' ' * term_len(prefix)
+            indent = ' ' * term_len(usage_prefix)
             self.write(wrap_text(args, text_width,
-                                 initial_indent=prefix,
+                                 initial_indent=usage_prefix,
                                  subsequent_indent=indent))
         else:
             # The prefix is too long, put the arguments on the next line.
-            self.write(prefix)
+            self.write(usage_prefix)
             self.write('\n')
-            indent = ' ' * (self.current_indent + 12)
+            indent = ' ' * (max(self.current_indent, term_len(prefix)) + 4)
             self.write(wrap_text(args, text_width,
                                  initial_indent=indent,
                                  subsequent_indent=indent))
--- a/click/formatting.py
+++ b/click/formatting.py
@@ -18,12 +18,6 @@ def iter_rows(rows, col_count):
         yield row + ('',) * (col_count - len(row))
 
 
-def add_subsequent_indent(text, subsequent_indent):
-    lines = text.splitlines()
-    lines = lines[:1] + [subsequent_indent + line for line in lines[1:]]
-    return '\n'.join(lines)
-
-
 def wrap_text(text, width=78, initial_indent='', subsequent_indent='',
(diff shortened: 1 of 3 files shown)


click/formatting.py
```diff
@@ -133,14 +123,23 @@ def write_usage(self, prog, args='', prefix='Usage: '):
...
         """
-        prefix = '%*s%s' % (self.current_indent, prefix, prog)
-        self.write(prefix)
-
-        text_width = max(self.width - self.current_indent - term_len(prefix), 10)
-        indent = ' ' * (term_len(prefix) + 1)
-        self.write(wrap_text(args, text_width,
-                             initial_indent=' ',
-                             subsequent_indent=indent))
+        prefix = '%*s%s ' % (self.current_indent, prefix, prog)
```
untitaker: Why the extra space here?

click/formatting.py
```diff
@@ -133,14 +123,23 @@ def write_usage(self, prog, args='', prefix='Usage: '):
...
+            # The arguments will fit to the right of the prefix.
+            indent = ' ' * term_len(prefix)
+            self.write(wrap_text(args, text_width,
+                                 initial_indent=prefix,
+                                 subsequent_indent=indent))
+        else:
+            # The prefix is too long, put the arguments on the next line.
+            self.write(prefix)
+            self.write('\n')
+            indent = ' ' * (self.current_indent + 12)
```
untitaker: This seems hardcoded, shouldn't this somehow depend on the length of `prefix`?

untitaker:
This somehow reverts all of https://github.com/mitsuhiko/click/pull/240 FWIW

click/formatting.py
markbt: This is the space from initial_indent (old line 142).

click/formatting.py
untitaker: By that I mean the *initial* prefix, `"Usage: "`.

markbt: At this point we know that prefix is too long, so there won't be enough space if we base it on prefix. We need a new indent value.

I picked 12 as that gives a nice level of indent relative to the "Usage: ..." line, so it appears visually indented.

markbt:
It should behave the same as #240 - but without the TextWrapper problems that @mitsuhiko was uncomfortable with.

click/formatting.py
untitaker: Right, I meant the original meaning of prefix, which is just the `"Usage: "` part, not including the program name.

This gives the variable a new meaning:
    prefix = '%*s%s' % (self.current_indent, prefix, prog)

markbt: Ah, right.  Then yes, it probably should be - maybe new-`prefix` should be renamed `initial_indent` and this should be `max(self.current_indent, term_len(prefix)) + 5`.

untitaker: I'm curious why you'd picked a number for the additional indent that is not a multiple of 4?

markbt: It was arbitrary (well, `12 - term_len("Usage: ")`).  I'm updating the branch now, and will use 4.

untitaker:
Excellent, could you add a changelog too?

untitaker:
Awesome, thanks!

Issue #362, opened 2015-06-19
CliRunner.invoke returns "wrong" error code when using sys.exit


torfuspolymorphus:
`CliRunner.invoke` returns the "wrong" error code when the target function uses `sys.exit(error_string)`. Instead of a numerical code (as expected), `Result.error_code` contains the string message of the `SystemExit` exception raised by `sys.exit`. The reason is that `SystemExit.code` (which is [used by `CliRunner.invoke`](https://github.com/mitsuhiko/click/blob/master/click/testing.py#L280)) also contains that string:
```
In [1]: import sys

In [2]: try:
   ...:     sys.exit('error')
   ...: except SystemExit as e:
   ...:     print e.code
   ...:     
error
```
The [documentation of `sys.exit`](https://docs.python.org/2/library/sys.html#sys.exit) says that

> The optional argument arg can be an integer giving the exit status (defaulting to zero), or another type of object. [...] If another type of object is passed, None is equivalent to passing zero, and any other object is printed to stderr and results in an exit code of 1.

However, the actual exit code seems to be set somewhere else (and only if the `SystemExit` exception is not caught, obviously).

In my opinion, `CliRunner` should try to report the exit code that is reported to the OS. Hence, if `SystemExit.code` is not an integer, `Result.exit_code` should be 1.

I'd be happy to provide a corresponding PR if that helps.

torfuspolymorphus:
It seems that `Result.output` also doesn't contain the error message in that case: Compare the output of 
```
import sys
import click
from click.testing import CliRunner

@click.command()
def main():
    sys.stderr.write('error')
    sys.exit(1)

runner = CliRunner()
result = runner.invoke(main, [])
print 'Output: %r' % result.output
```
with that of 
```
import sys
import click
from click.testing import CliRunner

@click.command()
def main():
    sys.exit('error')

runner = CliRunner()
result = runner.invoke(main, [])
print 'Output: %r' % result.output
```
In my opinion, `Result.output` should be the same in both cases, since both cases yield the same console output when not run via `CliRunner`.

untitaker:
`sys.exit("...")` is IMO not a thing people ought to do, but I'd agree that `CliRunner`'s isolation layer is too thin there. I'd be happy to merge a PR:

untitaker:
Fixed with https://github.com/mitsuhiko/click/pull/363

PR #363, opened 2015-06-19
Fix issue #362.

torfsen:
Fixes issue #362 (invalid `Result.return_code` and `Result.output` when using `CliRunner` and `sys.exit`).

### DIFF
--- a/click/testing.py
+++ b/click/testing.py
@@ -270,6 +270,7 @@ class CliRunner(object):
         with self.isolation(input=input, env=env, color=color) as out:
             exception = None
             exit_code = 0
+            error_message = ''
 
             try:
                 cli.main(args=args or (),
@@ -278,6 +279,9 @@ class CliRunner(object):
                 if e.code != 0:
                     exception = e
                 exit_code = e.code
+                if not isinstance(exit_code, int):
+                    error_message = '%s\n' % exit_code
+                    exit_code = 1
                 exc_info = sys.exc_info()
             except Exception as e:
                 if not catch_exceptions:
@@ -287,7 +291,9 @@ class CliRunner(object):
                 exc_info = sys.exc_info()
             finally:
                 sys.stdout.flush()
-                output = out.getvalue()
+                if not PY2:
+                    error_message = error_message.encode(self.charset)
+                output = out.getvalue() + error_message
 
         return Result(runner=self,
                       output_bytes=output,
--- a/CHANGES
+++ b/CHANGES
@@ -12,6 +12,8 @@ Version 5.0
 - Atomic files now only accept the `w` mode.
 - Change the usage part of help output for very long commands to wrap
   their arguments onto the next line, indented by 4 spaces.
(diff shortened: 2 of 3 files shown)


click/testing.py
```diff
@@ -278,6 +279,14 @@ def invoke(self, cli, args=None, input=None, env=None,
                 if e.code != 0:
                     exception = e
                 exit_code = e.code
+                # Fix for issue #362: `SystemExit.code` sometimes contains
+                # the error message, not the exit code.
+                try:
+                    if exit_code != int(exit_code):
+                        raise ValueError
+                except ValueError:
```
untitaker: You might as well use `isinstance(exit_code, int)` here.

click/testing.py
```diff
@@ -287,7 +296,9 @@ def invoke(self, cli, args=None, input=None, env=None,
                 exc_info = sys.exc_info()
             finally:
                 sys.stdout.flush()
-                output = out.getvalue()
+                if not PY2:
+                    error_message = bytes(error_message, self.charset)
```
untitaker: Use `error_message.encode(charset)` here.

click/testing.py
```diff
@@ -278,6 +279,14 @@ def invoke(self, cli, args=None, input=None, env=None,
                 if e.code != 0:
                     exception = e
                 exit_code = e.code
+                # Fix for issue #362: `SystemExit.code` sometimes contains
```
untitaker: I don't think you have to reference that issue here, it's obstructive.

torfuspolymorphus: Fixed.

click/testing.py
torfuspolymorphus: Fixed.

click/testing.py
torfuspolymorphus: I'd rather not tie this to a specific data type (and if so, `numbers.Integral` would probably be a better choice). The current implementation works for any kind of number, including integers stored in a `float`.

untitaker: Floating-point exit codes don't make sense, also this will trigger probably unexpected behavior with strings like "1" or even "0xfff".

We should mimick the behavior `sys.exit` exposes.

torfuspolymorphus: Here's the behavior of `sys.exit` for some corner cases:
```
$ python -c "import sys; sys.exit(1)"; echo $?
1
$ python -c "import sys; sys.exit(1.0)"; echo $?
1.0
1
$ python -c "import sys; sys.exit('1')"; echo $?
1
1
```
So integer float values are treated as non-integers.

So you were right after all, `isinstance(exit_code, int)` is the right choice. It's fixed now.

untitaker: Floats are simply not integers.

Thanks for this PR, if you have time, please squash all commits into one with rebase. Otherwise I will when I merge it.

torfuspolymorphus: Commits have been squashed.

untitaker:
Thanks!

PR #376, opened 2015-07-13
Click testing crashes in case of non existing name attribute.

tian2992:
Should fix error when calling and no name attribute.

```
  File "/usr/lib/python2.7/site-packages/click/testing.py", line 276, in invoke
    prog_name=self.get_default_prog_name(cli), **extra)
  File "/usr/lib/python2.7/site-packages/click/testing.py", line 128, in get_default_prog_name
    return cli.name or 'root'
AttributeError: 'module' object has no attribute 'name'
```

### DIFF
--- a/click/testing.py
+++ b/click/testing.py
@@ -125,7 +125,10 @@ class CliRunner(object):
         for it.  The default is the `name` attribute or ``"root"`` if not
         set.
         """
-        return cli.name or 'root'
+        try:
+            return cli.name
+        except AttributeError:
+            return 'root'
 
     def make_env(self, overrides=None):
         """Returns the environment overrides for invoking a script."""


untitaker:
I think you're using Click wrong. `cli` shouldn't be a module.

tian2992:
This is my use case, main is the module.
https://github.com/fashiontec/knitlib/blob/master/src/knitlib/__main__.py

I had the exception when setting a logger on a test case.

untitaker:
This is your problem: https://github.com/fashiontec/knitlib/blob/master/tests/test_knitlib.py

You're importing the `__main__` module as CLI, but you need to do `from knitlib.__main__ import main`.

tian2992:
Thanks, can you please explain why use one over another? I based this test on Cookiecutter's so if that can be fixed, it would be good too...

untitaker:
Could you link which file you've based it on?

tian2992:
@untitaker https://github.com/ionelmc/cookiecutter-pylibrary/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/tests/test_%7B%7Bcookiecutter.package_name%7D%7D.py 

here it is.

untitaker:
That file doesn't have this problem.

Issue #293, opened 2015-02-05
Implement threadlocks on CLI functions.


untitaker:
This is a continuation of https://github.com/mitsuhiko/click/issues/292. It has also been discussed with @mitsuhiko in IRC.

The problems:

- If you use threads in your CLI app, and they all use ``click.echo``, you get garbage on stdout (written data is mixed up).
- If one thread uses ``click.prompt``, and some other thread uses ``click.echo``, you get garbage.
- The list goes on...

The solution: Implement a global UI lock that prevents multiple threads from simultaneously writing to stdout.

See IRC log: https://botbot.me/freenode/pocoo/2015-01-24/?msg=30379697&page=6

untitaker:
I'm not sure if we should implement the fast path for `click.echo` as mentioned in IRC.

Basically I'd like the locking support to be non-transparent, i.e. make the user *provide* a lock rather than having automatic support for threading. The primary reason is that the user might use `multiprocessing` (or anything else) instead, which would require a different lock.

This could happen in combination with https://github.com/mitsuhiko/click/issues/292: The UI lock could live on the context in the form of an attribute `ctx.ui_lock`, with a default value of `None`, which equates to the current behavior (no locking at all). If the user then uses threading, they could set this value to their custom lock.

The user has to explicitly "initiate" locking, but as a consequence, the lock is exposed to the user as well. This is useful if one thread is launching e.g. a "configuration wizard" (a series of prompts) and not only wants to avoid overlapping prompts, but also wants to prevent other threads from writing *between* the wizard's prompts. This problem is easily solved by setting Click's lock to an `RLock` instance and holding that lock while the wizard is running.

We could still have simple thread locking with a fast path as default, but I don't the special casing is worth it. For anything nontrivial, the user almost certainly wants to override it.

untitaker:
What we also could do is to provide an interface for making the *acquiring function* known to the lock. This would allow for a custom threadlock wrapper that decides on a per-function basis whether to use an eventual fast path or not.

untitaker:
And I just realized that providing the same ctx in each subprocess of `multiprocessing`-based systems is impossible :( So I suppose the object can't live on the context if we want to support `multiprocessing`.

untitaker:
Closing this because it's a bad idea: `click.echo` blocking on prompts is a bad idea performance-wise. I'm not sure how else to solve this problem, but any effort should probably happen in a contrib package.

PR #380, opened 2015-07-15
First implementation of UI locks

untitaker:
See #293


untitaker:
cc @mitsuhiko

untitaker:
Regarding r/w-locks, I'm not sure if the performance gains aren't cancelled out by the string concatenation that (probably) becomes necessary in `echo`.

PR #401, opened 2015-08-07
progressbar label should not be shown when the output is not a tty

pombredanne:
If you pass a label to a progressbar, the label is incorrectly included when you redirect the command output to a file. 

The docstring states: "By default, this progress bar will not be rendered if the file is not a terminal."
However, the label is always displayed even when the output is not a tty.

The `render_finish` function does this correctly:
https://github.com/mitsuhiko/click/blob/36f83d91e27b5b88ee806cb40f284658e2f9cdb5/click/_termui_impl.py#L106

In contrast `render_progress` always prints a label even if `self.is_hidden`, and should just return instead without any echo:
https://github.com/mitsuhiko/click/blob/36f83d91e27b5b88ee806cb40f284658e2f9cdb5/click/_termui_impl.py#L183

This simple command shows the issue:

```
import click

@click.command()
def mycli():
    with click.progressbar(range(2), label='xyz') as pb:
        for _ in pb:
           pass

if __name__ == '__main__':
    mycli()
```

This is correct:

```
$ python click_checker.py
xyz  [------------------------------------]    0%
```

This is not:

```
$ python click_checker.py > tst
$ cat tst
xyz
```

I would expect the redirected stdout to be empty in this case.

This commit fixes this.

### DIFF
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -181,8 +181,6 @@ class ProgressBar(object):
         from .termui import get_terminal_size
 
         if self.is_hidden:
-            echo(self.label, file=self.file, color=self.color)
-            self.file.flush()
             return
 
         # Update width in case the terminal has been resized
--- a/tests/test_termui.py
+++ b/tests/test_termui.py
@@ -12,3 +12,17 @@ def test_progressbar_strip_regression(runner, monkeypatch):
 
     monkeypatch.setattr(click._termui_impl, 'isatty', lambda _: True)
     assert label in runner.invoke(cli, []).output
+
+
+def test_progressbar_should_not_show_in_non_tty(runner, monkeypatch):
+    label = 'progress label'
+
+    @click.command()
+    def cli():
+        with click.progressbar(range(3), label=label) as progress:
+            for thing in progress:
+                pass
+
+    monkeypatch.setattr(click._termui_impl, 'isatty', lambda _: False)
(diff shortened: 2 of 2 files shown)


untitaker:
Sorry for responding so late.

>The docstring states: "By default, this progress bar will not be rendered if the file is not a terminal."
However, the label is always displayed even when the output is not a tty.

The bar is not displayed, the label is. This is intended behavior I think.

pombredanne:
@untitaker I kindly disagree. The label is displayed as part of the bar and is as much part of the bar as the pos, the item show function or anything else. 

This does pollute a non-tty redirection for no good reason I can fathom. Furthermore, it prints it to a non-tty more than one time.

pombredanne:
and @untitaker it would be nice not to close the PR abruptly without a discussion first

untitaker:
@pombredanne You make this seem like a bug when it's not. This is intended behavior -- the label contains information about the work that is currently done by the launched program. Since you can't output anything else while the progressbar is running (without creating glitches), this is the only option to show any information at all, besides the progressbar itself. Your fix would make any well-behaved Click app effectively silent in non-TTY mode.

untitaker:
If you want to discuss the current behavior, please do open an issue for it.

pombredanne:
@untitaker fair enough I created #406 as an issue.

PR #419, opened 2015-08-30
Fix #418 ProgressBar should flush only when necessary

### LINKED ISSUE #418: ProgressBar should flush only when necessary
Current ProgressBar implementation enforces the output is flushed in every iteration step, regardless of whether the bar UI needs to be updated. This can result in significantly high IO overhead.

For example, the following code takes up to minutes to run while the same iteration without using a progress bar takes milliseconds.
```python
with click.progressbar(xrange(10 ** 6)) as pb:
    for i in pb:
        # do the work
        pass
```

The issue is worse in a remote ssh session, in which case bandwidth is also wasted.

ohwxyz:
This change is testable, for example, to count the number of flushes for the given file, not sure how useful it is though.

### DIFF
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -87,6 +87,7 @@ class ProgressBar(object):
         self.entered = False
         self.current_item = None
         self.is_hidden = not isatty(self.file)
+        self.last_rendered_line = None
 
     def __enter__(self):
         self.entered = True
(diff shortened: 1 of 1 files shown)


ohwang:
Close as #418 is fixed in commit c4c9eb23243174cd1ffc9fd0162caf31e93b49ee

PR #422, opened 2015-09-08
Option value separators

jvrsantacruz:
Adds _option_value_separators_ context setting to override default option value parsing

Currently `OptionParser` splits long option values using the '=' character.  eg: 

`command   --option=value`

This option extends the parser to admit value splitting via user defined text separator. eg: 

`command --option:value`

This is done by adding a new configuration variable named _option_value_separators_ at `Command` and `Context` level. This var holds a list of text separator that might be used to split long options into its name and associated parameter value.

``` python
CONTEXT_SETTINGS = dict(option_value_separators=set(['=', ':']))

@click.command(context_settings=CONTEXT_SETTINGS)
@click.option('--name')
def cli(name):
click.echo('Name: %s' % name)
```

The main purpouse of this change is to allow Windows Click users to compliment Microsoft [Command Line Standard](https://technet.microsoft.com/en-us/library/ee156811.aspx), which allows options to separate its value using a colon `:` instead of *nix convention which uses `=`.

> Consistent Syntax
> 
> Required:
> 
> ```
> Commands must use the “-“ character for parameter delimitation; a
> colon (:) or whitespace to separate the parameter name and its
> arguments and a comma (,) to separate multiple values within an
> argument.
> ```

To add a bit more of context to reason why someone would ever need this, I'm currently wrapping and extending an existing awfully big Windows C++ legacy application using Click. To my dismay, the operations group demanded the colon `--option:value` option syntax to be accepted, both because Windows CLI applications should accept it, and because the old application did it so. All options, overriden, redirected or new should admit this format.

If this does not get managed at Click parsing level, the only thing I could do would be to pre-parse `sys.argv` and replace the offending separators before passing it to Click.

Thanks for all the good work. :)

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -175,7 +175,8 @@ class Context(object):
                  resilient_parsing=False, allow_extra_args=None,
                  allow_interspersed_args=None,
                  ignore_unknown_options=None, help_option_names=None,
-                 token_normalize_func=None, color=None):
+                 token_normalize_func=None, color=None,
+                 option_value_separators=None):
         #: the parent context or `None` if none exists.
         self.parent = parent
         #: the :class:`Command` for this context.
@@ -250,6 +251,8 @@ class Context(object):
         #:
         #: .. versionadded:: 4.0
         self.ignore_unknown_options = ignore_unknown_options
+
+        self.option_value_separators = option_value_separators
 
         if help_option_names is None:
             if parent is not None:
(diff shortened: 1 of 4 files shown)


con-f-use:
I think this might be more of a case for [click-contrib](click.pocoo.org/latest/contrib/). As far as I know click tries to confirm to POSIX and your use case is rather obscure.

untitaker:
Yes, this is definetly out of scope for Click, sorry.

PR #459, opened 2015-11-19
Add tee capability to CliRunner.invoke().

pkrog:
Hi,

I'm a French software developer, working at CEA (http://www.cea.fr/english-portal). For a project using Galaxy (https://galaxyproject.org/), I've been looking at the related project Planemo (http://planemo.readthedocs.org/en/latest/index.html), which uses your library "click". One issue I have is that Planemo tests run a script using your CliRunner class, but this script takes a long time and, since stdout is hidden, no feedback is given to the user until the script terminates and Planemo tests finally print the output.
So I've decided to modify your library in order to be able to print continuously stdout and stderr while running. I've added a parameter called tee inside the invoke method, and I set it to false by default. When set to true, a class called MultOutput will be instantiated instead of StringIO, that will take for parameters the StringIO instance and also stdout (or stderr). Each output will then be printed into StringIO as well as into stdout (or stderr).

Don't hesitate to contact me if you have any question.

Regards,
Pierrick

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ build
 docs/_build
 click.egg-info
 .tox
+.swp
+.*.swp
+.cache
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 test:
-	@cd tests; PYTHONPATH=.. py.test --tb=short
+	@cd tests; PYTHONPATH=.. py.test -s --tb=short
 
 upload-docs:
 	$(MAKE) -C docs dirhtml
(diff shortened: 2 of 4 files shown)


pierrickrogermele:
My modification is only valid for Python 2 for the moment. I'm currently working on making it available for Python 3 too.

blueyed:
Interesting, related: https://github.com/django/django/pull/9743.

pkrog:
Discarded

PR #468, opened 2015-11-25
docs/custom multi-commands: prefer submodule import over eval

marthjod:
See #467.


davidism:
There's absolutely nothing wrong with `eval` in this case.  It's concise, and the files are from a trusted location.  You realize that importing allows arbitrary execution as well, right?

PR #483, opened 2015-12-10
Simplify MANIFEST.in commands

### LINKED ISSUE #480: DS_STORE in pypi tarball
Hello!

I am packaging click and I found .DS_STORE in the tarball. (```artwork/.DS_Store```).

also:
 - ```examples/.DS_Store```
 - ```examples/imagepipe/.DS_Store```

I guess that was uploaded by mistake!

nchammas:
The full MANIFEST.in syntax is here: https://docs.python.org/3/distutils/commandref.html

For what it's worth, many of the directories being packaged up here never get installed. They're just bundled up in the source distribution and ignored by pip on install. Only `click/` actually makes it onto the user's system.

Fixes #480.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,12 +1,9 @@
 include Makefile CHANGES LICENSE
-recursive-include artwork *
-recursive-include tests *
-recursive-include examples *
-recursive-include docs *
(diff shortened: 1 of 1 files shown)


untitaker:
LGTM, thanks!

PR #485, opened 2015-12-12
hide '0d' when eta<24h

naoyat:
t>0 is always true (and we always see '0d ...') if floor is not taken...


untitaker:
Please add tests.

PR #487, opened 2015-12-13
Show progressbar only if total execution time is noticable

con-f-use:
Currently `progressbar` displays its output immediately, even if the total execution time of what is being monitored is small. (Optionally) progress bar should only print something, if the total execution time is longer than e.g half a second or something equally unnoticeable.

**Reasoning:**
A user doesn't need progress displayed, if the whole execution monitored by the progress bar finishes within say half a second after its start. Here is an illustration:

```
import time
import click

if __name__ == '__main__':
    # Fast, should not generate output
    click.echo('fast:')
    with click.progressbar(range(10)) as bar:
        for data in bar:
            pass

    # Slow should generate output
    click.echo('slow:')
    with click.progressbar(range(100)) as bar:
        for data in bar:
            time.sleep(1)
```

**Output before pull:**

```
fast:
[####################################]  100% 
slow:
[####################################]  100% 
```

**Output after pull:**

```
fast:
slow:
[####################################]  100% 
```

**Discussion:**

In this proposed pull request, 0.5s are hard-coded as the time threshold. This is of course an arbitrary value and debatable. One could introduce another optional argument to `progressbar()` that lets the user specify the time threshold with 0.5s as default value (where 0 forces the display of progress bar no matter what = old behavior). Or zero could be the default value. I opted against an optional threshold, because I feel it is unnecessary clutter and 0.5-1s is what most people want, but you might disagree.

Since the added `time.time()` statements in `is_fast()` are quite quick and only executed conditionally, the impact on performance is negligible.


con-f-use:
Ah yes, the tests need to be updated to reflect the new changes. I didn't think about that. That's why the checks fail.

con-f-use:
Merged upstream changes into this fork.

dsully:
Any chance you could rebase & clean up the history?

Thanks

con-f-use:
Sure but maybe not this week.

con-f-use:
Okay I'm done. Sorry for last merge, no idea what made that necessary.

carlwgeorge:
Is this feature documented anywhere?  I spent a while debugging this behavior before searching through the issues and finding this pull request.

> One could introduce another optional argument to progressbar() that lets the user specify the time threshold with 0.5s as default value (where 0 forces the display of progress bar no matter what = old behavior).

Will this be implemented at some point?  I don't currently see a way to opt in to the old behavior.

con-f-use:
I have not tested it, but you could set `short_limit = -1`

carlwgeorge:
So something like this?

```
pb = progressbar(items)
pb.short_limit = -1
with pb as bar:
    for item in bar:
        do_something_with(item)
```

That's much less appealing than just being able to pass a keyword arg to `click.progressbar`.  And it's still completely undocumented.

con-f-use:
You are right, but I'm not sure if it's worth another pull request. Do you think you could explain, why you would want to display a progress bar in cases where progress is essentially instant?

carlwgeorge:
Isn't "because I want to" a good enough reason? :grinning: 

In my particular case, I was using progress bar labels to indicate the progression through the high level sections of my script.  When those labels weren't displayed, I was confused and wondered if the sections of my script were even running at all, and I questioned the logic in my own code.  Once I had triple checked my logic, I came to the GitHub issues to look and see if there were any reported issues about progress bars (and their labels) not being displayed.  As a work around in my code I'm using `click.echo` to indicated the high level progression through tasks in my script, so that they are always displayed even if the progress bar decides not to show.

At a minimum, please mention this in the documentation so others don't spend as much time as I did trying to find out why their progress bars or labels aren't being displayed.  And preferably, add an optional argument to `click.progressbar` to customize or disable the hiding time threshold.

con-f-use:
Well, it seems to me, that it is the intention of the progress bar to indicate progress of a lengthy on-goning progress and you'd be better off using `echo` or the logging module for your use case in the first place. However, I do think we need to improve documentation and, while at it, we could make the short limit a keyword argument to the constructor. Let's see first if there are similar complaints by other people.

davidism:
I'm going to revert this. #1648 reports that it's unexpected that the progress bar doesn't show until the second iteration for a long process. I ran into the issue too when fixing another progress bar test, and it took me a while to track it back to this. Reading over the rationale in this issue again, I'm not convinced that it's bad to show the progress bar if iterations are fast. And it actually seems to say that the progress bar should be hidden if the *full* iteration time is fast, but the implementation here just hides the first 0.25 seconds of iterations, regardless of the full time.

Maybe it would make sense to somehow hide the progress bar if all iterations finish quickly, but otherwise you end up in this weird situation where the bar displays late because the first iteration is always "fast". I don't see a simple way to implement that though. In general I want to avoid adding complexity to Click's progress bar at this point, and am pointing users at tqdm for advanced use cases.

PR #496, opened 2015-12-31
Added list of choices to termui prompt for user data if the data requested is a Choice

Rovanion:
I found it odd that the user had to give incorrect input in order to be presented with the list of available options, for example:

```
Group by: something
Error: invalid choice: something. (choose from none, day, week, month)
```

In the above example the user is asked to give the program some interval to group by but offers no help as to what the valid inputs are until the user has incorrectly entered 'something'. With the modifications included in this commit the prompt would look as follows:

```
Group by (none, day, week, month):
```

It is now clear to the user that it has a limited set of options to chose from. This of course works in combination with default options.

```
Group by (none, day, week, month) [week]:
```

It is possible to disable the display of the choices by passing show_choices=False just like passing show_default=False works.

Also added a test for the new behaviour.

The last test of the test_chain.py test suite xFails, whatever that means, but it does that on master too so I don't think it's related to my changes.

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -1461,7 +1461,7 @@ class Option(Parameter):
                  prompt=False, confirmation_prompt=False,
                  hide_input=False, is_flag=None, flag_value=None,
                  multiple=False, count=False, allow_from_autoenv=True,
-                 type=None, help=None, hidden=False, **attrs):
+                 type=None, help=None, hidden=False, show_choices=True, **attrs):
         default_is_missing = attrs.get('default', _missing) is _missing
         Parameter.__init__(self, param_decls, type=type, **attrs)
 
@@ -1507,6 +1507,7 @@ class Option(Parameter):
         self.allow_from_autoenv = allow_from_autoenv
         self.help = help
         self.show_default = show_default
+        self.show_choices = show_choices
 
         # Sanity check for stuff we don't support
         if __debug__:
@@ -1658,8 +1659,8 @@ class Option(Parameter):
         if self.is_bool_flag:
             return confirm(self.prompt, default)
 
-        return prompt(self.prompt, default=default,
-                      hide_input=self.hide_input,
+        return prompt(self.prompt, default=default, type=self.type,
+                      hide_input=self.hide_input, show_choices=self.show_choices,
                       confirmation_prompt=self.confirmation_prompt,
                       value_proc=lambda x: self.process_value(ctx, x))
 
--- a/click/termui.py
+++ b/click/termui.py
@@ -6,7 +6,7 @@ from ._compat import raw_input, text_type, string_types, \
      isatty, strip_ansi, get_winterm_size, DEFAULT_COLUMNS, WIN
 from .utils import echo
 from .exceptions import Abort, UsageError
-from .types import convert_type
+from .types import convert_type, Choice
 from .globals import resolve_color_default
 
 
@@ -24,22 +24,26 @@ def hidden_prompt_func(prompt):
     return getpass.getpass(prompt)
 
 
-def _build_prompt(text, suffix, show_default=False, default=None):
+def _build_prompt(text, suffix, show_default=False, default=None, show_choices=True, type=None):
     prompt = text
+    if type is not None and show_choices and isinstance(type, Choice):
+        prompt += ' (' + ", ".join(map(str, type.choices)) + ')'
     if default is not None and show_default:
         prompt = '%s [%s]' % (prompt, default)
     return prompt + suffix
 
 
-def prompt(text, default=None, hide_input=False,
-           confirmation_prompt=False, type=None,
-           value_proc=None, prompt_suffix=': ',
-           show_default=True, err=False):
+def prompt(text, default=None, hide_input=False, confirmation_prompt=False,
+           type=None, value_proc=None, prompt_suffix=': ', show_default=True,
+           err=False, show_choices=True):
     """Prompts a user for input.  This is a convenience function that can
     be used to prompt a user for input later.
 
     If the user aborts the input by sending a interrupt signal, this
     function will catch it and raise a :exc:`Abort` exception.
+
+    .. versionadded:: 7.0
+       Added the show_choices parameter.
 
     .. versionadded:: 6.0
        Added unicode support for cmd.exe on Windows.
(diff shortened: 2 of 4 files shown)


Rovanion:
This PR now has conflicts that need to be resolved. But other than that, is there any feedback you want to leave, anything you want to change in order merge these changes?

untitaker:
Depending on how many choices there are, this might become too verbose. Wouldn't it be better if this is explained through a custom helptext?

Rovanion:
> Depending on how many choices there are, this might become too verbose.

We can imagine some click application having more than twenty available options in a single choice, then it could become messy - iff the feature was enabled. But then such a large number of options in a choice is probably the exception rather than the norm amongst click programs.  And since the display of the available options in a choice has to be explicitly enabled by the developer I don't see how this is an issue in the context of the click framework providing this feature for developers to use. We have to be able to trust the program developer not to explicitly enable features of click which are unfit for his situation.

> Wouldn't it be better if this is explained through a custom helptext?

This is automatic. It changes with the code and stays consistent with the program at all times. A custom help text has to be explicitly written by the developer as well as kept up to date by the same developer.

Part of what got me into using click was that it did much of the user facing documentation and interaction "for free" based on the program's internal structure. This is another step in that direction.

click/termui.py
```diff
@@ -24,17 +24,18 @@ def hidden_prompt_func(prompt):
...
     if default is not None and show_default:
         prompt = '%s [%s]' % (prompt, default)
     return prompt + suffix
 
 
-def prompt(text, default=None, hide_input=False,
-           confirmation_prompt=False, type=None,
-           value_proc=None, prompt_suffix=': ',
-           show_default=True, err=False):
+def prompt(text, default=None, hide_input=False, confirmation_prompt=False,
```
untitaker: We already have a `type` parameter, which could be used to pass `click.Choice` in.

Ideally we would then be able to refactor `prompt_for_value` to have no choices-specific code, and just pass the type in.

untitaker:
I've just come across this myself, you convinced me.

click/termui.py
Rovanion: But we would still like to have the flag show_choices there right?

And what do you think it should default to& Right now it defaults to not show the available choices.

untitaker: Yes, I think it should be true by default.

Ideally it should also trunchate, but maybe for another time.

untitaker:
Hey @Rovanion any updates?

Rovanion:
@untitaker Haven't had any time to spend on the PR since last and since it took a couple of months for someone to get back to me in the first place I figured there was no rush.

I'll look into it when I get a free evening.

Rovanion:
@untitaker I've updated the code according to your suggestions and added documentation for show_choices to the docstring of `prompt`.

tests/test_termui.py
```diff
@@ -44,6 +44,16 @@ def cli():
...
 
 
+def test_choices_list_in_prompt(runner, monkeypatch):
+    @click.option('-g', type=click.Choice(['none', 'day', 'week', 'month']), prompt=True)
+    @click.command()
+    def cli():
+        pass
+
+    monkeypatch.setattr(click._termui_impl, 'isatty', lambda _: True)
+    assert '(none, day, week, month)' in runner.invoke(cli, [], input='none').output
```
untitaker: Could you assert the exact output here?

click/core.py
```diff
@@ -1642,7 +1642,8 @@ def get_default(self, ctx):
                 if param.name == self.name and param.default:
                     return param.flag_value
             return None
-        return Parameter.get_default(self, ctx)
+        else:
```
untitaker: Can we leave this change out? It's not really necessary and a matter of taste, also I'm a proponent of minimal, topical PRs

click/termui.py
```diff
@@ -24,23 +24,27 @@ def hidden_prompt_func(prompt):
     return getpass.getpass(prompt)
 
 
-def _build_prompt(text, suffix, show_default=False, default=None):
+def _build_prompt(text, suffix, show_default=False, default=None, show_choices=False, type=None):
```
untitaker: Why default value of `False` here?

click/core.py
```diff
@@ -1461,7 +1461,7 @@ def __init__(self, param_decls=None, show_default=False,
                  prompt=False, confirmation_prompt=False,
                  hide_input=False, is_flag=None, flag_value=None,
                  multiple=False, count=False, allow_from_autoenv=True,
-                 type=None, help=None, hidden=False, **attrs):
+                 type=None, help=None, hidden=False, show_choices=True, **attrs):
```
untitaker: What will happen with this param? Right now it gets discarded.

untitaker (changes requested):
Great stuff, but there's a few leftovers.

tests/test_termui.py
Rovanion: Why would that be clearer? The entire output is 'G (none, day, week, month): none\n' where the only part of this feature present in the output is '(none, day, week, month)'.

On the other hand, I only repeated the style I found in the progress bar test.

```python
    assert label in runner.invoke(cli, []).output
```

click/core.py
Rovanion: Done deal.

click/core.py
Rovanion: That doesn't even belong there. It should be in the Option class so that it can be toggled on an per option basis.

untitaker: Yeah but right now it appears to get ignored.

Rovanion:
Sorry if I wasn't clear enough in my earlier message: I have a fix in the
works. Just gotta test it a bit, commit and push.

2016-10-31 21:46 GMT+01:00 Markus Unterwaditzer <notifications@github.com>:

Rovanion:
With this latest version of the commit, hash 1f1e1a5, I've added an additional test to make sure that you can both turn the choices list on and off in the prompt. I could still rewrite the tests to assert against the exact string but then the tests would test for other features of click than that in this PR, they would stop passing even if code was changed which is unrelated to the choices list.

untitaker (approved):

untitaker (changes requested):
Cool stuff, please also add a changelog entry

Rovanion:
And now there's a changelog entry!

untitaker (approved):

PR #497, opened 2015-12-31
Added tests which show how show_default=False has no effect

Rovanion:
When I was writing my changes for the Prompt display of Choices, PR #496, I ended up discovering that passing show_default=False on an option has no effect. I've included two tests for this behaviour where the second one fails with the code currently on mitshuhiko/master.

### DIFF
--- a/tests/test_termui.py
+++ b/tests/test_termui.py
@@ -42,3 +42,23 @@ def test_progressbar_length_hint(runner, monkeypatch):
     monkeypatch.setattr(click._termui_impl, 'isatty', lambda _: True)
     result = runner.invoke(cli, [])
     assert result.exception is None
+
+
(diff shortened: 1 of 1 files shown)


Rovanion:
It should be noted that I do not have a fix for this. It is something better handled by someone more experienced with click.

untitaker:
I don't think this is a bug. `show_default`'s documentation for options:

>controls if the default value should be shown on the help page.

You're talking about the prompt. If you want to customize stuff like that, I'd rather suggest prompting manually.

I'm open to discussion about changing the behavior, but please open that discussion in a new issue. Also please understand that we can only change this in a major version.

Rovanion:
Yup, seems like I misunderstood the function of the flag.

PR #511, opened 2016-02-02
Bash complete hierarchicial cli

yxqd:
If the cli has a hierarchy of groups and commands, the bash complete did not work correctly. For example, the following cli

``` python
import click

@click.group()
def cli():
    return

@cli.group()
def g1():
    return

@cli.group()
def g2():
    return

@g1.command()
def subcmd1():
    return

@g1.command()
def subcmd2():
    return

@g2.command()
def subcmd3():
    return

@g2.command()
def subcmd4():
    return

cli()
```

when trying tab complete

``` bash
$ hcli g1 
```

(there is a space after "g1", and it is intended that the tab completion shows subcommands of "g1" group)
it will display choices "g1" and "g2" instead of the subcommands of "g1": "subcmd1" and "subcmd2".

This is an attempt to fix this issue.

### DIFF
--- a/click/_bashcomplete.py
+++ b/click/_bashcomplete.py
@@ -30,18 +30,21 @@ def get_completion_script(prog_name, complete_var):
 
 def resolve_ctx(cli, prog_name, args):
     ctx = cli.make_context(prog_name, args, resilient_parsing=True)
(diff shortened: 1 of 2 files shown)


untitaker:
Sorry, didn't see this and fixed it with https://github.com/pallets/click/pull/545

PR #522, opened 2016-02-18
Attempt to use an available locale

WhyNotHugo:
Click can detect if usable locales are available, but only exhibits this to users via a message.
I originally came across this via untitaker/vdirsyncer#335, where there was a possible sane environment, but click was still failing.

This patch attempts to use one of those good locales. It seems to fix click failing when both `LANG` and `LC_CTYPES` are unset, with no obvious drawbacks.

I haven't added any tests (yet) because I'd like to know if this patch is acceptable, or if, in my ignorance, I've overlooked something that makes this useless. It it's acceptable, I'll add some tests (and polish it a bit).

### DIFF
--- a/click/_unicodefun.py
+++ b/click/_unicodefun.py
@@ -70,11 +70,15 @@ def _verify_python3_env():
             rv = rv.decode('ascii', 'replace')
 
         for line in rv.splitlines():
-            locale = line.strip()
-            if locale.lower().endswith(('.utf-8', '.utf8')):
-                good_locales.add(locale)
-                if locale.lower() in ('c.utf8', 'c.utf-8'):
+            l = line.strip()
+            if l.lower().endswith(('.utf-8', '.utf8')):
+                good_locales.add(l)
+                if l.lower() in ('c.utf8', 'c.utf-8'):
                     has_c_utf8 = True
+
+        for l in good_locales:
+            locale.setlocale(locale.LC_ALL, l)
+            return
 
         extra += '\n\n'
         if not good_locales:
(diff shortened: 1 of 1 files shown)


untitaker:
Apparently this is already reported and fixed by https://github.com/mitsuhiko/click/issues/494

hobarrera:
Not really. Addressing the mentioned issue I had written code much like #494, but then had to add the code on this PR to fix a newly-uncovered error (I hadn't looked at master yet and was just patching in `site-packages`, heh).

#494 merely results in the proper error being displayed when a locale is not set, but does not attempt to set one (if available), which is what this PR does.

untitaker:
Sorry, posted on the wrong issue earlier.

I think this will leave things like `getfilesystemencoding` untouched, which is what Click actually cares about.

IIRC there is simply no way to set the locale from within the runtime as if it was set externally through the environment variable.

hobarrera:
I guess it's a matter of weighing best-effort vs correctness.

I honestly haven't gotten too much into click's internals, so I'm not sure what's preferred. If you feel this goes against it's spirit, then it's best to close this PR (I don't have a strong opinion either way, and don't feel in a position to pick one either).

untitaker:
>I guess it's a matter of weighing best-effort vs correctness.

In principle I opt for correctness here as encoding guessing comes with too many risks IMO (data corruption etc)

Even with best effort, `setlocale` has no effect on the locale-derived APIs Click actually uses.

also, locales affect language setting in a application, so this might actually change the language into something the user doesn't want?

PR #533, opened 2016-03-09
Non standalone exit

llllllllll:
I am using `Command.main` with `standalone=False` and found that `--help` would still call `ctx.exit()` which turned into a call to `sys.exit()`. This change makes `Context.exit` raise a custom click exception that is handled in `Command.main`. This allows us to check if we are in standalone mode and delegate to the proper behavior. In standalone mode we just pass the exit_code from the exception to `sys.exit` like before; however, when not in standalone mode we re-raise the exception when `exit_code != 0`, otherwise we just return None. This makes it possible for commands to early exit with `ctx.exit` without breaking non standalone usage.

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -7,7 +7,7 @@ from functools import update_wrapper
 from .types import convert_type, IntRange, BOOL
 from .utils import make_str, make_default_short_help, echo, get_os_args
 from .exceptions import ClickException, UsageError, BadParameter, Abort, \
-     MissingParameter
+     MissingParameter, Exit
 from .termui import prompt, confirm
 from .formatting import HelpFormatter, join_options
 from .parser import OptionParser, split_opt
@@ -480,7 +480,7 @@ class Context(object):
 
     def exit(self, code=0):
         """Exits the application with a given exit code."""
-        sys.exit(code)
+        raise Exit(code, self)
 
     def get_usage(self):
         """Helper method to get formatted usage string for the current
@@ -700,6 +700,12 @@ class BaseCommand(object):
             except (EOFError, KeyboardInterrupt):
                 echo(file=sys.stderr)
                 raise Abort()
+            except Exit as e:
+                if not standalone_mode:
+                    if e.exit_code:
+                        raise
+                    return None
+                sys.exit(e.exit_code)
             except ClickException as e:
                 if not standalone_mode:
                     raise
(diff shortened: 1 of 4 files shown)


llllllllll:
ping

llllllllll:
ping

tofumanchu:
ping

click/exceptions.py
```diff
@@ -52,7 +51,37 @@ def show(self, file=None):
...
+    status code.
+
+    :param code: the status code to exit with.
+    :param ctx: optionally the context that caused this error.  Click will
+                fill in the context automatically in some situations.
+    """
+    show_prefix = 'Exit: '
+
+    def __init__(self, code, ctx=None):
+        ContextAwareException.__init__(self, str(code), ctx=None)
```
Nzbuu: Did you mean to `ctx=ctx`, rather than `ctx=None` here?

click/exceptions.py
```diff
@@ -52,7 +51,37 @@ def show(self, file=None):
...
             echo(self.ctx.get_usage() + '\n', file=file, color=color)
-        echo('Error: %s' % self.format_message(), file=file, color=color)
+        echo(
+            '%s%s' % (self.show_prefix, self.format_message()),
+            file=file,
+            color=color,
+        )
+
+
+class Exit(ContextAwareException):
```
Nzbuu: I think this name could be improved. Maybe "ContextExit"?

click/core.py
```diff
@@ -700,6 +700,12 @@ def main(self, args=None, prog_name=None, complete_var=None,
             except (EOFError, KeyboardInterrupt):
                 echo(file=sys.stderr)
                 raise Abort()
+            except Exit as e:
```
Nzbuu: I would implement this so that the logic lives alongside "Abort". It's logically very similar.
This would mean it would be better if Exit didn't derive from ClickException

click/core.py
```diff
@@ -700,6 +700,12 @@ def main(self, args=None, prog_name=None, complete_var=None,
...
+                if not standalone_mode:
```
Nzbuu: I think this logic should be

```
if standalone_mode:
    sys.exit(e.exit_code)
else:
    return e.exit_code
```

Nzbuu:
This PR addresses https://github.com/pallets/click/issues/667

click/core.py
```diff
@@ -700,6 +700,12 @@ def main(self, args=None, prog_name=None, complete_var=None,
...
+                    if e.exit_code:
+                        raise
```
Nzbuu: Why re-raise the exception? The exit code should just be the return value from `invoke`, whatever it means

Issue #530, opened 2016-03-04
click bashcomplete falls short of one argument, shows wrong options


untitaker:
Using the following application: https://gist.github.com/untitaker/b05a2c321b6303974152
and this line to activate bash completion:

    eval "$(_LOL_COMPLETE=source lol)"

We find a few bugs:

* Typing `lol<space>`, and hitting tab N times, it completes to `lol sub sub sub`, where `sub` is repeated N times.
* Typing `lol sub --` and hitting tab will suggest options that are registered on the group, and are invalid in this location.

This patch fixes both problems:

```diff
diff --git a/click/_bashcomplete.py b/click/_bashcomplete.py
index 43feffb..fd62137 100644
--- a/click/_bashcomplete.py
+++ b/click/_bashcomplete.py
@@ -29,6 +29,7 @@ def get_completion_script(prog_name, complete_var):
 
 
 def resolve_ctx(cli, prog_name, args):
+    args = [''] + args
     ctx = cli.make_context(prog_name, args, resilient_parsing=True)
     while ctx.args and isinstance(ctx.command, MultiCommand):
         cmd = ctx.command.get_command(ctx, ctx.args[0])
```

and I have no idea why.

katakumpo:
I can confirm this works nicely :)

kuchi:
Appears this bug has also been documented in  #471.

untitaker:
Closing as duplicate.

untitaker:
6.5 is out.

Issue #471, opened 2015-11-27
Bash completion for MultiCommands returns subcommands twice until it stops.


rubot:
Python 2.7.10
broken in click >= 6
ok in click (5.1)

```
manage hello hello world world
```

danielsnider:
+1 
Still a problem in click v6.2

bb4242:
I've noticed this problem as well.  Some `git bisect` experiments indicate that the problem was introduced in commit 0a2919f34fcbc635d8530b4c5b60bf119b2bcedb on Nov 24th.

kuchi:
Still observing this issue with click v 6.3 and the current development branch  57c6f09. (Using python 2.7 on OS X in ZSH using bashcomplete (bashcompinit).

instead of getting tab completion providing:

    cmd sub_cmd1 cmd_1_options ...

Autocomplete endlessly completes at the sub command level and never moves into the sub_command itself for auto completion:

    cmd sub_cmd1 sub_cmd1 sub_cmd1 ....

Downgrading to click v 5.1 fixes the issue.

    pip uninstall click
    pip install -Iv click==5.1

untitaker:
@kuchi Thank you very much for the bisect, the fix is now obvious to me.

untitaker:
(edit because wrong user mentioned)

@bb4242 Thank you very much for the bisect, the fix is now obvious to me.

untitaker:
6.5 is out.

PR #543, opened 2016-03-23
fix for bug #530 / #471

west-rynes:
This is a fix for #530 / #471.

Fix was posted in #530 by @untitaker, I just created a pull request but @untitaker should get the credit for the fix!

I can confirm that it corrects the problem on my side. All tests still pass for me. I would write a py.test test but I'm not sure how to do a test for bash autocomplete Suggestions for how to write the test are welcome so we could avoid the problem down the road again. I just don't know how.

### DIFF
--- a/click/_bashcomplete.py
+++ b/click/_bashcomplete.py
@@ -29,6 +29,7 @@ def get_completion_script(prog_name, complete_var):
 
 
 def resolve_ctx(cli, prog_name, args):
+    args = [''] + args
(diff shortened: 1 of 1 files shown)


untitaker:
This is not a proper fix, sorry. It's more of a hack.

Proper fix will follow in a minute, I've figured it out.

kuchi:
Thanks @untitaker! You are awesome. I couldn't figure out what was going on. Much appreciated.

PR #546, opened 2016-03-30
Clarify behavior of resolve_path

buckket:
'resolve_path' suggests, that it will also call os.path.expanduser(), while in fact it does not.
Avoid confusion by stating this in the docstring of :class:`Path`.

### DIFF
--- a/click/types.py
+++ b/click/types.py
@@ -358,7 +358,9 @@ class Path(ParamType):
     :param readable: if true, a readable check is performed.
     :param resolve_path: if this is true, then the path is fully resolved
                          before the value is passed onwards.  This means
-                         that it's absolute and symlinks are resolved.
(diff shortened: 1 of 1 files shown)


click/types.py
```diff
@@ -358,7 +358,8 @@ class Path(ParamType):
     :param readable: if true, a readable check is performed.
     :param resolve_path: if this is true, then the path is fully resolved
                          before the value is passed onwards.  This means
-                         that it's absolute and symlinks are resolved.
+                         that it's absolute and symlinks are resolved.  It
+                         will not expand a tilde-prefix.
```
untitaker: Could we clarify that this is the job of the shell anyway?

buckket:
Clarified the clarification.

untitaker:
Excellent, thanks!

PR #548, opened 2016-04-01
Support options with const value if flag exists

mbrancato:
The flag_value option passes to the const value if is_flag=False

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -1593,7 +1593,10 @@ class Option(Parameter):
                                   **kwargs)
         else:
             kwargs['action'] = action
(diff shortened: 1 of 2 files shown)


dsully:
Could you add some tests & docs to this PR? Thanks.

davidism:
Thanks for working on this. We're working on a hopefully more general solution to options that can omit their value in #1618. This PR changes the meaning of `flag_value` a bit, since currently setting it implies `is_flag`. Instead, we're using an explicit `is_flag=False` (default is `None`) to indicate that providing the value is optional.

PR #557, opened 2016-04-13
issue #393: Usage error should show there is a --help option

rsiemens:
Went ahead and took a crack at this one. If you have suggestions for change please let me know.

PR #560, opened 2016-04-15
Add option for #557

untitaker:


Nudies:
nice :+1:

untitaker:
Armin wants to replace this with #561

PR #570, opened 2016-05-02
add support for fish completion

glehmann:


glehmann:
No test or doc yet — this is just to have some feedback before going further :-)

untitaker:
At a first glance the implementation seems to be solid. However, eventually we'd like to move tab-completion of any kind into a third-party package (or refactor it in a way that makes it easy to replace by third-party packages). The general consensus is that there are features in Click that don't belong into it.

untitaker:
See also http://click.pocoo.org/5/contrib/

glehmann:
click-contrib looks interesting for sure.

I'm not sure about what I should do with the proposed feature though.
Should I wait until the shell completion code has been moved to click-contrib?
Should I continue in click main repository, so the code can be moved to click-contrib with bash completion?

glehmann:
Just for the record:

The dynamic fish completion method comes from https://github.com/fish-shell/fish-shell/issues/1217#issuecomment-31441757

untitaker:
I think it's useful to just keep this PR incl. code on ice until we manage to clean Click up.

glehmann:
Ok. When you're ready, just let me know if you need some help with the fish completer!

glehmann:
For people in the hurry to get fish completion (I may be alone, but…), I've created a project which does just that:

  https://github.com/glehmann/click-fish

glehmann:
For the completion lovers: click-fish has evolved in click-completion and gained the support for zsh and powershell:

[https://github.com/glehmann/click-completion](https://github.com/glehmann/click-completion)

Don't fear about the 0.1.0 version — we have been using it intensively in a private project for several months, and it works very well!

untitaker:
That looks great, though I wonder if we could do this without monkeypatching
click (would require modifications in Click itself, open to that).

Do you want to add this repo to https://github.com/click-contrib ?

Thanks for your work!

glehmann:
It could be done without monkey patching, and without much changes in click:

* add a `complete` method to `ParamType` and `Choice`
* add a `get_command_short_help` method to `MultiCommand` — this allows the user to implement a cache in its own `MultiCommand` class
* add a way to set a user defined function instead of `click.core._bashcomplete` to implement the completion

The two first points are implemented there: https://github.com/glehmann/click-completion/blob/9f230bb3126b19fe896612332a1a776e4a6efe4d/click_completion/__init__.py#L216
Any preference on how to implement the last point?

glehmann:
And I'd be pleased to move click-completion to click-contrib!

untitaker:
I invited you to the organization, you should be able to move your repo.

For your proposed changes, please open new PRs.

PR #572, opened 2016-05-02
Added check for ctx.obj to make this work with setup tools entry points

ghost:
Hi Markus,

Here is the tiny PR for the doc update to make the example in the docs work with entry points. I felt that it went better here than in setup tools integration.

Best
-Kaushik

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -87,6 +87,9 @@ script like this:
     @click.option('--debug/--no-debug', default=False)
     @click.pass_context
     def cli(ctx, debug):
+        if ctx.obj is None:  # This is needed when we get here via a setup tools entry point
+            ctx.obj = {}
+
         ctx.obj['DEBUG'] = debug
 
     @cli.command()


docs/commands.rst
```diff
@@ -87,6 +87,9 @@ script like this:
     @click.option('--debug/--no-debug', default=False)
     @click.pass_context
     def cli(ctx, debug):
+        if ctx.obj is None:  # This is needed when we get here via a setup tools entry point
```
untitaker: On a newline please

untitaker:
This also makes the `obj={}` in `if __name__ == '__main__'` redundant.

untitaker:
@kghosesbg Hello, any update on this? I've left some inline comments that need addressing. Feel free to shoot any questions here or in IRC!

maethor:
Hi,

I'm not sure this is a generic solution. I have:

```
if __name__ == "__main__":                                                                                         
     cli(auto_envvar_prefix="SHINKEN", obj={})
```

And I don't now how to manage `auto_envvar_prefix` with setuptools.

maethor:
I ended up doing an intermediary `main()` function:

```
def main():
    cli(auto_envvar_prefix="SHINKEN", obj={})


if __name__ == "__main__":
    main()
```

And in setup.py:

```
    entry_points='''
        [console_scripts]
        shinkendo=shinkendo:main
    ''',
```

I think this is the cleanest workaround.

Issue #580, opened 2016-05-11
Editor.edit_file doesn't block if editor is 'gvim'


zblach:
Feel free to close if this is a design feature, but gvim by itself doesn't block execution when invoked via shell process. The captured exit code (_termui_impl.py:405, in click 6.4) only tracks whether the editor failed to be invoked, rather than the exit code of the editor itself as it would be seen with 'vim'

If gvim is invoked with the -f flag, then it behaves as expected.

blueyed:
AFAIK you are required to use `-f` when using `gvim`, and there is not much that click can do about it.
So this issue should get closed.

fdavis:
Thanks @blueyed 
looks like you'll need to set `-f` to prevent it from forking just like when running `git commit`
https://stackoverflow.com/posts/3764528/revisions

PR #585, opened 2016-05-18
Added support for gvim blocking

zblach:
Currently, if your EDITOR or VISUAL environmental values are set to 'gvim', subprocess.Popen will not block, and c.wait() immediately returns. This is different from the behavior with other editors, where this blocks until the file is closed. This pull request adds the '-f' flag to gvim invocations, which runs the editor in the foreground.

This is a fix for https://github.com/pallets/click/issues/580

### DIFF
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -394,15 +394,15 @@ class Editor(object):
     def edit_file(self, filename):
         import subprocess
         editor = self.get_editor()
+        if os.path.basename(editor) == 'gvim':
+            editor += " -f"
+
         if self.env:
             environ = os.environ.copy()
             environ.update(self.env)
         else:
             environ = None
         try:
-            if os.path.basename(editor) == 'gvim':
-                editor += " -f"
-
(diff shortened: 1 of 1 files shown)


amcgregor:
This sounds like a bug in your shell environment configuration, not a bug in a consumer of those variables.  Many other processes (such as git when committing without an explicit message on the command-line) will also utilize those variables, and most will expect EDITOR or VISUAL to block until done editing.  Instead of adding that flag to all tools that invoke EDITOR or VISUAL, why not just update your variables to include that switch?

As a note, recent Git, I believe, waits for the file to be opened, then closed.  This would seem to be a more universal solution on platforms where such functionality is available, and would correct this for not just `gvim`, but `mvim`, `vimr`, `tm`, and a wide range of other GUI programmers' editors' CLI invokers.

untitaker:
Yes, this PR is not a solution as-is.

>As a note, recent Git, I believe, waits for the file to be opened, then closed.

Can't reproduce that on my machine but would be interesting to explore for POSIX at least.

PR #610, opened 2016-07-14
Quick fix to get_winterm_size() returning (0, 0)

stxlvt:
The `get_winterm_size()` function will return `(0, 0)` when invoked
from inside a subprocess. This is not very useful and results in a very
cramped and ugly output, so instead provide a sensible default.


dsully:
#997 fixes.

PR #626, opened 2016-08-03
Handle SIGPIPE in click.echo().

### LINKED ISSUE #625: Handling of SIGPIPE
The simple example from the front page of click:

```
import click

@click.command()
@click.option('--count', default=1, help='Number of greetings.')
@click.option('--name', prompt='Your name',
              help='The person to greet.')
def hello(count, name):
    """Simple program that greets NAME for a total of COUNT times."""
    for x in range(count):
        click.echo('Hello %s!' % name)

if __name__ == '__main__':
    hello()
```

when piped to a command that closes the pipe prematurely, produces an ugly python stack trace:

```
$ python ./test-click.py --name ob --count 1000 | head -1
Hello ob!
  File "./test-click.py", line 13, in <module>
    hello()
  File "/Users/obonilla/o/click/click/core.py", line 716, in __call__
  File "/Users/obonilla/o/click/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/Users/obonilla/o/click/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/obonilla/o/click/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "./test-click.py", line 10, in hello
    click.echo('Hello %s!' % name)
  File "/Users/obonilla/o/click/click/utils.py", line 259, in echo
    file.write(message)
  File "/Users/obonilla/o/click/click/_compat.py", line 63, in write
    return io.TextIOWrapper.write(self, x)
IOError: [Errno 32] Broken pipe
```

Python maps the handling of SIGPIPE to an exception that I believe click.echo is not catching.
It seems it should silently exit in this particular case.

ob:
When the output of a click program is sent to another program via a
pipe, and the pipe is closed prematurely, click.echo() was printing an
exception backtrace instead of silently exiting like most command line
utilities.

Fixes: https://github.com/pallets/click/issues/625

### DIFF
--- a/click/utils.py
+++ b/click/utils.py
@@ -241,7 +241,7 @@ def echo(message=None, file=None, nl=True, err=False, color=None):
                 file.flush()
                 binary_file.write(message)
                 binary_file.flush()
-            except IOError, e:
+            except IOError as e:
                 if e.errno != errno.EPIPE:
                     raise
             return
@@ -264,7 +264,7 @@ def echo(message=None, file=None, nl=True, err=False, color=None):
     if message:
         try:
             file.write(message)
-        except IOError, e:
+        except IOError as e:
             if e.errno != errno.EPIPE:
                 raise
     file.flush()
--- a/click/utils.py
+++ b/click/utils.py
@@ -1,4 +1,3 @@
-import errno
 import os
 import sys
 
@@ -237,13 +236,9 @@ def echo(message=None, file=None, nl=True, err=False, color=None):
     if message and not PY2 and is_bytes(message):
         binary_file = _find_binary_writer(file)
         if binary_file is not None:
-            try:
-                file.flush()
-                binary_file.write(message)
-                binary_file.flush()
-            except IOError as e:
-                if e.errno != errno.EPIPE:
-                    raise
+            file.flush()
+            binary_file.write(message)
+            binary_file.flush()
             return
 
     # ANSI-style support.  If there is no message or we are dealing with
@@ -262,11 +257,7 @@ def echo(message=None, file=None, nl=True, err=False, color=None):
                 message = strip_ansi(message)
 
     if message:
-        try:
-            file.write(message)
-        except IOError as e:
-            if e.errno != errno.EPIPE:
-                raise
+        file.write(message)
     file.flush()
 
 
--- a/click/utils.py
+++ b/click/utils.py
@@ -1,6 +1,6 @@
+import errno
 import os
 import sys
-import errno
 
 from .globals import resolve_color_default
 
--- a/click/utils.py
+++ b/click/utils.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import errno
 
 from .globals import resolve_color_default
 
@@ -236,9 +237,13 @@ def echo(message=None, file=None, nl=True, err=False, color=None):
     if message and not PY2 and is_bytes(message):
         binary_file = _find_binary_writer(file)
         if binary_file is not None:
-            file.flush()
-            binary_file.write(message)
-            binary_file.flush()
+            try:
+                file.flush()
+                binary_file.write(message)
+                binary_file.flush()
+            except IOError, e:
+                if e.errno != errno.EPIPE:
+                    raise
             return
 
     # ANSI-style support.  If there is no message or we are dealing with
@@ -257,7 +262,11 @@ def echo(message=None, file=None, nl=True, err=False, color=None):
                 message = strip_ansi(message)
 
     if message:
-        file.write(message)
+        try:
+            file.write(message)
+        except IOError, e:
+            if e.errno != errno.EPIPE:
+                raise
     file.flush()
 
 
--- a/click/core.py
+++ b/click/core.py
@@ -694,13 +694,7 @@ class BaseCommand(object):
         try:
             try:
                 with self.make_context(prog_name, args, **extra) as ctx:
-                    try:
-                        rv = self.invoke(ctx)
-                    except IOError as e:
-                        if e.errno == errno.EPIPE:
-                            sys.exit(errno.EPIPE)
-                        else:
-                            raise
+                    rv = self.invoke(ctx)
                     if not standalone_mode:
                         return rv
                     ctx.exit()
@@ -712,6 +706,11 @@ class BaseCommand(object):
                     raise
                 e.show()
                 sys.exit(e.exit_code)
+            except IOError as e:
+                if e.errno == errno.EPIPE:
+                    sys.exit(1)
+                else:
+                    raise
         except Abort:
             if not standalone_mode:
                 raise
--- a/click/core.py
+++ b/click/core.py
@@ -1,3 +1,4 @@
+import errno
 import os
 import sys
 from contextlib import contextmanager
@@ -693,7 +694,13 @@ class BaseCommand(object):
         try:
             try:
                 with self.make_context(prog_name, args, **extra) as ctx:
-                    rv = self.invoke(ctx)
+                    try:
+                        rv = self.invoke(ctx)
+                    except IOError as e:
+                        if e.errno == errno.EPIPE:
+                            sys.exit(errno.EPIPE)
+                        else:
+                            raise
                     if not standalone_mode:
                         return rv
                     ctx.exit()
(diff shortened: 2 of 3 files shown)


ob:
Any interest in this fix?

click/utils.py
```diff
@@ -1,5 +1,6 @@
 import os
 import sys
+import errno
```
ThiefMaster: e < o, should be the first import

ThiefMaster:
LGTM, but I'd rather have someone who did some actual click development look over it. General suggestion: Squash those three commits into a single one. The last two just fix issues in the initial one after all, no need for that to be part of the project's permanent history ;)

ob:
I thought commit squashing was up to whoever merged the branch... 

https://github.com/blog/2141-squash-your-commits

I can squash the commits if that is preferable.

ThiefMaster:
True, we can do it too. Forgot about that ;)

untitaker:
This doesn't silently exit, it swallows the exception and lets the application continue. IMO SIGPIPE should not be catched on this level but much further up (somewhere near subcommand dispatch) and then be swallowed.

ob:
Oops... I'm missing a `sys.exit()` there. I couldn't find a place to add a test so I ended up screwing up the fix ;-/

I'll take a look at subcommand dispatch and see if I can figure out where you mean.

untitaker:
I think here: https://github.com/pallets/click/blob/master/click/core.py#L700

@mitsuhiko Do you agree this is the proper way to handle `SIGPIPE`, or should the `signal` module be used?

ob:
Pushed something better...

ob:
Ah, and now I read your comment and I see your suggestion is probably better...

ob:
Something like this? 

https://gist.github.com/ob/2fb860d5aea6a75c2ef3863c7b4b255d

untitaker:
Yeah, but is `sys.exit(EPIPE)` the right thing to do?

ob:
I was just googling that... it seems exiting -1 might be better... I tried a few system utilities and they seem to exit 0. I was about to go read POSIX :)

untitaker:
I think we're supposed to raise Abort.

ob:
The problem with raising Abort is that it prints "Aborted()" to stderr. This is undesirable for things like piping the output to `head` for example.

untitaker:
Fair enough, then `sys.exit(1)` is probably the way to go.

ob:
Okay, I pushed that. BTW, I'm so counting on you squashing all this mess ;)

ob:
Is this acceptable now?

untitaker:
I'll probably merge in a few days. It looks fine to me, but code review at
conferences is a bad idea.

sirosen:
Although I agree that `click.echo` should capture EPIPE by default, could we add an option somewhere to have it raise `Abort` or reraise the `IOError` instead?

I have a bit of code that's doing
```python
try:
    click.echo(message, nl=newline, err=write_to_stderr)
except IOError as err:
    if err.errno is errno.EPIPE:
        pass
    else:
        raise
```

So that I can maintain my promises about exit codes.

This would be, to me, an acceptable alternative:

```python
try:
    click.echo(message, nl=newline, err=write_to_stderr, handle_epipe=False)
except IOError:
    # a reraised IOError from inside of click.echo! Must be EPIPE
    pass
```

What I object to is being asked to do this:

```python
try:
    click.echo(message, nl=newline, err=write_to_stderr)
except SystemExit:
    # I guess this was an EPIPE?
    pass
```

As a caller into `click.echo`, having it decide to raise `SystemExit` seems wrong.
Consider:
```bash
set -o pipefail
mycommand | head -n 20 | grep 'importantstring'
```

If supporting a pipeline like that is a priority for `mycommand`, it shouldn't be deprived of the features of `click.echo`.

untitaker:
Your code doesn't have to change since SystemExit is raised somewhere completely different than from within echo. The PR title is now wrong.

On 17 September 2016 19:39:29 CEST, Stephen Rosen <notifications@github.com> wrote:
>Although I agree that `click.echo` should capture EPIPE by default,
>could we add an option somewhere to have it raise `Abort` or reraise
>the `IOError` instead?
>
>I have a bit of code that's doing
>```python
>try:
>    click.echo(message, nl=newline, err=write_to_stderr)
>except IOError as err:
>    if err.errno is errno.EPIPE:
>        pass
>    else:
>        raise
>```
>
>So that I can maintain my promises about exit codes.
>
>This would be, to me, an acceptable alternative:
>
>```python
>try:
>click.echo(message, nl=newline, err=write_to_stderr,
>handle_epipe=False)
>except IOError:
>    # a reraised IOError from inside of click.echo! Must be EPIPE
>    pass
>```
>
>What I object to is being asked to do this:
>
>```python
>try:
>    click.echo(message, nl=newline, err=write_to_stderr)
>except SystemExit:
>    # I guess this was an EPIPE?
>    pass
>```
>
>As a caller into `click.echo`, having it decide to raise `SystemExit`
>seems wrong.
>Consider:
>```bash
>set -o pipefail
>mycommand | head -n 20 | grep 'importantstring'
>```
>
>If supporting a pipeline like that is a priority for `mycommand`, it
>shouldn't be deprived of the features of `click.echo`.
>
>-- 
>You are receiving this because you commented.
>https://github.com/pallets/click/pull/626#issuecomment-247792481

-- 

untitaker:
Thanks! Please add yourself to AUTHORS if you want!

sirosen:
@untitaker Thanks for pointing out that this handling is, in fact, safe for callers of `click.echo`.

Please correct me if I'm wrong, but as I understand it this catches an EPIPE that makes its way up to `click.Command.main()`, so it handles `click.echo()`, `print()`, and anything else that may encounter a broken pipe inside of the running command.
If you want re-title the PR, I'd suggest "Handle EPIPE by exiting silently with status 1".

Anyway, as always, thanks for this excellent CLI toolkit.

untitaker:
Sorry, yes, I forgot to rename the PR and a same-named commit is now in master :(

mitsuhiko:
@untitaker i just discovered this merge and I'm not particularly happy with this because this captures *all* sigpipes, so even on sockets. Also not sure if click should handle this by default like this. At least maybe make it optional?

ThiefMaster:
Any way of checking if SIGPIPE was caused by stdin/stdout/stderr? I don't think getting a traceback in these cases is something you'd want

mitsuhiko:
No, not really. You can only catch down individual writes (like in echo) and rethrow it.

untitaker:
In which situation would you have a SIGPIPE from e.g. a socket?

I'd vote for optional since I think this is still reasonable default behavior
(assuming more fine-grained handling isn't possible).

mitsuhiko:
You get a sigpipe when the socket is not connected and you write to it.

untitaker:
Ah damn. Feel free to revert then.

In general I would say that Click could use some sort of exception handling
like it's known from webframeworks, but that's a different topic.

mitsuhiko:
@untitaker what about us wrapping both text and binary streams and augment the exception with `from_stdio = True`? Then we can capture that only by default.

untitaker:
I thought about that too but that's monkeypatching. Having it in click.echo would be less surprising I think.

On 21 September 2017 09:14:17 GMT+02:00, Armin Ronacher <notifications@github.com> wrote:
>@untitaker what about us wrapping both text and binary streams and
>augment the exception with `from_stdio = True`? Then we can capture
>that only by default.
>
>-- 
>https://github.com/pallets/click/pull/626#issuecomment-331072493

-- 

mitsuhiko:
We could also raise a subclass of `IOError`.

untitaker:
Whatever you think is best, however I dont think people will apprechiate us patching stdio even if it behaves almost the same.

On 21 September 2017 10:35:18 GMT+02:00, Armin Ronacher <notifications@github.com> wrote:
>We could also raise a subclass of `IOError`.
>
>-- 
>https://github.com/pallets/click/pull/626#issuecomment-331090373

-- 

mitsuhiko:
@untitaker the proposal is not to patch stdio but to change our wrappers. We already use wrappers for the unicode support and there are already accessors for getting binary streams. We can trivially further customize those streams.

untitaker:
Sure, I'm totally fine with echo rewrapping this exception.

On 23 September 2017 11:37:20 GMT+02:00, Armin Ronacher <notifications@github.com> wrote:
>@untitaker the proposal is not to patch stdio but to change our
>wrappers. We already use wrappers for the unicode support and there are
>already accessors for getting binary streams. We can trivially further
>customize those streams.
>
>-- 
>https://github.com/pallets/click/pull/626#issuecomment-331622961

-- 

PR #629, opened 2016-08-10
Add ability to truncate help text in docstrings.

Robpol86:
I use docstrings in all of my functions to document function arguments
so it makes me sad that I couldn't exclude :param: and other tags from
my command help texts.

Here I am introducing the ability for Click to truncate help text in
docstrings using the \f form feed escaped character. I chose it
randomly.

I've been accomplishing this in my project using this decorator:

``` python
def click_truncate(click_func, **kwargs):
    """Truncate function docstrings from Click's perspective on the form-feed character.

    :param function click_func: Either click.command or click.group.
    :param dict kwargs: Extra keyword arguments to pass to click.option().

    :return: The wrapped function.
    :rtype: function
    """
    def wrapper(func):
        original = func.__doc__
        if original:
            func.__doc__ = original.split('\f', 1)[0].rstrip()
        wrapped = click_func(**kwargs)(func)
        if original:
            wrapped.__doc__ = func.__doc__ = original
        return wrapped
    return wrapper
```

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -752,6 +752,8 @@ class Command(BaseCommand):
         #: should show up in the help page and execute.  Eager parameters
         #: will automatically be handled before non eager ones.
         self.params = params or []
+        if help and '\f' in help:
+            help = help.split('\f', 1)[0]
         self.help = help
         self.epilog = epilog
         self.options_metavar = options_metavar
--- a/docs/documentation.rst
+++ b/docs/documentation.rst
@@ -77,6 +77,40 @@ And what it looks like:
     invoke(cli, args=['--help'])
 
 
+Truncating Help Texts
+---------------------
+
+Click gets command help text from function docstrings.  However if you
+already use docstrings to document function arguments you may not want
+to see :param: and :return: lines in your help text.
+
+You can use the ``\f`` escape marker to have Click truncate the help text
+after the marker.
+
+Example:
+
+.. click:example::
+
+    @click.command()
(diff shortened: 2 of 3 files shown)


Robpol86:
To anyone reading this if this PR doesn't get accepted, a better implementation is:

```python
class ClickGroup(click.Group):
    """Truncate docstrings at form-feed character for click.group()."""

    def __init__(self, *args, **kwargs):
        """Constructor."""
        if 'help' in kwargs and kwargs['help'] and '\f' in kwargs['help']:
            kwargs['help'] = kwargs['help'].split('\f', 1)[0]
        super(ClickGroup, self).__init__(*args, **kwargs)

@click.group(cls=ClickGroup)
@click.option('-v', '--verbose', help='Debug logging.', expose_value=False, is_flag=True)
@click.pass_context
def cli(ctx):
    """Some command.
    
    This command does stuff.
    \f

    :param click.core.Context ctx: Click context.
    """
```

click/core.py
```diff
@@ -752,6 +752,8 @@ def __init__(self, name, context_settings=None, callback=None,
         #: should show up in the help page and execute.  Eager parameters
         #: will automatically be handled before non eager ones.
         self.params = params or []
+        if help and '\f' in help:
```
dsully: Can you add a comment here, describing what this is for and why? Will be good to merge after that.

Thanks for the PR!

dsully (changes requested):

Issue #636, opened 2016-08-19
Progress bar doesn't work with codecs.open when using encoding


mina-asham:
Progress bar crashes while trying to get a length hint of a file that is opened using codecs.open function.
Python version: 2.7.11 (64bit) on Windows
Output:
```
raceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Program Files\Python27\lib\site-packages\click\termui.py", line 306, in progressbar
    width=width, color=color)
  File "C:\Program Files\Python27\lib\site-packages\click\_termui_impl.py", line 73, in __init__
    length = _length_hint(iterable)
  File "C:\Program Files\Python27\lib\site-packages\click\_termui_impl.py", line 33, in _length_hint
    return len(obj)
  File "C:\Program Files\Python27\lib\codecs.py", line 728, in __getattr__
    return getattr(self.stream, name)
AttributeError: 'file' object has no attribute '__len__'
```
To reproduce:
```python
import codecs
from click import progressbar

file_path = 'my_file.txt'

with progressbar(codecs.open(file_path, encoding='utf-8')) as progress:
    pass
```
The same code works as expected when not using `encoding` parameter, mainly because codecs.open return the basic `open()` method result if no encoding is specified.

The `_length_hint` function catches the `TypeError` exception but this one produces an `AttributeError` exception, should be modified to include the `AttributeError` type as well.

untitaker:
Fixed with #637

PR #637, opened 2016-08-19
Bugfix: Make click.progressbar work with codecs.open files

mina-asham:
- codecs.open returns a StreamReaderWriter when used with encoding, this object doesn't have a `__len__` attribute, so calling the `len` function on it result in an `AttributeError` instead of a `TypeError`
- this will catch the AttributeError as well as the TypeError instead of only catching the later
- Fixes: https://github.com/pallets/click/issues/636

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,5 +17,11 @@ install:
 
 script: make test
 
+branches:
+  only:
+    - master
+    - auto
+    - /^.*-maintenance$/
+
 notifications:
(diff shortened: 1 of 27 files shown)


untitaker:
Oops, sorry, I thought I could actually cherry-pick your commit with this new GitHub feature. I'll merge manually.

untitaker:
Manually applied, thanks!

PR #639, opened 2016-08-19
Fix #628: Bash completion for nested commands broken

### LINKED ISSUE #628: Bash completion for arguments not working
Option arguments don't seem to show up after hitting tabs after a dash, while the documentation says option arguments get completed. Here is a stripped down example:

`$ cd test`
`$ virtualenv .test`
`$ . ./.test/bin/activate`
`$ cat tester.py`

```python
import click

@click.command()
@click.option('--hi', help='hi')
def cli(hi):
    click.echo(hi)
```

`$ cat setup.py`

```python
from setuptools import setup

setup(
    name = 'tester',
    version = '0.1',
    py_modules = ['tester'],
    install_requires = [
        'Click',
    ],
    entry_points = """
    [console_scripts]
    tester=tester:cli
    """,
)
```

`$ pip install --editable .`
`$ _TESTER_COMPLETE=source tester > tester-complete.sh`
`$ . tester-complete.sh`
`$ tester --<TAB><TAB>` <-- doesn't show the `hi` option.

`$ python --version`
`Python 3.4.3`

`$ pip list | grep click`
`click (6.6)`

`$ bash --version`
`GNU bash, version 4.3.11(1)-release (x86_64-pc-linux-gnu)`

`$ cat tester-complete.sh`

```bash
_tester_completion() {
    COMPREPLY=( $( env COMP_WORDS="${COMP_WORDS[*]}" \
                   COMP_CWORD=$COMP_CWORD \
                   _TESTER_COMPLETE=complete $1 ) )
    return 0
}

complete -F _tester_completion -o default tester;
```

The command otherwise works fine.
`$ tester --help`

```
Usage: tester [OPTIONS]

Options:
  --hi TEXT  hi
  --help     Show this message and exit.
```

`$ tester --hi there`
`there`

pgkelley4:
When concatenating protected_args and args, protected_args
should be prepended. Otherwise the arguments are out
of order and an incorrect subcommand is choosen.

### DIFF
--- a/tests/test_bashcomplete.py
+++ b/tests/test_bashcomplete.py
@@ -3,7 +3,18 @@
 import click
 from click._bashcomplete import get_choices
 
-def test_basic():
+
+def test_single_command():
+    @click.command()
+    @click.option('--local-opt')
+    def cli(local_opt):
+        pass
+
+    assert list(get_choices(cli, 'lol', [], '-')) == ['--local-opt']
+    assert list(get_choices(cli, 'lol', [], '')) == []
+
+
+def test_small_chain():
     @click.group()
     @click.option('--global-opt')
     def cli(global_opt):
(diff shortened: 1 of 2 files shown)


pgkelley4:
This is the same fix as xificurC did here: https://github.com/pallets/click/pull/630, but with added unit tests to demonstrate the issue.

tests/test_bashcomplete.py
```diff
@@ -3,7 +3,18 @@
...
-def test_basic():
+
+def test_single_command():
+    @click.command()
+    @click.option('--local-opt')
+    def cli(local_opt):
+        pass
+
+    assert list(get_choices(cli, 'lol', [], '-')) == ['--local-opt']
+    assert list(get_choices(cli, 'lol', [], '')) == []
```
untitaker: This test doesn't fail if I revert the fix though.

pgkelley4: The other one: 'def test_long_chain():' does. I just added the test_single_command for completeness. See my comment here: https://github.com/pallets/click/issues/628#issuecomment-241047886

PR #643, opened 2016-08-26
add support for custom bash completers and choices

ericfrederich:
After writing this I noticed #428  which mentioned #241 but thought I'd submit this pull request anyway.

This one uses `completer` and `choices` as keyword arguments using the same terminology as [argcomplete](https://github.com/kislyuk/argcomplete)

Dunno how hacky this is, but I got it to work for me.  I basically pass some more context (the last parsed argument) to `get_choices` which it uses to look up either `completer` or `choices` attributes on the option.

### DIFF
--- a/click/_bashcomplete.py
+++ b/click/_bashcomplete.py
@@ -40,18 +40,30 @@ def resolve_ctx(cli, prog_name, args):
 
 
 def get_choices(cli, prog_name, args, incomplete):
+    # resolve_ctx will change args in place
+    # need to grab the last completed one here and now
+    last_complete = [a for a in (args if args else ['']) if a != '='][-1]
+    if incomplete == '=':
+        incomplete = ''
+
     ctx = resolve_ctx(cli, prog_name, args)
     if ctx is None:
         return
 
     choices = []
+
+    if last_complete:
+        for param in ctx.command.params:
+            if last_complete in param.opts:
+                choices.extend(param.get_completion())
+                break
     if incomplete and not incomplete[:1].isalnum():
         for param in ctx.command.params:
             if not isinstance(param, Option):
                 continue
             choices.extend(param.opts)
             choices.extend(param.secondary_opts)
-    elif isinstance(ctx.command, MultiCommand):
+    if isinstance(ctx.command, MultiCommand):
         choices.extend(ctx.command.list_commands(ctx))
 
     for item in choices:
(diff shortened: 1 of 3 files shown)


ericfrederich:
updated now with passing tests; added some tests for this new feature too

click/_bashcomplete.py
```diff
@@ -39,12 +39,21 @@ def resolve_ctx(cli, prog_name, args):
...
         return
 
     choices = []
+
+    if last_complete:
+        for param in ctx.command.params:
+            if last_complete in param.opts:
+                if param.completer is not None:
+                    choices.extend(str(c) for c in param.completer())
+                if param.choices is not None:
```
untitaker: I think this extra logic for choices could be moved somewhere else by using the custom completer API you just introduced.

click/_bashcomplete.py
```diff
@@ -68,7 +77,12 @@ def do_complete(cli, prog_name):
     except IndexError:
         incomplete = ''
 
-    for item in get_choices(cli, prog_name, args, incomplete):
+    try:
+        last_complete = cwords[cword - 1]
```
untitaker: This is the same as `args[-1]`, is this extra parameter for `get_choices` really necessary?

untitaker:
I suspect this command is slightly buggy when completing a option with `=` like `--local-opt=`.

click/_bashcomplete.py
ericfrederich: @untitaker yeah... it would look like `args[0]` would be the same as `cwords[cword - 1]`

I got confused because within `get_choices` it wasn't, but was because I was checking after the line `ctx = resolve_ctx(cli, prog_name, args)`.
It seems that line actually modifies args!

I'll fix this up now and re-submit

ericfrederich:
@untitaker This patch should fix all the issues you mentioned with the previous one.

untitaker:
Could you add a testcase that uses `--local-opt=value` instead of `--local-opt
value`? As said I suspect that there's a bug in your code with invocations such
as:

    cli subcommand --local-opt=<tab><tab>

Thanks!

ericfrederich:
@untitaker as I'm getting more familiar with Click's internals I'm going to close this request and do a better implementation later... hopefully soon!

untitaker:
Okay cool. Feel free to email me if you're unclear about something.

PR #644, opened 2016-08-29
Fix test runner not saving caller env correctly

pmezard:
isolate() captures current env variable values, overwrites them, yields,
then restores them. Unfortunately the capturing code was broken and the
env variables were wiped out during the restore phase.

### DIFF
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -1,3 +1,4 @@
+import os
 import sys
 
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_testing.py
```diff
@@ -183,3 +184,21 @@ def cli_no_error():
     result = runner.invoke(cli_no_error)
     assert result.exit_code == 0
     assert result.output == 'hello world\n'
+
+
+def test_env():
+    @click.command()
+    def cli_env():
+        click.echo('ENV={}'.format(os.environ['TEST_CLICK_ENV']))
```
untitaker: We have to run under Python 2.6, so please change this to `"ENV=%s" % os.environ['TEST_CLICK_ENV']`

untitaker:
Applied to `6.x-maintenance` and master. Thanks!

PR #648, opened 2016-09-03
added pass_fixture decorator

ghost:
I like to create the my commands context with a generator:
http://stackoverflow.com/questions/39307764/how-to-have-a-common-setup-dispose-for-commands

It seems like there is no obvious answer so I added a decorator to do the work.


markfink:
I found a good way to monkey-patch that in. If you are not interested this can be closed...

untitaker:
@markfink Sorry for the late response.

* I don't know how this is supposed to be used. Docs and tests are missing
* No idea why you need to monkeypatch anywhere, it's just an added function

PR #649, opened 2016-09-04
Added a spinner utility

yoavram:
Sometimes you would just like to show the user some progress, but a progress bar is not suitable because you don’t know how much longer it would take. In these cases you might want to display a simple spinner using the `spinner()` function.

Example usage:

``` py
with click.spinner():
        do_something()
        do_something_else()
```

It looks like this:

spinner

Spinner class based on on a [gist by @cevaris](https://gist.github.com/cevaris/79700649f0543584009e).

### DIFF
--- a/click/__init__.py
+++ b/click/__init__.py
@@ -37,7 +37,7 @@ from .utils import echo, get_binary_stream, get_text_stream, open_file, \
 # Terminal functions
 from .termui import prompt, confirm, get_terminal_size, echo_via_pager, \
      progressbar, clear, style, unstyle, secho, edit, launch, getchar, \
-     pause
+     pause, spinner
 
 # Exceptions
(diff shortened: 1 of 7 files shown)


untitaker:
Thanks for your contribution, however, I think this is more appropriate as a third-party package, see http://click.pocoo.org/5/contrib/

yoavram:
OK, I understand, I've created a contrib repo called "click-spinner", could I put it under click-contrib?

untitaker:
I've invited you, you should be able to just transfer your repo to the org. Your access to the repo will be preserved, but you won't have write-access to other repos.

yoavram:
Done. Code moved to [click-contrib/click-spinner](https://github.com/click-contrib/click-spinner). 10x.

PR #650, opened 2016-09-05
Recognise cygwin platform and open urls accordingly

TDaglis:
Currently, click does not recognise when it is run inside a [cygwin](https://www.cygwin.com/) terminal.
This PR fixes [launch](http://click.pocoo.org/5/utils/#launching-applications). Instead of doing nothing when run inside a cygwin terminal, it properly opens the default application (picture viewer, browser, etc)

I did not find a way to make `locate` option work though so I just let it pass to the default case (where nothing happens).

### DIFF
--- a/click/_compat.py
+++ b/click/_compat.py
@@ -8,6 +8,7 @@ from weakref import WeakKeyDictionary
 
 PY2 = sys.version_info[0] == 2
 WIN = sys.platform.startswith('win')
+CYGWIN = sys.platform.startswith('cygwin')
 DEFAULT_COLUMNS = 80
 
 
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -14,7 +14,8 @@ import sys
 import time
 import math
 from ._compat import _default_text_stdout, range_type, PY2, isatty, \
-     open_stream, strip_ansi, term_len, get_best_encoding, WIN, int_types
+     open_stream, strip_ansi, term_len, get_best_encoding, WIN, int_types, \
+     CYGWIN
 from .utils import echo
 from .exceptions import ClickException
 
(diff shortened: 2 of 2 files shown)


untitaker:
According to http://stackoverflow.com/a/577609 we could also just replace `start` with `cmd /c start` in the Windows section for proper behavior within cygwin.

TDaglis:
:+1: will test and update accordingly

TDaglis:
@untitaker I updated it to `cmd /c start` and merged Cygwin case with Windows case.
The only issue I faced was the following:
With `cygstart`, to open a file via its absolute path, the "cygwin" path was used
With `cmd /c start`, to open a file by providing its absolute path, its "Windows" path has to be used.
For example, to open `test.png` in `C:\\cygwin64\\home\\Thanasis\\mydir` I'll have to:
- `cygstart "/home/Thanasis/mydir/test.png"` or
- `cmd /c start "C:\cygwin64\home\Thanasis\mydir\test.png"`

I'm not sure which should be considered correct. `os.path.realpath('test.png')` run in a cygwin terminal would produce the first path.

TDaglis:
I also just noticed that attempting to open files in any cygwin subdirectory with `cmd /c start` results to failure with `Access is denied.`, while `cygstart` opens files no matter the location.
I have no idea why this happens... I'll have to  look into it

untitaker:
I see, in that case the previous version (with the special conditional) is probably better, although the codepath for `locate` is problematic.

TDaglis:
Reverted to previous version.
Also updated codepath for `locate`.
Now, when `locate=True`, instead of opening the url, the parent directory opens. What I could not achieve was selecting the file in the new the File Explorer.

untitaker:
Excellent, thanks!

PR #656, opened 2016-09-16
Add exception assert to test_multicommand_chaining

joshleeb:
Add runtime exception assertion to `test_multicommand_chaining`.

I believe that this makes the test more expressive in conveying:
- that a runtime exception is expected to be thrown by the test;
- which section of the test is expected to throw the exception.


joshleeb:
Closed due to inactivity. Happy to reopen later if necessary.

Issue #659, opened 2016-09-20
Alter warning message in _verify_python3_env


jrobeson:
This warning could easily cause confusion on py3 only programs.

``` python
raise RuntimeError('Click will abort further execution because Python 3 '
                       'was configured to use ASCII as encoding for the '
                       'environment.  Either run this under Python 2 or '
                       'consult http://click.pocoo.org/python3/ for '
'mitigation steps.' + extra)
```

Simply changing the last sentance to `'Consult http://click.pocoo.org/python3/ for mitigation steps.' + extra'`  should make it work well enough as a last resort error message.

jrobeson:
It might be good to even just split the function up so that the output is separate from the input. That'd make it easier for folks to use the locale detection, but still provide links to their own documentation.

untitaker:
The part you are referring to is more targeted towards developers than end users, and the click docs page is too. I think it would make sense to do the change you propose and split the docs page into sections for devs and users each.

jrobeson:
I think the message should be changed either way, because some users are gonna run into it because the devs forgot to implement their own logic (or call the new split function I talked about) to handle this case.

untitaker:
Fair enough, please open a new PR

On 22 September 2016 21:46:36 CEST, Johnny Robeson <notifications@github.com> wrote:
>I think the message should be changed either way, because some users
>are gonna run into it because the devs forgot to implement their own
>logic (or call the new split function I talked about) to handle this
>case.
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/pallets/click/issues/659#issuecomment-249008290

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

shamsimam:
Fixed in commit PR #661

PR #661, opened 2016-09-23
Don't suggest to run application under Python 2

ghost:
This message is invalid for Python 3 only apps. I know that developers
should do their own error handling, but this error could confusing
to users where such error handling was not done.


untitaker:
Thanks!

dposada:
Fixes https://github.com/pallets/click/issues/659

PR #662, opened 2016-09-26
Allow empty responses from prompts

djmattyg007:
This adds a new optional parameter to click.prompt() that allows a
developer to specify that it's okay for a user to type nothing into a
prompt, and to receive None back to indicate that that is what happened.

I believe a similar thing can be achieved by setting the default value
for the prompt to an empty string (`default=''`) as the code appears to
check to specifically check if the default is `None` before using it.
I believe it is better to make this situation explicit, but if other
developers feel as though this is unwarranted, I am happy for this
patch to be rejected.

### DIFF
--- a/click/termui.py
+++ b/click/termui.py
@@ -34,7 +34,8 @@ def _build_prompt(text, suffix, show_default=False, default=None):
 def prompt(text, default=None, hide_input=False,
            confirmation_prompt=False, type=None,
            value_proc=None, prompt_suffix=': ',
-           show_default=True, err=False):
+           show_default=True, err=False,
+           allow_empty=False):
     """Prompts a user for input.  This is a convenience function that can
     be used to prompt a user for input later.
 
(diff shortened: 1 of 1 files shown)


untitaker:
I'm not sure which usecase this is trying to solve. `default` is explicitly for cases where empty input is ok (because there exists a default).

djmattyg007:
It isn't possible to specify `None` as the default value, which is what I was ideally after. I can make do with an empty string, but it would be nice to have the choice of allowing `None` too.

untitaker:
That's fair though there has to exist a value for "there is no default", and `Ǹone` seems natural for that. I don't think that a new parameter is warranted for this special case, especially as the workaround is trivial and IMO sacrifices very little elegance. There exist hacks that would allow us to make `None` a non-special value (by defining a internal `real_none = object()` and using that), but I don't consider usage of those to be good API design.

djmattyg007:
Fair enough. Thanks for the feedback :)

PR #663, opened 2016-09-27
Add direction to change default invoke for groups

timwis:
I'm relatively new to python and it took me a while to figure out why I couldn't get my second command to work. Feel free to edit this PR to add more explanatory verbiage (I don't actually know why this code block is necessary), unless you think this is enough.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -225,6 +225,14 @@ script can instead be written like this:
     def dropdb():
         click.echo('Dropped the database')
 
+When using :func:`group`, the block at the bottom should change to invoke the
+group command:
+
+.. click:example::
+
+    if __name__ == '__main__':
+        cli()
+
 Adding Parameters
 -----------------
 


untitaker:
It would be helpful if you describe what exactly you assumed/did instead, and how your code didn't work. I currently don't know which misunderstanding this PR is trying to resolve.

timwis:
Sure - I followed the [quickstart](http://click.pocoo.org/5/quickstart/) which had me start with:

```python
import click

@click.command()
def hello():
    click.echo('Hello World!')

if __name__ == '__main__':
    hello()
```

I called it something other than `hello` of course, but I started building out the functionality of that command and it worked great. I expected to run `python main.py <command-here> <arg-here>` but that didn't work; so I tried `python main.py <arg-here>` and it worked. So I kept going.

Then I wanted to add a second command, so I added another `@click.command()` and `def`, and ran `python main.py <new-command>` but it ran the original command. I figured it was because of that last line defaulting to the first command, so I tried setting that to `pass`. I eventually found someone else's python code where they used click and saw that apparently you need groups just to have more than one command. I had assumed groups were only necessary for things like nested commands, not basic 2-command CLI scripts.

Does that make it more clear?

untitaker:
Perhaps:

>Make sure that you update your setuptools entrypoints or other invocations when wrapping your existing command with a group. When you previously invoked a command from your `if __name__ == '__main__'`-block, you now need to invoke the group, as that is your new "root".

PR #664, opened 2016-09-28
Add support to invoke args as string

menegazzo:
The idea is providing a more "directly" way to use `CliRunner.invoke`. Depending on how many arguments/options you have it is really annoying writing all of them as a list

```
CliRunner().invoke(mycmd, ["mygroup", "foo", "bar", "--hey", "ya"])
```

since all I want is

```
CliRunner().invoke(mycmd, "mygroup foo bar --hey ya")
```

Sure, I can just add a `.split()` at the end (and this PR is all about this) but why not adding this as default?

Also took the chance and added tests related to `invoke(args)` since there was none (at least as far as I can see).

### DIFF
--- a/click/testing.py
+++ b/click/testing.py
@@ -3,8 +3,9 @@ import sys
 import shutil
 import tempfile
 import contextlib
+import shlex
 
-from ._compat import iteritems, PY2
+from ._compat import iteritems, PY2, string_types
 
 
 # If someone wants to vendor click, we want to ensure the
@@ -260,7 +261,10 @@ class CliRunner(object):
            The ``color`` parameter was added.
 
         :param cli: the command to invoke
-        :param args: the arguments to invoke
+        :param args: the arguments to invoke. It may be given as an iterable
+                     or a string. When given as string it will be interpreted
+                     as a Unix shell command. More details at
+                     :func:`shlex.split`.
         :param input: the input data for `sys.stdin`.
         :param env: the environment overrides.
         :param catch_exceptions: Whether to catch any other exceptions than
@@ -273,6 +277,9 @@ class CliRunner(object):
         with self.isolation(input=input, env=env, color=color) as out:
             exception = None
             exit_code = 0
+
+            if isinstance(args, string_types):
+                args = shlex.split(args)
 
             try:
                 cli.main(args=args or (),
--- a/CHANGES
+++ b/CHANGES
@@ -14,6 +14,8 @@ Version 7.0
 - `secho`'s first argument can now be `None`, like in `echo`.
 - Usage errors now hint at the `--help` option.
 - ``launch`` now works properly under Cygwin. See #650.
+- `CliRunner.invoke` now may receive `args` as a string representing
+  a Unix shell command. See #664.
 
 Version 6.7
 -----------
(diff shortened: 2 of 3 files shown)


untitaker:
What about arguments that have strings in them? What about all the escaping and quoting rules from bash? Splitting by whitespace is insufficient, explicit is also better than implicit.

On 28 September 2016 20:05:47 CEST, Fabio Menegazzo <notifications@github.com> wrote:
>The idea is providing a more "directly" way to use `CliRunner.invoke`.
>Depending on how many arguments/options you have it is really annoying
>writing all of them as a list
>
>    CliRunner().invoke(mycmd, ["mygroup", "foo", "bar", "--hey", "ya"])
>
>since all I want is
>
>    CliRunner().invoke(mycmd, "mygroup foo bar --hey ya")
>
>Sure, I can just add a `.split()` at the end (and this PR is all about
>this) but why not adding this as default?
>
>Also took the chance and added tests related to `invoke(args)` since
>there was none (at least as far as I can see).
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/pallets/click/pull/664
>
>-- Commit Summary --
>
>  * Add support to invoke args as string
>
>-- File Changes --
>
>    M click/testing.py (7)
>    M tests/test_testing.py (20)
>
>-- Patch Links --
>
>https://github.com/pallets/click/pull/664.patch
>https://github.com/pallets/click/pull/664.diff
>
>-- 
>You are receiving this because you are subscribed to this thread.
>https://github.com/pallets/click/pull/664

-- 

menegazzo:
@untitaker you have a point about arguments with strings but the idea here is bringing invoke closer to how it will be used on terminal. Yes, the support must be improved but a don't think this is an invalid idea.

Anyway, if you guys think this PR is unreasonable let's just drop it. :smile:

untitaker:
That's totally fair. I'm unsure whether I want this PR, but in any case I'd recommend using shlex over split. 

On 28 September 2016 20:24:52 CEST, Fabio Menegazzo <notifications@github.com> wrote:
>@untitaker you have a point about arguments with strings but the idea
>here is bringing invoke closer to how it will be used on terminal. Yes,
>the support must be improved but a don't think this is an invalid idea.
>
>Anyway, if you guys think this PR is unreasonable let's just drop it.
>:smile: 
>
>-- 
>https://github.com/pallets/click/pull/664#issuecomment-250255337

-- 

menegazzo:
@untitaker done. Feel free to request more changes or, as I said later, drop it. :tada:

click/testing.py
```diff
@@ -260,7 +261,7 @@ def invoke(self, cli, args=None, input=None, env=None,
            The ``color`` parameter was added.
 
         :param cli: the command to invoke
-        :param args: the arguments to invoke
+        :param args: the arguments to invoke. May be an iterable or string.
```
untitaker: We need more information in the docs than that. For example, say that it will be parsed as Unix shell command, possibly mentioning `shlex`.

We need to make clear that you can't e.g. pass a string like you would write it on the Windows commandline, and that the way you write your tests using this method doesn't necessarily represent how users would call it.

untitaker (changes requested):
Ok, this looks good. Some nitpicks inline, and also please add a changelog entry.

untitaker (approved):

untitaker:
Cool thanks!

PR #682, opened 2016-11-07
Added deprecation flag to commands

FooBarQuaxx:
I worked on a code base using click and I noticed that I need a way to signal deprecation warning to the user so I thought of adding it into click. 
Please check and advice how appropriate this commits to add to click.


LokiHokie:
Something went wrong with GitHub. I will make a new PR for this with your commits and preserve the author information. Sorry.

PR #683, opened 2016-11-07
Guard against attribute error when getting default program name. 

andricDu:
Apologies for not opening up an issue first but it seemed a simple fix. 

I'm sure this is an extreme edge case, but I inherited a project built with click and I was putting some work in getting the testing to actually run when I ran into this issue.

My tests showed:
```
E           assert '' == 0
E            +  where '' = <Result AttributeError("'module' object has no attribute 'name'",)>.output
```

When I dug a little deeper I found the culprit to be the line 129 in `testing.py`. 

Ran tests with `tox`, everything seems okay.

### DIFF
--- a/click/testing.py
+++ b/click/testing.py
@@ -126,7 +126,7 @@ class CliRunner(object):
         for it.  The default is the `name` attribute or ``"root"`` if not
         set.
         """
-        return getattr(cli, 'name', 'root')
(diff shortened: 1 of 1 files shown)


untitaker:
You're passing a module object somewhere where a command object is expected.

PR #684, opened 2016-11-15
improve the doc

didicout:
another way to realize Option-Like Arguments

### DIFF
--- a/docs/arguments.rst
+++ b/docs/arguments.rst
@@ -243,3 +243,21 @@ And from the command line:
 .. click:run::
 
     invoke(touch, ['--', '-foo.txt', 'bar.txt'])
+
+If you don't like the '--' marker, you can set ignore_unknown_options to
+True to avoid checking unknown options:
+
+.. click:example::
+
+    @click.command(context_settings={"ignore_unknown_options": True})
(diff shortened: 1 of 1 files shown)


docs/arguments.rst
```diff
@@ -243,3 +243,21 @@ And from the command line:
 .. click:run::
 
     invoke(touch, ['--', '-foo.txt', 'bar.txt'])
+
+If you don't like the '--' marker, you can set ignore_unknown_options to
```
untitaker: Please wrap Python code (and the `--`) in `````` to make it appear as monospace.

didicout:
Thanks for the suggestion. In fact, after I read the doc, I thought using -- is the only way to realize Option-Like Arguments. Then I started to edit the source code😂... Finally I found "ignore_unknown_options". So I want to tell other devs 😅

didicout:
One of my subcommand is actually a proxy of another command-line tool. For example:  `my-tool hdfs fs -ls /xxx `  is to list files in hdfs. So I need to get all the args of this subcommand without checking.

untitaker:
Fair enough, but please add a note saying that this may have unintended side effects.

PR #695, opened 2016-12-04
Fix path type

skozlovf:
File type is never reached in original code, I believe it's a typo.

### DIFF
--- a/click/types.py
+++ b/click/types.py
@@ -425,7 +425,7 @@ class Path(ParamType):
         if self.file_okay and not self.dir_okay:
             self.name = 'file'
             self.path_type = 'File'
-        if self.dir_okay and not self.file_okay:
+        elif self.dir_okay and not self.file_okay:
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Doesn't really make a difference whether it's `if` or `elif` here since the body of the initial `if` does not affect the conditions of the second `if`, so the behavior is identical for `if` and `elif`

skozlovf:
`else` overwrites first `if`.

Fast test:

```python
    import click


    @click.command()
    @click.option('-o', type=click.Path(dir_okay=False))
    def test(mypath):
        """Path test"""
        click.echo(mypath)


    test()
```

Original behaviour:
```
Options:
  -o PATH
```

Fixed behaviour:
```
Options:
  -o FILE
```

ThiefMaster:
true, makes sense. didn't look at the else for some reason

PR #706, opened 2016-12-22
Update progress after iteration (fix #651)

### LINKED ISSUE #651: Progress bar should advance at end of loop
I'm using the progress bar with a piece of code like this:

``` python
with click.progressbar(some_list) as bar:
    for element in bar:
        element.long_calculation
```

It seems that the progress bar updates every time it enters the loop, which for the last element means the bar shows everything has been completed, although the last computation just started. It would be great if it would show the progress in terms of results, not as an iterator.

I looked briefly at the implementation in [_termui_impl.py](https://github.com/pallets/click/blob/master/click/_termui_impl.py), but the rendering seems to be closely linked to the `pos` attribute and I don't feel comfortable to mess with that. 😒

jojje:
Here's an initial stab at #651, which I'm submitting for discussion.
By using a generator instead of an iterator, doing post-iteration work (progress bar update) becomes trivial. With an iterator, not so much, at least no obvious way that I could think of. 

Delaying the status update to _after_ the iteration also fixed the ETA.
E.g. the following won't show a proper ETA using 0.6.6 nor master
```python
from time import sleep
from click import progressbar

with progressbar(range(10)) as bar:
    for _ in bar:
        sleep(1)
```
    =[output]==>
    before first  iteration: [###---------------------------------]   10%
    before second iteration: [#######-----------------------------]   20%  00:00:04

Using a generator to update the progress _after_ the iteration rounds fixes the ETA, and the progress starts at 0% _until_ the first iteration is done, as was requested in the ticket. Only when the last item has been iterated over, and processed by the user's block of code, will the statusbar show 100%.

```python
with progressbar(range(10)) as bar:
    for _ in bar:
        sleep(1)
```
    =[output]==>
    before first iteration  [------------------------------------]    0%
    before second iteration [###---------------------------------]   10%  00:00:09

Any style or design notions that are broken in this draft change?
-----
[commit message]
Update progress after iteration (fix #651)

Reported "off by one" error is due to the fact that an interator is used
instead of a generator. As such there is no way for the bar to know when
it has iterated over the user's code block.

This commit swaps the iterator for a generator, and changes _when_ the
status report is updated, from before each iteration of a user's code
block, to _after_.

Updating the status report after the user's code has run allows the
report to convey how much work has been _done_ so far.

Accidental gain:
The associated unit test performs a black-box test on the generated
progress output. This test revealed a bug in the ETA time calculation
were the behavior differed between python 2 and 3, causing the unit test
to fail on python 3. The remedy employed to allow passing of the test is
to use integer division for the ETA calculation, which should also
address a related issue (fixes #485).

### DIFF
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -102,7 +102,7 @@ class ProgressBar(object):
         if not self.entered:
             raise RuntimeError('You need to use progress bars in a with block.')
         self.render_progress()
-        return self
+        return self.generator()
 
     def render_finish(self):
         if self.is_hidden:
@@ -130,13 +130,13 @@ class ProgressBar(object):
 
     def format_eta(self):
         if self.eta_known:
-            t = self.eta + 1
+            t = int(self.eta)
             seconds = t % 60
-            t /= 60
+            t //= 60
             minutes = t % 60
-            t /= 60
+            t //= 60
             hours = t % 24
-            t /= 24
+            t //= 24
             if t > 0:
                 days = t
                 return '%dd %02d:%02d:%02d' % (days, hours, minutes, seconds)
@@ -253,23 +253,25 @@ class ProgressBar(object):
         self.current_item = None
         self.finished = True
 
-    def next(self):
+    def generator(self):
+        """
+        Returns a generator which yields the items added to the bar during
+        construction, and updates the progress bar *after* the yielded block
+        returns.
+        """
+        if not self.entered:
+            raise RuntimeError('You need to use progress bars in a with block.')
+
         if self.is_hidden:
-            return next(self.iter)
-        try:
-            rv = next(self.iter)
-            self.current_item = rv
-        except StopIteration:
+            for rv in self.iter:
+                yield rv
+        else:
+            for rv in self.iter:
+                self.current_item = rv
+                yield rv
+                self.update(1)
             self.finish()
             self.render_progress()
-            raise StopIteration()
-        else:
-            self.update(1)
-            return rv
-
-    if not PY2:
-        __next__ = next
-        del next
 
 
 def pager(text, color=None):
--- a/tests/test_termui.py
+++ b/tests/test_termui.py
@@ -1,4 +1,5 @@
 import click
+import time
 
 
 def test_progressbar_strip_regression(runner, monkeypatch):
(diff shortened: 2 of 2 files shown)


ThiefMaster:
I think you could simply adapt `__iter__` to use `progress` itself as an iterator. i.e. simply move the logic from your `items()` method into `__iter__`

click/_termui_impl.py
```diff
@@ -130,7 +130,7 @@ def eta(self):
 
     def format_eta(self):
         if self.eta_known:
-            t = self.eta + 1
+            t = int(self.eta)
```
jojje: * offset removed as it's not needed when update order has been swapped (update after instead of before).
* The int is here since the code below operates on integer logic, and eta (time) is decimal.
* Note: Touching this line will cause the change to conflict with a PR for issue 485, so depending on merging order, some rebasing will likely be needed.

tests/test_termui.py
```diff
@@ -69,3 +70,42 @@ def test_secho(runner):
...
+            return self.now
+
+    fake_clock = FakeClock()
+
+    @click.command()
+    def cli():
+        with click.progressbar(range(4)) as progress:
+            for _ in progress.items():
+                fake_clock.advance_time()
+                print("")
```
jojje: Allows capture and assertion of the status bar update, after each iteration, by using _line splitting and "grep"_

jojje: Allows capture and assertion of the status bar update, after each iteration, by using _line splitting and "grep"_

jojje:
@ThiefMaster I'll have to sleep on it and a way may appear. Will take a stab at it tomorrow after the final christmas shopping.

click/_termui_impl.py
```diff
@@ -130,13 +130,13 @@ def eta(self):
...
             seconds = t % 60
-            t /= 60
+            t //= 60
```
jojje: On python 2, division was integer, but with pyton 3, the result was floating point. Since the algorithm clearly relies on an assumption of integer division, making it explicit was the solution to fix the travis python3 failures.
Why this hasn't been discovered earlier is probably because this code path hasn't had any testing.. Which it now does, to a lesser extent.

jojje:
@ThiefMaster couldn't sleep 'till this was sorted. Was less painful than I initially thought and you pointing me in the right direction helped immensely. thanks!

I killed off all but the __iter__ method from the iterator interface, and with the tests passing it doesn't seem to have done any harm. Only risky thing I can imagine with this _slaughter_  is if some user for some reason is relying on there being explicit methods available from the iterator interface.  In such case that code will break if this change is merged.

With this update, any other reflections on the PR?

wreed4:
How soon can this get merged?  The progress bar updating too soon is kinda odd behavior. Particularly with a small number of long-running tasks.

Issue #598, opened 2016-06-13
Argument metavar parameter not used with exception message


cdunklau:
#273 exists (again/still) in Click 6.6:
# Code

``` python
import click
print click.__version__

@click.command()
@click.argument('to_say', metavar='message')
def foo(to_say):
    print to_say


if __name__ == '__main__':
    foo()
```
# Expected Output

```
6.6
Usage: foo.py [OPTIONS] message

Error: Missing argument "message".
```
# Observed Output

```
6.6
Usage: foo.py [OPTIONS] message

Error: Missing argument "to_say".
```

cdunklau:
Still an issue in 6.7

haizaar:
Indeed an issue - should've been `type` instead of `backup_type`

```python
$ python /tmp/bug.py
Usage: bug.py [OPTIONS] type

Error: Missing argument "backup_type".

$ cat /tmp/bug.py
import click

@click.command('mssql')
@click.argument('backup_type', metavar='type')
def mssql(backup_type):
        click.echo('Running the backup...')
mssql()

$ pip show click
Name: click
Version: 6.7

$ python --version
Python 3.6.4
```

LokiHokie:
Fixed #709

PR #709, opened 2016-12-30
Derive "param_hint" in errors from param itself

### LINKED ISSUE #704: "Missing argument" message uses arg name, not metavar
If one defines the metavar different from the argument name, like in:
```
@adapter_group.command('list')
@click.argument('CPC-NAME', type=str, metavar='CPC')
@click.pass_obj
def adapter_list(cmd_obj, cpc_name):
```
the usage line in the help text correctly uses the metavar:
```
Usage: zhmc adapter list [OPTIONS] CPC
```
but the error message when invoked without arguments incorrectly uses the argument name:
```
Error: Missing argument "CPC-NAME".
```

I suggest that the metavar is used consistently in all display to the user.

sirosen:
Rather than trying to apply the same logic to all param types, trying to extract a meaningful value from Parameter.opts even for Arguments, make this a function computed against the Parameter itself. Then, when it's time to specialize it to use the metavar for Arguments, we just need to override the function.
Also unify a small amount of logic being shared among exception classes.
Fully backwards compatible, except for the change to error text on Arguments, which is the intended effect.

Grabs a test out of #612 which checks that this works as intended.

Resolves #704, #598, and #612

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -1431,6 +1431,13 @@ class Parameter(object):
     def get_usage_pieces(self, ctx):
         return []
 
+    def get_error_hint(self, ctx):
+        """Get a stringified version of the param for use in error messages to
+        indicate which param caused the error.
+        """
+        hint_list = self.opts or [self.human_readable_name]
+        return ' / '.join('"%s"' % x for x in hint_list)
+
 
 class Option(Parameter):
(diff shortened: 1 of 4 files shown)


andy-maier:
+1

PR #728, opened 2017-02-10
Disabled sys._getframe() on Python interpreters that don't support it

jhwgh1968:
sys._getframe() is an implementation detail of CPython, which has been copied by some but not all Python interpreters. Notably IronPython does not support it.

This simple patch allows the library to function on IronPython with minimal side effects. It disables a unicode check from the future module (which is not likely to happen on that platform), and disables autodetecting module version strings in the version option decorator (which could fail anyway).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,14 @@ Click Changelog
 
 This contains all major version changes between Click releases.
 
+Version 6.8
+-----------
+
+(bugfix release; yet to be released)
+
+- Disabled sys._getframes() on Python interpreters that don't support it. See
+  #728.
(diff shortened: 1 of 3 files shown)


untitaker:
Could you add a changelog entry? Thanks!

untitaker:
Did you rebase this against 6.x-maintenance?

jhwgh1968:
I originally did rebase against 6.x-maintenance, I just picked the wrong option when opening the pull request.

Fixed.

untitaker:
Neat, thanks!

Issue #732, opened 2017-02-11
[patch] autosave with nano when calling click.edit(require_save=False)


elegantandrogyne:
GNU nano has a nice feature of saving a file automatically on exit (i.e. not asking user to do it) when "-t [FILE]" (or "--tempfile [FILE]") is supplied in cmdline args. Let's make a use of that.

elegantandrogyne:
#733

elegantandrogyne:
wontfix

PR #733, opened 2017-02-11
fix for #732

keriszafir:
makes click.edit(require_save=False) make nano use the "-t" / "--tempfile" parameter and not ask for saving on exit

### DIFF
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -372,10 +372,12 @@ def _nullpager(stream, text, color):
 class Editor(object):
 
     def __init__(self, editor=None, env=None, require_save=True,
+                 nano_autosave=False,
                  extension='.txt'):
         self.editor = editor
         self.env = env
         self.require_save = require_save
+        self.nano_autosave = nano_autosave
         self.extension = extension
 
     def get_editor(self):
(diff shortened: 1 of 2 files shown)


untitaker:
Nano already skips the save prompt if not required.

elegantandrogyne:
@untitaker testcase on a current (unpatched) click:
1. open python3, enter:
`import click`
`string = click.edit(require_save=False)`

2. nano opens
3. enter some characters
4. `ctrl-X` to exit

what happens: nano asks if you want to save the file (if so - give filename)
expected behavior: no prompt, just exit

untitaker:
Ah, I see why you want to have that flag now. But I don't understand why you make it dependent on `require_save`.

elegantandrogyne:
Okay, let's add another option and we'll be fine.

untitaker:
I think it's acceptable to make this behavior not configurable.

elegantandrogyne:
OK, I found a workaround: https://gist.github.com/elegantandrogyne/c1363a62654c2414f0fe90fe35dacf7d

untitaker:
Yeah, something like that. Could you therefore remove the `nano_autosave` option and just add `-t` everytime? Thanks!

untitaker:
Why @elegantandrogyne? I would've been happy to merge this PR if it just unconditionally added the `-t` option!

elegantandrogyne:
Unconditionally = for all editors? No way, it's too risky. Every editor has its own set of command line parameters and there's a high chance of our argument breaking things.

untitaker:
No, just for nano I mean. Unconditionally in the sense that there's no dependency on any option.

PR #739, opened 2017-02-21
fix exit_code for click.testing.CliRunner.invoke

timfeirg:
the `SystemExit.code` could be None, and when that happens, `CliRunner.invoke` would return `1` as the exit code while it's actually 0, 

https://docs.python.org/2/library/exceptions.html#exceptions.SystemExit

### DIFF
--- a/click/testing.py
+++ b/click/testing.py
@@ -1,9 +1,9 @@
-import os
-import sys
-import shutil
-import tempfile
 import contextlib
+import os
 import shlex
+import shutil
+import sys
+import tempfile
 
 from ._compat import iteritems, PY2, string_types
 
@@ -197,6 +197,7 @@ class CliRunner(object):
             return char
 
         default_color = color
+
         def should_strip_ansi(stream=None, color=None):
             if color is None:
                 return not default_color
@@ -285,21 +286,24 @@ class CliRunner(object):
                 cli.main(args=args or (),
                          prog_name=self.get_default_prog_name(cli), **extra)
             except SystemExit as e:
-                if e.code != 0:
+                exc_info = sys.exc_info()
+                exit_code = e.code
+                if exit_code is None:
+                    exit_code = 0
+
+                if exit_code != 0:
                     exception = e
 
-                exc_info = sys.exc_info()
-
-                exit_code = e.code
                 if not isinstance(exit_code, int):
                     sys.stdout.write(str(exit_code))
                     sys.stdout.write('\n')
                     exit_code = 1
+
             except Exception as e:
                 if not catch_exceptions:
                     raise
                 exception = e
-                exit_code = -1
+                exit_code = 1
                 exc_info = sys.exc_info()
             finally:
                 sys.stdout.flush()
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ build
 docs/_build
 click.egg-info
 .tox
+.cache
+.ropeproject


untitaker:
This still prints `"None"` though.

timfeirg:
I'm not sure why `CliRunner.invoke` would write the exit_code to stdout in the first place, this is not standard linux behavior.

So I can't decide what to write to stdout, I can change the `None` to 0, but I think a bigger issue is at hand.

@untitaker

untitaker:
>I'm not sure why CliRunner.invoke would write the exit_code to stdout in the first place, this is not standard linux behavior.

This is all specified behavior of `sys.exit`, see the link you posted in the OP :)

untitaker:
Sorry, but I don't think writing the traceback to stdio is a good idea. This is information that can best be accessed using `catch_exceptions=False` or something similar. Stick to your original plan :)

timfeirg:
I may have misunderstood the purpose of `CliRunner` here, from the code of `CliRunner.invoke`, I can see that the intention is not to fully mimic the standard CLI behavior, but to conveniently determine the exit code and stdout (but not the stderr) of `click.command`s, am I right?

In that case, I've revert a few lines of code and remove the float exit code in the output, because linux exit code  is 8-bit int?

@untitaker

untitaker:
I think you've dug yourself into a hole here. Your original code was almost perfect!

>but to conveniently determine the exit code and stdout (but not the stderr) of click.commands, am I right?

We do want to mimic the CLI behavior, but whenever `exception` is something other than `SystemExit` I really don't think we should print the traceback, because those situations indicate an abnormal crash where the output should be irrelevant. If you actually want to print a traceback as part of "normal" program output, you should do so explicitly and not "rely on the Python interpreter" to do it for you.

So please no extra print-traceback code here. Just, as said, stick to your original bugfix for situations where `sys.exit(None)` is called.

timfeirg:
Your explanation has been very educating, thanks. 
I didn't check the diff carefully enough, the code which prints stack trace has been removed.

click/testing.py
```diff
@@ -292,14 +294,19 @@ def invoke(self, cli, args=None, input=None, env=None,
 
                 exit_code = e.code
                 if not isinstance(exit_code, int):
+                    if exit_code is None:
+                        exit_code = 0
+                    else:
+                        exit_code = 1
```
untitaker: This breaks the usecase where somebody calls `sys.exit("Error message")`. This is why the `exit_code` was overridden *after* we print the message

click/testing.py
```diff
@@ -292,14 +294,19 @@ def invoke(self, cli, args=None, input=None, env=None,
 
                 exit_code = e.code
                 if not isinstance(exit_code, int):
+                    if exit_code is None:
```
untitaker: You need to do this outside of the outer if, because otherwise we will print `"0"`.

timfeirg: done

click/testing.py
timfeirg: fixed

click/testing.py
```diff
@@ -285,21 +286,24 @@ def invoke(self, cli, args=None, input=None, env=None,
                 cli.main(args=args or (),
                          prog_name=self.get_default_prog_name(cli), **extra)
             except SystemExit as e:
```
untitaker: Hmm, if we instead write `exit_code = e.code; if exit_code is None: exit_code = 0` and replace `e.code` with `exit_code` further down, we can spare the double-check for `None`

timfeirg: done

untitaker:
Thanks! Sorry for the very "iterative" review.

PR #755, opened 2017-03-27
Dynamic bash autocompletion

stopthatcow:
This PR enables arguments to supply an auto completion callback that dynamically generates completion options. This allows Click to complete arguments that are dynamic in nature like git branch names, database variables, and any other number of things. Much thanks to @bb4242 for spearheading this change. Brett, feel free to take a look! Feedback is welcome!


click/core.py
```diff
@@ -1706,6 +1706,8 @@ def __init__(self, param_decls, required=None, **attrs):
                 required = False
             else:
                 required = attrs.get('nargs', 1) > 0
+        if attrs.get('autocompletion') is not None:
+            self.autocompletion =  attrs.pop('autocompletion')
```
brianbruggeman: I think this should actually show up under Parameters rather than explicitly Argument.  Presumably, an Option might also benefit from autocompletion.

stopthatcow: Ok, I added that as well

stopthatcow:
@pgkelley4 Please feel free to take a look as you've recently been in this code.

click/core.py
```diff
@@ -1266,7 +1266,8 @@ class Parameter(object):
 
     def __init__(self, param_decls=None, type=None, required=False,
                  default=None, callback=None, nargs=None, metavar=None,
-                 expose_value=True, is_eager=False, envvar=None):
+                 expose_value=True, is_eager=False, envvar=None,
+                 autocompletion=None):
```
pgkelley4: If this is only available for arguments, shouldn't it be defined on that class instead?

click/core.py
```diff
@@ -1289,6 +1290,8 @@ def __init__(self, param_decls=None, type=None, required=False,
         self.is_eager = is_eager
         self.metavar = metavar
         self.envvar = envvar
+        if autocompletion is not None:
```
pgkelley4: Nitpick but I would just allow this to be set to None here and then check `if cmd_param.autocompletion`  in _bashcomplete.py instead of hasattr. Or was there a reason that you did it this way that I'm missing?

pgkelley4 (commented):
Great addition!

click/core.py
pgkelley4: Ahh I see you had it there originally. I think it makes sense as you have it now.

docs/bashcomplete.rst
```diff
@@ -32,6 +29,33 @@ least a dash has been provided.  Example::
     $ repo clone -<TAB><TAB>
     --deep     --help     --rev      --shallow  -r
 
+Additionally, custom suggestions can be provided for arguments with the
```
pgkelley4: It's both arguments and parameters as you have it now.

click/core.py
pgkelley4: Or maybe more explicitly:
`if cmd_param.autocompletion is not None`

Then the behavior is consistent if they incorrectly set autocompletion=false. Although that seems unlikely.

pgkelley4:
LGTM @untitaker @stopthatcow

bb4242:
Hi @stopthatcow! This looks great to me. :)

stopthatcow:
@untitaker any chance you could take a look at this? I'd love your feedback.

click/_bashcomplete.py
```diff
@@ -97,6 +96,22 @@ def is_incomplete_argument(current_params, cmd_param):
...
 
+def get_user_autocompletions(ctx, args, incomplete, cmd_param):
+    """
+    :param ctx: context associated with the parsed command
+    :param args: full list of args
+    :param incomplete: the incomplete text to autocomplete
+    :param cmd_param: command definition
+    :return: all the possible user-specified completions for the param
+    """
+    choices = []
```
untitaker: You can just use `yield` here.

untitaker:
Yeah LGTM.

click/_bashcomplete.py
stopthatcow: Ok, is this is what you had in mind?

untitaker: Ah, hmm. I'm trying to avoid instantiation of a list here. More comments will follow.

click/_bashcomplete.py
```diff
@@ -97,6 +96,23 @@ def is_incomplete_argument(current_params, cmd_param):
...
+    if isinstance(cmd_param.type, Choice):
+        choices = cmd_param.type.choices
```
untitaker: You could just `return cmd_param.type.choices` here (same for the elif branch) and add an else branch that does `return []`. Then you don't have to create a list everytime.

stopthatcow: Sounds good.

untitaker:
Lgtm, will merge tomorrow

On April 15, 2017 9:25:05 PM GMT+02:00, Nicholas Wiles <notifications@github.com> wrote:
>stopthatcow commented on this pull request.
>
>
>
>> @@ -97,6 +96,23 @@ def is_incomplete_argument(current_params,
>cmd_param):
>         return True
>     return False
> 
>+def get_user_autocompletions(ctx, args, incomplete, cmd_param):
>+    """
>+    :param ctx: context associated with the parsed command
>+    :param args: full list of args
>+    :param incomplete: the incomplete text to autocomplete
>+    :param cmd_param: command definition
>+    :return: all the possible user-specified completions for the param
>+    """
>+    choices = []
>+    if isinstance(cmd_param.type, Choice):
>+        choices = cmd_param.type.choices
>
>Sounds good.
>
>-- 
>https://github.com/pallets/click/pull/755#discussion_r111671637

-- 

untitaker:
Great thanks!

click/core.py
adamtheturtle: This is not documented in the `:param:` list above.

Issue #756, opened 2017-03-27
Click should provide means to customize write


Kentzo:
Sometimes the output machinery embedded in click is not desirable. At least the function that performs write should be customizable to give a user a chance to handle and process all errors that may happen.

untitaker:
This is not very specific. Could you name a particular situation where the current behavior is undesirable?

Kentzo:
With current design it's impossible to control how errors are being handled in `echo` and all other functions that end up using it. 

In particular, we have a problem when message cannot be represented in `latin-1`. Since this is part of the output that _must_ be printed and never cause an exception (at least not due to encoding) I need to customize the behavior. In particular, I'd change [from](https://github.com/pallets/click/blob/c8e2110/click/utils.py#L260):
```
if message:
    file.write(message)
```
to:
```
if message:
    try:
        file.write(message)
    except UnicodeEncodeError as e:
        file.write(message.encode(e.encoding, errors='backslashreplace').decode(e.encoding))
```

(I'm using Python 3, Python 2 compatibility may require adjustments)

untitaker:
If you encounter encoding errors like that you really should set your system locale to a correct (i.e. UTF-8) value.

Kentzo:
@untitaker Not always you can control the encoding of stdout. And that's a different (and not a simple) story of why it's that restricted in my case.

My point is that you can have a UTF-16-only string and you may want to still print it (handling arbitrary logging, data stream, whatever?) even when you have crippled 'latin-1'. Granted, one could use custom printing, but then it'll require to give up all the conveniences of `click`.

KiYugadgeter:
Hello, I think `errors` should use `replace` instead of `backslashreplace`.
In python2, `echo` prints `?` at #705 .

untitaker:
We could provide a possibility for the user to replace `click.echo` with a custom function (that lives on the context), but stuff like #757 is not sensible.

KiYugadgeter:
sys.stdout.encoding returns system default encoding of stdout by default.

untitaker:
I am talking about backslashreplace, not the encoding.

KiYugadgeter:
Is it possible to add `errors` argument to echo function?
Or should not do this for backward compatibility?

    click.echo(message, file, nl, errors="strict", error, color)

untitaker:
Sure, but that solves only a subset of all problems. It doesn't affect echo()
calls outside of your code.

KiYugadgeter:
[...]

PR #757, opened 2017-03-29
Make message printable even when it cannot be encoded

Kentzo:
In case when file's encoding cannot represent message,
`backslashreplace` problematic characters.

Refs #756


untitaker:
"keep chugging along at all costs" is not a sensible default.

KiYugadgeter:
`sys.stdout.encoding` returns a encoding name of stdout by default.

PR #681, opened 2016-11-04
Issue #535: Bash completion for click.Choice()

pgkelley4:
Added support for bash compeletion of
type=click.Choice Options and Arguments.

Added comprehensive tests for bash completion.
There are many scenarios present for arguments.

PR #761, opened 2017-04-05
Handle completion of Choices in Tuple Options

bobwhitelock:
Completion of Choice option values was recently added, in https://github.com/pallets/click/pull/681. However, this does not cover Choice values within Tuple options. This commit adds similar handling so completion will cover these option values as well.

### DIFF
--- a/click/_bashcomplete.py
+++ b/click/_bashcomplete.py
@@ -6,7 +6,7 @@ import re
 from .utils import echo
 from .parser import split_arg_string
 from .core import MultiCommand, Option, Argument
-from .types import Choice
+from .types import Choice, Tuple
 
 
 WORDBREAK = '='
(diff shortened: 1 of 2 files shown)


JDFagan:
I was wondering if this issue was for click (version 6) to support having a click.option of type Tuple whose one tuple is itself a click.Choice type like so?:

```
@click.option('--route', '-r', default=None, multiple=True, 
              type=(click.Choice(['Accounts', 'Listing', 'Summary'], click.STRING))
```

davidism:
Completion has changed significantly since this PR was posted. Would you or @stopthatcow review this to see if it's still needed and can be rebased?

davidism:
Thanks for working on this. We're rewriting the whole completion system, if this is still and issue after that we'll need to find a new solution that fits with the new system. See #1484 and #1622.

Issue #270, opened 2014-12-12
is raise SystemExit(exit_code) the right way to return a non-zero exit status?


aconrad:
I would like to control the exit status of a command. I saw that different exceptions are available from Click which are mostly used for bad command line usage.

I just want a clean exit for a properly ran command (with no `Abort!` message). I ended up using `raise SystemExit(1)`. Is this the proper way of doing it? I was thinking that it'd be nice if the return value of the command would be the exit status.

```python
@pycobertura.command()
@click.argument('cobertura_file1')
@click.argument('cobertura_file2')
def diff(cobertura_file1, cobertura_file2):
    """compare two coverage files"""
    cobertura1 = Cobertura(cobertura_file1)
    cobertura2 = Cobertura(cobertura_file2)
    reporter = DeltaReporter(cobertura1, cobertura2)

    report = reporter.generate()
    click.echo(report)

    # non-zero exit code if line rate worsened
    exit_code = cobertura1.line_rate() > cobertura2.line_rate()
    raise SystemExit(exit_code)  # or `return exit_code`?
```

scharissis:
Why not use the regular python site module exit()?
It works great for me.

https://docs.python.org/2/library/constants.html#exit

untitaker:
@scharissis:

>They are useful for the interactive interpreter shell and should not be used in programs.

Use ``sys.exit`` instead, or just raise the exception manually, either way is fine.

gst:
Hi,

@scharissis  @untitaker 

I don't create a new issue as my question is totally linked to this one : 

why does click so apparently "force" the return code of the function to 0 and ignore the given return value?

well actually it's not that it force the return value to 0, it doesn't return but executes (force?) sys.exit() itself. with 0 whatever is the return value you actually returned from the function so..

I mean : wouldn't it be ok/better that it uses the returned value from the function to sys.exit() ?

Thx for any info.

untitaker:
@gst You're mixing up raising `SystemExit` with function return values here, so it's not really clear what you're asking. The return value of the function was never something that Click honored.

gst:
hi,

well.. I wasn't 100% clear effectively.

My question is so:

couldn't `click` use the return value of a _click.Command_ decorated function when it returns, like `diff` here above,  and to sys.exit() that value ? That is just `return cobertura1.line_rate() > cobertura2.line_rate()` for the above.

[...]

PR #765, opened 2017-04-18
Command.main() now calls exit() with returnvalue of invoke()

NiklasRosenstein:
see pallets/click#270.

> @unittake  
> I tend to agree, if you open a complete PR with doc updates and tests and changelog, I will merge it.

Testcase results are the same before and after the change ([before](https://bpaste.net/show/4b45e8b59f35), [after](https://bpaste.net/show/99189519432e)).

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -638,9 +638,16 @@ class BaseCommand(object):
     def main(self, args=None, prog_name=None, complete_var=None,
              standalone_mode=True, **extra):
         """This is the way to invoke a script with all the bells and
-        whistles as a command line application.  This will always terminate
-        the application after a call.  If this is not wanted, ``SystemExit``
-        needs to be caught.
+        whistles as a command line application.
+
+        If *standalone_mode* is ``True`` (default), the application will
+        terminate immediately after the command was executed. The return value
+        of :meth:`invoke` will be passed to :func:`sys.exit` and thus be used
+        as the application's exit-code.
+
+        Otherwise, if *standalone_mode* is ``False``, the return value of
+        :meth:`invoke` is returned instead and no ``SystemExit`` will be
+        raised.
 
         This method is also available by directly calling the instance of
         a :class:`Command`.
@@ -662,7 +669,7 @@ class BaseCommand(object):
                                 handle exceptions and convert them into
                                 error messages and the function will never
                                 return but shut down the interpreter.  If
-                                this is set to `False` they will be
+                                this is set to ``False`` they will be
                                 propagated to the caller and the return
                                 value of this function is the return value
                                 of :meth:`invoke`.
@@ -697,7 +704,7 @@ class BaseCommand(object):
                     rv = self.invoke(ctx)
                     if not standalone_mode:
                         return rv
-                    ctx.exit()
+                    ctx.exit(rv)
             except (EOFError, KeyboardInterrupt):
                 echo(file=sys.stderr)
                 raise Abort()
--- a/click/core.py
+++ b/click/core.py
@@ -640,12 +640,13 @@ class BaseCommand(object):
         """This is the way to invoke a script with all the bells and
         whistles as a command line application.
 
-        If *standalone_mode* is ``True`` (default), the application will
+        If ``standalone_mode`` is ``True`` (default), the application will
         terminate immediately after the command was executed. The return value
         of :meth:`invoke` will be passed to :func:`sys.exit` and thus be used
-        as the application's exit-code.
+        as the application's exit-code. Note that a return-value of ``None``
+        is mapped to ``0`` in this case.
 
-        Otherwise, if *standalone_mode* is ``False``, the return value of
+        Otherwise, if ``standalone_mode`` is ``False``, the return value of
         :meth:`invoke` is returned instead and no ``SystemExit`` will be
         raised.
 
@@ -653,7 +654,11 @@ class BaseCommand(object):
         a :class:`Command`.
 
         .. versionadded:: 3.0
-           Added the `standalone_mode` flag to control the standalone mode.
+           Added the ``standalone_mode`` flag to control the standalone mode.
+
+        .. versionchanged:: 7.0
+           ``standalone_mode`` set to ``True`` causes the generated
+           ``SystemExit`` exception to carry the result of :meth:`invoke()`.
 
         :param args: the arguments that should be used for parsing.  If not
                      provided, ``sys.argv[1:]`` is used.
@@ -704,6 +709,8 @@ class BaseCommand(object):
                     rv = self.invoke(ctx)
                     if not standalone_mode:
                         return rv
+                    if rv is None:
+                        rv = 0
                     ctx.exit(rv)
             except (EOFError, KeyboardInterrupt):
                 echo(file=sys.stderr)


click/core.py
```diff
@@ -638,9 +638,16 @@ def invoke(self, ctx):
     def main(self, args=None, prog_name=None, complete_var=None,
              standalone_mode=True, **extra):
         """This is the way to invoke a script with all the bells and
-        whistles as a command line application.  This will always terminate
-        the application after a call.  If this is not wanted, ``SystemExit``
-        needs to be caught.
+        whistles as a command line application.
+
+        If *standalone_mode* is ``True`` (default), the application will
```
untitaker: Please use codetags instead of italics here.

untitaker:
Please add a changelog entry and perhaps a `versionchanged` rst block (like `versionadded` in the same docstring)

click/core.py
```diff
@@ -697,7 +704,7 @@ def main(self, args=None, prog_name=None, complete_var=None,
                     rv = self.invoke(ctx)
                     if not standalone_mode:
                         return rv
-                    ctx.exit()
+                    ctx.exit(rv)
```
untitaker: We should map `None` to `0` because otherwise all tests will fail (see Travis)

NiklasRosenstein: Interesting that I get different testing results when using `cd tests && pytest` and `make test`. I can reproduce the test case failures on Travis using `make test`.

NiklasRosenstein:
@untitaker Unfortunately the tests for chained commands still fail, I couldn't find out where to solve it yet. Could you take a look?

untitaker:
Ok. Here's the problem:

    @click.command()
    def foo():
       return 0

If you run `foo.main(['foo'], standalone_mode=False)`, you will get a `0` as return value. However, if you have a command group with `chain=True` and the subcommands `a`, `b`, `c`, and you invoke them like this:

    foo.main(['foo', 'a', 'b', 'c'], standalone_mode=False)

You currently get `[0, 0, 0]` back, and your current code tries to run `sys.exit([0,0,0])`.

It's much less obvious what to do here. Should we use the first non-zero, non-none value as exit code? Or...?

untitaker:
Also it appears you can interrupt the chain by calling `sys.exit(0)` in one subcommand. If I raise `SystemExit(0)` from `a`, then `b` and `c` will not run but the exit code will be `0`. How should we behave if one returns `0` from a command? Should we interrupt the chain too?

NiklasRosenstein:
Thanks for taking a look. Imho, interrupting the chain and returning the last command's exit code makes the most sense.

untitaker:
So if we `return 0` we should interrupt the chain?

untitaker:
TBH at this point I think this is too much of compatibility breakage to add, for very little benefit.

NiklasRosenstein:
> So if we return 0 we should interrupt the chain?

No, on non-zero return, non-zero sys.exit() or any other exception. Just as if the command chain was invoked via the shell `&&` operator, like `cli foo && cli bar && cli baz`. I don't think that behaviour would affect compatibility much. Alternatively, the `chain` parameter could support an `||` and `&&` mode. Making `||` the default (and mapping `chain=True` to `||`) would maintain full backwards compatibility except for the change of the return value to only return the result of the last command, which I deem a necessary and logical change.

untitaker:
The thing is that sys.exit(0) breaks the chain already.

On April 18, 2017 9:25:03 PM GMT+02:00, Niklas Rosenstein <notifications@github.com> wrote:
>> So if we return 0 we should interrupt the chain?
>
>No, on non-zero return, non-zero sys.exit() or any other exception.
>Just as if the command chain was invoked via the shell `&&` operator,
>like `cli foo && cli bar && cli baz`. I don't think that behaviour
>would affect compatibility much. Alternatively, the `chain` parameter
>could support an `||` and `&&` mode. Making `||` the default (and
>mapping `chain=True` to `||`) would maintain full backwards
>compatibility except for the change of the return value to only return
>the result of the last command, which I deem a necessary and logical
>change.
>
>-- 
>https://github.com/pallets/click/pull/765#issuecomment-294953301

-- 

NiklasRosenstein:
You can catch it as `SystemExit` and check if the exit code is any of `(0, None)` and dismiss it in that case if a nother command is to be executed in the chain.

untitaker:
Right, but at that point we would be breaking backwards compat because raising SystemExit already unconditionally interrupts the chain

On April 19, 2017 7:43:14 AM GMT+02:00, Niklas Rosenstein <notifications@github.com> wrote:
>You can catch it as `SystemExit` and check if the exit code is any of
>`(0, None)` and dismiss it in that case if a nother command is to be
>executed in the chain.
>
>-- 
>https://github.com/pallets/click/pull/765#issuecomment-295105275

-- 

untitaker:
I will close this because this behavior would be just too complex for the little value added. Sorry for noticing this problem so late.

Kamekameha:
I might be missing your point or something but what's wrong with simply raising a `ClickException` instance? Like you can look literally 4 lines bellow your change and see that `click` already catches `ClickException` exceptions and exit the interpreter with the exit code associated with the exception (i.e. with the `ClickException.exit_code` attribute), so maybe you could subclass and customize said exception to your needs and you could technically exit your CLI with virtually any exit code.

Bringing this up 'cause nobody mentioned this here nor in #270 when it's already in my opinion a viable approach. Also functions' return values are already used for chained commands that relies on `MultiCommand.resultcallback()` so basically any modification with regards to commands' return values is already an API breakage for `click` (see, for example, http://click.pocoo.org/6/commands/#multi-command-pipelines).

untitaker:
Your approach is entirely valid, but solves a different problem. This PR doesn't add new capabilities or enable a new usecase, it would've been a cosmetic change.

On April 21, 2017 5:58:35 AM GMT+02:00, Kamekameha <notifications@github.com> wrote:
>I might be missing your point or something but what's wrong with simply
>raising a `ClickException` instance? Like you can look literally 4
>lines bellow your change and see that `click` already catches
>`ClickException` exceptions and exit the interpreter with the exit code
>associated with the exception (i.e. with the `ClickException.exit_code`
>attribute), so maybe you could subclass and customize said exception to
>your needs and you could technically exit your CLI with virtually any
>exit code.
>
>Bringing this up 'cause nobody mentioned this here nor in #270 when
>it's already in my opinion a viable approach. Also functions' return
>values are already used for chained commands that relies on
>`MultiCommand.resultcallback()` so basically any modification with
>regards to commands' return values is already an API breakage for
>`click` (see, for example,
>http://click.pocoo.org/6/commands/#multi-command-pipelines).
>
>-- 
>You are receiving this because you modified the open/close state.
>https://github.com/pallets/click/pull/765#issuecomment-296049052

-- 

Kamekameha:
Oh, no no, my suggestion isn't trying to solve any problem at all, but neither is this PR. As you've noticed, this breaks chained multi-commands, and exit codes by function return value don't make that much sense for those either. How would you discern a valid return value of, for example, `1` from an actual exit code?

Just to be clear about my previous comment, what I meant is that return values on commands have already a documented behavior and design, and signaling errors is what I believe exceptions are for. To me, this looks more like a behavioral change rather than cosmetics.

untitaker:
I agree with you, this is why I closed this PR.

PR #771, opened 2017-04-29
Fix bash completion on chained command arguments

### LINKED ISSUE #754: Bash auto-complete of options not working properly in command chains
Example:

    $ cli cmd1 cmd2 --<TAB><TAB>

Auto-completes options of `cmd1`, not `cmd2`.

Observed in Python version 3.5.2 and click version 6.6.

### LINKED ISSUE #769: Dynamic bash autocompletion should escape blanks
Thanks for #755, wonderful feature! I found an issue when my completion callback returns entries that have (legitimate) blanks in them. In this case, entries get split into separate arguments, whereas I would have expected that blanks are escaped with a backslash (as bash does by default).

stopthatcow:
Enables bash completion on chained command arguments, allows completions with spaces in them, makes bash completion about 10ms faster by exiting python using os._exit() rather than sys.exit(), and updates documentation that was incorrect from PR #755

Fixes #754 
Fixes #769

### DIFF
--- a/click/_bashcomplete.py
+++ b/click/_bashcomplete.py
@@ -169,6 +169,11 @@ def get_choices(cli, prog_name, args, incomplete):
         # completion for any subcommands
         choices.extend(ctx.command.list_commands(ctx))
 
+    if not start_of_option(incomplete) and ctx.parent is not None and isinstance(ctx.parent.command, MultiCommand) and ctx.parent.command.chain:
+        # completion for chained commands
+        remaining_comands = set(ctx.parent.command.list_commands(ctx.parent))-set(ctx.parent.protected_args)
+        choices.extend(remaining_comands)
+
     for item in choices:
         if item.startswith(incomplete):
             yield item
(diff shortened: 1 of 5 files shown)


stopthatcow:
@Gig77 Feel free to see if this solves your issue.

Gig77:
It does, thanks a lot!

stopthatcow:
@untitaker I'd love to get your thoughts on this.

untitaker:
That's a bit crowded, I'm unsure if we want `fast_exit`. It potentially makes this a pain to test.

stopthatcow:
@untitaker good call. I broke this into #772, #773, #774, and #775. I think `fast_exit` in this specific case is ok because it is only used when bash completion is invoked (something we don't test directly). If we did want to test it we'd just need to launch the test using subprocess. Bzr actually uses this 'trick' in all exit code paths I believe to try to be more responsive as a CLI. I got the inspiration from [here](https://files.bemusement.org/talks/OSDC2008-FastPython/)

untitaker:
>we'd just need to launch the test using subprocess.

not a huge fan of this, but we'll see when we get to test it. I was thinking more about monkeypatching `fast_exit`.

PR #773, opened 2017-05-02
Allow spaces in bash completion results

### LINKED ISSUE #769: Dynamic bash autocompletion should escape blanks

stopthatcow:
Allows spaces in returns from bash autocompletion.

Fixes #769

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -7,6 +7,7 @@ Version 7.0
 -----------
 
 (upcoming release with new features, release date to be decided)
+- Added support for bash completions containing spaces. See #773.
 - Added support for dynamic bash completion from a user-supplied callback.
   See #755
 - Added support for bash completion of type=click.Choice for Options and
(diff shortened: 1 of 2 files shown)


untitaker:
I'm not entirely sure how this is related to #769, on the tin they look completely unrelated.

stopthatcow:
They are related. IFS needs to be set to only break apart completion results based on newline, rather than space as well. This allows spaces to be returned as a completion result. Previously bash would regard a completion result with spaces as multiple possible completions.

untitaker:
Ok, could you add a changelog entry for this as well (but keep it targeted at master)? Thanks!

stopthatcow:
Done. Thanks!

untitaker:
Thanks!

PR #775, opened 2017-05-02
Use fast_exit when invoking bash completion to save about 10ms.

stopthatcow:
Makes bash completion about 10ms faster by exiting python using os._exit() rather than sys.exit().


stopthatcow:
Some concerns have been raised about testability but I think `fast_exit` in this specific case is ok because it is only used when bash completion is invoked (something we don't test directly). If we did want to test it we'd just need to launch the test using subprocess. Bzr actually uses this 'trick' in all exit code paths I believe to try to be more responsive as a CLI. I got the inspiration from [here](https://files.bemusement.org/talks/OSDC2008-FastPython/)

untitaker:
I suppose we can use it as long as we don't actually test anything.

Issue #780, opened 2017-05-06
Tab completion behavior for click.File and click.Path types


gtristan:
Lets say this is related to #241 ...

Recently support was added to have argument autocompletion, custom completion functions/lists and a special case for click.Choice options too.

This is great ! I would like to see better support for filenames and directories, have been looking at the code and wanted to share my thoughts here... as far as I can see the problems with filename/directory are multiple but hopefully addressed with a couple of simple patches.

### Undesirable Autocompletion of paths

In the case that the user is completing an option which takes one or more arguments, and those arguments cannot be completed (i.e. they are just a string or integer type), then the builtin completion mechanism correctly reports no results, however the bash completion glue ([here](https://github.com/pallets/click/blob/master/click/_bashcomplete.py#L22)) specifies *"-o default"*

This parameter to the bash *complete* function tells bash to default to completing paths when the resulting COMPREPLY array is empty, and the side effect is undesirable filenames being completed.

### Filenames reported for directory arguments

Because click does not do anything for filenames or directories itself, it instead falls back on the undesirable *complete -o default* behavior mentioned above for filenames and all arguments which did not supply any custom autocompletion list or function.

This means that filenames are suggested when the argument is in fact a directory, and also directory names are suggested for filenames.

### Filename and directory completion has different behavior

The default shell behavior is to take the reported list in COMPREPLY, and if there is only one element, it will append a space and the next TAB is ready to complete the next token, but this is not the desired behavior when completing paths, where you want to complete only one directory at a time.

This can be addressed however by using the *"complete -o nospace ..."* option, which will just cause the shell to not append any space for a single available completion, allowing the program to continue completing the last completion on the next iteration.

In this case, click can:
* Append a space to the reported completion in the case that only one completion was available
* Append a slash (if none is already there), in the case that the completion is a directory name

### Proposed Approach

Instead of handling click.Choice explicitly in _bashcompletion.py, we should instead have the base ParamType cooperate in the process of completion, so that paths and filenames can have an opportunity to decide if the parameter is entirely complete or could be completed further (i.e. whether a space should be appended to the reported completion, or whether it should remain, as in the case for a filename which is a directory that could be completed further).

click.Choice would implement completions by overriding an abstract method of click.ParamType, click.BOOL could also implement this, because why not.

The new *autocompletion* attribute would be checked before ever consulting the click.ParamType abstract method, so the user can always override the default completions for any parameter.

[...]

PR #782, opened 2017-05-08
Enhancements to bash completions for better file/directory handling (take 2)

gtristan:
This addresses the issues outlined in #780 

* Changes the bash entry point to use *"complete -o nospace"*
* Adds abstract method to ParamType allowing the parameters to participate in the completion
* Migrates Choice completion implementation into types.py on the actual Choice type
* Implement the `completions()` method for BoolParamType, File and Path parameter types too

This behaves all around much better:
* No more completing of filenames when that is undesirable
* Completes filenames or directories selectively, depending on the parameter

**Note**: This is take two of pull request #781 ... I rebased my branch to have a more coherent history and did not expect github to delete the PR as a result, sorry.

### DIFF
--- a/click/_bashcomplete.py
+++ b/click/_bashcomplete.py
@@ -19,7 +19,7 @@ COMPLETION_SCRIPT = '''
     return 0
 }
 
-complete -F %(complete_func)s -o default %(script_names)s
+complete -F %(complete_func)s -o nospace %(script_names)s
 '''
 
 _invalid_ident_char_re = re.compile(r'[^a-zA-Z0-9_]')
--- a/click/_bashcomplete.py
+++ b/click/_bashcomplete.py
@@ -109,14 +109,12 @@ def get_user_autocompletions(ctx, args, incomplete, cmd_param):
     :param cmd_param: command definition
     :return: all the possible user-specified completions for the param
     """
-    if isinstance(cmd_param.type, Choice):
-        return cmd_param.type.choices
-    elif cmd_param.autocompletion is not None:
+    if cmd_param.autocompletion is not None:
         return cmd_param.autocompletion(ctx=ctx,
                                         args=args,
                                         incomplete=incomplete)
     else:
-        return []
+        return cmd_param.type.completions(incomplete) or []
 
 def get_choices(cli, prog_name, args, incomplete):
     """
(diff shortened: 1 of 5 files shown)


gtristan:
Addressed some corner cases regarding path name completions in some follow up commits instead of squashing those and rebasing the branch, as that seems to cause a merge request to be automatically closed.

I can rebase the branch locally so that these changes are squashed into the one commit implementing completions for file & directory paths, if that is desirable.

davidism:
Completion has changed significantly since this PR was posted. Would you or @stopthatcow review this to see if it's still needed and can be rebased?

stopthatcow:
There are definitely some attractive aspects of this PR. I like the idea of delegating some completion to the type. I can look into rebasing it on top of master this weekend.

gtristan:
> There are definitely some attractive aspects of this PR. I like the idea of delegating some completion to the type, and path completion would be very nice to have. I can look into rebasing it on top of master this weekend, unless of course @gtristan wants to take a stab at it himself!

It's great to see some activity on this front !

I would certainly love to upstream this stuff, as we're doing some [pretty ugly hacks](https://gitlab.com/BuildStream/buildstream/blob/master/buildstream/_frontend/cli.py#L165) which [reimplement the completions](https://gitlab.com/BuildStream/buildstream/blob/master/buildstream/_frontend/complete.py) mechanism in order to get these features in BuildStream.

The most important part for us is the ability to perform *partial completions*, since we do completions of filenames which are intentionally not in the `CWD`, and soon we will be doing completions of entity names within a database/cache, these entities are also *"pathy"*. If partial completions are not supported, then it is rather unacceptably slow (one would need to recursively list every leaf in an unpredictably large tree in order to report a full completion list in one go; for every `<TAB>`).

As far as I know, this can only be done by enabling `-o nospace` in the bash completion script; so that the program handling completions has the decision to append a space (fully complete) or not (partially complete, next `<TAB>` will show deeper results).

I'll have to refresh my memory on how compatible my proposed changes were, but I'm pretty confident that by delegating completions to the type and to delegated user functions; we can make sure that the behavior of allowing partial completions is *opt in*.

jsmedmar:
@gtristan Kudos to great work! Hope you can figure out this last bit so that we can see file completions in master. Again thanks so much for such thoughtful work in #780 and here.

davidism:
@gtristan thanks for working on this. I'm going to close it in favor of the updated work in #1403.

gtristan:
@davidism Thanks for reviewing and for everyone who picked up my abandoned patches :)

I will try to find time to upgrade click in BuildStream and drop our custom code in favor of this :)

PR #787, opened 2017-05-15
Don't arbitrarily strip an argument from make_pass_decorator

### LINKED ISSUE #786: make_pass_decorator strips argument
The inner most wrapper of make_pass_decorator invokes a function with `*args[1:]` which strips an argument.

A smaller example could be made but the following is the [invoking other commands](http://click.pocoo.org/6/advanced/#invoking-other-commands) example from the docs modified to also take in a custom config object.

The only difference between `dist_working` and `dist_broken` is the order of the decorators and consequently the parameters they take.

In my opinion they should both be valid and work.  What happens in `dist_broken` is, like the title says, the `make_pass_decorator` wrapper invokes the function with `*args[1:]` which removes the context object put there by the previous decorator.

```python
import click


class MyConfig(object):
    def __init__(self, verbose):
        self.verbose = verbose


pass_config = click.make_pass_decorator(MyConfig)


@click.group()
@click.option('--verbose', '-v', is_flag=True)
@click.pass_context
def cli(ctx, verbose):
    ctx.obj = MyConfig(verbose=verbose)


@cli.command()
@click.option('--count', default=1)
def test(count):
    click.echo('Count: %d' % count)


@cli.command('dist-working')
@click.option('--count', default=1)
@pass_config
@click.pass_context
def dist_working(ctx, cfg, count):
    if cfg.verbose:
        click.echo('about to forward')
    ctx.forward(test)
    if cfg.verbose:
        click.echo('about to invoke')
    ctx.invoke(test, count=42)


@cli.command('dist-broken')
@click.option('--count', default=1)
@click.pass_context
@pass_config
def dist_broken(cfg, ctx, count):
    if cfg.verbose:
        click.echo('about to forward')
    ctx.forward(test)
    if cfg.verbose:
        click.echo('about to invoke')
    ctx.invoke(test, count=42)
```

ericfrederich:
This fixes #786

### DIFF
--- a/click/decorators.py
+++ b/click/decorators.py
@@ -61,7 +61,7 @@ def make_pass_decorator(object_type, ensure=False):
                 raise RuntimeError('Managed to invoke callback without a '
                                    'context object of type %r existing'
                                    % object_type.__name__)
-            return ctx.invoke(f, obj, *args[1:], **kwargs)
+            return ctx.invoke(f, obj, *args, **kwargs)
         return update_wrapper(new_func, f)
     return decorator
 
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -203,3 +203,39 @@ def test_close_before_pop(runner):
     assert not result.exception
     assert result.output == 'aha!\n'
     assert called == [True]
+
+
+def test_make_pass_decorator_args(runner):
+    """
+    Test to check that make_pass_decorator doesn't consume arguments based on
+    invocation order.
+    """
+    class Foo(object):
+        title = 'foocmd'
+
+    pass_foo = click.make_pass_decorator(Foo)
+
+    @click.group()
+    @click.pass_context
+    def cli(ctx):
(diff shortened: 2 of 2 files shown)


ericfrederich:
I'd like this merged.  Should this be based on `master` or `6.x-maintenance`?
When I rebase I'll also write a test.

dsully:
@ericfrederich Could you turn the example in #786 into unit tests?

This is a very subtle change, so it'd be good to ensure it is covered.

Thanks

sirosen:
As this PR is fairly old, I took the liberty of rebasing it and adding the requested test.
Apologies if people feel like this is overstepping, but it's all meant in good faith.

davidism:
This seems correct, but I'd really like to know why it was stripping one argument, which would only come up when using `pass_context` and `make_pass_decorator` together.

Since neither `pass_context` nor `pass_obj` strip an argument, I'll merge this.

PR #794, opened 2017-05-18
Bug fix #793; use deterministic option name; can't rely on list.sort

### LINKED ISSUE #793: Incorrect names being picked
Docs state `By default, the name of the parameter is the first long option defined; otherwise the first short one is used`.

The implementation of this is in `_parse_decls` method of `Option` which has a line...

```python
possible_names.sort(key=lambda x: len(x[0]))
name = possible_names[-1][1].replace('-', '_').lower()
```

... currently [here](https://github.com/pallets/click/blob/master/click/core.py#L1570)

This invocation of `list.sort` with `key=` essentially makes every short option equal to each other and every long option equal to each other.  The original order they were defined in is subject to change.

In fact... even if it were to maintain that order, the code would actually be grabbing the __*last*__ long parameter.

Fix coming soon... I'll submit it.  I'll change the code to match the docs.

ericfrederich:
Please merge; this matches the implementation to the documentation.

Currently an option defined as...
```python
@click.option('--ambitious', '--all', '-a', is_flag=True)
```
...will target `all` instead of `ambitious`

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -13,6 +13,7 @@ Version 6.8
 - Fix bug in test runner when calling ``sys.exit`` with ``None``. See #739.
 - Fix crash on Windows console, see #744.
 - Fix bashcompletion on chained commands. See #754.
+- Fix option naming routine to match documentation.  See #793
 
 Version 6.7
 -----------
--- a/click/core.py
+++ b/click/core.py
@@ -1548,8 +1548,8 @@ class Option(Parameter):
                     opts.append(decl)
 
         if name is None and possible_names:
-            possible_names.sort(key=lambda x: len(x[0]))
-            name = possible_names[-1][1].replace('-', '_').lower()
+            possible_names.sort(key=lambda x: -len(x[0]))  # group long options first
+            name = possible_names[0][1].replace('-', '_').lower()
             if not isidentifier(name):
                 name = None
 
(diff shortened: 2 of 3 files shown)


untitaker:
Thanks, please:

* retarget at 6.x-maintenance branch
* Write a changelog and testcase

Also I think the fix you have in mind can have a more minimal diff. In particular all that needs to be done is to replace `-1` with `0` I think.

ericfrederich:
@untitaker taking 0 would take from the beginning of the list which would be one of the short options.

I looked into it and it appears `sorted()` and `list.sort()` are documented as __*stable*__ in that *it guarantees not to change the relative order of elements that compare equal*.  So because this is documented, and not just an implementation detail, it is okay to rely on.  We don't have to do comprehension like the current pull request is showing.

... but it's not as simple as taking 0 either.  I'll amend this commit with one with a smaller diff.

ericfrederich:
Hows that for a __*more minimal diff*__?  If not for the comment the size of the code would remain unchanged ;-)... just moved the minus sign up a line and changed a 1 to a 0 😄

untitaker:
That looks really good, please add a changelog + test though

ericfrederich:
Okay @untitaker Got the changelog and test as well.

ericfrederich:
@untitaker whenever this is merged I'd like to create another pull request adding more testing and documentation around the explicit naming of options.  I'd prefer to keep them as separate commits and base it on this one.

untitaker:
Thank you very much! Feel free to open a new PR regarding new tests (preferrably targeted at master).

ericfrederich:
@untitaker I created #798 and #799

PR #799, opened 2017-05-22
Option naming: add test and documentation for existing functionality

### LINKED ISSUE #798: Missing tests and docs for existing functionality
Explicit option naming is not documented or tested for.  A lot of [Python keywords](https://docs.python.org/2/reference/lexical_analysis.html#keywords) make for great option names.  To do this you must explicitly choose another param name (or have your command accept `**kwargs`)

ericfrederich:
This fixes #798


untitaker:
Great, thanks!

PR #804, opened 2017-05-30
added test and code

iluxame:
I've rebased https://github.com/pallets/click/pull/502
@jdell64 Thanks for the feature! We need this feature badly so I've rebased your PR. No worries, you are still the author of your code.
Feel free to rebase your code and submit it in the original PR.
In such case we will not need this PR

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -762,8 +762,8 @@ class Command(BaseCommand):
 
     def __init__(self, name, context_settings=None, callback=None,
                  params=None, help=None, epilog=None, short_help=None,
-                 options_metavar='[OPTIONS]', add_help_option=True, hidden,
-                 no_args_is_help=None):
+                 options_metavar='[OPTIONS]', add_help_option=True,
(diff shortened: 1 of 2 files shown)


iluxame:
@untitaker can you please take a look?

untitaker:
@iluxame I have proposed a few alternatives in that thread which I consider better since they add no extra API, but haven't recieved any feedback.

iluxame:
@untitaker Aha, sorry, somehow I missed it.
In my opinion its better to show help message with empty command.
in case of a sentence like "See command subcommand --help" user will need 2 commands instead of one. However from other point of view it will be clear to user that he didn't entered any meaningful input.

This patch actually aligns command user experience with group user experience. It seems that similar experience is less confusing. So I would prefer @jdell64 approach.

iluxame:
@untitaker 
What do you think?

davidism:
Decision made to keep current "see --help" message in #502.

PR #806, opened 2017-06-01
Fix overzealous completion when required options/arguments are being completed

### LINKED ISSUE #790: Autocomplete should not complete subcommands while expecting required argument or option
Suppose my `cli` has two subcommands, `sub1` and `sub2`. `sub1` has an option `-o` that requires a value. This option has an autocomplete function assigned to it. 

When I enter

`$ cli sub1 -o <tab><tab>
`

autocomplete not only suggests my allowed values for `-o`, but also `sub2`. However, since an option value is required after `-o`, `sub2` is an invalid suggestion at this point. 

The same happens with required arguments.

Is there a possibility to suppress autocompletion of subcommands while still in the context of required arguments or options?

Python version is `Python 3.5.2 :: Anaconda 4.2.0 (64-bit)` and Click version is `7.0-dev`.

stopthatcow:
Don't offer bash completions for subcommands or chained commands when completing a required argument or option.
Fixes #790

### DIFF
--- a/tests/test_bashcomplete.py
+++ b/tests/test_bashcomplete.py
@@ -86,13 +86,20 @@ def test_long_chain():
     COLORS = ['red', 'green', 'blue']
     def get_colors(ctx, args, incomplete):
         for c in COLORS:
-            yield c
+            if c.startswith(incomplete):
+                yield c
+
+    def search_colors(ctx, args, incomplete):
+        for c in COLORS:
+            if incomplete in c:
+                yield c
 
     CSUB_OPT_CHOICES = ['foo', 'bar']
     CSUB_CHOICES = ['bar', 'baz']
     @bsub.command('csub')
     @click.option('--csub-opt', type=click.Choice(CSUB_OPT_CHOICES))
     @click.option('--csub', type=click.Choice(CSUB_CHOICES))
+    @click.option('--search-color', autocompletion=search_colors)
     @click.argument('color', autocompletion=get_colors)
     def csub(csub_opt, color):
         pass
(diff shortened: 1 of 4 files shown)


stopthatcow:
@Gig77 feel free to comment and suggest additional test cases.

Gig77:
Just tested it on my CLI and it works. Thanks a lot!

stopthatcow:
@untitaker could you take a look?

tests/test_bashcomplete.py
```diff
@@ -125,15 +125,16 @@ def asub(asub_opt):
...
-    @click.argument('arg', type=click.Choice(['arg1', 'arg2']))
+    @click.argument('arg', type=click.Choice(['arg1', 'arg2']), required=True)
     def bsub(bsub_opt, arg):
         pass
 
     assert list(get_choices(cli, 'lol', [], '-')) == ['--cli-opt']
     assert list(get_choices(cli, 'lol', [], '')) == ['asub', 'bsub']
     assert list(get_choices(cli, 'lol', ['asub'], '-')) == ['--asub-opt']
     assert list(get_choices(cli, 'lol', ['asub'], '')) == ['bsub']
-    assert list(get_choices(cli, 'lol', ['bsub'], '')) == ['arg1', 'arg2', 'asub']
```
untitaker: Since this is a chained group, I don't see a reason why it would be zealous in any way to suggest another subcommand in this situation?

stopthatcow: In general yes, though note that bsub *requires* the "arg" argument, so we know that the smart thing to do is to complete that before offering subcommands.

stopthatcow: What do you think?

stopthatcow: @untitaker any thoughs on this?

stopthatcow: @untitaker ?

untitaker: I'm happy to merge this if you add a test that has optional arguments, for that case the other subcommands shouldn't be excluded.

Also please try to split up `get_choices`.

stopthatcow: Ok, I added a test case and simplified get_choices. After this goes in I also have a PR [here](https://github.com/pallets/click/pull/865) to add native ZSH completion support.

untitaker:
Awesome thanks!

PR #812, opened 2017-06-27
Fixed URL with python3 unicode error description

mliubchych:
Originally this string looks like `http://click.pocoo.org/python3/for` (without space between actual link and "for mitigation steps").

I hope this PR's so small that test running isn't necessary.


nlyubchich:
Seems that small fixes aren't supposed to be merged (because of half-year silence).

Closing PR.

PR #830, opened 2017-07-12
Wrap stdout/stderr to avoid "Not enough space" in Python 2 on Windows 7

### LINKED ISSUE #825: "IOError: [Errno 12] Not enough space" on Windows 7 and Python 2
This is caused by [click/_winconsole.py:268-272](https://github.com/pallets/click/blob/5701c4000c39b5a90c45f1083c0cbf4e00731e81/click/_winconsole.py#L268-L272). It seems that switching the standard streams to binary mode reveals the underlying limit of the Windows console of limiting a single write to 64K (I think this limit was lifted in Windows 8). This affects `print`, `sys.stdout.write` and `click.echo`.

It seems that many libraries and Python itself limit a single write to 32767 bytes to workaround this. Python 3 for example: [Python/fileutils.c:1244-1250](https://github.com/python/cpython/blob/f7eae0adfcd4c50034281b2c69f461b43b68db84/Python/fileutils.c#L1244-L1250). It doesn't seem like Python 2 has this hack though.

We will have to either:
1. Don't set the streams to binary and read/write to dupped stream fds (and file objects) that are set to binary.
2. Wrap the standard streams and workaround this.
3. Submit a patch to Python 2 which I'm not sure will ever be accepted.
4. Other?

To reproduce:
```py
import sys
import click

click.echo("spam")

print('A' * 100000) # or
sys.stdout.write('A' * 100000) # or
click.echo('A' * 100000)
```

Windows 7 SP1 x64
Python: 2.7.10
Click: 6.7

segevfiner:
When you set the standard streams to binary you seem to only be able to write up to some limit in a single call to write. So we wrap the streams and write in chunks to avoid this.

A similar hack was also implemented in Python 3 to work around this. But that hack doesn't seem to exist in Python 2.

This might have unintended side effects due to replacing the standard stream objects. See the issue for other possible fixes.

I didn't test that this really fixes the issue yet since I don't have access to Windows 7 currently, if anyone could test I will be glad. Just do (with and without this fix):
```py
import click
click.echo('A' * 100000)
```

Fixes #825

### DIFF
--- a/click/_winconsole.py
+++ b/click/_winconsole.py
@@ -201,7 +201,7 @@ class ConsoleStream(object):
         )
 
 
-class StreamWrapper(object):
+class WindowsChunkedWriter(object):
     """
     Wraps a stream (such as stdout), acting as a transparent proxy for all
     attribute access apart from method 'write()' which we wrap to write in
@@ -225,6 +225,16 @@ class StreamWrapper(object):
             written += to_write
 
 
+_wrapped_std_streams = set()
+
+
+def _wrap_std_stream(name):
+    # Python 2 & Windows 7 and below
+    if PY2 and sys.getwindowsversion()[:2] <= (6, 1) and name not in _wrapped_std_streams:
+        setattr(sys, name, WindowsChunkedWriter(getattr(sys, name)))
+        _wrapped_std_streams.add(name)
+
+
 def _get_text_stdin(buffer_stream):
     text_stream = _NonClosingTextIOWrapper(
         io.BufferedReader(_WindowsConsoleReader(STDIN_HANDLE)),
@@ -233,10 +243,6 @@ def _get_text_stdin(buffer_stream):
 
 
 def _get_text_stdout(buffer_stream):
-    if PY2:
-        buffer_stream = StreamWrapper(buffer_stream)
-        sys.stdout = StreamWrapper(sys.stdout)
-
     text_stream = _NonClosingTextIOWrapper(
         io.BufferedWriter(_WindowsConsoleWriter(STDOUT_HANDLE)),
         'utf-16-le', 'strict', line_buffering=True)
@@ -244,10 +250,6 @@ def _get_text_stdout(buffer_stream):
 
 
 def _get_text_stderr(buffer_stream):
-    if PY2:
-        buffer_stream = StreamWrapper(buffer_stream)
-        sys.stderr = StreamWrapper(sys.stderr)
-
     text_stream = _NonClosingTextIOWrapper(
         io.BufferedWriter(_WindowsConsoleWriter(STDERR_HANDLE)),
         'utf-16-le', 'strict', line_buffering=True)
--- a/click/_winconsole.py
+++ b/click/_winconsole.py
@@ -201,6 +201,30 @@ class ConsoleStream(object):
         )
 
 
+class StreamWrapper(object):
+    """
+    Wraps a stream (such as stdout), acting as a transparent proxy for all
+    attribute access apart from method 'write()' which we wrap to write in
+    limited chunks due to a Windows limitation on binary console streams.
+    """
+    def __init__(self, wrapped):
+        # double-underscore everything to prevent clashes with names of
+        # attributes on the wrapped stream object.
+        self.__wrapped = wrapped
+
+    def __getattr__(self, name):
+        return getattr(self.__wrapped, name)
+
+    def write(self, text):
+        total_to_write = len(text)
+        written = 0
+
+        while written < total_to_write:
+            to_write = min(total_to_write - written, MAX_BYTES_WRITTEN)
+            self.__wrapped.write(text[written:written+to_write])
+            written += to_write
+
+
 def _get_text_stdin(buffer_stream):
     text_stream = _NonClosingTextIOWrapper(
         io.BufferedReader(_WindowsConsoleReader(STDIN_HANDLE)),
@@ -209,6 +233,10 @@ def _get_text_stdin(buffer_stream):
 
 
 def _get_text_stdout(buffer_stream):
+    if PY2:
+        buffer_stream = StreamWrapper(buffer_stream)
+        sys.stdout = StreamWrapper(sys.stdout)
+
     text_stream = _NonClosingTextIOWrapper(
         io.BufferedWriter(_WindowsConsoleWriter(STDOUT_HANDLE)),
         'utf-16-le', 'strict', line_buffering=True)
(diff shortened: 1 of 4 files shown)


davidism:
I was about to merge this, then I realized I just merged #819 which fixed a 16k character limit on Windows. I want to make sure I understand what the difference between the 16k limit there and the 64k limit here is, since I'm not very familiar with Windows. It appears you're implementing a buffered writer, which was handled in the other PR by just using `io.BufferedWriter`.

davidism:
I guess the immediate difference is that the other one was for the text stream, and this is the binary stream, but why not use `io.BufferedWriter` here?

segevfiner:
> I was about to merge this, then I realized I just merged #819 which fixed a 16k character limit on Windows. I want to make sure I understand what the difference between the 16k limit there and the 64k limit here is, since I'm not very familiar with Windows. It appears you're implementing a buffered writer, which was handled in the other PR by just using `io.BufferedWriter`.

The other PR was due to `TextIOWrapper` requiring a buffered stream and it not handling partial writes. 

This one is due to some Windows versions (e.g. Windows 7) failing to write to the console past a certain size. `StreamWrapper` isn't exactly like `BufferedWriter` as it does no buffering, it merely makes multiple calls to write so as to avoid the arbitrary limit in writing to the console.

But now that I think about it, if the `BufferedWriter` doesn't ever write more than the limit in one call, it's possible that this is redundant, but I have no Windows 7 installation available to test this.

davidism:
You can get a VM for testing from MS at https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/. I can try this out later if you can't get to it before me.

davidism:
I confirmed that this still happens on Windows 7 with the previous PR, and is fixed by this PR. But I'm confused why `BufferedWriter` didn't take care of it.

click/_winconsole.py
```diff
@@ -209,13 +233,21 @@ def _get_text_stdin(buffer_stream):
 
 
 def _get_text_stdout(buffer_stream):
+    if PY2:
+        buffer_stream = StreamWrapper(buffer_stream)
+        sys.stdout = StreamWrapper(sys.stdout)
```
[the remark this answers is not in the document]
segevfiner: The issue here is triggered by calling `_setmode` which also affects `sys.stdout` and `sys.stderr`.

The user code might still use a normal print or just write to them directly. Which makes it necessary to wrap `sys.stdout` and `sys.stderr`. Alternatively it's possible to use a duplicate/separate file descriptor for the unicode streams, but that can lead to issues with internal buffering in the CRT.

segevfiner:
> I confirmed that this still happens on Windows 7 with the previous PR, and is fixed by this PR. But I'm confused why `BufferedWriter` didn't take care of it.

Reading the code again more carefully. `ConsoleStream` has *two* streams. One for text, and another one for anything else. There is no issue with the text stream because `_WindowsConsoleWriter` already has the limit in place. The other stream, on the other hand, is just the original object and will happily write whatever size you give it.

click/_winconsole.py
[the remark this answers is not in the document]
davidism: `buffer_stream` will always be `sys.stdout`, so we should be able to assign the same wrapper to both, right?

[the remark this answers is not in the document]
davidism: `_compat.get_binary_stream` also calls `setmode`, might need to think more about where this patch happens.

davidism: `buffer_stream` will always be `sys.stdout`, so we should be able to assign the same wrapper to both, right?

davidism: `_compat.get_binary_stream` also calls `setmode`, might need to think more about where this patch happens.

davidism: Is there a way to do this only for Windows versions that need it? Tested this on a Windows 10 VM and it wasn't necessary.

click/_winconsole.py
```diff
@@ -201,6 +201,30 @@ def __repr__(self):
         )
 
 
+class StreamWrapper(object):
```
davidism: This should have a more descriptive name.

segevfiner:
> 
> 
> It seems like we should just be able to wrap the original stream with a buffered writer as well. I tried
> 
> ```python
> if PY2:
>     buffer_stream = io.BufferedWriter(io.FileIO(buffer_stream, "wb"))
> ```
> 
> but that didn't seem to fix it.

I'd guess than, that when passed a large buffer, `BufferedWriter` just writes it to the file, without splitting it up to smaller writes.

click/_winconsole.py
segevfiner: Wrapping `sys.{stdout,stderr}` should probably be moved elsewhere actually... I think this can be called multiple times? Wrapping `sys.{stdout,stderr}` should be done only once. But I'm not sure where. It starts to become necessary to do so once you call `setmode(..., O_BINARY)`. And if this can be called multiple times it's possible `buffer_stream` will already be a `StreamWrapper`...

This needs some place that will only be called once or otherwise we should remember if we already applied the patch.

segevfiner: Yeah. Probably need to rethink where to apply the patch. See https://github.com/pallets/click/pull/830/files#r219271929.

segevfiner: [`sys.getwindowsversion`](https://docs.python.org/2.7/library/sys.html#sys.getwindowsversion) should do the trick. Assuming we don't need to detect Windows 8.1 (See [GetVersionEx](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724451(v=vs.85).aspx), because Microsoft, never understood why they did that). I guess the underlying limitation was fixed by Windows 8 condrv. Need to test if that's the right cutoff.

click/_winconsole.py
segevfiner: Anything in mind? How about `WindowsConsoleStreamWrapper`?

davidism: `WindowsChunkedWriter`?

click/_winconsole.py
davidism: I think multiple calls are either cached or exit early, but I'd have to check again.

Maybe just extract the `O_BINARY` part to a common function that checks `isinstance(stream, StreamWrapper)`.

davidism: I'm downloading an 8.1 VM (8 isn't available from MS), I can report on this later today.

davidism: The issue does not occur in 8.1 or 10.

davidism:
This is the last issue marked for the 7.0 milestone. It would be great to get this done over the weekend. Let me know if you need anything.

click/_winconsole.py
segevfiner: `isinstance(stream, StreamWrapper)` will probably interact badly with other libraries that touch `sys.std*`. Calling `get_text_stream` from the utils module probably runs this again.

There are two places we call `setmode`: In `_get_windows_console_stream` here, and in `set_binary_mode` in _compat.py. But those only get the file object and won't know which global to patch. They are only called by functions in _compat.py which are per std stream. I can add a function, and call it from these places, which get's the name of the stream and if it is the right Windows version, Python version and we didn''t patch that stream already (Module level variable), then it will apply the patch.

segevfiner: So I'd bet it also doesn't happen in 8 and is likely fixed by condrv that was added in 8.

segevfiner:
> 
> 
> This is the last issue marked for the 7.0 milestone. It would be great to get this done over the weekend. Let me know if you need anything.

I reworked the PR and checked that it works on Windows 7.

PR #842, opened 2017-08-17
add bool conversions for t and f

axu2:
Just thought this would be nice. Love Click!


untitaker:
Thanks, could you add a changelog entry and perhaps a test?

axu2:
Would this go under 6.8 or 7.0?

untitaker:
Sorry for the slow response, you made the right choice. Could you add a small test too? There is a test for bool options in `test_basic.py`.

axu2:
How's that?

untitaker:
Excellent, thanks!

PR #853, opened 2017-09-25
minor: do not wrap exception messages

blueyed:
Apart from that: could this be supported?
It is useful with a custom `callback` that should get applied always (even for the default).


JosiahDub:
There are a lot more wrapped exception methods that would have to be changed along with this one.

PR #859, opened 2017-09-28
Add Executable param type

### LINKED ISSUE #856: [RFC] new Executable param type
It would be derived from `Path` (which would get a new flag `executable`), and then `Executable` would look for if a command exists in `$PATH` - using `shutil.which` (https://github.com/python/cpython/blob/3.6/Lib/shutil.py#L1093), which would need to be vendored then for older Python versions.

This would allow for validation of params that are then meant to be run, e.g. via `subprocess.call`.

blueyed:
Fixes https://github.com/pallets/click/issues/856.

### DIFF
--- a/click/types.py
+++ b/click/types.py
@@ -395,9 +395,10 @@ class Path(ParamType):
     :param dir_okay: controls if a directory is a possible value.
     :param writable: if true, a writable check is performed.
     :param readable: if true, a readable check is performed.
+    :param executable: if true, a executable check is performed.
     :param resolve_path: if this is true, then the path is fully resolved
                          before the value is passed onwards.  This means
-                         that it's absolute and symlinks are resolved.  It
+                         that it is absolute and symlinks are resolved.  It
                          will not expand a tilde-prefix, as this is
                          supposed to be done by the shell only.
     :param allow_dash: If this is set to `True`, a single dash to indicate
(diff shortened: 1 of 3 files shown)


click/types.py
```diff
@@ -479,10 +481,51 @@ def convert(self, value, param, ctx):
...
+    def __init__(self, absolute_path=True, resolve_path=False, writable=False,
+                 readable=False):
+        self.absolute_path = absolute_path
+        super(Executable, self).__init__(
+            exists=True, file_okay=True, dir_okay=False,
+            writable=writable, readable=readable, executable=True,
+            resolve_path=False, allow_dash=False, path_type=None)
+
+    def convert(self, value, param, ctx):
+        from shutil import which
```
ThiefMaster: AFAIK `shutil.which()` only exists in recent Python 3 versions.

click/types.py
```diff
@@ -479,10 +481,51 @@ def convert(self, value, param, ctx):
...
+
+    For more information see :ref:`tuple-type`.
+
+    :param absolute_path: use the absolute (but not necessarily resolved path).
+    :param resolve_path: if this is true, then the path is fully resolved
+                         before the value is passed onwards.  This means
+                         that it is absolute and symlinks are resolved.  It
+                         will not expand a tilde-prefix, as this is
+                         supposed to be done by the shell only.
+    :param writable: if true, a writable check is performed.
```
ThiefMaster: is this really useful here?

blueyed: well..  I can imagine that you want to look for a readable/writable exe.

click/types.py
blueyed: yes, would need to be vendored, see https://github.com/pallets/click/issues/856.

click/types.py
ThiefMaster: hrm at least i've never seen any tool that looks up something from $PATH except to execute it - which needs neither of these permissions

blueyed: Yeah.. but I've wanted for this to be generic.. you might want to look for an exe "true" to overwrite then.  Corner case, but it made sense.  The defaults should reflect what you expect from it though.

click/types.py
blueyed: Should I do this?
I would add it to click/_compat.py, as `which`?!

davidism:
Thanks for working on this! However, I don't think I'll merge it for now. It's fairly niche, the number of checks available to `Path` are getting out of hand, and this check can be done really easily in a callback if needed.

PR #860, opened 2017-09-29
Pass ctx to NoSuchOption exceptions

gar1t:
This is a naive change without tests or any real understanding of the implications, but it seems correct on the surface.

To provide the user with better help (e.g. "Try 'CMD --help' for help.") the context needs to come along with the exception, unless there's another approach I'm missing.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -51,6 +51,7 @@ Unreleased
 -   Update doc to match arg name for ``path_type``. (`#801`_)
 -   Raw strings added so correct escaping occurs. (`#807`_)
 -   Add bool conversion for "t" and "f". (`#842`_)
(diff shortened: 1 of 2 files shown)


click/formatting.py
```diff
@@ -170,7 +170,8 @@ def write_text(self, text):
                              preserve_paragraphs=True))
         self.write('\n')
 
-    def write_dl(self, rows, col_max=30, col_spacing=2):
+    def write_dl(self, rows, col_max=30, col_spacing=2,
+                 preserve_paragraphs=False):
```
dsully: Can you add :param: docstring for this new kwarg?

Thanks

sirosen:
Doing another pass over old PRs, I realized that this was carrying a bunch of unnecessary cruft (which made the builds fail) and that the change to the help formatter wasn't even necessary.

I've squashed it down to one commit and pushed. I still have the old commits tagged and handy if someone wants the old version of this branch restored.

PR #861, opened 2017-09-29
Fix some issues reported by flake8

blueyed:


davidism:
Thanks, but we don't do large scale style changes like this. If you'd like to address style around code being edited, that's fine, but otherwise it's too easy to affect other PR merges.

PR #863, opened 2017-10-01
ProgressBar should use stderr by default

georgeyk:
A minor change (I think), but it makes sense when we have commands that we want to pipe or redirect the "real" output, not the  progressbar.

If accepted I'll add a note in documentation.
I'm not sure if this change is testable, but I can try to add a test too.

I saw other PR targeting master branch instead of the maintenance branches pointed in the contributing guideline, tell me if I need to change the target branch.

### DIFF
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -192,43 +192,40 @@ class ProgressBar(object):
 
     def render_progress(self):
         from .termui import get_terminal_size
-        nl = False
 
         if self.is_hidden:
-            buf = [self.label]
-            nl = True
-        else:
-            buf = []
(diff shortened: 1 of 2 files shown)


blueyed (commented):
Makes total sense IMHO, but should have a test indeed.

OTOH, shouldn't it just get skipped maybe in case of not being a TTY in the first place?

georgeyk:
@blueyed yes, maybe just changing the logic here: https://github.com/pallets/click/blob/master/click/_termui_impl.py#L197-L199

It should skip the label or just `return` I guess. I'll try to write a test and validate this idea.

PR #865, opened 2017-10-10
Add native ZSH completion script

### LINKED ISSUE #323: Autocomplete support for zsh
Bash autocompletion works well, but I would really like zsh to work as well

stopthatcow:
Add native ZSH completion script. This allows completion in ZSH without the need for `autoload bashcompinit` and `bashcompinit` Fixes #323

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -33,6 +33,7 @@ Version 7.0
   See #790.
 - Allow autocompletion function to determine whether or not to return
   completions that start with the incomplete argument.
+- Add native ZSH autocompletion support See #323.
 - Subcommands that are named by the function now automatically have the
   underscore replaced with a dash.  So if you register a function named
   `my_command` it becomes `my-command` in the command line interface.
--- a/click/_bashcomplete.py
+++ b/click/_bashcomplete.py
@@ -10,7 +10,7 @@ from .types import Choice
 
 WORDBREAK = '='
 
-COMPLETION_SCRIPT = '''
+COMPLETION_SCRIPT_BASH = '''
 %(complete_func)s() {
     local IFS=$'\n'
     COMPREPLY=( $( env COMP_WORDS="${COMP_WORDS[*]}" \\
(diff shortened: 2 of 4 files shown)


stopthatcow:
@mitsuhiko or @untitaker fell free to take a look.

stopthatcow:
@untitaker any thoughts?

pseyfert:
well, as variation of my suggestion one can avoid creating `_click_CI` and put

```
_click_CI() {
    emulate -L zsh
    local IFS=$'\n'
    local completions=( $( env COMP_WORDS="${words[*]}" \\
                        COMP_CWORD=$((CURRENT-1)) \\
                        SHELL_COMPLETE="complete" \\
                        ${words[1]} ) )
    compadd -M 'r:|=* l:|=* r:|=*' -a -- completions 
}
```

into the `.zshrc`

BUT, for both bash and zsh, defining completion functions in individual files
is pretty common (see `/usr/share/bash-completion/completions` or
`/usr/share/zsh/4.3.11/functions`(lxplus),  which is why I think collecting
them in the `$fpath`, rather than spreading between `$fpath` and rc-files makes
sense.

That aside, I agree, not calling the CI for every tab press makes sense, BUT
that happens in click already for bash, so that would be a big-ish rework of
the existing functionality.


Not sure what happens in openshift (never used, seems I don't have it installed
anywhere to check) but running the CI on every shell startup, as through
`source <(oc completion zsh)` in the `.zshrc`, might actually cause more CI
calls than upon every TAB stroke in a line containing `oc `.

lbenka:
I tried it and it works :)

eli-b:
It looks like this uses a "source_zsh" command, but I don't have such a command in my zsh, nor can I find mentioned anywhere else. Could somebody give me a hint please?

pseyfert:
if i understand what you mean correctly, then no it doesn't use a `source_zsh` command:
`_FOO_BAR_COMPLETE=source_zsh` sets the variable `_FOO_BAR_COMPLETE` to the string `source_zsh`. So
```
_FOO_BAR_COMPLETE=source_zsh foo-bar > foo-bar-complete.sh
```
executes `foo-bar` with the variable `_FOO_BAR_COMPLETE` set to `source_zsh` and writes the output to the file `foo-bar-complete.sh` (and replaces an existing `foo-bar-complete.sh` should there be one).

eli-b:
Thanks for explaining, @pseyfert !

I didn't realize that when click sees this specially-named variable it outputs those arcane shell configuration commands and exits, instead of running the rest of my program. Those commands can then be either eval'ed in the new shell or saved for later.

Issue #371, opened 2015-07-03
Result object should have .stdout and .stderr in addition to .output


omribahumi:
I'm writing a click application that sends execution information to stderr. The interesting output is on stdout, which I want to make assertions on.

Right now I have to use a regular expression to extract stdout from the `Result.output` property.
Is it possible to save them separately in addition to the aggregated value?

untitaker:
Definetly possible, but I've yet to see a case where testing whether something goes to stderr instead of stdout caught any actual errors.

omribahumi:
It's not what I wanna test. The stderr output is irrelevant for the testing. Think about verbose logging where the log output goes to stderr. You don't want that in your output assertions.

untitaker:
Ahh, but `stderr` might contain serious (non-fatal) error messages as well. In that case, perhaps you actually want to stay with regex-filtering, or provide and use a switch to turn the logging messages off.

omribahumi:
Fair enough. I'll add a `--quiet` flag. Though I still think this might be a nice feature.

bryder:
I do stderr vs stdout error testing all the time. My tools write python
exceptions to stderr - so I usually have

assert stderr == ""

Sometimes my stdout is supposed to be empty (ie a selection didn't match
anything) - and not checking stderr means I'll miss an exception but the
test will pass because stdout is empty.

So without that separation I can not properly test arguments which are
supposed to produce no result.

I run my cli tests  in a  wrapper which use pexpect because my tools behave
differently if stdin is /dev/tty or a file/pipe (I warn the user that input
is expected on stdin if they start the tool without naming files - and
stdin is /dev/tty - otherwise people run things and wonder why nothing is
happening). So I have to test the case when stdin is a file or /dev/tty.

This is why for me the prompt functions in click are not useful because I
read data off stdin all the time - at which point prompt is useless. Maybe
I'lll make a pull request but no one seems to need a proper separation of
/dev/tty and stdin and it'll take a bit of work.

geowurster:
@untitaker Another argument for `.stdout` and `.stderr` is for ensuring serialized output like JSON is de-serializable.  Warnings can be triggered inside an external dependency and may not be filtered with a regex but end up in `.output`.  We can of course add `warnings.simplefilter('ignore')` at the beginning of tests affected by this scenario, but they can be difficult and time consuming to identify.

[...]

PR #868, opened 2017-10-13
CliRunner: add stdout / stderr, allow separating standard streams

thecubic:
attempt resolution of #371 (adding stdout / stderr)

This changes the internal API such that a singular out is never returned - it's either `(stdout, stderr)` when not mixed or `(stdout_and_stderr, False)` when mixed

This defaults to mixed to preserve old use cases (and tests); the runner switches to non-mixed via `mix_stderr`

### DIFF
--- a/click/testing.py
+++ b/click/testing.py
@@ -73,12 +73,14 @@ def make_input_stream(input, charset):
 class Result(object):
     """Holds the captured result of an invoked CLI script."""
 
-    def __init__(self, runner, output_bytes, exit_code, exception,
-                 exc_info=None):
+    def __init__(self, runner, stdout_bytes, stderr_bytes, exit_code,
+                 exception, exc_info=None):
         #: The runner that created the result
         self.runner = runner
-        #: The output as bytes.
-        self.output_bytes = output_bytes
+        #: The standard output as bytes.
+        self.stdout_bytes = stdout_bytes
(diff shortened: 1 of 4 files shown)


zsimic:
When will this be released?

thecubic:
(the failures are 2.6 and 3.3 not being able to install test coverage, not this code)
@untitaker care for a review of this, please?  Thanks

tests/test_testing.py
```diff
@@ -204,6 +204,30 @@ def cli_env():
     assert os.environ == env_orig
 
 
+def test_stderr():
```
dsully: Can you add a comment / docstring describing what this is testing & why?

click/testing.py
```diff
@@ -88,12 +90,27 @@ def __init__(self, runner, output_bytes, exit_code, exception,
...
     def output(self):
-        """The output as unicode string."""
-        return self.output_bytes.decode(self.runner.charset, 'replace') \
+        """The (standard) output as unicode string."""
+        return self.stdout
+
+    @property
+    def stdout(self):
+        """The standard output as unicode string."""
+        return self.stderr_bytes.decode(self.runner.charset, 'replace') \
```
dsully: Shouldn't this be 'return self.stdout_bytes' ?

PR #869, opened 2017-10-14
ZSH completion auto-documentation

### LINKED ISSUE #866: Allow documentation in ZSH auto completion.
After #865 is merged we can start using the extended autocompletion benefits of ZSH such as offering documentation for arguments.

stopthatcow:
Implements zsh descriptions for autocompletions based on the help parameter.

Fixes #866

### DIFF
--- a/click/_bashcomplete.py
+++ b/click/_bashcomplete.py
@@ -24,13 +24,28 @@ complete -F %(complete_func)s %(script_names)s
 
 COMPLETION_SCRIPT_ZSH = '''
 %(complete_func)s() {
-    emulate -L zsh
-    local IFS=$'\n'
-    local completions=( $( env COMP_WORDS="${words[*]}" \\
+    local -a completions
+    local -a completions_with_descriptions
+    local -a response
+    response=("${(@f)$( env COMP_WORDS=\"${words[*]}\" \\
                         COMP_CWORD=$((CURRENT-1)) \\
-                        %(autocomplete_var)s="complete" \\
-                        %(script_names)s ) )
-    compadd -M 'r:|=* l:|=* r:|=*' -a -- completions 
+                        %(autocomplete_var)s=\"complete_zsh\" \\
+                        %(script_names)s )}")
+
+    for key descr in ${(kv)response}; do
+      if [[ "$descr" == "_" ]]; then
+          completions+=("$key")
+      else
+          completions_with_descriptions+=("$key":"$descr")
+      fi
+    done
+
+    if [ -n "$completions_with_descriptions" ]; then
+        _describe '' completions_with_descriptions
+    fi
+    if [ -n "$completions" ]; then
+        compadd -M 'r:|=* l:|=* r:|=*' -a completions
+    fi
 }
 
 compdef %(complete_func)s %(script_names)s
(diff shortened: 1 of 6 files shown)


stopthatcow:
@rowoflo take a look if you like.

click/_bashcomplete.py
```diff
@@ -203,15 +237,19 @@ def do_complete(cli, prog_name):
         incomplete = ''
 
     for item in get_choices(cli, prog_name, args, incomplete):
-        echo(item)
+        echo(item[0])
+        if include_descriptions:
+            # ZSH has trouble dealing with empty array parameters when returned from commands, so use a well defined character '_' to indicate no description is present.
```
blueyed: What exactly is the problem here?
Shouldn't echoing an empty string should work?
(but have not looked at the details)

docs/bashcomplete.rst
```diff
@@ -56,6 +55,33 @@ Here is an example of using a callback function to generate dynamic suggestions:
...
+Completion help strings (ZSH only)
+----------------------------------
+
+ZSH supports showing documentation strings for completions. These are taken
+from the help parameters of options and subcommands. For dynamically generated
+completions a help string can be provided by returning a tuple instead of a
+string. The first element of the tuple is the completion and the second is the
+help string to display.
+
+Here is an example of using a callback function to generate dynamic suggestions with a help strings:
```
blueyed: s/with a/with/

docs/bashcomplete.rst
```diff
@@ -56,6 +55,33 @@ Here is an example of using a callback function to generate dynamic suggestions:
...
+
+.. click:example::
+
+    import os
+
+    def get_colors(ctx, args, incomplete):
+        colors = [('red', 'help string for the color red'),
+                  ('blue', 'help string for the color blue'),
+                  ('green', 'help string for the color green')]
+        return [c for c in colors if incomplete in c[0]]
```
blueyed: `return [c for c in colors if c[0].startswith(incomplete)]` ?

blueyed (commented):
Awesome, thanks!
Left some minor comments - looks good to me in general, but have not reviewed it thoroughly.

click/_bashcomplete.py
blueyed: Would be simpler to output it as `'%s:%s' % item` - ready for Zsh to consume.
Your approach is more universal, but `%s:%s` could also be parsed for other shells later, and avoids issues with empty descs.

blueyed:
In general (also for Bash) it might be better to generate all the completions at once, which would not require to call the program over and over again.
This does not allow for dynamic completions then though (but I have not checked if this is currently supported after all).

davidism:
When I try to rebase this against master, doc completion stops working. Not sure why. Can you rebase this and make sure it's still working for you?

stopthatcow:
@davidism Should be fixed. Looks like there was another CL that changed the behavior or the short_help command property.

davidism:
If I try to tab complete when there's nothing more to complete, I get the following error:

```
# press tab after this:
bashcompletion cmd1 EDITOR
# get this:
_bashcompletion_completion:4: bad set of key/value pairs for associative array
```

stopthatcow:
Reproduced the issue, fixed and tested. It looks like zsh doesn't love empty associative arrays. Please take another look.

Issue #870, opened 2017-10-14
click.prompt kwargs `default` and `value_proc` seem to be behaving inconsistently


kunalbhagawati:
The `default` argument seems to behave inconsistently with `value_proc`.

Lets take a file

```python
from __future__ import absolute_import, unicode_literals
import click

def run():
    def test_proc(val):
        print('HERE')
        return '* {} *'.format(val)

    inp = click.prompt('test val', default=1, value_proc=test_proc)
    print(inp)


if __name__ == '__main__':
    run()
```

Running this script we get:

```sh
test val [1]: 1
HERE
* 1 *
```

But using the default, 
```sh
test val [1]:
1
```

It seems that the inner function is not invoked at all (verified using debugger)

The culprit seems to be here
```python

# click/termui.py:87

    while 1:
        while 1:
            value = prompt_func(prompt)
            if value:  # <<< since the user did not enter anything, `value` is ''
                break
            # If a default is set and used, then the confirmation
            # prompt is always skipped because that's the only thing
            # that really makes sense.
            elif default is not None:  # <<< since default is not `None` this is returned 
                return default
        try:
            result = value_proc(value)  # and this is never called
```

macOS Sierra (10.12.6)
Python 2.7.10
click==6.7

PR https://github.com/pallets/click/pull/872

davidism:
Based on discussion in #872, going to close this. The default should be the final value, not an intermediate value. `value_proc` seems intended for user input only.

PR #872, opened 2017-10-14
Fix/default value proc

kunalbhagawati:
https://github.com/pallets/click/issues/870

# Changes
- Passes the `default` value through `value_proc` and then returns it. At the moment, the default value is just returned. Thus the input to the function call cannot determine if the value is changed.

### DIFF
--- a/click/termui.py
+++ b/click/termui.py
@@ -101,7 +101,7 @@ def prompt(text, default=None, hide_input=False, confirmation_prompt=False,
             # prompt is always skipped because that's the only thing
             # that really makes sense.
             elif default is not None:
-                return default
+                return value_proc(default)
         try:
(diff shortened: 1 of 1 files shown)


dsully:
@stefreak Can you rebase this PR? The changes list above doesn't seem quite right.

Thanks

stefreak:
@dsully  this is not my PR, just some commit that i commented on a long time ago happened to be part of it ;) 

@kunalbhagawati please rebase this PR!

stefreak:
@kunalbhagawati maybe you wanted to merge into `master` instead of `6.x-maintenance`?

davidism:
I agree with @sirosen, going to close this. Thanks for working on it though!

PR #874, opened 2017-10-14
Fix variable precedence

### LINKED ISSUE #873: Variable precedence
Playing with `click`'s ability to read environment variables and provide default values, I noticed that the order in which the variables are looked up seems wrong.

Here is what I expected (first variable found wins):

1. Retrieve the value from the CLI
2. Retrieve the value from the environment
3. Retrieve the default value

But here is what happened:

1. Retrieve the value from the CLI
2. Retrieve the default value
3. Retrieve the value from the environment

This seems wrong because the fact that I set default values should not prevent the environment lookup to happen.

## How to reproduce

Create the `precedence.py` file:

```python
import click

@click.command()
@click.option('--name', help="Specify a name")
def prec(name):
    """Greet the world."""
    click.echo(f'Hey {name}!')


if __name__ == '__main__':
    prec(auto_envvar_prefix='CLICK', default_map={'name': 'Stranger'})
```
Then run:
```
$ export CLICK_NAME=Tiger
$ python precedence.py --name Joe
Hey Joe!
$ python precedence.py
Hey Stranger!
```

Then remove the `default_map` argument from the command call and you get:
```
$ export CLICK_NAME=Tiger
$ python precedence.py --name Joe
Hey Joe!
$ python precedence.py
Hey Tiger!
```

## Versions
Python: 3.6.3
Click: 6.7

rgreinho:
This patch fixes the variable precedence in order to have the default
values read last:

1. Values from CLI
2. Values from the environment
3. Default values

Tests were added accordingly.

Drive-by:
* Update `.gitignore` file
* Update `tox.ini` to skip missing environments

Fixes: #873

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ dist
 build
 docs/_build
 click.egg-info
+venv/
 .tox
 .cache
 .ropeproject
--- a/click/core.py
+++ b/click/core.py
@@ -1328,12 +1328,13 @@ class Parameter(object):
     def add_to_parser(self, parser, ctx):
         pass
 
+
     def consume_value(self, ctx, opts):
         value = opts.get(self.name)
         if value is None:
-            value = ctx.lookup_default(self.name)
-        if value is None:
             value = self.value_from_envvar(ctx)
+        if value is None:
+            value = ctx.lookup_default(self.name)
         return value
 
     def type_cast_value(self, ctx, value):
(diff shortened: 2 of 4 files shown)


ThiefMaster:
Something went very wrong here. Please rebase to master

zacbir:
Yeah, my bad. Will do.

PR #896, opened 2017-12-04
Fix issue #892

sscherfke:


sscherfke:
210 is the number of open issues, not the ID of the issue …

This PR fixes issue #892 .

PR #903, opened 2017-12-15
Add context to Commands and Groups example...

jtpereyda:
This way it works without needing to guess at the context; more
helpful for newbs.

If you read the docs straight through, you'll know this, but I ended up at this page searching by topic before I learned about `add_command`.

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -22,6 +22,8 @@ when an inner command runs:
 
 .. click:example::
 
+    import click
+
     @click.group()
     @click.option('--debug/--no-debug', default=False)
     def cli(debug):
@@ -30,6 +32,11 @@ when an inner command runs:
     @cli.command()
     def sync():
         click.echo('Synching')
+
+    if __name__ == '__main__':
+        cli.add_command(sync)
+        cli()
+
 
 Here is what this looks like:
 


ThiefMaster:
I think this just adds noise to a concise example. This page is not a quickstart guide.

jtpereyda:
@ThiefMaster thanks for the quick response in any case!

docs/commands.rst
```diff
@@ -31,6 +33,11 @@ when an inner command runs:
     def sync():
         click.echo('Synching')
 
+    if __name__ == '__main__':
+        cli.add_command(sync)
```
ThiefMaster: btw, this isn't necessary anyway since `@cli.command()` already registers the command

jtpereyda: @ThiefMaster Nice! Actually, this is exactly the kind of reason I would like this sort of context! I wrote what I got by trying to piece together what I saw in different places.

Specifically, I got the add_command from [quickstart nesting commands](http://click.pocoo.org/5/quickstart/#nesting-commands), but I didn't read far enough to see the shortcut.

To explain my reasoning: What people almost always want when they look through documentation (except perhaps advanced users who are using it for a reference) is examples. That doesn't mean one should sacrifice quality reference material for an untamed example quagmire, but it does mean that the more examples are present, the better.

So significant is the need for examples that StackOverflow has become massively popular in no small part because it provides so many beginner friendly examples (there are other reasons of course). In fact, providing more examples was the motivating factor behind [Stack Overflow Documentation](https://stackoverflow.blog/2016/07/21/introducing-stack-overflow-documentation-beta/). The project failed for other reasons -- but there was a kernel of truth in its observation: too often documentation provides a thorough technical reference but not the examples that people are looking for.

In my case, I was trying to evaluate click and understand how easy it is to do nested commands. When I came to the docs page, I understood what the example was saying but not how to use it. You are right that the couple extra lines adds a small amount of noise. However, it adds a large amount of helpful context for user most in need of documentation: the noob.

If one argues that the point of the documentation is nevertheless only to be a strict technical reference, then the whole example should be removed carte blanche. But if the example is already there, why not go all the way and fill it out?

That said, the documentation is overall pretty great, and the half hour I lost on this confusion is quite small compared to what I've lost working with other modules. In my opinion the added context would save time for many people like myself, but I wouldn't have made this PR or still be discussing it if I didn't thick `click` was the best module in its class. I've been wanting something like this for years, and every competing module I've seen has fallen short, so thank you for keeping it such a high quality tool!

PR #912, opened 2018-01-07
Add AppVeyor testing for Python 3.5 and 3.6

jdufresne:


jdufresne:
It looks like AppVeyor isn't even enabled. Should I just remove the file instead?

PR #922, opened 2018-02-01
fix example error. it is `parse`,not `parser`

9nix00:


wangwenpei:
BTW, thanks Click, it's awesome! 
I am planing add logger feature to click. but I am not sure it can be merged. 
maybe we should discuss it first?

I am the author of [Cliez](https://github.com/wangwenpei/cliez).  it's a CLI tool like Click with same goal. we‘ve use it in our projects for years . Recently, we are migrating our projects from Django to Flask, and we found `Click`. it's really a good idea use decorator instead parsers.

after I read the `Click` code. I found it should add a logger feature like `Cliez`. it may look like this:

```
@click.command()
@click.option('--swarm')
@click.option('--k8s')
def deploy(swarm, k8s):
    click.echo('hi')

    #--------> new feature code    
    click.logger.debug('debug')
    click.logger.info('info')
    #<--------
    pass

```
and of course, we may need set logger level feature too.
I am thinking about it. it may looks like this(not final version).

```
@click.settings('logging', replace=True, level=10)
def logging():
    pass


@click.settings('logging', replace=False, level=100)
def logging():
    pass

```

`level` and `replace` to make sure user can set logging level in right way to work with other code.

PR #930, opened 2018-02-12
Overhaul bash completion to mirror invoke logic

### LINKED ISSUE #919: Completion of subcommand options after last argument does not work
* Install latest `HEAD` version of `click`:
```
$ pip3 install -e git+git://github.com/pallets/click.git@132d66ac7d69a2b0e8f218a4cd39e50be3e0bcb9#egg=click
```
* Save the following in a file called `foo` on your `PATH`, ensuring it's executable:
```python
#!/usr/bin/env python3

import click

def _complete(ctx, args, incomplete):
    return ['abc', 'def', 'ghi', 'jkl', 'mno', 'pqr', 'stu', 'vwx', 'yz']

@click.group()
def entrypoint():
    pass

@click.command()
@click.option('--opt', autocompletion=_complete)
@click.argument('arg', nargs=-1)
def subcommand(opt, arg):
    pass

entrypoint.add_command(subcommand)

entrypoint()
```
* Source completion code
```sh
source <(_FOO_COMPLETE=source foo)
```
---
**Expected Behaviour**
```
$ foo subcommand --opt <TAB>
abc  def  ghi  jkl  mno  pqr  stu  vwx  yz

$ foo subcommand whatever --opt <TAB>
abc  def  ghi  jkl  mno  pqr  stu  vwx  yz
```
**Actual Behaviour**
```
# WORKS
$ foo subcommand --opt <TAB>
abc  def  ghi  jkl  mno  pqr  stu  vwx  yz

# BROKEN
$ foo subcommand whatever --opt <TAB>
<list of files in current directory (i.e. fallback when no results)>
```

### LINKED ISSUE #925: Bash completion is broken when a default value is specified for a Choice Argument
Adding a default value to an argument of the Choice type will hose autocompletion. Can be tested as follows:
 
Slightly modifying test_argument_choice test example in test_bashcomplete.py to add a default argument:
```
def test_argument_choice():
    @click.command()
    @click.argument('arg1', required=False, type=click.Choice(['arg11', 'arg12']))
    @click.argument('arg2', required=False, type=click.Choice(['arg21', 'arg22']), default='arg21')
    @click.argument('arg3', required=False, type=click.Choice(['arg', 'argument']))
    def cli():
        pass

    assert list(get_choices(cli, 'lol', [], '')) == ['arg11', 'arg12']
    assert list(get_choices(cli, 'lol', [], 'arg')) == ['arg11', 'arg12']
    assert list(get_choices(cli, 'lol', ['arg11'], '')) == ['arg21', 'arg22']
    assert list(get_choices(cli, 'lol', ['arg12', 'arg21'], '')) == ['arg', 'argument']
    assert list(get_choices(cli, 'lol', ['arg12', 'arg21'], 'argu')) == ['argument']
```

Result:
```
  File "test_compl.py", line 19, in <module>
    test_argument_choice()
  File "test_compl.py", line 14, in test_argument_choice
    assert list(get_choices(cli, 'lol', ['arg11'], '')) == ['arg21', 'arg22']
AssertionError
```

stopthatcow:
Overhaul the bash completion code to mirror the invoke logic in core.py. Removed flawed assumption about handling of arguments with required=false in chained scenarios. Now completion search will halt on the first incomplete argument as the parser would.

More test case suggestions are welcome.

Fixes #925 
Fixes #919


stopthatcow:
@untitaker  or @mitsuhiko feel free to take a look.

stopthatcow:
Ping @davidism

davidism:
Would you merge master in and ensure this is all working, now that some other completion PRs are merged?

stopthatcow:
Merged and tested. Thanks!

CHANGES.rst
```diff
@@ -106,7 +106,14 @@ Version 7.0
   path so that changing working directories does not harm it.
 - Force stdout/stderr writable.  This works around issues with badly patched
   standard streams like those from jupyter.
+- Add support for auto-completion documentation. See (`#866`_).
```
davidism: This is already mentioned earlier in the log. Keep all the completion logs together.

click/core.py
```diff
@@ -315,6 +315,12 @@ def __init__(self, command, parent=None, info_name=None, obj=None,
         #: will do its best to not cause any failures.
         self.resilient_parsing = resilient_parsing
 
+        #: Indicates that default values should be ignored.
+        #: Useful for completion.
+
+        #: .. versionadded:: 7.0
+        self.ignore_default_values = ignore_default_values
```
davidism: This doesn't seem like an option that should be exposed to users. Is there a better way to handle this only for completions without making it something users can mess with easily?

davidism:
It looks like there's a lot of commits from previous PRs. Would you squash this in to a reasonable number of commits?

    git fetch upstream
    git rebase -i upstream/master

click/core.py
stopthatcow: The only other thing that comes to mind would be when we check that we are in completion mode based on the presence of the completion environment variable we could set a module local variable in core.py. These doesn't seem to be a precedence for that though. Let me know if you had other ideas and what you prefer.

sirosen: If this is the only sticking point for this PR, what about using `resilient_parsing` mode for this purpose instead? It would be a subtle change in what `resilient_parsing` does, but the whole point of that flag is to turn off a bunch of normal parsing behaviors to make completion work.

Is there a use for setting `resilient_parsing=True` and ensuring that parameters are populated with their defaults? I can't imagine one.

davidism: Good idea, let's roll this into `resilient_parsing`. Just be sure to document the change in behavior.

davidism:
I would like to get this into 7.0. Would you address the last review comment and fix the merge conflict with CHANGES.rst (sorry for the messy merge there).

stopthatcow:
Will make requested changes this evening.

stopthatcow:
Ok @davidism PTAL. It would be great if #1059 and #1047 could get in for 7.0 too. They just need to be re-syned with master once this PR is merged.

davidism:
Looks good, I'll merge soon.

Do we need to eventually rename `resilient_parsing` to `completion_parsing` or something? I'm not sure what the history of the option is or where it would be used besides completion.

Just to be clear in case we want to work on this more in the future, what about defaults is breaking completion?

stopthatcow:
We could, I didn't want to break any existing users of the flag, though this change will change behavior of the flag, so might be better to rename it. This change is needed to ensure that we offer completion on items that have defaults as the parser will fill values in as if they were entered.

Issue #824, opened 2017-07-09
ValueError: I/O operation on closed file in testing


ResidentMario:
I am getting a `ValueError` when using `CliRunner` to test my CLI tool. Unfortunately I can't isolate, but I can give a reproduction:

```bash
pip install git+git://github.com/ResidentMario/datablocks.git@d6601c8321e7fb050bb62760067dc63ca4600003
pip install pytest
cd datablocks/tests
pytest cli_tests.py
```

Which raises the following traceback:

```
=================================== FAILURES ===================================
_________________________ TestLink.test_depositor_link _________________________

self = <cli_tests.TestLink testMethod=test_depositor_link>

    def test_depositor_link(self):
>       result = runner.invoke(cli.link, ["test_depositor.py", "--outputs", "['bar2.txt']"])

cli_tests.py:41: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <click.testing.CliRunner object at 0x7f2c833a6cc0>
cli = <click.core.Command object at 0x7f2c82e7a5f8>
args = ['test_depositor.py', '--outputs', "['bar2.txt']"], input = None
env = None, catch_exceptions = True, color = False, extra = {}
exc_info = (<class 'SystemExit'>, SystemExit(0,), <traceback object at 0x7f2c82e1eb08>)
out = <_io.BytesIO object at 0x7f2c82e0a5c8>, exception = None, exit_code = 0

    def invoke(self, cli, args=None, input=None, env=None,
               catch_exceptions=True, color=False, **extra):
        """Invokes a command in an isolated environment.  The arguments are
            forwarded directly to the command line script, the `extra` keyword
            arguments are passed to the :meth:`~clickpkg.Command.main` function of
            the command.
    
            This returns a :class:`Result` object.
    
            .. versionadded:: 3.0
               The ``catch_exceptions`` parameter was added.
    
            .. versionchanged:: 3.0
               The result object now has an `exc_info` attribute with the
               traceback if available.
    
            .. versionadded:: 4.0
               The ``color`` parameter was added.
    
            :param cli: the command to invoke
            :param args: the arguments to invoke
            :param input: the input data for `sys.stdin`.
            :param env: the environment overrides.
            :param catch_exceptions: Whether to catch any other exceptions than
                                     ``SystemExit``.
[...]

Issue #654, opened 2016-09-14
click.testing is not properly proxying stdout.


abetkin:
For example, you cannot set `import pdb; pdb.set_trace()` in your code, or drop into debugger on exceptions.

This works:

``` python
from io import StringIO
import sys

from contextlib import contextmanager

class StreamWrapper(object):

    def __init__(self, stream):
        self.stream = stream
        self._stringio_ = StringIO()

    def __getattr__(self, name):
        return getattr(self.stream, name)

    def write(self, s, *args, **kw):
        self.stream.write(s)
        self._stringio_.write(s)

    @classmethod
    @contextmanager
    def stdout(cls):
        __stdout__ = sys.stdout
        sys.stdout = cls(sys.stdout)
        yield sys.stdout._stringio_
        sys.stdout = __stdout__

def invoke(cmd, args=()):
    with StreamWrapper.stdout() as c:
        cmd(standalone_mode=False, args=args)
    return c.getvalue()
```

stevekuznetsov:
@abetkin You can get around this issue by using `mkfifo` as described in [this](http://stackoverflow.com/a/26975795) StackOverflow answer:

>Another option is to create you own Pdb object, and set there the stdin and stdout. My proof of concept involves 2 terminals, but for sure some work can be merged some kind of very unsecure network 

>1. Create two FIFOs:

>    ```sh
    $ mkfifo stdin
    $ mkfifo stdout
    ```

>2. In one terminal, open `stdout` on background, and write to `stdin`:

>    ```sh
    $ cat stdout &
    $ cat > stdin
    ```

>3. In your Python code/console create the `pdb` object, and use it:

>    ```py
    import pdb
    mypdb=pdb.Pdb(stdin=open('stdin','r'), stdout=open('stdout','w'))
    mypdb.set_trace()
    ```

>4. You should be able to use `pdb` on the first console. The only drawback is having to use your custom `pdb`, but some monkey patching at `init` (`PYTHONSTARTUP` or similar) can help:

>    ```py
    import pdb
    mypdb=pdb.Pdb(stdin=open('stdin','r'), stdout=open('stdout','w'))
    pdb.set_trace=mydbp.set_trace
    ```

blueyed:
@abetkin 
Thanks for filing this, and suggesting some fix.
I've noticed this myself and reported it for pytest - since I hope it can be fixed there even/additionally, given that pytest also does some magic regarding capturing: https://github.com/pytest-dev/pytest/issues/3344.

[...]

PR #951, opened 2018-03-26
WIP: testing: runner: wrap stdin+stderr

blueyed:
This also helps with failures during testing, e.g. when using
`pdb.set_tracce()` somewhere inside the runner's isolation - likely
since it handles `write` already.

Ref: https://github.com/pallets/click/issues/824

### DIFF
--- a/click/testing.py
+++ b/click/testing.py
@@ -73,12 +73,12 @@ def make_input_stream(input, charset):
 class Result(object):
     """Holds the captured result of an invoked CLI script."""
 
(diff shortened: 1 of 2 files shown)


blueyed:
https://github.com/pallets/click/issues/654 has a patch also (only for stdout), that might be simpler.
And see https://github.com/pallets/click/issues/371 for discussion about API changes/adjustments.

dsully:
@blueyed I merged PR #868 - let's follow on with any changes from that. Thanks.

PR #953, opened 2018-03-27
Set blocking I/O flags for FreeBSD

juice500ml:
FreeBSD's threading library automatically sets stdin and stdout to be non-blocking.
So have to ensure that stdio is in blocking mode.

Similar fix in other project
https://trac.edgewall.org/ticket/2066
https://trac.edgewall.org/changeset/2260


juice500ml:
Oops wrong branch. My bad.

PR #959, opened 2018-04-01
Fix #958, check errors param in _force_correct_text_reader

### LINKED ISSUE #958: The `errors` argument in `click.open_file` is ignored on python3
The `errors` argument of `click.open_file` is not propagated to `sys.stdin` on python 3.x if there is no `encoding` argument in couple with `errors`.

How to reproduce:

```bash
> head -c64 /dev/urandom | python3 -c 'import click
print(click.open_file("-", errors="ignore").read())'
  File "/usr/lib/python3.6/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xcd in position 1: invalid continuation byte
```

It is because `_force_correct_text_reader` function checks that `_stream_is_misconfigured(text_reader)` call returns `False` and returns the `text_reader` as is.

ei-grad:


ei-grad:
Oops. It doesn't fix the problem, actually.

PR #969, opened 2018-04-12
Document project as stable and ready for use in production

jdufresne:
For a full list of trove classifiers, see:

https://pypi.python.org/pypi?%3Aaction=list_classifiers


zacbir:
Already in master. Thanks for submitting.

PR #985, opened 2018-05-03
Backport hidden commands and options from master 

sstaszkiewicz-copperleaf:
Seems like there are a number of users who would like to see #500 added to the 6.x branch.

Taking a stab at backporting the feature from master


sstaszkiewicz-copperleaf:
Awesome! I'm glad to hear that, cause there are folks that have been waiting 2+ years for this functionality.

mikenerone:
Are we getting close on 7.0?

PR #993, opened 2018-05-14
Add Py2/3 unicode / str compat for documentation tools.

### LINKED ISSUE #719: Can't build documentation with python 3
```
> make html
...
Exception occurred:
  File "/home/user/click-6.7/docs/clickdoctools.py", line 73, in isolation
    if isinstance(input, unicode):
NameError: name 'unicode' is not defined
```

dsully:
Fixes #719


dawranliou (commented):
Would it be a better idea to reuse the `click._compat.text_type`?

dsully:
I looked at that, but docs/*.py doesn't use anything from src/

PR #997, opened 2018-05-14
Rebased fix to get_winterm_size() returning (0, 0)

H1SWarrior:
Rebased #610 against current master.

### DIFF
--- a/click/termui.py
+++ b/click/termui.py
@@ -176,6 +176,8 @@ def get_terminal_size():
             sz = shutil_get_terminal_size()
             return sz.columns, sz.lines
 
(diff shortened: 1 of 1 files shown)


click/termui.py
```diff
@@ -177,7 +177,11 @@ def get_terminal_size():
             return sz.columns, sz.lines
 
     if get_winterm_size is not None:
-        return get_winterm_size()
+        size = get_winterm_size()
+        if size == (0, 0):
```
dsully: Can you add an inline comment that reflect the same info as the commit message?

Thanks

PR #998, opened 2018-05-14
format commands normalize format -> remove short_help option

fdavis:
The short_help option is only used in the commands view and seems to have caused some confusion about its interaction with the actual help for a command.

I propose we deprecate it and allow the format_commands to handle this


fdavis:
closing in favor of #1002

PR #436, opened 2015-10-10
Introduce raw_terminal

untitaker:

PR #1007, opened 2018-05-15
Recreated pull request #436

JosiahDub:
Added a fix where fd was undefined. Added "as fd" to the with statement.


dsully:
A docstring here would be great as a follow on PR.

PR #1008, opened 2018-05-15
Recreate #446

zacbir:
This branch recreates a1exsh's branch for formatting short help.


dsully:
Looks like this needs to be rebased first, as to not lose the deprecation message PR that went in prior.

zacbir:
Hmm, I did a rebase off a fresh pull of master, unless something got slotted in. Yay, race condition sprinting!

Issue #447, opened 2015-10-31
progressbar does not allow update of 0


nchammas:
``` python
with click.progressbar(length=5) as progress_bar:
    progress_bar.update(0)
```

This throws a `float division by zero` error, which I believe is caused by [this line here](https://github.com/mitsuhiko/click/blob/89d3bfda52d151762e732738b828312a240bf89e/click/_termui_impl.py#L229).

The reason I ran into this is because I'm trying to use `progressbar` to track the status of remote resources which do work at unpredictable intervals. I pass in something like `work_remaining_start - work_remaining_end` to `progress_bar.update()` in a loop that periodically checks these remote resources, and sometimes this difference is 0.

Does it make sense for `update()` to accept a step of 0?

nchammas:
One good reason to allow this, at least for my use case, is that you can update the "current item" while you are waiting for work to be done, like a heartbeat.

For example:

```python
with click.progressbar(length=5, item_show_func=lambda x: current_time()) as progress_bar:
    wait_some_time()
    work_done = check_remote_resource_for_progress()
    progress_bar.update(work_done)
```

So even if no work was done on a given iteration, you can still update the progress bar with something like the current time to indicate that things are still working.

If I understand correctly, `show_eta` doesn't kick in until after the first step has been made, and it probably is unreliable for this situation where you are waiting on remote resources to do work.

An alternative solution for this "heartbeat" use case may be to add an option similar to `show_eta` which shows a timer that simply counts up.

fdavis:
I think @nchammas is right and we can just default to 0 instead of doing that division if this is called with 0

https://github.com/mitsuhiko/click/blob/89d3bfda52d151762e732738b828312a240bf89e/click/_termui_impl.py#L229

PR #1012, opened 2018-05-15
Fixes issue #447

rpgraham84:
Fixes a ZeroDivisionError in ProgressBar.make_step when the first
arg passed to the first call of ProgressBar.update is 0.

### DIFF
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -245,7 +245,16 @@ class ProgressBar(object):
             return
 
         self.last_eta = time.time()
-        self.avg = self.avg[-6:] + [-(self.start - time.time()) / (self.pos)]
+
+        # self.avg is a rolling list of length <= 7 of steps where steps are
+        # defined as time elapsed divided by the total progress through
+        # self.length.
+        if self.pos:
+            step = (time.time() - self.start) / self.pos
+        else:
+            step = time.time() - self.start
+
+        self.avg = self.avg[-6:] + [step]
 
         self.eta_known = self.length_known
 


click/_termui_impl.py
```diff
@@ -245,7 +245,16 @@ def make_step(self, n_steps):
             return
 
         self.last_eta = time.time()
-        self.avg = self.avg[-6:] + [-(self.start - time.time()) / (self.pos)]
+
+        # self.avg is a rolling list of length <= 7 of steps where steps are
+        # defined as time elapsed divided by the total progress through
+        # self.length.
+        if not self.pos == 0:
```
ThiefMaster: `if not foo == bar` is bad. `if self.pos != 0` is better, but you can just do `if self.pos`.

rpgraham84: Fixed ;)

dawranliou (commented):
Looks like this commit 832a3ab belongs to another issue :P

davidism:
I'm going to split the docs change into another PR.

rpgraham84:
Thanks, it's like my first PR and thought it would probably just sit here rejected forever. I think the moral of the story for me is to always make a new issue branch for each issue. Someone at PyCon recommended just using master but I think I misunderstood what they were telling me. They meant make the PR into click's master from my own issue branch.

davidism:
It's just taking me a while to get to all the notifications. :-)

When making PRs, create a branch off of master, push it to your fork, then make the merge request from that branch into master in this repository (GitHub sets that last part up up by default).

rpgraham84:
Thank you!! I didn't mean that you were taking too long, btw :) Happy to have contributed something, I'll fix some more things as time allows. With tests, even!

PR #1013, opened 2018-05-15
don't add newlines by default for progressbars

kporangehat:
Fixes a regression introduced by 61abbef7cec63a9dcd15ec471d866271933e45bb which changed the default behavior of progressbars to rendering with newlines producing output like this:

```
Downloading reference  [#######-----------------------------]   20%
Downloading reference  [##########--------------------------]   30%
Downloading reference  [##############----------------------]   40%  00:00:01
Downloading reference  [##################------------------]   50%  00:00:01
Downloading reference  [#####################---------------]   60%  00:00:01
Downloading reference  [#########################-----------]   70%  00:00:00
Downloading reference  [############################--------]   80%  00:00:00
Downloading reference  [################################----]   90%  00:00:00
Downloading reference  [####################################]  100%
```

Now it looks like this:
```
Downloading reference  [####################################]  100%
```

### DIFF
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -233,7 +233,7 @@ class ProgressBar(object):
 
         if line != self._last_line and not self.is_fast():
             self._last_line = line
-            echo(line, file=self.file, color=self.color, nl=True)
+            echo(line, file=self.file, color=self.color, nl=False)
             self.file.flush()
 
     def make_step(self, n_steps):


kporangehat:
Yep. 

```python
with click.progressbar(things,
                       label='Modifying things') as bar:
    for thing in bar:
        modify_the_thing(thing)
click.echo("And we're done!")
```

```
Modifying things  [####################################]  100%
And we're done!
```

PR #1023, opened 2018-05-24
Treats empty strings as false

jpwatts:
I found it surprising that empty strings weren't parsed as `False` when interpreting environment variables for default values. I understand this may have been an intentional design decision, but here's a patch in case you're open to making a change.

### DIFF
--- a/click/types.py
+++ b/click/types.py
@@ -295,7 +295,7 @@ class BoolParamType(ParamType):
         value = value.lower()
         if value in ('true', 't', '1', 'yes', 'y'):
             return True
-        elif value in ('false', 'f', '0', 'no', 'n'):
+        elif value in ('false', 'f', '0', 'no', 'n', ''):
             return False
         self.fail('%s is not a valid boolean' % value, param, ctx)
 
--- a/docs/parameters.rst
+++ b/docs/parameters.rst
@@ -49,7 +49,7 @@ different behavior and some are supported out of the box:
     A parameter that accepts boolean values.  This is automatically used
     for boolean flags.  If used with string values ``1``, ``yes``, ``y``, ``t``
     and ``true`` convert to `True` and ``0``, ``no``, ``n``, ``f`` and ``false``
-    convert to `False`.
+    convert to `False`. An empty string will also convert to `False`.
 
 :data:`click.UUID`:
     A parameter that accepts UUID values.  This is not automatically
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -192,7 +192,7 @@ def test_boolean_conversion(runner):
             assert not result.exception
             assert result.output == 'True\n'
 
-        for value in 'false', 'f', '0', 'no', 'n':
+        for value in 'false', 'f', '0', 'no', 'n', '':
             result = runner.invoke(cli, ['--flag', value])
             assert not result.exception
             assert result.output == 'False\n'


davidism:
When would you be passing an empty value to a bool? My experience with empty vars is that they mean "no value" or `None`.

jpwatts:
I agree with you that I was thinking that my variable had no value and expected it to be interpreted as "falsy" in accordance with Python conventions. I wish I'd said "falsy" in my comment above, instead of `False`. I apologize for any confusion I created by being imprecise.

Here's an example CLI app:

```python
#!/usr/bin/env python3

import click

@click.command()
@click.option('--enable-feature', envvar="ENABLE_FEATURE", is_flag=True)
def main(**options):
    print("Feature enabled:", options['enable_feature'])

if __name__ == '__main__':
    main()
```

And a usage example:

```bash
# make sure this isn't set
$ unset ENABLE_FEATURE

# case 1: not set
$ python3 example.py
Feature enabled: False

# case 2: true, t, 1, yes, y
$ ENABLE_FEATURE="true" python3 example.py 
Feature enabled: True

# case 3: false, f, 0, no, n
$ ENABLE_FEATURE="false" python3 example.py 
Feature enabled: False

# case 4: empty
$ ENABLE_FEATURE="" python3 example.py 
Usage: example.py [OPTIONS]
Try "example.py --help" for help.

Error: Invalid value for "--enable-feature":  is not a valid boolean

# case 5: invalid
$ ENABLE_FEATURE="asdf" python3 example.py 
Usage: example.py [OPTIONS]
Try "example.py --help" for help.

Error: Invalid value for "--enable-feature": asdf is not a valid boolean
```

In case 1 the environment variable isn't set. Click pulls the value out of the environment using `os.environ.get` and we end up with `None` because it isn't there. In case 4 the environment variable is set, so Python finds the value and returns an empty string. Python considers both `None` and `""` to be "falsy", so I'd expect the app to run with the feature disabled. Case 1 works as expected, but case 4 crashes. Also, please note that Click is already coercing `None` to `False` in the context of a boolean flag.

Before reading the docs, I also found cases 3 and 5 a little suprising because all environment variable values are strings and in Python all strings are "truthy" except empty ones. I had previously assumed that all non-empty strings passed to a boolean flag would be evaluated as `True`, but never hit any issues in Click because I had always set my flags to `"1"`. I understand why it is convenient in a CLI app to have some additional non-empty strings interpreted as `False`, and I'm certainly not suggesting that you change documented behavior there, but I don't understand why it's helpful to disallow the one string that all other Python applications understand to be "falsy".


-----


In case it's helpful, I'll also describe our specific use case. We have a configuration management system that builds `docker-compose.yml` files from Jinja templates. The template context contains environment variables that can be defined as part of a generic global environment or overriden at the node level.

For a new feature we're testing, the normal workflow is to add a line to the template, but only define the variable for the node we're working with.

Here's the relevant part of the template:

```yml
ENABLE_FEATURE: "{{ ENABLE_FEATURE }}"
```

Here's how that renders for nodes with the flag enabled:

```yml
ENABLE_FEATURE: "1"
```

And here's how it looks for nodes that don't define the variable:

```yml
ENABLE_FEATURE: ""
```

Our app runs as expected in the first case, with the feature enabled. In the second case, our app crashes instead of running without the feature.

We could work around the problem by setting the variable to `"false"` in the global environment, but this would create confusion because we use other tools that follow the Python convention of all non-empty strings being "truthy", even strings like `"false"`. If Click would adopt the convention that empty strings are "falsy", even while preserving the other "falsy" strings it supports, we could define a safe subset of values that work across all the tools we use: empty/missing for `False` and any of Click's valid strings for `True`.

davidism:
Sorry, I don't think this change makes sense. If you need different behavior, you can use a subclass of the type. Or output a default in the template.

PR #1037, opened 2018-06-12
feature: support customisation of "Commands:" section label.

thruflo:
The formatted text output currently defaults to a hardcoded `Commands` string.
This patch adds a kwarg that can override the formatted text outputed for a
given MultiCommand or Group.

So, for example, initialising `Group(section_label='Resources')` would print
`Resources:` as the formatted section label for a command group that represents
a list of resources with nested sub commands.

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -955,7 +955,8 @@ class MultiCommand(Command):
 
     def __init__(self, name=None, invoke_without_command=False,
                  no_args_is_help=None, subcommand_metavar=None,
-                 chain=False, result_callback=None, **attrs):
+                 chain=False, result_callback=None,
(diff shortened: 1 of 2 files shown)


dsully:
Thanks for the PR. Could you add a documentation update for this functionality as well?

thruflo:
@dsully added the kwarg to the class api docs.  Not sure if or where you'd want to add this to any narrative documentation.

davidism:
Going to close this as I think it may be too limited as we move towards a more customizable formatting system (#561). For now, you can use https://github.com/janLuke/cloup if you need named sections for options or commands.

PR #1050, opened 2018-06-20
Fix in `click.edit()` for fast editors

giovannipizzi:
If a fast editor is used and the filesystem
only provides a resolution of 1s, `click.edit()`
returns None if the file is edited and saved very quickly.

This is a workaround that should work on all systems.

The added test would fail on my system before the fix.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -41,6 +41,8 @@ Unreleased
     name instead of ``True`` or ``False``. :issue:`1538`
 -   Non-string objects passed to ``style()`` and ``secho()`` will be
     converted to string. :pr:`1146`
+-   ``edit(require_save=True)`` will detect saves for editors that exit
+    very fast on filesystems with 1 second resolution. :pr:`1050`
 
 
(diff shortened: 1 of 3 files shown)


giovannipizzi:
Also, I understand that this is the right approach ()merge into 6.x-maintenance,  and you take care  of merging into `master`?

giovannipizzi:
Sorry, I realised there were a couple of tests not passing, I close this PR to fix them and I open it again ASAP.

giovannipizzi:
Tests should pass now (at least they pass in my fork) so for me the PR is ready to be merged.

giovannipizzi:
Hello, will this get a chance to be merged? Please let me know if I should change anything

davidism:
Retargeted this to master. This won't be in 7.0 but could be targeted at 7.1. Needs review before I can say for sure.

russkel:
Am I wrong in thinking that this test will fail on machines that do not have sed? Of course we're talking about the disfigured step child Windows.

Issue #587, opened 2016-05-19
add help to click argument


borismod:
I would like to be able to specify help on positional arguments. 

``` Python

@click.group()
def cli():
    click.echo(u'shellfoundry - CloudShell shell command-line tool')
    pass

@cli.command()
@click.argument(u'name', help='Shell name to be created')
@click.argument(u'template', default=u'default', help='Template to be used')
def new(name, template):
    """
    Create a new shell based on a template.\r\n
    """
    NewCommandExecutor().new(name, template)
```

The above code throws the below exception:

> Traceback (most recent call last):
>   File "C:\Python27\lib\runpy.py", line 162, in _run_module_as_main
>     "__main__", fname, loader, pkg_name)
>   File "C:\Python27\lib\runpy.py", line 72, in _run_code
>     exec code in run_globals
>   File "c:\work\GitHub\shellfoundry\shellfoundry__main__.py", line 6, in <module>
>     from bootstrap import cli
>   File "shellfoundry\bootstrap.py", line 36, in <module>
>     @click.argument(u'template', default=u'default', help='Template to be used')
>   File "C:\Python27\lib\site-packages\click\decorators.py", line 151, in decorator
>     _param_memo(f, ArgumentClass(param_decls, **attrs))
>   File "C:\Python27\lib\site-packages\click\core.py", line 1693, in __init__
>     Parameter.**init**(self, param_decls, required=required, **attrs)
> TypeError: **init**() got an unexpected keyword argument 'help'

tony:
I just encountered this.

http://click.pocoo.org/6/documentation/

> Arguments cannot be documented this way. This is to follow the general convention of Unix tools of using arguments for only the most necessary things and to document them in the introduction text by referring to them by name.

Is this to say the best practice would be to document the argument in the docstring itself?

Evidlo:
I think they mean to put a reference to the argument in the command long-help.  For example, heres a script with command `show` that accepts an argument `PATH`.
```
[evan@blackbox ~] ph show -h
Usage: ph show [OPTIONS] PATH

show the contents of an entry, where PATH is the entry path

optional arguments:
  -h, --help     show this help message and exit
  --field FIELD  show the contents of a specific field as plaintext
```

So in this case you'd want to use both `help` and `short_help`, where `short_help` is a shorter command description that doesn't contain information about `PATH`.

hadsed:
Please do, I've also just been hit with this.

Nukesor:
Would love to see this feature as well!

[...]

PR #1051, opened 2018-06-21
Support help message for argument

### LINKED ISSUE #2983: Add help kwarg to argument parameter
We are now open to a PR for adding a help kwarg to arguments in the same way that there is for options and adjusting the help pages to account for this. 

[Help Pages](https://click.palletsprojects.com/en/stable/documentation/#help-pages)

```
@click.command()
@click.argument('name', help='The name to print')
@click.option('--count', default=1, help='number of greetings')
def hello(name: str, count: int):
    """This script prints hello and a name one or more times."""
    for x in range(count):
        if name:
            click.echo(f"Hello {name}!")
        else:
            click.echo("Hello!")
```

Ralph-Wang:
1. Support help message for argument
2. Add UT case for message
3. Fix other failed cases as there are more messages printed.

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -888,13 +888,11 @@ class Command(BaseCommand):
         opts = []
         args = []
         for param in self.get_params(ctx):
-            if isinstance(param, Option):
-                rv = param.get_help_record(ctx)
-                if rv is not None:
+            rv = param.get_help_record(ctx)
+            if rv is not None:
+                if isinstance(param, Option):
                     opts.append(rv)
-            elif isinstance(param, Argument):
-                rv = param.get_help_record(ctx)
-                if rv is not None:
+                elif isinstance(param, Argument):
                     args.append(rv)
 
         if opts:
--- a/click/core.py
+++ b/click/core.py
@@ -1837,10 +1837,8 @@ class Argument(Parameter):
                             obj=self)
 
     def get_help_record(self, ctx):
-        if not self.help:
-            return None
         key = self.metavar if self.metavar else self.name.upper()
-        return (key, self.help)
+        return (key, self.help or '')
 
 
 # Circular dependency between decorators and core
--- a/click/core.py
+++ b/click/core.py
@@ -886,14 +886,24 @@ class Command(BaseCommand):
     def format_options(self, ctx, formatter):
         """Writes all the options into the formatter if they exist."""
         opts = []
+        args = []
         for param in self.get_params(ctx):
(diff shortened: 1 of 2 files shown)


mwgamble:
This looks really good. I'm not sure if it should be part of this pull request, but another great addition would be automatically displaying all available choices for parameters of type `Choice`.

davidism:
I'm fairly sure this isn't supported on purpose. I think the docs mention that you should document required args in the docstring (the help text for the command itself.

mwgamble:
I have a program with multiple commands where one of the required arguments is of type `Choice`, and can take over 20 different values. I don't want to have to list those out manually, I'd much prefer it if the help text generator could do that for me given it should already have access to the values.

dsully:
I'm in favor of this patch. It's always struck me as odd that arguments couldn't have help specified like options could.

davidism:
http://click.pocoo.org/5/documentation/#help-texts

> Arguments cannot be documented this way. This is to follow the general convention of Unix tools of using arguments for only the most necessary things and to document them in the introduction text by referring to them by name.

@untitaker opinion?

untitaker:
I agree with the docs. If the arguments are not documented as part of the command's action then they should probably be options as they appear to be inessential to the core functionality

Ralph-Wang:
If we don't want independent help text on screen, how about support the argument in the decorator but merge the message to the command description just like
```
# command ARGUMENT

command description and ARGUMENT(help message for argument)
```

untitaker:
So @mwgamble I am pretty sure that you can generate the docstring programmatically in your case. I am not sure if you'd be satisfied with how click would format your 20 possible values in any case

@Ralph-Wang and @mwgamble could you give me a concrete example where you would need this feature? I am pretty sure that in such cases custom formatting is better than relying on click.

mwgamble:
@untitaker That's a fair point. I was envisaging some kind of default output when no explicit help text is provided, for example:

    param1              Possible choices: choice1, choice2, choice3, choice4

Ralph-Wang:
@untitaker It's pretty fair to generate docstring ourselves. But if we support help message in argument, we can make the message more close to the argument on the code side.

davidism:
At this point I'm unwilling to change Click's handling of this.

briancappello:
It's possible to extend `click` yourself if you want to support this feature. [Example code](https://github.com/pallets/click/issues/587#issuecomment-410097642)

PR #1055, opened 2018-06-27
Document automatic lowercasing of parameter names

ZenithDK:
Be explicit in documenting that parameter names are lowercased as this can cause confusion otherwise.


docs/parameters.rst
```diff
@@ -91,11 +91,14 @@ If all names for a parameter contain dashes, the internal name is generated
 automatically by taking the longest argument and converting all dashes to
 underscores.
 
+All parameter names are converted to lowercase.
```
davidism: This should be "the internal name" to match the other paragraphs.

PR #1061, opened 2018-07-04
Ignore completions for partially typed hidden commands and options. (Fixes #1058)

### LINKED ISSUE #1058: New upcoming hidden option in Click.Option is not hidden from bash completion
Thanks for wonderful package.

I don't know whether this is intended behavior or not but I was just trying out new upcoming release 7.0 for hidden option and autocompletion to Click.Option and hidden option hides option from help message but it shows up in bash completion.
However it will be good if we can hide it from bash completion also, so that hidden option is actually hidden from everywhere.

stopthatcow:
Ignore completions for partially typed hidden commands and options.

Based on #930 which should be merged first.
Fixes #1058


stopthatcow:
@hirenvadalia can you see if this behaves as you'd expect?

hirenvadalia:
@stopthatcow Thanks for quickly working on this issue. This changes works fine for me.

Issue #1068, opened 2018-07-13
Click cannot be packaged as a submodule on Windows


brianbruggeman:
The windows shim references `click` within its import system, making it impossible to actually submodule or vendor click within another package if that package is intended to be run on Windows.

See:  https://github.com/pallets/click/blob/master/click/_winconsole.py#L18

davidism:
I don't understand the issue. Can you give an example of how you're trying to vendor this, and what error it's causing?

ThiefMaster:
My guess is he's importing click as `import mypkg.vendor.click` so there's no top-level `click` available for import

davidism:
Oh I see, it looks like most of the code is using relative imports, and the PR fixes a line that uses absolute.

davidism:
Closed with merge.

PR #1069, opened 2018-07-13
Allows click to be vendored on Windows

brianbruggeman:
Fixes: Github issue #1068
 
    * Removes import references to click within click package and 
      uses relative imports


davidism:
What does this have to do with Windows? Also, assuming you set up vendoring correctly, there's no issue with abbreviated imports, they'd still refer to the correct level of import.

PR #1070, opened 2018-07-13
additional param for Parameters to specify complex default_map lookup

mortbauer:
I added an additional keyword for any parameters to make a complex lookup in the default_map.

I find this useful since the configuration file structure might not be the same as the command structure, quite certainly it won't be. 

To understand what I mean please have a look at the new example `default_map.py` I added under examples/complex/complex/commands.

Since I used a new key this shouldn't break anything.

Would be really cool if this could be merged.

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -484,6 +484,26 @@ class Context(object):
                 rv = rv()
             return rv
 
+    def lookup_default_path(self, path):
+        """Looks up the default for a parameter path.  This by default
+        looks into the :attr:`default_map` if available.
+        """
+        if not isinstance(path, (tuple, list)):
+            path = path.split('.')
+        rv = None
+        if self.default_map is not None:
+            rv = self.default_map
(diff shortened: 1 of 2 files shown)


davidism:
Why does this need to be done in Click? If you have some other data structure besides a flat map, why can't you do the transformation on your end to pass as the default map?

mortbauer:
Well sure it can be done, but then I have to ship a separate module for this, thought it is a neat idea since I almost always have the same problem on how to merge the the config file structure and allow flexible overrides form the commandline. Thought I'm not the only one to have such a use case.

But if you say this should go to a separate module then I will go that road since it is certainly possible.

mortbauer:
But I think I can't make it as clean and easy in an extension as with the proposed approach.

davidism:
Thanks for working on this. At this time I don't want to add this to Click.

PR #1083, opened 2018-08-01
Progress tweaks

danchr:
This PR contains a few tweaks to the progress support to:

* Let me print out something during the run.
* Throttle the output so that it doesn't waste CPU resources on terminal rendering.
* Print out total time when a run is complete.

### DIFF
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -94,6 +94,8 @@ class ProgressBar(object):
         self.throttle = throttle
 
     def __enter__(self):
+        if self.entered:
+            raise RuntimeError('This progress bar is already active.')
         self.entered = True
         self.render_progress()
         return self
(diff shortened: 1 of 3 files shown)


dsully:
This looks good to me overall. The appveyor tests aren't passing though.

Does anyone have any concerns to merge this after the test issue is addressed?

danchr:
Not sure why it's failing on one location, but not the other — is there anything special about the AppVeyor tests?

(The only thing that comes to mind is that it could be a timing issue, but the tests override the system clock.)

davidism:
Thanks for working on this. While the code you've written seems good, after dealing with many progress bar issues and requests I've come to the conclusion that I don't want to add to Click's progress bar at this point. Users are better served by choosing [tqdm](https://tqdm.github.io/) if they need complex progress bar support. Click's is still a good choice for simple cases, but I'd like to avoid continuing to add complexity to it.

PR #1087, opened 2018-08-11
Clarify the "lazy" File behaviors

### LINKED ISSUE #1056: LazyFile isn't lazy for mode 'r'
The documentation for the `File` type states:

> The lazy flag controls if the file should be opened immediately or upon first IO. The default is to be non lazy for standard input and output streams as well as files opened for reading, lazy otherwise.

However, this isn't true if the file-mode is `r`. There is an explicit check
https://github.com/pallets/click/blob/ef93b49174540e72c16ad194c39947974f3d5c29/click/utils.py#L89-L93
that tries to open and close the file in lazy mode. 

This prevents using a `lazy` `File` to access files that may only come into existence later.

sirosen:
#1056 notes a documentation ambiguity which seems pretty easy to clean up.

Be more explicit about the fact that `lazy` is ignored entirely if the file is opened for reading.

close #1056

### DIFF
--- a/click/types.py
+++ b/click/types.py
@@ -331,9 +331,12 @@ class File(ParamType):
     opened in binary mode or for writing.  The encoding parameter can be used
     to force a specific encoding.
 
-    The `lazy` flag controls if the file should be opened immediately or
-    upon first IO.  The default is to be non lazy for standard input and
-    output streams as well as files opened for reading, lazy otherwise.
+    The `lazy` flag controls if the file should be opened immediately or upon
+    first IO. The default is to be non-lazy for standard input and output
+    streams as well as files opened for reading, `lazy` otherwise. When opening a
+    file lazily for reading, it is still opened temporarily for validation, but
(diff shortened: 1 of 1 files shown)


davidism:
The original wording was more accurate. `lazy` is not ignored for reads and standard streams, it defaults to `False`. 

Appending some sentences can clarify what "lazy" means for read mode. Something like:

> The `lazy` flag controls if the file should be opened immediately or upon first IO. The default is to be non-lazy for standard input and output streams as well as files opened for reading, lazy otherwise. When opening a file lazily for reading, it is still opened temporarily for validation, but will not be held open until first IO. `lazy` is mainly useful when opening for writing to avoid creating the file until it is needed.

sirosen:
Sorry, I got thrown off by the fact that it *is* ignored when handling stdin/stdout.

Since your suggested rewording reads correctly to me, I've amended to use that wording.

PR #1098, opened 2018-08-26
Allow non-standalone ctx.exit() calls to return values rather than call sys.exit()

### LINKED ISSUE #667: standalone_mode does not affect the help option
If we use `standalone_mode=True` passing arguments to command execution will not call `sys.exit`, except if the argument to the command is `["--help"]`.

From inside ipython:

> In [12]: import click
> 
> In [13]: @click.command()
>     ...: def hello():
>     ...:     print("hello, world")
>     ...:
> 
> In [14]: hello.main([], standalone_mode=False)
> hello, world
> 
> In [15]: hello.main(['wrong_arg'], standalone_mode=False)
> - --------------------------------------------------------------------------
> UsageError                                Traceback (most recent call last)
> <ipython-input-15-c2bf17950330> in <module>()
> ...
> UsageError: Got unexpected extra argument (wrong_arg)
> 
> In [16]: hello.main(['--help'], standalone_mode=False)
> Usage: ipython [OPTIONS]
> 
> Options:
>   --help  Show this message and exit.
> An exception has occurred, use %tb to see the full traceback.
> 
> SystemExit: 0
> 
> /Users/malkarouri/Library/Python/2.7/lib/python/site-packages/IPython/core/interactiveshell.py:2889: UserWarning: To exit: use 'exit', 'quit', or Ctrl-D.
>   warn("To exit: use 'exit', 'quit', or Ctrl-D.", stacklevel=1)

Expected:
`--help` parameter in standalone mode does not call `sys.exit`.

sirosen:
Based on the work in #533 

However, I think that PR did more than was necessary and got itself into trouble. Exits don't need to pretty print themselves, and that would actually be a destructive or at least annoying change for many `click` consumers.
Rather than squashing that work away, I've preserved those commits but made further changes to pare things back.

This makes a new class of exception `Exit(RuntimeError)`, which is just a container for an exit code.
`Context.exit` raises an `Exit` instead of calling `sys.exit` directly, and `BaseCommand.main` looks at `Exit` exceptions a lot like it looks at `Abort`s, either returning the result (when non-standalone) or calling `sys.exit` (when standalone).
(This is all very much in line with @jamesmyatt's feedback on that original work, which I largely agree with.)

closes #533
closes #667

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ Version 7.0
 
 (upcoming release with new features, release date to be decided)
 
+- Non-standalone calls to Context.exit return the exit code, rather than
+  calling ``sys.exit`` (`#667`_)(`#533`_)
 - Updated test env matrix. (`#1027`_)
 - Fixes a ZeroDivisionError in ProgressBar.make_step,
   when the arg passed to the first call of ProgressBar.update is 0. (`#1012`_)(`#447`_)
--- a/click/core.py
+++ b/click/core.py
@@ -714,9 +714,14 @@ class BaseCommand(object):
                     rv = self.invoke(ctx)
                     if not standalone_mode:
                         return rv
-                    if not rv:
-                        rv = 0
-                    ctx.exit(rv)
+                    # it's not safe to `ctx.exit(rv)` here!
+                    # note that `rv` may actually contain data like "1" which
+                    # has obvious effects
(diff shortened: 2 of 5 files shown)


dsully:
I like this change, thanks!

@davidism - what do you think?

jamesmyatt:
Thanks @sirosen

PR #1106, opened 2018-09-05
Prevent error messages/exit code 120 after exiting on EPIPE

### LINKED ISSUE #712: BrokenPipeError when piping over stdin
I was very excited to see the contributions come in over #626 -- thank you for that. However, it looks like a `SIGPIPE` issue is still possible when piping in over stdin. An example is shown below. I want to note that I have not been able to absolutely isolate the problem as can be seen below where a small test is fine but a larger test results in the exception.

```bash
$ python --version
Python 3.5.2 :: Continuum Analytics, Inc.
$ python -c "import click; print(click.__version__)"
6.7
$ uname -a
Darwin codinator.local 16.3.0 Darwin Kernel Version 16.3.0: Thu Nov 17 20:23:58 PST 2016; root:xnu-3789.31.2~1/RELEASE_X86_64 x86_64
$ cat pipeerror
#!/usr/bin/env python

import click

@click.command()
@click.option('--bar', type=click.File('r'))
def foo(bar):
    for i in bar:
        click.echo(i)

if __name__ == '__main__':
    foo() 
$ seq 1 10 | python pipeerror --bar - | head -n 2
1

$ seq 1 100 | python pipeerror --bar - | head -n 2
1

Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>
BrokenPipeError: [Errno 32] Broken pipe
```

frankier:
Fixes https://github.com/pallets/click/issues/712 . Compared to the workarounds there, this is a little more careful about the possibility of odd interactions if sys.stdout/sys.stderr being replaced and then some kind of other cleanup running after sys.exit(...) or the possibility of EPIPE not being related to stdout/stderr (in which case they really should be flushed at program exit).

Side point: I'm not sure if the current approach is the best possible. In particular I think the choice of exiting with an exit code of 1 is somewhat questionable. At least the sigpipe handler should probably be customisable "at the click level" now the current behaviour has been released.

### DIFF
--- a/click/utils.py
+++ b/click/utils.py
@@ -414,3 +414,27 @@ def get_app_dir(app_name, roaming=True, force_posix=False):
     return os.path.join(
         os.environ.get('XDG_CONFIG_HOME', os.path.expanduser('~/.config')),
         _posixify(app_name))
+
+
+class PacifyFlushWrapper(object):
+    """This wrapper is used to catch and suppress BrokenPipeErrors resulting
+    from ``.flush()`` being called on broken pipe during the shutdown/final-GC
+    of the Python interpreter. Notably ``.flush()`` is always called on
+    ``sys.stdout`` and ``sys.stderr``. So as to have minimal impact on any
+    other cleanup code, and the case where the underlying file is not a broken
+    pipe, all calls and attributes are proxied.
+    """
+
+    def __init__(self, wrapped):
+        self.wrapped = wrapped
+
+    def flush(self):
+        try:
+            self.wrapped.flush()
+        except IOError as e:
+            import errno
+            if e.errno != errno.EPIPE:
+                raise
+
+    def __getattr__(self, attr):
+        return getattr(self.wrapped, attr)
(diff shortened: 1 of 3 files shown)


click/utils.py
```diff
@@ -414,3 +414,17 @@ def get_app_dir(app_name, roaming=True, force_posix=False):
     return os.path.join(
         os.environ.get('XDG_CONFIG_HOME', os.path.expanduser('~/.config')),
         _posixify(app_name))
+
+
+class PacifyFlushWrapper:
```
dsully: Please add a doc string describing what this class is used for and why.

Since we're still on Python 2 & 3, have it inherit from 'object' to be a new style class as well.

click/utils.py
```diff
@@ -414,3 +414,25 @@ def get_app_dir(app_name, roaming=True, force_posix=False):
...
+    pipe, all calls and attributes are proxied.
+    """
+
+    def __init__(self, wrapped):
+        self.wrapped = wrapped
+
+    def flush(self):
+        try:
+            self.wrapped.flush()
+        except BrokenPipeError:
```
davidism: Needs to be `IOError` and check `e.errno` like above for Python 2 compatibility.

davidism (changes requested):

davidism (approved):

sirosen:
I would merge, but this is missing from the changelog.

@frankier, if you add a changelog entry, I think this is good to move forward.

Issue #822, opened 2017-07-09
Inconsistent return type from click.getchar()


pkch:
It [seems](https://github.com/python/typeshed/pull/1473#issuecomment-313962504) that `click.getchar()` will return `bytes` under python 3 / Windows, and `str` under any other combination of python version and platform. Is this the intended behavior or a bug?

I wanted to update the stub for `click` but if it's a bug and can be fixed soon, I can leave the typeshed stub unchanged.

ScottSturdivant:
The [documentation ](http://click.pocoo.org/6/api/#click.getchar) states: 

> This will always return a unicode character

I've witnessed the same behavior (``bytes`` w/Python3 and Windows, ``str`` elsewhere) and to me, this inconsistency is a bug.

kip-13:
The python 3 [doc's](https://docs.python.org/2/library/msvcrt.html#msvcrt.getch) is clear with this:
> Read a keypress and return the resulting character as a byte string

In python 2 the [doc](https://docs.python.org/2/library/msvcrt.html#msvcrt.getch) say this:
> Read a keypress and return the resulting character

JelleZijlstra:
Those are docs for the standard library, not for click.

kip-13:
@JelleZijlstra I know, but if you see the code of [getchar](https://github.com/pallets/click/blob/master/click/_termui_impl.py#L519) you understand why I mentioned the module `msvcrt`

sirosen:
Closed by #1108

Issue #821, opened 2017-07-09
Handling of special characters in getchar


pkch:
When the user presses arrow up, the `click.getchar()` returns b'\xe0'. This behavior is probably good, but I guess it should be documented (along with other special characters), since it's not a valid utf-8 value even though the terminal encoding is utf-8. (The naive `click.getchar().decode('utf-8')` would crash on this example with `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe0 in position 0: unexpected end of data`.)

fdavis:
hey @pkch is this still an issue for you? I don't seem to get this running on master (different python3 invocation because I seem to get a str type back in py3)

```
(.clickve) [~/git/click]$ python2 -c 'import click;print(click.getchar().decode("utf-8").__repr__())'
u'\x1b[A'
(.clickve) [~/git/click]$ python -c 'import click;print(bytes(click.getchar(),"utf-8").__repr__())'
b'\x1b[A'
```

Issue #537, opened 2016-03-15
click.getchar does not return Unicode on Windows


intrinseca:
Test Script (Python 3.4.3 on Windows 7 x64):

```
import click
print(click.getchar()) # press 'a' key, prints "b'a'" i.e. a byte string
```

When calling `click.getchar()` on Windows `mscvrt.getch` is used in `_termui_impl.getchar` rather than `mscvrt.getwch`.

This means you get a byte string back, which defeats the Ctrl-C/Ctrl-D processing in `_translate_ch_to_exc` and breaks comparisons with the return value. The [documentation for msvcrt](https://docs.python.org/3.4/library/msvcrt.html) advises that the wide char functions should be used wherever possible.

I've patched this over by changing `getchar` to the following (also with a tweak to use the built in echoing rather than `putchar`, which is a typo for `putch` anyway):

```
def getchar(echo):
    if echo:
        rv = msvcrt.getwche()
    else:
        rv = msvcrt.getwch()

    _translate_ch_to_exc(rv)
    if PY2:
        enc = getattr(sys.stdin, 'encoding', None)
        if enc is not None:
            rv = rv.decode(enc, 'replace')
        else:
            rv = rv.decode('cp1252', 'replace')
    return rv
```

But I don't really want to embark on rigorously testing this (e.g. on Python 2), given the complexity of all the encoding handling. I'm not sure how you would even write a test for this, since it's so tied to the console I/O?

(if click uses the wide char API, then does it even need the `PY2` branch?)

sirosen:
Closed by #1108

PR #1108, opened 2018-09-12
make `getchar` work on Windows.

### LINKED ISSUE #1088: Characters getting from terminal using getchar returns bytes
Hello,
it is said in the documentation that _getchar_ function **always returns unicode** [there](http://click.pocoo.org/6/utils/#getting-characters-from-terminal), but when i tried to use the code snippet of the documentation on my **windows 10** machine, i always fell on `"invalid input :("`

I inspected the type of value returned and I realized that it was _**bytes**_. So either you have to change the documentation and the code like this:
```python
    click.echo('Continue? [yn] ', nl=False)
    c = click.getchar()
    if c == b'y':
        click.echo('We will go on')
    elif c == b'n':
        click.echo('Abort!')
    else:
        click.echo('Invalid input :(')
```

either the problem have to be solved..

I'm using click 6.7, python 3.6 and I'm working under windows 10.

matejcik:
Fixes #1088, #822, #821, and #537.

No tests because the test suite can't properly test `_termui_impl` functions, AFAICT at least. Manually tested on Windows 10 with Python 2.7 and 3.6.

The implemented behavior is not the best (see the comment; it's impossible to read French letter "à" as a single character), but it does fix a long-standing issue in a reasonable way.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Version 7.0
 
 Unreleased
 
+-   ``click.getchar()`` now returns unicode in Python 3 on Windows, same as other platforms. (`#1108`_, `#1088`_, `#822`_, `#821`_, `#537`_)
 -   Drop support for Python 2.6 and 3.3. (`#976`_)
 -   Wrap ``click.Choice``'s missing message. (`#202`_, `#1000`_)
 -   Add native ZSH autocompletion support. (`#323`_, `#865`_)
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -529,7 +529,9 @@ def open_url(url, wait=False, locate=False):
 def _translate_ch_to_exc(ch):
     if ch == '\x03':
         raise KeyboardInterrupt()
-    if ch == '\x04':
+    if ch == '\x04' and not WIN:  # Unix-like, Ctrl+D
+        raise EOFError()
+    if ch == '\x1a' and WIN:      # Windows, Ctrl+Z
         raise EOFError()
 
 
@@ -541,16 +543,46 @@ if WIN:
         yield
 
     def getchar(echo):
-        rv = msvcrt.getch()
+        # The function `getch` will return a bytes object corresponding to
+        # the pressed character. Since Windows 10 build 1803, it will also
+        # return \x00 when called a second time after pressing a regular key.
+        #
+        # `getwch` does not share this probably-bugged behavior. Moreover, it
+        # returns a Unicode object by default, which is what we want.
+        #
+        # Either of these functions will return \x00 or \xe0 to indicate
+        # a special key, and you need to call the same function again to get
+        # the "rest" of the code. The fun part is that \u00e0 is
+        # "latin small letter a with grave", so if you type that on a French
+        # keyboard, you _also_ get a \xe0.
+        # E.g., consider the Up arrow. This returns \xe0 and then \x48. The
+        # resulting Unicode string reads as "a with grave" + "capital H".
+        # This is indistinguishable from when the user actually types
+        # "a with grave" and then "capital H".
+        #
+        # When \xe0 is returned, we assume it's part of a special-key sequence
+        # and call `getwch` again, but that means that when the user types
+        # the \u00e0 character, `getchar` doesn't return until a second
+        # character is typed.
+        # The alternative is returning immediately, but that would mess up
+        # cross-platform handling of arrow keys and others that start with
+        # \xe0. Another option is using `getch`, but then we can't reliably
+        # read non-ASCII characters, because return values of `getch` are
+        # limited to the current 8-bit codepage.
+        #
+        # Anyway, Click doesn't claim to do this Right(tm), and using `getwch`
+        # is doing the right thing in more situations than with `getch`.
         if echo:
-            msvcrt.putchar(rv)
+            func = msvcrt.getwche
+        else:
+            func = msvcrt.getwch
+
+        rv = func()
+        if rv in (u'\x00', u'\xe0'):
+            # \x00 and \xe0 are control characters that indicate special key,
+            # see above.
+            rv += func()
         _translate_ch_to_exc(rv)
-        if PY2:
-            enc = getattr(sys.stdin, 'encoding', None)
-            if enc is not None:
-                rv = rv.decode(enc, 'replace')
-            else:
-                rv = rv.decode('cp1252', 'replace')
         return rv
 else:
     import tty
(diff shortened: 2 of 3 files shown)


matejcik:
hmm, this is a change of behavior though, so maybe not appropriate for a maintenance branch?

To reiterate: with Python 3 on Windows, `click.getchar()` would return `bytes` objects, contrary to the documentation (that [states](http://click.pocoo.org/6/utils/#getting-characters-from-terminal) that it always returns an Unicode character), and contrary to behavior in Python2+Windows, or Any+POSIX. This PR changes it so that `getchar()` returns Unicode everywhere.

In addition, `getchar()` on Windows (regardless of Python version) would fail to read non-ASCII characters properly, and since Win10 build 1803, it would return `\x00` on every other call. Both of these problems are fixed here.

davidism:
Thank you for the pull request and the detailed code comments.

This function recurses indefinitely, but the Windows docs for `getwch` and your comment seem to indicate that it only needs to be called twice.

Is the issue with à that it will show up as two bytes rather than one, not that it can't be input at all? Is that something that can / should be solved with Unicode normalization?

matejcik:
Hmmm, apparently I'm wrong about what `getch` returns for special characters; it's their code from the DOS codepage. Perhaps using `sys.stdin.encoding` would be the way to go after all? I'll have to fix the comment in either case.

I don't see what you mean by "recursing indefinitely". The second call calls `func` which is `getwch/e`, so it's called at most twice.

The issue with "à" is that if you press it, `getchar()` doesn't return because it's waiting for a second character. Then you'll get two characters at once.

davidism:
@pallets/click or @mitsuhiko, would another maintainer using Windows check this out? I'm not familiar with Windows terminal behavior.

davidism:
I'm worried that the `encoding` handling was there for a reason and this removes that. Also, I don't think it's good that `getchar` might wait for a second character with no indication why; it's still not clear to me why that even happens.

sirosen:
I've been testing this on a Windows 2016 Server and Python 3.7 (it's what I have most readily available). The results make sense of the code, but don't jive perfectly with the comments.
I can retest on py3.6 if we think that will make a difference.

Encoding handling was needed to deal with the fact that `msvcrt.getch` returns bytes, but `msvcrt.getwch` returns a unicode. That old code can/should be removed unless someone can point out a case where `sys.stdin.encoding` has a value which gives us pause.

On a French keyboard, "e with grave accent" works with a single call to `msvcrt.getwch`. From the comments it sounds like this is expected to require two calls? @matejcik, can you clarify what you mean by that comment? Are you seeing `é` require multiple calls?

The msvcrt docs seem to say that two calls are only needed to handle function keys and other exotic keycodes, not general multiple-byte characters. That's the behavior I see as well -- if I hit a function key, I need to call `getwchar` twice to get the results.

Regardless of the disagreement between what I see and the comment about `é'`, the new behavior for `click.getchar` appears to be correct as far as I'm able to test and push it.

I'd like to hear back about that one discrepancy, and then I'd push ahead and merge this into `master` and get it into the 7.0 release. @matejcik, do you feel strongly that you'd like to see this bug fixed in 6.x ? Or would you be okay with it getting fixed in 7.0 ?
Making the change in a major version bump just provides a little bit of CYA, in case the fix breaks someone.

matejcik:
The problem is with a single particular character, `U+00E0` "latin small letter a with grave", or "à". It does not affect other non-ascii or accented characters. And it stems from the fact that the codepoint, `\u00E0`, represented in Python's Unicode type as character `\xe0`, aliases to the special signaling byte `\xe0`.

Try this to reproduce:
```python
import msvcrt
first = msvcrt.getwch()
second = msvcrt.getwch()
print(first, second)
```
For first attempt, press the Up arrow. The result is `\xe0` and `H`:
```
à H
```
For second attempt, use a French keyboard and type "àH":
```
à H
```
The way I wrote this PR, signaling bytes `\x00` and `\xe0` are caught, so that when you use `getchar()` and press a special key (e.g., the Up arrow), you will get a single return value. In this case, your return value is the Unicode string `"àH"`.
However, due to the aliasing, this is _impossible to distinguish_ from the situation when the user actually types the letter "à". The code will think that the `\xe0` it received is signaling byte, and calls `getwch()` again.
When a user is typing text that randomly contains "à", this doesn't matter. The next letter is returned in the next `getwch()` call, and in the end you get all the letters that were typed (unless "à" is the last one).

However, when you ask the user to "press any key", and they choose to press "à", nothing will happen until they also press a second key.

I welcome suggestions for improving the comment; apparently this problem is convoluted enough to warrant a clearer explanation?

---

I'm ok with waiting for 7.0, after all this is a potentially breaking change in behavior.

sirosen:
I've changed the PR base to `master`, not 6.x-maintenance.

I don't think the comment needs improvement. Last night I was testing this with letters with acute accents (you can see my `é`s above) and getting myself confused. Thanks for explaining one more time.
If you _want_ to improve it, you could note specifically that `àH` and up-arrow are indistinguishable, as an example. That helped me have the "aha!" moment of realizing I had tested this incorrectly.

I tried `msvcrt.kbhit` in the hope that it would maybe give us a way to return immediately on a true `à`, but it does not. 😢

There's another reasonable behavior available: to have `click.getchar` _always_ return after the first character read by `msvcrt.getchw`, even if it's `\xe0` or `\x00`. However, I think this leaves a minefield, especially for inexperienced developers, when a user hits up-arrow or similar, so the currently proposed behavior is better.

There are two documentation changes I'd like to see before this merges:

1. Add a note to the documentation something like "in very rare cases, `click.getchar` will wait for a second character of input, as some unicode characters look like parts of special keycodes". That way, we won't get bug reports in the future about `click.getchar` blocking on `à` on Windows.
2. Add an entry to the top of the 7.0 changelog.

davidism:
I'm fine with merging this along with @sirosen's requests.

How do other libraries handle this? Do they just block on à too?

Maybe we could add a a `single=False` switch to always return a single character on Windows in case the user wants "press any key" behavior. Or would a very short timeout of some sort be reasonable, since the special char code should come in at the same time?

sirosen:
I also considered the idea of a flag, e.g. `single=False`. It would come with the caveat that you might be leaving unread characters in the console buffer.
I'd like to leave that as a potential improvement.

A cursory look at other projects on github shows that most people are either catching two chars like we will, or blissfully unaware of this snakepit. I'll keep looking around in case I see something clever, but I think this PR might represent the current state-of-the-art.

davidism:
Marked this for 7.0, let's get the docs in and merge it.

matejcik:
@sirosen done

Issue #1112, opened 2018-09-17
Separate Group's subcommands and commands in help


AuHau:
Coming from the Argparse background, I would appreciate having a possibility to differentiate between additional sub-commands (sub-groups) and commands in help page to increase its clarity when comes to big CLI programs.

Example how I see the help page to look like:
```
Usage: toggl_runner.py [OPTIONS] COMMAND [ARGS]...

Options:
  -q, --quiet    don't print anything
  --help         Show this message and exit.

Sub-Commands:
  clients          clients management
  projects         projects management
  users            users management
  workspaces       workspaces management

Commands:
  add  adds finished time entry
```

janLuke:
Still interested in this? I've just added the possibility of organizing subcommands of a Group in multiple (titled) help sections in an external package called `cloup`. I originally wrote the package to support only "option groups".

I opened a PR on my own repository because I'd like to receive some feedback: https://github.com/janLuke/cloup/pull/1. Also, I invite click mainteiners to give a look and tell me if they would be interested in merging this feature into click. 

In `cloup` you can manually organize your commands in arbitrary sections, but it's easy to write a reusable decorator that monkey-patch a `cloup.Group` to do exactly what you want automatically:
```python
import click
import cloup


# Re-usable decorator
def group(*args, **kwargs):
    """ Returns a monkey-patched cloup.Group with two "default section":
    one for MultiCommand's and one for simple Command's. """
    SUB_COMMANDS_SECTION = cloup.GroupSection.sorted('Sub-commands')
    COMMANDS_SECTION = cloup.GroupSection.sorted('Commands')

    def get_section(cmd):
        if isinstance(cmd, click.MultiCommand):
            return SUB_COMMANDS_SECTION
        return COMMANDS_SECTION

    def decorator(f):
        # Pass "sections" to control the order of sections
        grp = cloup.group(*args, sections=[SUB_COMMANDS_SECTION, COMMANDS_SECTION], **kwargs)(f)

        def add_command(cmd, section=None):
            if section is None:
                section = get_section(cmd)
            cloup.Group.add_command(grp, cmd, section=section)

        grp.add_command = add_command
        return grp

    return decorator

""" ===========================================================
Use the decorator to create your root group, then use it as normal. 
Below, subcommands are cloup.Command/Group but they don't need to be, 
they can be normal click.Command/Group
"""

@group()
def main():
    pass

# Use main.add_command to add commands defined in this or other modules
def f(): return 0
[...]

PR #1113, opened 2018-09-17
Sub-commands help group

AuHau:
Example how could #1112 be implemented. Usage then is:
```
@click.group(cls=SubCommandsGroup)
def cli(...):
     pass
```

It is a question if not to implement it directly in the `Group` class and have a switch parameter which would declare the style of help to use. I am open for guideline regarding that.

This is an example PR. After direction is set I will provide full-fledged PR with test coverage, docs etc.

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -1252,6 +1252,54 @@ class Group(MultiCommand):
         return sorted(self.commands)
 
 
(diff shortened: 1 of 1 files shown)


AuHau:
Closing for no interest.

PR #1129, opened 2018-10-01
Added `-h` alias for `--help` option.

thechief389:


davidism:
This is intentional. It's common to have options such as `--host` where `-h` would be more useful. As shown in the example you removed, Click already provides a way to customize this.

PR #1146, opened 2018-10-14
Handle non text in style/secho by converting to text

### LINKED ISSUE #944: No color support for Exceptions
Right now, the click module gives us the option of using colors only through click.echo and click.secho commands. I was thinking that if we can tweak the classes inside the exceptions.py file and use secho inside them in place of echo, then we can give the users an option of colorizing error/exception messages.

segevfiner:
Call `str()` on any non-string type. `echo()` also supports printing `bytes`, but this PR doesn't support that, ending up printing its repr (`b'bytes'`). I'm not sure how to handle it, since making `style()` support bytes might be rather ugly.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,6 +39,8 @@ Unreleased
     :issue:`1475`
 -   Help for boolean flags with ``show_default=True`` shows the flag
     name instead of ``True`` or ``False``. :issue:`1538`
+-   Non-string objects passed to ``style()`` and ``secho()`` will be
+    converted to string. :pr:`1146`
 
 
 Version 7.1.2
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -482,11 +482,6 @@ def style(
     * ``bright_white``
     * ``reset`` (reset the color code only)
 
-    .. versionadded:: 2.0
-
-    .. versionadded:: 7.0
-       Added support for bright colors.
-
     :param text: the string to style with ansi codes.
     :param fg: if provided this will become the foreground color.
     :param bg: if provided this will become the background color.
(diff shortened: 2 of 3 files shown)


davidism:
Within `echo`, there's already a check to not try to strip bytes of color information. This makes sense, as it wouldn't make sense to color raw bytes being output (even if they do represent text). So I think it makes sense to only do `str(message)`.

That said, I did try an implementation for decoding bytes passed to `secho`, but decided against it. It had to be in `secho` and not `style` since only `secho` has information about the stream being written to.

```python
from ._compat import get_best_encoding

# extracted from echo for use in secho as well
def _get_echo_file(file, err):
    if file is not None:
        return file

    if err:
        return _default_text_stderr()
    else:
        return _default_text_stdout()

def secho(message, file=None, err=False, ...):
    file = _get_echo_file(file, err)

    if message is not None:
        if is_bytes(message):
            file = _get_echo_file(file, err)
            encoding = get_best_encoding(file)
            message = message.decode(encoding)
        elif not isinstance(message, str):
            message = str(message)

        message = style(message, **styles)

    return echo(message, file=file, err=err, ...)
```

This isn't completely correct, it doesn't pull out the `errors` attribute of the stream for decoding. In the end, this seemed a bit too complicated. Since 8.0 is Python 3-only, it also makes more sense to require text for output, requiring users to explicitly decode bytes being passed is the common pattern now.

davidism:
I don't think this addresses #944, that seems to be about allowing `ClickException.show()` to use `secho()` instead of `echo()`.

davidism:
I guess we could also do `not is_bytes(message)` before applying `style()` in `secho()`. That way bytes get passed directly to `echo()`.

Issue #1140, opened 2018-10-09
Longest option argument not used as internal/variable name


altendky:
It seems that with 7.0 the longest option argument isn't being used as the internal and variable name.

https://github.com/pallets/click/blob/2c622eed402c64e1645db3904c2410f6c05fbf19/docs/parameters.rst#L93-L95
> the internal name is generated automatically by taking the longest argument

```python3
import click
import click.testing


print(click.__version__)

@click.command()
@click.option(
    '--temp',
    '--temporary',
    'temporary',
)
def good(temporary):
    pass


@click.command()
@click.option(
    '--temp',
    '--temporary',
)
def bad(temporary):
    pass


def click_runner(command):
    print(command.name)
    runner = click.testing.CliRunner()
    result = runner.invoke(
        command,
        [],
        catch_exceptions=False,
    )
    print(result.output)

click_runner(command=good)
click_runner(command=bad)

```
https://repl.it/@altendky/click-67-unexpected-keyword-argument-1
```python3
6.7
good

bad

```

https://repl.it/@altendky/click-70-unexpected-keyword-argument-1
```python3
7.0
good

bad
Traceback (most recent call last):
  File "python", line 37, in <module>
  File "python", line 32, in click_runner
TypeError: bad() got an unexpected keyword argument 'temp'
```

https://repl.it/@altendky/click-2e856a5-unexpected-keyword-argument
```python3
7.0
good

bad
Traceback (most recent call last):
  File "python", line 37, in <module>
  File "python", line 32, in click_runner
TypeError: bad() got an unexpected keyword argument 'temp'
```

altendky:
Seems to relate to #793/#794 where a different part of the docs were referenced.  So I guess the question is which one do we want?

options:
https://github.com/pallets/click/blob/b1b8c48345d8accc4dea262b8a41ac5ac1de3af2/docs/options.rst#L41-L43

vs parameters:
https://github.com/pallets/click/blob/2c622eed402c64e1645db3904c2410f6c05fbf19/docs/parameters.rst#L93-L95

amiryal:
@altendky I would suggest to revert to the old (pre #794) behaviour to fix the regression, and align the documentation with actual implementation instead.

davidism:
Thansk! Seems better to revert to the old behavior and update the docs.

altendky:
@davidism should the PR be against `7.x` or `master`?  Am I supposed to modify `CHANGES.rst`?

davidism:
[...]

PR #1149, opened 2018-10-16
[WIP] [1140] use longest option argument again

altendky:
#1140

### DIFF
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -462,14 +462,15 @@ def test_aliases_for_flags(runner):
     result = runner.invoke(cli_alt, ['-w'])
     assert result.output == 'True\n'
 
+
 @pytest.mark.parametrize('option_args,expected', [
     (['--aggressive', '--all', '-a'], 'aggressive'),
-    (['--first', '--second', '--third', '-a', '-b', '-c'], 'first'),
-    (['--apple', '--banana', '--cantaloupe', '-a', '-b', '-c'], 'apple'),
+    (['--first', '--second', '--third', '-a', '-b', '-c'], 'second'),
+    (['--apple', '--banana', '--cantaloupe', '-a', '-b', '-c'], 'cantaloupe'),
     (['--cantaloupe', '--banana', '--apple', '-c', '-b', '-a'], 'cantaloupe'),
     (['-a', '-b', '-c'], 'a'),
     (['-c', '-b', '-a'], 'c'),
-    (['-a', '--apple', '-b', '--banana', '-c', '--cantaloupe'], 'apple'),
+    (['-a', '--apple', '-b', '--banana', '-c', '--cantaloupe'], 'cantaloupe'),
     (['-c', '-a', '--cantaloupe', '-b', '--banana', '--apple'], 'cantaloupe'),
     (['--from', '-f', '_from'], '_from'),
     (['--return', '-r', '_ret'], '_ret'),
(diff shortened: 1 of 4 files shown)


altendky:
WIP for:
* [ ] Doc fixes

amiryal (approved):
My only feedback, other than LGTM 👍, is regarding commit messages. Maybe you could provide more context, i.e. what the change actually does and why, and possibly squash some or all of them together.

amiryal:
Just wanted to echo your earlier https://github.com/pallets/click/issues/1140#issuecomment-432291544 here as well:
> It's kind of tempting to deprecate this 'feature' and just say 'if you have multiple of the longest type of parameter, specify the name'. Or maybe provide a marker so you don't have to repeat. Maybe `'+--the-default'` or somesuch? I dunno, explicit over implicit and all. This just seems like a bit of a fiddly mess with little benefit.
> Sorry, make it backwards compatible but fix the situation by deprecating the implicit order-based selection.

:+1: 
I fully agree, Zen of Python and all.

tests/test_options.py
```diff
@@ -462,14 +462,15 @@ def cli_alt(warnings):
     result = runner.invoke(cli_alt, ['-w'])
     assert result.output == 'True\n'
 
+
 @pytest.mark.parametrize('option_args,expected', [
     (['--aggressive', '--all', '-a'], 'aggressive'),
-    (['--first', '--second', '--third', '-a', '-b', '-c'], 'first'),
-    (['--apple', '--banana', '--cantaloupe', '-a', '-b', '-c'], 'apple'),
+    (['--first', '--second', '--third', '-a', '-b', '-c'], 'second'),
```
sirosen: This makes me so dizzy to read. 😂 

I understand that we're going back to documented behavior, but this is pretty much the perfect example of why I think [this comment of yours is spot on](https://github.com/pallets/click/issues/1140#issuecomment-432291544). We should really just tell people to write down what they mean...

sirosen (commented):
@altendky, is this still a work in progress?

Since this is the course of action which you and @davidism agreed on, and the changes appear to be complete to me, I'd like to merge.

I agree that your commit messages are a bit minimal -- personally, I'd prefer them either squashed or fleshed out with explanation -- but all of the important discussion is in GitHub anyway.

altendky:
After https://github.com/pallets/click/issues/1140#issuecomment-432304387 and https://github.com/pallets/click/issues/1140#issuecomment-432307145 I figured we should figure out where we are going since we have no totally obvious thing to revert to.  If we want to 'just revert' for now I would think we would want the previous functionality which would be 'last long' not 'longest' like this implements.

sirosen:
Ah, I thought after [this comment that we should revert the behavior](https://github.com/pallets/click/issues/1140#issuecomment-429924902) that this reverts #794.

Reading that PR again, it looks like we switched from "last long opt" to "first long opt" but never updated the docs? I didn't realize we never had the also-documented "longest option" behavior.

Sorry for jumping in on this without properly grasping the situation.

Now that I do, I actually think this PR is probably a bad call -- we should either keep the new behavior in 7.0 and document it, or revert to pre #794 behavior and note it as a bugfix in the 7.1 changelog.

IMO, keeping the new behavior and just updating the docs is the cheapest/easiest solution. If someone showed up as a new user of 7.0, then any behavior change appears to be breaking.

altendky:
I only left this open to avoid thrash just in case... But yeah, this is likely just totally dead.

PR #1151, opened 2018-10-19
Fix get_current_context typo in docstring

moshe:


sirosen:
Thanks for spotting this.

PR #1169, opened 2018-11-15
Draw attention to an easy mistake to make

Alphare:
Hi.

I just spent 10 minutes trying to figure out why my very simple code would not work, and this change would have helped me a lot. :smiley: 

There is an explanation in the "Nesting commands" section, but if you - like me - tried directly going for the "Commands" section, you might have ran into the same issue I did.

 Thanks for `click`, it's awesome!

### DIFF
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -27,7 +27,7 @@ when an inner command runs:
     def cli(debug):
         click.echo('Debug mode is %s' % ('on' if debug else 'off'))
 
-    @cli.command()
+    @cli.command()  # @cli, not @click!
     def sync():
         click.echo('Syncing')
 


ThiefMaster:
You need to rebase this PR against 7.x or change it to master - currently it tries to merge all the stuff from master into 7.x.

davidism:
From your branch you *should* be able to do `git rebase --onto 7.x master` to fix this. After that `git log` should show that it's close to the 7.x branch, then you can do `git push -f` to update the PR.

davidism:
Rebased onto 7.x.

Alphare:
@davidism Thanks, I tried doing so as soon as I opened the PR and saw the other commits, but you were faster.

PR #1172, opened 2018-11-20
click.edit accepts binary data (bytes)

### LINKED ISSUE #1171: click.edit doesn't support binary
I am using python 3.7.1 and click 7, and passing a bytes object to edit causes an exception on 
https://github.com/pallets/click/blob/b4d25f807e8a98f87bc60b76839e83302a5d9f3d/click/_termui_impl.py#L439

Where bytes.endswith() cannot accept a str as an argument.

From looking at the code we could add a "binary" flag to edit() or check to see if the object is bytes and skip all the encoding business and write bytes, then read and return bytes.

russkel:
This fixed #1171. click.edit did not handle binary data and treated it as if it were a string. This checks to see if it is a bytes object and then treats it accordingly by not appending newlines or changing encoding. This enables using a hexeditor such as hexedit with binary data.

### DIFF
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -436,17 +436,20 @@ class Editor(object):
         import tempfile
 
         text = text or ''
-        if text and not text.endswith('\n'):
+        binary_data = type(text) is bytes
+
+        if not binary_data and text and not text.endswith('\n'):
             text += '\n'
 
         fd, name = tempfile.mkstemp(prefix='editor-', suffix=self.extension)
         try:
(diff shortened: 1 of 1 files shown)


dsully:
Looks good to me.

davidism:
@dsully needs a changelog entry

davidism:
Also needs to be cherry-picked to 7.x branch, unless you think it should wait for 8.0.

dsully:
Sure, will do.

PR #1174, opened 2018-11-25
Support generator as choice.

corvofeng:
## When I want dynamic parameters

I think if I could use iterator as an array. So, I wrote the following code, you know, it's work, but the help message is not appropriate. Hope my commits help it.

```python
def iter_arr():
    ret_arr = ['man', 'woman', 'unknow']
    for item in ret_arr:
        yield item


@click.command()
@click.option('--gender', type=click.Choice(iter_arr()))
def cli(gender):
    click.echo(gender)
```

image

Best wishes.

### DIFF
--- a/click/types.py
+++ b/click/types.py
@@ -142,7 +142,7 @@ class Choice(ParamType):
     name = 'choice'
 
     def __init__(self, choices, case_sensitive=True):
-        self.choices = choices
+        self.choices = list(choices)
         self.case_sensitive = case_sensitive
 
     def get_metavar(self, param):


sirosen:
I might be alone in this opinion, but I think the current click behavior is correct and shouldn't be changed.
The choice type takes an iterable of choices. If you give it a generator, it will only be able to iterate on that generator once.

Consider code like this:

```python
choices = []

@click.command()
@click.option('--foo', type=click.Choice(choices))
def cli(foo):
    click.echo('foo: {}'.format(foo))

choices.append('val1')
choices.append('val2')
```

Today this works, but if `click.Choice` instantiation forces evaluation of the choices by listifying it, it will not.
Handling this in fancy ways is possible (e.g. use a `@property` and call `list` on first access), but I don't think it's worth making the interface to `click.Choice` more complex.

If another maintainer agrees, I'd like to close this as a wontfix.

PR #1188, opened 2018-12-20
Formatted code as per PEP8

pravarag:
Formatted code in click/ for core.py, decorators.py & few other code files as per PEP8 guidelines.

### DIFF
--- a/click/_compat.py
+++ b/click/_compat.py
@@ -628,7 +628,7 @@ if WIN:
             def _safe_write(s):
                 try:
                     return _write(s)
-                except:
+                except Exception:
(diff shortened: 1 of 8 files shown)


davidism:
Thanks for your interest, but we don't accept these types of prs.

click/_compat.py
```diff
@@ -628,7 +628,7 @@ def auto_wrap_for_ansi(stream, color=None):
             def _safe_write(s):
                 try:
                     return _write(s)
-                except:
+                except Exception:
```
ThiefMaster: fyi, when re-raising at the end, a bare `except` is fine

PR #1190, opened 2018-12-21
superfluous word ;-)

m-u-xyz:
still not quite right …


ThiefMaster:
i think your second commit went to the wrong branch/PR

smurfix:
Yep. Already fixed.

PR #1191, opened 2018-12-21
Add an internal ExitStack to click's Context object

### LINKED ISSUE #1245: Run subcommand in context manager
If would like to run a subcommand inside of a context manager that would be setup in a higher command. How is it possible to do that with click? My pseudo-code looks something like:

```python

import click

from contextlib import contextmanager

@contextmanager
def database_context(db_url):
    try:
        print(f'setup db connection: {db_url}')
        yield
    finally:
        print('teardown db connection')


@click.group
@click.option('--db',default='local')
def main(db):
    print(f'running command against {db} database')
    db_url = get_db_url(db)
    connection_manager = database_context(db_url)
    # here come the mysterious part that makes all subcommands
    # run inside the connection manager

@main.command
def do_this_thing()
    print('doing this thing')

@main.command
def do_that_thing()
    print('doing that thing')

```

And this would be called like:

```
> that_cli do_that_thing
running command against local database
setup db connection: db://user:pass@localdb:db_name
doing that thing
teardown db connection

> that_cli --db staging do_this_thing
running command against staging database
setup db connection: db://user:pass@123.456.123.789:db_name
doing this thing
teardown db connection
```

m-u-xyz:
This patch allows adding context managers to a command group's common code that are then usable in the group's statements. This reduces code duplication and lets us move code using the group's options closer to the options themselves.

fixes #1245

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -60,6 +60,11 @@ Unreleased
     -   Core objects use ``super()`` consistently for better support of
         subclassing.
 
+-   Use ``Context.with_resource()`` to manage resources that would
+    normally be used in a ``with`` statement, allowing them to be used
+    across subcommands and callbacks, then cleaned up when the context
+    ends. :pr:`1191`
+
 
 Version 7.1.2
 -------------
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -406,3 +406,77 @@ method can be used to find this out.
     println()
     invoke(cli, prog_name='cli', args=[])
     println()
+
+
+Managing Resources
+------------------
+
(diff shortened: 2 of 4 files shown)


davidism:
Renamed the method to `with_resource` because I felt `enter_context` was confusing as a method on `Context`. The new name reflects that the resource would be managed with the `with` statement. Moved the docs to `advanced.rst` and rewrote them, along with the doc strings for `with_resource` and `call_on_close`. `ExitStack.__enter__` doesn't have to be called, and doesn't return a different object, so now the implementation only tracks a single attribute called `_exit_stack`, and calls the `close` method instead of `__exit__`.

PR #1195, opened 2018-12-28
Remove unused compat shim for 'bytes'

jdufresne:
Never used. Both Python 2.7 and Python 3 have the type bytes. On Python
2.7 it is an alias for str, same as what was previously defined.

### DIFF
--- a/click/_compat.py
+++ b/click/_compat.py
@@ -155,7 +155,6 @@ class _FixupStream(object):
 
 if PY2:
     text_type = unicode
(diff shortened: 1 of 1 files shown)


Julian:
Not a click author, but seems to me that it's unlikely that the point of that line was to be "useful", otherwise the next line (`raw_input = raw_input`) is even more transparently pointless.

I suspect instead that the point is to explicitly list each cross-compatible name that is to be made available.

PR #1196, opened 2018-12-28
Meaningful filename for old globals.py

julianogouveia:
Reading the click project code, I found a poor meaning for the globals.py file. I believe that the name context.py is better, so I renamed the file for it, so with the get_current_context function for only get_context, which makes the code more understandable and cleaner.

I hope you guys enjoy the contribution.

Thks.


davidism:
Thanks, but renames like this are not backwards compatible and don't really improve the code.

PR #1200, opened 2019-01-02
Add `cli()` call to Nesting Commands section

bbkane:
As a `click` beginner, I got stuck wondering why running the copy-pasted example code wouldn't parse my nested command but also wouldn't throw any errors. Adding this `cli()` call makes it work.


davidism:
Thanks for looking at this. Due to the way the docs work, you can't invoke the command in `click:example::` blocks. Additionally, we don't typically add this type of boilerplate to every example (and there would be many more places to add it if we were being consistent).

PR #1221, opened 2019-02-04
Explicit explanation of the breaking change in 7.0

jjmaestro:
After the discussion in #1123 and seeing how this has bitten people (even if it can be expected because it's a major version and people should pin versions, etc) I think it would be nice to be explicit about such a change and list it in the "Upgrade" section.

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -6,6 +6,12 @@ this is not entirely possible.  In case we need to break backwards
 compatibility this document gives you information about how to upgrade or
 handle backwards compatibility properly.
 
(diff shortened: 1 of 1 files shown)


jjmaestro:
@davidism could you give it a quick check? It's just a change to docs, not code. Just to highlight the issue with v7.0.

Thanks!!

davidism:
Thanks for writing this. I rewrote it to mention the different options for upgrading from https://github.com/pallets/click/issues/1123#issuecomment-589989721 and rebased it to the 7.x branch.

Issue #1150, opened 2018-10-18
Pass option to group main function if such option in child.


eromoe:
Hello,

I want add a `-d` option to all child command, I want some global config work for child command like below:

```
@click.group()
@click.option('-d', '--debug', is_flag=True, default=False)
def cli(debug):
    if debug:
        logger.setLevel(logging.DEBUG)


@cli.command()
def simple():
    logger.debug('xxxx')
```

eromoe:
Could you provide a special option type for this ?

Issue #1104, opened 2018-08-30
Common options for all subcommands 


halloleo:
I have an application with two sub commands `read` and `write`:

```
@click.group()
@click.pass_context
def cli(ctx):
    pass

@cli.command()
@click.pass_context
def read(ctx):
    print("read")

@cli.command()
@click.pass_context
def write(ctx):
    print("read")
```

I want to declare a common option `--format`. I know I can add is as a root option via

```
@click.group()
@click.option('--format', default='json')
@click.pass_context
def cli(ctx, format):
    ctx.obj['format'] = format
```

But then I cannot give the option _after_ the subcommand, which in my use case is a lot more natural. So how can I add a common option in one go to the subcommands instead of the root command?

davidism:
This is intentional. It makes parsing more predictable and callback order predictable.

halloleo:
I know. But is there any way to still achive this?

Issue #108, opened 2014-05-30
Support for shared arguments?


mahmoudimus:
What are your thoughts on shared argument support for click? Sometimes it is useful to have simple inheritance hierarchies in options for complex applications. This might be frowned upon since click's raison d'etre is dynamic subcommands, however, I wanted to explore possible solutions. 

A very simple and trivial example is the [verbose example](http://click.pocoo.org/options/#counting). Assume you have more than one subcommand in a CLI app. An ideal user experience on the CLI would be:

``` bash
python script.py subcmd -vvv
```

However, this wouldn't be the case with click, since subcmd doesn't define a `verbose` option. You'd have to invoke it as follows:

``` bash
python script.py -vvv subcmd
```

This example is very simple, but when there are many subcommands, sometimes a root support option would go a long way to make something simple and easy to use. Let me know if you'd like further clarification.

Stiivi:
I'm giving my vote for this feature, as it makes the CLI experience less complex. Even though technically `cmd -a subcmd -b subsubcmd -c` is correct, `cmd subcmd subsubcmd -a -b -c` is analogous to have `cmd_subcmd_subsubcmd -a -b -c`.

Think of PostgreSQL tools `pg_dump`, `pg_restore` and `pg_ctl` which might be implemented using Click as a single command `pg` with subcommands `dump`, `restore` and `ctl` and they both share options such as `--host`, `--port` or `-D` which come after the whole logical command.

It would be nice if there was a `Group` argument `tail_options = True`, `share_options` or something similar, meaning that all the options will be put at the "tail" of the command chain:

```python
@click.group(tail_options=True)
@click.pass_context
@click.option('--host', nargs=1)
@click.option('--port', nargs=1)
def pg(ctx, host, port):
    # initialize context here

@pg.command()
@click.pass_context
def dump(ctx, ...):
    # ...

@pg.command()
@click.pass_context
def restore(ctx, ...):
    # ...
```

Another problem with current behavior is that `pg restore --help` will not display the `pg` options, which might be confusing to the users. In this case, user will see all the options, including the shared ones, available to the `pg restore`.

*Caveat:* group defines namespace and in this case all options would share a single "command tail" namespace. There should be either a well described behavior in a case of two options (such as "the latest definition overrides the previous one") with the same name or to raise an error.

pjz:
:+1: for easy support for this, either built-in or via a well-marked contrib library

itsmemattchung:
+1 :+1:

[...]

PR #1222, opened 2019-02-05
Loose ordering of options around groups and commands

theodoregoetz:
There are several closed (wont-fix) issues that center around weakening the strict grouping of options to their subcommands: #108 #245 #1104 #1150 . This proposal introduces a new context setting called ``defer_unknown_options`` that when ``True`` along-side having ``allow_interspersed_args == True`` would allow for a loose ordering of options around subcommands. For example if ``--verbose`` is a top-level option, it could be placed after a subcommand and you'd still have the ability to use options associated with the subcommand. Loose ordering, in this case, can be summarized as: the first command to the left consumes the option, if no command to the left knows about the option, the first one to the right will consume the option. If no command understands the option and ``ignore_unknown_options == False`` on the last command issued, then an error is raised.

I realize this is a feature that has been called-out as counter to click's design, but note that the default ``defer_unknown_options == False`` is exactly existing behavior and no unittests were changed as a result of this feature. Further, this is a rather small delta where most of the logic is contained in ``_process_args_for_options()``.

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -196,6 +196,9 @@ class Context(object):
     :param ignore_unknown_options: instructs click to ignore options it does
                                    not know and keeps them for later
                                    processing.
+    :param defer_unknown_options: defer unknown options to subcommands, known
+                                  options will be deferred only if a subcommand
+                                  has overridden it.
     :param help_option_names: optionally a list of strings that define how
                               the default help parameter is named.  The
                               default is ``['--help']``.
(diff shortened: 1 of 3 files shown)


davidism:
How does this interact with autocompletion?

davidism:
I'm not particularly on board with this. It seems that this would be even *more* confusing as now options will work in some places and not others, rather than exactly one place.

davidism:
Thanks for working on this. However, I still think this is against Click's design goals. #938 suggests making the various classes, which might make it easier to implement what you want to do in your project.

Issue #1018, opened 2018-05-17
Allow showing default values globally.


mchwalisz:
Feature request, as in #646, we already had a discussion how to easily show default values for options. As `context_settings` is already used to configure global settings anyway, the idea would be to allow the following:

```python
@click.command(
    help='Command helptext',
    context_settings=dict(show_default=True))
@click.option("--test", default="foobar", help="Option helptext")
def cli(test):
    pass
```

Which should enable `show_default` for all options and give the following help text:

```bash
$ cli --help
Usage: cli [OPTIONS]

  Command helptext

Options:
  --test TEXT   Option helptext. Default: 'foobar'
  --help        Show this message and exit.
```

mchwalisz:
If implemented correctly (haven't looked into the code). It should also work for `@click.group()`

lukeyeager:
Temporary workaround:
```Python
import functools

import click

click_option = functools.partial(click.option, show_default=True)

@click.command()
@click_option('--name', default='Luke')
def say_hi(name):
    print(name)

if __name__ == '__main__':
    say_hi()
```

drorata:
Thanks @lukeyeager for the nice idea. 

Is this enhancement aligns with the package's developers agenda?

jcrotts:
This seems to be fixed in https://github.com/pallets/click/pull/1225 .

drorata:
The merged #1225 is not yet part of version 7.0, is that right? When is it planned to be released?

ehles:
we have a lot of cli commands and a huge more params for them and going to show default, so we very appreciate for releasing this feature.
anyway a year has passed since release 7.0, do you have any plans for releasing 7.1? 
Thanks.

pohmelie:
Spent some times to dig this. It confusing that [documentation](https://click.palletsprojects.com/en/7.x/api/#click.Context) promotes this feature, but it is actually not there.

altendky:
The lack of a `latest` selection for version, or any actual releases, is a bit odd.

![image](https://user-images.githubusercontent.com/543719/73092006-b973f180-3ea9-11ea-80a4-9ffadee1f97d.png)

davidism:
Just like git, `master` contains the latest development code, and 7.x contains the latest 7.x branch code. I'm not sure what's odd about that.

altendky:
![image](https://user-images.githubusercontent.com/543719/73093106-db6e7380-3eab-11ea-8c42-5454641a6570.png)

[...]

PR #1225, opened 2019-02-11
Showing defaults globally

antimirov:
Added show_default parameter to Context, enabling all defaults to be displayed globally. Implementing feature request #1018

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -208,6 +208,9 @@ class Context(object):
                   codes are used in texts that Click prints which is by
                   default not the case.  This for instance would affect
                   help output.
+    :param show_default: if True, shows defaults for all options.
+                    Even if an option is later created with show_default=False,
+                    this command-level setting overrides it.
     """
 
     def __init__(self, command, parent=None, info_name=None, obj=None,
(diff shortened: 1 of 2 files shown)


click/core.py
```diff
@@ -208,6 +208,9 @@ class Context(object):
                   codes are used in texts that Click prints which is by
                   default not the case.  This for instance would affect
                   help output.
+    :param show_default: if True, shows defaults for all options.
+                    Even if an option is later created with show_default=False,
+                    this command-level setting overrides it.
```
davidism: I don't think this should unconditionally override the per-option parameter, unless that's how other context args are working now.

davidism:
It looks like you don't have Git configured correctly. If it matters to you, you should configure `user.email` to match one associated with your GitHub account. Then amend the commit to set the author.

antimirov:
> It looks like you don't have Git configured correctly. If it matters to you, you should configure `user.email` to match one associated with your GitHub account. Then amend the commit to set the author.

Thanks! Fixed it now.

click/core.py
antimirov: But the option's show_default is False by default, so this global command argument does not make sense then. Because that's what it does - shows all defaults without the need to set it explicitly for each option. I can simply drop the last sentence here because it confuses things, in my opinion.

P.S. I did not include any documentation to the .rst files in this PR, do you think it's necessary?

davidism: Ah, ok, that's fine for now then.

sirosen:
Due to the spurious merge commit, I'm going to squash merge this. Hope that's alright.
Also, retargeting to the 7.x branch.

Issue #549, opened 2016-04-01
Support click.option with default values / optional arguments


mbrancato:
Since the default value always applies, there should be a way to differentiate between the following options usage:

file.py
file.py --foo
file.py --foo 2

where '2' is an argument to --foo. This should produce 3 different values for the destination of 'foo'. This is implemented in argparse with nargs='?' but has no functional equivalent in click (that I can identify). I've created a pull request that implements a potential solution. I'm not asking for a specific way of implementing the solution, and what I provided is mostly for a functional example. It does need test cases.
#548

untitaker:
How would you know if `2` is an argument to `file.py` or `--foo`?

ZenithDK:
I had a similar issue.
I want to check that the input values are different from the defaults, and I ended up with this (in case others want the same):
```
@cli.command()
@click.option('--usb')
@click.option('--adk')
@click.pass_context
def local(ctx, usb, adk):
....
    default_dict = dict(zip([command.name for command in ctx.command.params], [command.default for command in ctx.command.params]))
```
Then you can look up the defaults in default_dict - and compare to what you received.

anarcat:
@untitaker ordering. first comes first served.

anarcat:
could this be merged if unit tests are added to #548? this is a feature i'm often missing from click...

is there a way to monkeypatch our way around this limitation?

pombredanne:
The thing here is that there is no way to tell if an option was entered by a user or if the default value has been used. A user-entered option that has the same value as the default is the same as a non user-entered option with a default. As a result, the simple nargs='?' pattern in argparse cannot be implemented and instead two options must be used: a flag and a value. See https://gist.github.com/pombredanne/aeab7f5b4783919dfe5265b2c553cdb8 for some examples.

Evidlo:
@pombredanne argparse provides a `const` option in addition to `default`.  

    add_parser = subparsers.add_parser('add')
    add_parser.add_argument('-w', '--words', metavar='length', type=int, nargs='?', const=5, default=None)

So `foo add` means `args.words == None`, `foo add -w` means `args.words == 5`

pombredanne:
@untitaker would you like to see this in Click or not?

mbrancato:
How does Click conform to say the POSIX command line argument syntax standard without supporting optional option-arguments?

anarcat:
[...]

PR #1229, opened 2019-02-13
refactor code to get option value from cli state

jpassaro:
Prior to now, the logic for assigning an option value based on the
remaining arguments on the command line has been handled directly by
OptionParser. This refactor accomplishes two things:

1: Bringing two very similar procedures into a single method;
2: Placing them in an Option class that can be easily inherited.

This clears the way for users or plugins to easily address #549 without
affecting core code.

### DIFF
--- a/click/parser.py
+++ b/click/parser.py
@@ -73,12 +73,6 @@ def _unpack_args(args, nargs_spec):
     return tuple(rv), list(args)
 
 
-def _error_opt_args(nargs, opt):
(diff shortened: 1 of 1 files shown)


jpassaro:
I was mistaken: I don't think this solves anything with respect to the linked issue.

It might lead the way toward a solution in core, but until that's fleshed out this is probably not worthwhile.

jpassaro:
see #1230

Issue #764, opened 2017-04-14
Flags with optional = argument


andyneff:
Is it possible to have a flag argument with an optional argument when using the `=` operator?

For example, on `grep` on GNU systems, it has

```
--color[=WHEN], --colour[=WHEN]
              Surround the matched (non-empty) strings, matching lines, context lines, file names, line numbers, byte offsets, and separators (for fields and groups of context lines) with escape sequences to display them in color on
              the terminal.  The colors are defined by the environment variable GREP_COLORS.  The deprecated environment variable GREP_COLOR is still supported, but its setting does not have priority.   WHEN  is  never,  always,  or
              auto
```

- `grep --color ...` results in color being set to "auto"
- `grep --color=always ...` results in color being set to "always"
- `grep --color always ...` does NOT result in color being set to "always", but still "auto" and always is interpreted as what ever arguments come next

Slightly more complicated situation (`rsync` for example) combining multiple single letter versions of these together.

- `rsync -ravB=1000`, so flags "r", "a", and "v" are "set" and "B" is set to 1000.

jpassaro:
See #1229 , which if accepted would allow users to do this pretty easily by subclassing Option

PR #1230, opened 2019-02-13
Teach Option to accept only explicit values

jpassaro:
Add a new attribute/argument `explicit_only` to `Option` class. When set
to true, this option will cause Options with nargs=1 to only accept
input that is explicitly associated to it.

This makes the following valid:

    command
    command --option
    command --option=value

This addresses #549 and #764.


davidism:
Thanks for working on this. While it's an interesting solution to the issue, I don't think we should be adjusting how the basic syntax works for individual options. Addressing this with #1618 instead.

PR #1234, opened 2019-02-19
Allow path_type of the Path type to accept a callable (`#405`_)

termim:
This would allow to convert path names to pathlib.Path instances
for example.

closes #405

PR #1240, opened 2019-02-21
Allow path_type of the Path type to accept a callable (`#405`_) (7.x branch)

termim:
Cherry-pick #1234  to 7.x branch:

This would allow to convert path names to pathlib.Path instances
for example.


davidism:
This is a new feature, so it would be in 8.0, not 7.x.

Issue #1239, opened 2019-02-20
setuptools integration using "Scripts in Packages" example - not working


be-rock:
I am following the setuptools example using a python3.7.1 conda virtual env from the documentation [here](https://click.palletsprojects.com/en/7.x/setuptools/) and continue to receive an error:

`ModuleNotFoundError: No module named 'yourpackage.scripts'`

```
yourpackage/
├── __init__.py
├── main.py
├── scripts
│   ├── __init__.py
│   └── yourscript.py
├── setup.py
└── utils.py
```
```
#scripts/yourscript.py

import click

@click.command()
def cli():
    """Example script."""
    click.echo('Hello World!')
```

```
# setup.py

from setuptools import setup, find_packages
setup(
    name='yourpackage',
    version='0.1',
    packages=find_packages(),
    include_package_data=True,
    install_requires=[
        'Click',
    ],
    entry_points='''
        [console_scripts]
        yourscript=yourpackage.scripts.yourscript:cli
    ''',
)
```

```
$ pip install -e .
Looking in indexes: https://InternalArtifactoryUrl
Obtaining file:///private/tmp/yourpackage
Requirement already satisfied: Click in /Users/myusername/miniconda3/envs/test_20190220/lib/python3.7/site-packages (from yourpackage==0.1) (7.0)
Installing collected packages: yourpackage
  Found existing installation: yourpackage 0.1
    Uninstalling yourpackage-0.1:
      Successfully uninstalled yourpackage-0.1
  Running setup.py develop for yourpackage
Successfully installed yourpackage

$ which yourscript
/Users/myusername/miniconda3/envs/test_20190220/bin/yourscript

$ yourscript
Traceback (most recent call last):
  File "/Users/myusername/miniconda3/envs/test_20190220/bin/yourscript", line 11, in <module>
    load_entry_point('yourpackage', 'console_scripts', 'yourscript')()
  File "/Users/myusername/miniconda3/envs/test_20190220/lib/python3.7/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/Users/myusername/miniconda3/envs/test_20190220/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2793, in load_entry_point
    return ep.load()
  File "/Users/myusername/miniconda3/envs/test_20190220/lib/python3.7/site-packages/pkg_resources/__init__.py", line 2411, in load
    return self.resolve()
[...]

PR #1241, opened 2019-02-24
Fix setuptools code snippet (dict, module path)

bittner:
I can't help but I find using a dictionary-style entry_points definition more pythonic.

### DIFF
--- a/docs/setuptools.rst
+++ b/docs/setuptools.rst
@@ -126,14 +126,15 @@ These would be the modified contents of ``setup.py``::
 
     setup(
         name='yourpackage',
-        version='0.1',
+        version='0.1.0',
         packages=find_packages(),
         include_package_data=True,
         install_requires=[
             'Click',
         ],
-        entry_points='''
-            [console_scripts]
-            yourscript=yourpackage.scripts.yourscript:cli
-        ''',
+        entry_points={
+            'console_scripts': [
+                'yourscript = yourpackage.scripts.yourscript:cli',
+            ],
+        },
     )


docs/setuptools.rst
```diff
@@ -126,14 +126,15 @@ These would be the modified contents of ``setup.py``::
 
     setup(
         name='yourpackage',
-        version='0.1',
+        version='0.1.0',
```
ThiefMaster: please don't make unrelated changes

bittner: Really? Quite common? Not with semver.

You're semi-right: It's a code snippet. I'm not changing the behavior of the application. I'm adapting the code snippet.

bittner: BTW, I would also try to fix #1239 with this PR. I think it makes sense to do it in one go. Still a code snippet change.

ThiefMaster: 0.1 and 0.1.0 have the same meaning even with semver, but i just checked the spec and they indeed expect you to always put x.y.z and not omit z if z==0...

bittner: Thank you for verifying. :+1: So, it's okay that I leave it in there?

ThiefMaster: Ideally it should be in a separate commit (it's a different change), but the change is trivial so I'll just go ahead and merge it as it is

ThiefMaster (approved):

bittner:
Wow! That was fast! -- I've just noticed, there is another code snippet on the same page. I'll place another PR.

docs/setuptools.rst
davidism: Click has always been 2 numbers, not 3. Please revert this.

ThiefMaster: @davidism it's not the click version but the example for people's own packages - otherwise i wouldn't have accepted it

davidism: Yeah, I noticed right after, never mind.

PR #1242, opened 2019-02-24
Use code-block markup, align snippets, add setup.py

bittner:
This PR fixes a few things in the docs on setuptools integration:

1. Instead of simple typewriter markup we can use code-blocks with explicit language markup
1. The `entry_points` block in the setuptools code snippets differed and are now aligned -- relates to #1241
1. Also show `setup.py` in the directory listing to make copy+paste less error prone (for newbies to setuptools) -- fixes #1239

PR #1243, opened 2019-02-24
Fix broken markup (setuptools chapter)

bittner:
Fixes a left-over from changing the code snippet markup in PR #1242. (Sorry for that!)


bittner:
Thanks for merging!

PR #1192, opened 2018-12-21
Add Python 3.7 to test matrix

chbrown:
Python 3.7 was released on 2018-06-27 and is the current version of Python: https://www.python.org/downloads/release/python-370/

PR #1258, opened 2019-03-13
Test on Python 3.7

kevin-brown:
This requires switching to Xenial which requires specifiying exact versions of PyPy2 and PyPy3.


kevin-brown:
Closing this off as the exact same PR exists already: https://github.com/pallets/click/pull/1192

PR #1167, opened 2018-11-10
Add no_args_is_help to click.Command

sirosen:
The functionality for no_args_is_help, available on MultiCommands, can easily be added on Commands as well. The result is not much more code, and a more consistent interface between MultiCommands and Commands.
It's also nice to be able to write a click command `foo` and then just hit `foo` for help, regardless of whether its a Group or Command.

closes #1156

Will add changelog edit in a moment, after this PR is assigned a number.

PR #1260, opened 2019-03-15
Document adding no_args_is_help to Command for 7.1

### LINKED ISSUE #1259: no_args_is_help for Command: documentation vs code is wrong
I am using Python 3.7 and have Click 7.0 installed via `pip install click`.

In the current [API documentation](https://click.palletsprojects.com/en/7.x/api/) the documentation for the `command` decorator says the following:

> Creates a new `Command` and uses the decorated function as callback. This will also automatically attach all decorated `option()`s and `argument()`s as parameters to the command.
> ...
> **All keyword arguments are forwarded to the underlying command class.**

This is my first time using click, so when I click on the link to get the `Command` class documentation, it tells me it has this option:

> `no_args_is_help` – this controls what happens if no arguments are provided. This option is disabled by default. If enabled this will add --help as argument if no arguments are passed

However when I try to add this as an option to the `command` decorator I get an error:

> `TypeError: __init__() got an unexpected keyword argument 'no_args_is_help'`

Looking into the code in `decorators` and `core` it is apparent that this error is because `Command` doesn't actually have the `no_args_is_help` argument, only `MultiCommand` does.

Why is there such a big discrepancy between the documentation and the actual implementation? I see that there are other issues here about this same option, but they are all either merged or closed as far as I can tell. As a new user this is extremely frustrating because I'm trying to 'rtfm' but that doesn't work when the manual is wrong.

pwlandoll:
In #1167, a parameter was added to the `Command` class, changing the documentation accordingly. However, it was not specified that it was a change for an unreleased version. This change indicates that the `no_args_is_help` parameter is added in 7.1. 

Tests are passing. Closes #1259.

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -776,6 +776,8 @@ class Command(BaseCommand):
 
     .. versionchanged:: 2.0
        Added the `context_settings` parameter.
+    .. versionchanged:: 7.1
+       Added the `no_args_is_help` parameter.
 
     :param name: the name of the command to use unless a group overrides it.
     :param context_settings: an optional dictionary with defaults that are


sirosen:
Sorry for the confusion that this generated!

This looks good. Merging.

Ricyteach:
I was trying to use `no_args_is_help` today and then saw this note on the docs- thanks for adding it!

Can anyone say when 7.1 might drop?

PR #1273, opened 2019-04-02
Annotate required choices with braces

### LINKED ISSUE #1272: Choices.get_metavar makes all choices look optional
Creating a required parameter with a `Choice` type leads to confusing helpstrings.

From https://en.wikipedia.org/wiki/Command-line_interface#Command_description_syntax 

> Built-in usage help and man pages commonly employ a small syntax to describe the valid command form:
> * angle brackets for required parameters: `ping <hostname>`
> * square brackets for optional parameters: `mkdir [-p] <dirname>`
> * ellipses for repeated items: `cp <source1> [source2...] <dest>`
> * vertical bars for choice of items: `netstat {-t|-u}`

0xDEC0DE:
Display the metavar for Choice options with curly braces on required
params, to reduce user confusion.

fixes #1272

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -67,6 +67,9 @@ Unreleased
 -   The result object returned by the test runner's ``invoke()`` method
     has a ``return_value`` attribute with the value returned by the
     invoked command. :pr:`1312`
+-   Required arguments with the ``Choice`` type show the choices in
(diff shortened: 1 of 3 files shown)


davidism:
Made it only apply to required arguments, not options. Might be able to do something with optional option values once #1618 is in, but that can be addressed later and might be more general than the `Choice` type. Added a changelog.

Issue #1277, opened 2019-04-16
Add coercion/normalization of Choice values


altendky:
I want my CLI to be case insensitive for certain `Choice` options but I don't want to deal with that variability internally as it has no meaning.  I think an option like `coerce_case` or `normalize_case` would be a reasonable addition.

PR to follow.

davidism:
It appears #887 already added this.

davidism:
Never mind, this is about what value gets returned, that was about what got matched.

sirosen:
> I don't want to deal with that variability internally as it has no meaning.

Sounds like you should just be handling things in all-lowercase? #887 set things to return the `normed_value` -- so when `case_sensitive=False`, `click.Choice` should always be returning a lowercase string.
(Yes, this should be improved in the documentation for `case_sensitive=False`.)

Reading #1278, it seems this is about allowing `--foo [Apple|Orange]` and always returning title-case `Apple` or `Orange` even if `case_sensitive=False`.

Unless there's some other rationale, I don't think this is well-justified. It adds another parameter to `click.Choice` which is conditional on `case_sensitive=False` and "undoes" part of that behavior -- I think that will be confusing for most users.

I'm not saying there isn't a reason! In #569 I argued that I really have a use-case where case-sensitivity is driven by external constraints. But I'd need to know what the use-case is before agreeing that this should be part of `click`.

[I'm reminded of this comment on the initial request for case_sensitive](https://github.com/pallets/click/issues/569#issuecomment-216515603). If there is a clear use-case here, it's worth considering a generic parameter which changes the `token_normalize_func` on a per-option basis.

altendky:
@sirosen, yes, #1278 retains the original (not necessarily title) case specified by the choices.  It seems to me this may have been a better approach to begin with as opposed having case insensitivity force the introduction of a third casing (lower as opposed to original choice and user-entered).  But, for backwards compatibility I introduced this as an option with default behavior remaining as it is now.

My actual use case is in https://github.com/altendky/romp where I have choices for platform and Python interpreter (and more, but specifically those).  I stuck a copy of the modified choice in there which I am using for now.

https://github.com/altendky/romp/blob/742d4ac5ff0b918e33002af1f374b8ce6938367a/src/romp/_matrix.py#L11-L18
```python
vm_images = collections.OrderedDict((
    ('Linux', 'ubuntu-16.04'),
    ('macOS', 'macOS-10.13'),
    ('Windows', 'vs2017-win2016'),
))


[...]

PR #1278, opened 2019-04-16
Add coerce_case for click.Choice

altendky:
#1277


altendky:
Replaced by #1318 which has been merged.

PR #1280, opened 2019-04-19
Add show_index flag for Choice type.

### LINKED ISSUE #1411: Allow a type to customize the prompt
As of Click 7, `click.Choice` appears only to support choice prompting by string, i.e. the user must specify the exact string (modulo case sensitivity) from the choices. It would be nice if Click could support two additional modes of selection:

Numeric:
```bash
Choose from:
[1] Choice A...
[2] Choice B...
[3] Choice C...
Enter the index for one of the above choices: 2
"Choice B..." was selected
```

or partial matches as long as what the user enters is unique:

```bash
Choose from:
This thing
That thing
Those things
Enter choice: thi
"This thing" was selected
```
("thi" is unique amongst the three strings so is chosen)

I thought it would be easy to implement this by subclassing `click.Choice` but unfortunately not: `click.Choice` does not provide the initial prompt message (only the prompt after an invalid choice) - this is provided by `click.prompt` from the specified choices list. The prompt message in at least the first example above would require to be changed to add numeric indexes. It seems to me therefore that this functionality requires more structural changes to click and so I thought I'd submit this issue.

ChenglongWang:
I added show_index flag for Choice.
Show_index enables users to use index of choice instead of choice item itself.
I think this would help a little when users have a bunch of choice items.

Usage example:
```
@click.command()
@click.option('--v', prompt=True, type=click.Choice(['a','bb','ccc'], show_index=True), default=1)
def main(v):
    print("Choosed:", v)
```

> $ python test_click.py
> V (0:a, 1:bb, 2:ccc) [bb]: 0
> Choosed: a

To avoid confusion of indices and choices, when show_index is True, only index (integer) is allowed.
Code was tested on Ubuntu 18.04 with zsh: 196 passed, 1 xfailed in 2.02 seconds

### DIFF
--- a/click/termui.py
+++ b/click/termui.py
@@ -46,7 +46,8 @@ def hidden_prompt_func(prompt):
 def _build_prompt(text, suffix, show_default=False, default=None, show_choices=True, type=None):
     prompt = text
     if type is not None and show_choices and isinstance(type, Choice):
-        prompt += ' (' + ", ".join(map(str, type.choices)) + ')'
+        prompt += ' (' + ", ".join(map(str, type.choices.values())) + ')' if not type.show_index else \
+                  ' (' + ", ".join(['%s:%s'%(idx,ch) for idx,ch in type.choices.items()]) + ')' 
     if default is not None and show_default:
         prompt = '%s [%s]' % (prompt, default)
     return prompt + suffix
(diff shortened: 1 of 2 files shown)


davidism:
See #1411, I'd like to design a way for types to override the prompt, but I don't want to modify the built-in `Choice` type.

PR #1298, opened 2019-05-06
click.open_file respects "errors" parameter.

### LINKED ISSUE #958: The `errors` argument in `click.open_file` is ignored on python3

joewledger:
Fixes https://github.com/pallets/click/issues/958

Signed-off-by: Joe Ledger <joewledger@gmail.com>

### DIFF
--- a/click/_compat.py
+++ b/click/_compat.py
@@ -329,80 +329,88 @@ else:
         # but this at least will force Click to recover somehow.
         return is_ascii_encoding(getattr(stream, 'encoding', None) or 'ascii')
 
+    def _is_compat_stream_attr(stream, attr, value):
+        """A stream attribute is compatible if it is equal to the
+        desired value or the desired value is unset and the attribute
+        has a value.
+        """
+        stream_value = getattr(stream, attr, None)
+        return stream_value == value or (value is None and stream_value is not None)
+
     def _is_compatible_text_stream(stream, encoding, errors):
-        stream_encoding = getattr(stream, 'encoding', None)
-        stream_errors = getattr(stream, 'errors', None)
+        """Check if a stream's encoding and errors attributes are
+        compatible with the desired values.
+        """
+        return (
+            _is_compat_stream_attr(stream, "encoding", encoding)
+            and _is_compat_stream_attr(stream, "errors", errors)
+        )
 
-        # Perfect match.
-        if stream_encoding == encoding and stream_errors == errors:
-            return True
-
-        # Otherwise, it's only a compatible stream if we did not ask for
-        # an encoding.
-        if encoding is None:
-            return stream_encoding is not None
-
-        return False
-
-    def _force_correct_text_reader(text_reader, encoding, errors,
-                                   force_readable=False):
-        if _is_binary_reader(text_reader, False):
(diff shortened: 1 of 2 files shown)


joewledger:
Fixes https://github.com/pallets/click/issues/958

click/_compat.py
```diff
@@ -482,9 +482,15 @@ def open_stream(filename, mode='r', encoding=None, errors='strict',
...
-        return io.open(filename, mode, encoding=encoding, errors=errors), True
+        open_kwargs = {}
+
+        if encoding is not None:
+            open_kwargs['encoding'] = encoding
+
+        if errors is not None:
+            open_kwargs['errors'] = errors
+
+        return codecs.open(filename, mode, **open_kwargs), True
```
davidism: Looks like `codecs.open` has an early return that ignores `errors`, so we need to continue using `io.open`. I'll add back your check for `"b" not in mode`.

click/_compat.py
```diff
@@ -482,9 +482,15 @@ def open_stream(filename, mode='r', encoding=None, errors='strict',
 
     # Non-atomic writes directly go out through the regular open functions.
     if not atomic:
-        if encoding is None:
-            return open(filename, mode), True
-        return io.open(filename, mode, encoding=encoding, errors=errors), True
+        open_kwargs = {}
```
davidism: This needs to happen in atomic mode too, I'll move the kwargs outside the `if` block.

davidism:
This is failing for `stdin` (`"-"`) because `_force_correct_text_reader` returns early if `encoding` isn't passed. I'll move those checks around.

davidism:
Making the same changes to `_force_correct_text_writer` that I did for `reader` caused `test_termui.test_progressbar_update` to fail. For some reason it decided to write `\r` instead of `\n` newlines. Leaving a TODO in case someone works on that later.

davidism:
A test is failing on Python 2 right now because `io.open(mode="w")` requires passing unicode to `write`. I'll see what I can do about that tomorrow.

davidism:
I finally have this consistent. It applies to both read and write streams, makes standard and is consistent about using `io.open` and wrapping text mode on Python 2.

Thanks for adding the extra tests, they helped me ensure this was finished, I don't think I would have been nearly as thorough at refactoring it otherwise.

I was also able to identify why two tests were marked xfail for Windows Python 3: they were incorrectly writing `os.linesp` manually instead of `\n` with universal newline mode.

Issue #1212, opened 2019-01-25
Document Python 3.7 behaviour in "Python 3 Surrogate Handling" documentation


adamtheturtle:
http://click.palletsprojects.com/en/7.x/python3/#python-3-surrogate-handling rightly warns about issues when using Python 3.

A ``RuntimeError`` is documented.

The good news is that this `RuntimeError` no longer is a problem in Python 3.7, as demonstrated below.

This took some experimentation to find out, and as a user I would have liked to see this in the documentation.

Thank you!

```
~/Desktop ~> echo $LANG

~/Desktop ~> echo $LC_ALL

~/Desktop ~> cat example.py
import click

@click.command()
def hello():
    pass

hello()
~/Desktop ~> vf activate python-3.5.6
~/Desktop (python-3.5.6) ~> python --version
Python 3.5.6
~/Desktop (python-3.5.6) ~> python example.py
Traceback (most recent call last):
  File "example.py", line 7, in <module>
    hello()
  File "/Users/adam/.virtualenvs/python-3.5.6/lib/python3.5/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/adam/.virtualenvs/python-3.5.6/lib/python3.5/site-packages/click/core.py", line 696, in main
    _verify_python3_env()
  File "/Users/adam/.virtualenvs/python-3.5.6/lib/python3.5/site-packages/click/_unicodefun.py", line 124, in _verify_python3_env
    ' mitigation steps.' + extra
RuntimeError: Click will abort further execution because Python 3 was configured to use ASCII as encoding for the environment. Consult https://click.palletsprojects.com/en/7.x/python3/ for mitigation steps.

This system lists a couple of UTF-8 supporting locales that
you can pick from.  The following suitable locales were
[...]

PR #1302, opened 2019-05-06
Adds a note that there is a behavior change after Python 3.7

paxnovem:
This notes what PEP's led to a change to an error that no longer occurs
in python 3.7 and later

Closes: #1212


jcrotts:
I think it would be nice if you added direct links to the PEPS.

PR #1303, opened 2019-05-06
allow open intervals on float range

### LINKED ISSUE #1110: FloatRange: add options that exclude each endpoint
Thanks for developing this beautiful program.

At this time, FloatRange assumes that the passing argument value is always in *closed* interval `[a, b]`, not *open* one. Sometimes, this may be harsh when the user want to accept a floating number exclude two endpoints, `a` and `b`. A user who uses IntRange can solve this problem fairly easily, but FloatRange cannot.

So, I think FloatRange should have the options which can exclude each endpoint. May I PR for this?

robintully:
fixes #1110

This allows for open intervals (exclusive ranges) for the FloatRange.  I believe that without this you would have to specify arbitrary precision of the intervals to emulate open intervals (for example you would have to do 4.999999 in the initialization of FloatRange).


jcrotts:
Looks okay to me. Could you add a changelog entry.

robintully:
Thanks for the feedback, added the changelog

davidism:
I think it's clearer as two boolean options, `min_inclusive=True, max_inclusive=True`. That way it can also be configured to match `range`, where the start is inclusive but the end is exclusive.

davidism:
I decided the entire implementation needed to be refactored. I've left your initial commit, then added my refactor. `min` and `max` should be able to be set to open separately. While open bonds don't mean as much for `IntRange`, it should probably be implemented for consistency. The `clamp` behavior was not accounting for open bounds. #1586 by @IamCathal added the range to the help text, but didn't add it for `FloatRange`, and it needed to be updated for open ranges and unbounded ranges.

I created a `_NumberParamTypeBase` as the base for `IntParamType` and `FloatParamType`, and `_NumberRangeBase(_NumberParamTypeBase)` as the base for `IntRange` and `FloatRange`. All the range behavior except for clamping is implemented by the base class. Help text shows the range like `1<=x<=32`, `1<=x`, `x<=32`, `x<32`, etc. For `IntRange`, clamping is easy for open and closed bounds. For `FloatRange`, clamping open bounds means figuring out the next representable float value away from the bound. `math.nextafter` isn't implemented until Python 3.9. For < 3.9, I used the following, which *seems* right based on the docs, but you probably shouldn't be combining clamping with open ranges to begin with.

```python
m, e = math.frexp(bound)
return math.ldexp(m + math.copysign(sys.float_info.epsilon, dir), e)
```

I need to add more tests, update the error messages, and write docs.

davidism:
I think I might just disallow clamp if either bound is open in `FloatRange` on Python < 3.9. I'm not confident that my implementation is correct.

PR #1304, opened 2019-05-06
1285 fix path does not exist

### LINKED ISSUE #1285: Path(exists=True, envvar='FOO') tries to interpret FOO="" as an existing path
Example:
```python
@click.command()
@click.option('--mypath', type=click.Path(exists=True), envvar='MYPATH')
def cli(mypath):
    click.echo(f'mypath: {mypath}')
```

Test case:
```
➜ env MYPATH= ./mycli.py
```

Observed output:
```
Usage: mycli.py [OPTIONS]
Try "mycli.py --help" for help.

Error: Invalid value for "--mypath": Path "" does not exist.
```

Expected output:
```
➜ ./mycli.py
mypath: None
```

i.e. Shouldn't Click special-case `envvar == ""` and treat it as though the option wasn't passed? This is more useful than failing with "`Path "" does not exist`", as it gives users who are inheriting an unwanted env var, and who can only set it to empty string rather than unset, a way to clear it.

jzengg:
Fixes #1285

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,7 +25,8 @@ Unreleased
     False :pr:`1167`
 -   Add ``show_defaults`` parameter to ``Context`` to enable showing
     defaults globally. :issue:`1018`
-
+-   Handle ``env MYPATH=''`` as though the option were not passed.
+    :issue:`1196`
 
 Version 7.0
 -----------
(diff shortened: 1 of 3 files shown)


twosigmajab (approved):
LGTM, fwiw... Thanks for working on this @jzengg!

PR #1306, opened 2019-05-06
Unprintable MissingParameter exception

### LINKED ISSUE #1139: Unprintable MissingParameter exception
My code is (correctly) hitting https://github.com/pallets/click/blob/master/click/core.py#L1444 and raising a MissingParameter exception. However, there's no `message` passed there, which means `exc.message` is defaulted to `None` by the base class (`ClickException`) constructor here https://github.com/pallets/click/blob/master/click/exceptions.py#L23.

With other `ClickException`s I've seen (e.g. `NoSuchOption` https://github.com/pallets/click/blob/master/click/exceptions.py#L165), if message is set to None then it gets reset as something more sensible e.g. `"No such option 'foo'"`. This means that doing `str(exc)` works nicely. However, `MissingParameter` doesn't have this, which means any attempt to stringify (e.g. print or log) a `MissingParameter` actually causes a `TypeError: __str__ returned non-string (type NoneType)` and then things go badly wrong.

Is it expected that `MissingParameter` is an unprintable exception when caused e.g. by missing out a required argument? It seems odd that no one else has come across this before. I'm using click in a slightly unusual context in that I'm using the group/command/arg/option parsing stuff outside of a traditional command line script, which might be why no one has seen this if click does it's own special handling for `MissingParameter` errors.

TL;DR It would be nice if `MissingParameter` was printable when raised via `Parameter.full_process_value` (or ideally more generally).

It might be enough to have something along the lines of what `NoSuchOption` does?

Ketzalkotal:
Resolves #1139

I attempted to resolve the issue by setting message to a default string if `message is None` is true. However, this broke multiple tests in test_formatting.py by appending the 'missing parameter ...' default message set in the MissingParameter init method.

Overriding __str__ seems like the clearer and more direct fix to allow casting of MissingParameter objects to strings without breaking existing behavior.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,8 @@ Unreleased
     ``type=bool``. :issue:`1287`
 -   Newlines in option help text are replaced with spaces before
     re-wrapping to avoid uneven line breaks. :issue:`834`
+-   ``MissingParameter`` exceptions are printable in the Python
+    interpreter. :issue:`1139`
 
 
 Version 7.0
--- a/click/exceptions.py
+++ b/click/exceptions.py
@@ -152,6 +152,19 @@ class MissingParameter(BadParameter):
             msg or '',
         )
 
+    def __str__(self):
+        if self.message is None:
+            param_name = self.param.name if self.param else None
+            return "missing parameter: {}".format(param_name)
+        else:
+            return self.message
+
+    if PY2:
+        __unicode__ = __str__
+
+        def __str__(self):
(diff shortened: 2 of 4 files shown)


tests/conftest.py
```diff
@@ -6,3 +6,12 @@
 @pytest.fixture(scope='function')
 def runner(request):
     return CliRunner()
+
+
+@pytest.fixture(scope='function')
+def ctx():
```
Ketzalkotal: Is there a better way to mock a context?

PR #1307, opened 2019-05-07
Fix to #1066 text wrapping

### LINKED ISSUE #1066: When wrapping text, the first line of a subsequent paragraph is wrong.
In click 6.7, python 3.6.4

This help text for an option:
`'x\n\nI am going to get broken in the wrong place.\n'`

Looks fine:
```
  --test TEXT                     x
                                  
                                  I am going to get broken in the wrong
                                  place.
```

But if we make our first paragraph last line a bit longer:
`'This will break the next paragraph.\n\nI am going to get broken in the wrong place.\n'`

The line position for the first paragraph is carried forward incorrectly when wrapping the first line of the following paragraph so it breaks the line in the wrong place:
```
  --test TEXT                     This will break the next paragraph.
                                  
                                  I am
                                  going to get broken in the wrong place.
```
The result should be:
```
  --test TEXT                     This will break the next paragraph.
                                  
                                  I am going to get broken in the wrong
                                  place.
```

Zarad1993:
### Problem
Closes #1066 

**To Reproduce:** 

Add the python snippet below on a new file using `master` branch.
On your terminal.
Run
```sh
python sample.py --help
```
```python
import click
TEXT = 'This will break the next paragraph.\n\nI am going to get broken in the wrong place.'
@click.command()
@click.option('--test', help=TEXT)
def hello(test):
    """This script prints hello NAME COUNT times."""
    pass
if __name__ == '__main__':
	hello()
```

Result

issue-produce

--------

### Proof of fix

Follow the same reproduce-able steps above on this branch

1066-fix

### DIFF
--- a/click/formatting.py
+++ b/click/formatting.py
@@ -198,7 +198,19 @@ class HelpFormatter(object):
                 self.write(' ' * (first_col + self.current_indent))
 
             text_width = max(self.width - first_col - 2, 10)
-            lines = iter(wrap_text(second, text_width).splitlines())
+
+            if '\n' in second:
+                lines = []
+                broken_lines = second.split('\n')
+                for broken_line in broken_lines:
+                    if broken_line == '':
+                        lines.append(broken_line)
+                    else:
+                        lines += list(wrap_text(broken_line, text_width).splitlines())
+                lines = iter(lines)
+            else:
+                lines = iter(wrap_text(second, text_width).splitlines())
+
             if lines:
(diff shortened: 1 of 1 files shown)


davidism:
Thanks for working on this, sorry it took so long for me to get back to.

Turns out the underlying `wrap_text` function already takes a `preserve_paragraphs` option, but it's not enabled for option help text right now. That basically does what this does, by preserving `\n\n`, so I'll create a different PR to enable that instead.

PR #1309, opened 2019-05-07
Formatting file option prompting to path only

### LINKED ISSUE #914: File option - prompting with pretty default?
Hello,
I'm developing simple script, but I have found horrible default when using a file option with prompting with default value...
I have this code:
```python
import click
import os.path
@click.command()
@click.option('-f', default=os.path.realpath("subdir/file.txt"), prompt="Please provide a file", type=click.File('rt'))
def func(f):
    pass
func()
```
But when I call it, it produces prompt like this:

    me@computer:~/directory$ python3 script.py
    Please provide a file [<_io.TextIOWrapper name='/home/me/directory/subdir/file.txt' mode='rt' encoding='UTF-8'>]:
Is this solved anywhere or how difficult would it be to fix it, so it would show `[/home/me/directory/subdir/file.txt]`?

`Ubuntu 16.04 LTS 32-bit, Python 3.5.2, click 6.7`

Please excuse any imperfections of this issue, I'm quite new to GitHub and haven't written any issues yet...

icreatedanaccount:
fixes #914


icreatedanaccount:
#914

jcrotts:
Could you add an entry to the change log please.

tomgoren (approved):

tomgoren:
Thank you @icreatedanaccount !!!

icreatedanaccount:
Added to change log @jcrotts

hg-pyun (approved):

jcrotts:
Sorry could you also point this to 7.x

davidism:
After #1298, Click always returns some subclass of `io.IOBase`, even on Python 2, no need to check for `file`. Also need to account for `LazyFile` type. The `name` attribute actually comes from `text_file.buffer.raw.name`, which might not exist if some other `IOBase` type was returned by a custom type, so check that the attribute exists.

PR #1314, opened 2019-05-08
Set up CI with Azure Pipelines

t-eckert:
I have at least a first pass for moving work over to Azure Pipelines. I have attached the logs I get from running the tests.

[logs_66.zip](https://github.com/pallets/click/files/3158703/logs_66.zip)


jcrotts:
Awesome, thanks for setting it up.

davidism:
Thanks for starting this, sorry it took so long to get to. I've updated the Azure and Tox configs based on the other Pallets projects that have moved over. Once I merge this I'll enable the project in Azure.

davidism:
A few more tweaks and the tests pass: https://dev.azure.com/pallets/click/_build/results?buildId=431

Issue #1116, opened 2018-09-19
Expand unit testing around termui, especially getchar


sirosen:
#1115, #583, and a couple of other issues seem to show that there are some subtle cases to handle in `click.getchar`.
But these are hard to test and therefore, presently, untested. I wouldn't be surprised to find other gaps in testing of termui.

I think it would be nice to have some more testing around the termui module, limited to what we can do safely and definitely correctly -- e.g. feed `getchar` a `u'\x04'` and make sure it raises an exception.

The best way to tackle this is probably to start writing more tests and see what's needed and possible.
I'd like to step up and do it myself, but if someone else has the cycles they should feel more than welcome.

icreatedanaccount:
Taking a stab at it

icreatedanaccount:
I am able to expand the coverage by about 15% and test the win-specific version of `getchar` as I develop on Windows. How do you suggest we go about the Linux half (keep the issue open, new one, etc.) ? @sirosen

sirosen:
My big worry was about Windows coverage. For the issues I mentioned, I was firing up Windows VMs and poking and prodding at it, without any kind of tests as a safety net.

I would say that #1315 closes this. At a glance it seems fine/good, but I want to do a more careful read after work before merging.

PR #1315, opened 2019-05-09
Expand unit testing around termui (getchar testing for Windows)

icreatedanaccount:
#1116

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,11 +11,13 @@ Unreleased
 - Fix link in ``unicode_literals`` error message. (`#1151`_)
 - Add support for colored output on UNIX Jupyter notebooks. (`#1185`_)
 - Remove unused compat shim for ``bytes``. (`#1195`_)
+- Expand unit testing around termui, especially getchar (Windows). (`#1116`_)
 
 .. _#1167: https://github.com/pallets/click/pull/1167
 .. _#1151: https://github.com/pallets/click/pull/1151
 .. _#1185: https://github.com/pallets/click/issues/1185
 .. _#1195: https://github.com/pallets/click/pull/1195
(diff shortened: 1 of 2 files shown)


sirosen:
Sorry it took me so long to look through this (even after [I said I would](https://github.com/pallets/click/issues/1116#issuecomment-491396566))!

This looks really good. I can't think of a better way of testing getchar. 👍 

I have a vague worry that these tests could turn out to be flaky or hard to maintain over the long term, but _if_ they do we can mark skip or remove. That's not a reason to block and wait.

I want to add a time limit to these tests so that if something changes and the tests hang and wait for input, pytest will abort. But I'll open that up as follow-up work and we can discuss there.

I'm going to retarget this to `7.x-maintenance` instead of `master`, on the grounds that we want to test future 7.x releases with this. And I'll resolve any changelog conflict in a rebase-merge, since it's by my doing -- slow review -- that conflicts may have arisen.

sirosen (approved):

davidism:
I recently added pytest-timeout to a test, it was really straightforward.

sirosen:
Yeah, I didn't want to fold that in here because it doesn't look like `click` currently has `pytest-timeout` in the dependencies. So it means some tox.ini editing, etc. It should be easy to add though.

davidism:
@sirosen would you be able to create a ticket for adding these things? I'm running a sprint next week and it sounds like a good task for someone to work on.

sirosen:
Sorry I didn't get to this faster -- I was off-grid a bit the past few days.
@davidism, I just opened https://github.com/pallets/click/issues/1326 and marked it `save-for-sprint`. I hope it's in time to be useful, but worst-case it's written up now.

PR #1317, opened 2019-05-09
Preventing prompts without tty from echoing their outputs

### LINKED ISSUE #906: prompt + hide_input shows a GetPassWarning if no tty
getpass requires a tty so it can ensure output won't be echoed.  If it doesn't have one, it will warn.  The [docs](https://pymotw.com/2/getpass/) suggest checking for a tty before use so it would be nice if prompt() did that for us.

Simple example:
```
import click, sys

print sys.stdin.isatty()
click.prompt('test', hide_input=True)
```

Execution:
```
[vagrant@test ~]$ python a.py 
True
test: 
[vagrant@test ~]$ ssh localhost python a.py
False
test: /usr/lib64/python2.7/getpass.py:83: GetPassWarning: Can not control echo on the terminal.
  passwd = fallback_getpass(prompt, stream)
Warning: Password input may be echoed.
asdf

[vagrant@test ~]$
```

icreatedanaccount:
fixes #906

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,10 @@ Version 8.0
 
 Unreleased
 
+-   Preventing prompts without tty from echoing their outputs. (`#906`_)
+
+.. _#906: https://github.com/pallets/click/issues/906
+
 
 Version 7.1
 -----------
--- a/click/termui.py
+++ b/click/termui.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import os
 import sys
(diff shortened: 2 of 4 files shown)


davidism:
Thanks for working on this. However, after looking at the issue again, I'm not clear why we want to hide the error. This fix just hides the error but enables the same insecure echoing. I'm going to close this as I don't think it needs to be fixed. See my comment in the linked issue.

PR #1318, opened 2019-05-11
Always return originally passed choice for Choice()

### LINKED ISSUE #1277: Add coercion/normalization of Choice values

altendky:
closes #1277
replaces #1278

### DIFF
--- a/click/types.py
+++ b/click/types.py
@@ -133,6 +133,10 @@ class Choice(ParamType):
     You should only pass a list or tuple of choices. Other iterables
     (like generators) may lead to surprising results.
 
+    The resulting value will always be one of the originally passed choices
+    regardless of ``case_sensitive`` or any ``ctx.token_normalize_func``
+    being specified.
+
     See :ref:`choice-opts` for an example.
 
     :param case_sensitive: Set to false to make choices case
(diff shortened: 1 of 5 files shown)


click/types.py
```diff
@@ -152,29 +156,34 @@ def get_missing_message(self, param):
...
+            normed_choices = {
+                ctx.token_normalize_func(normed_choice): original
+                for normed_choice, original in normed_choices.items()
+            }
 
         if not self.case_sensitive:
-            normed_value = normed_value.lower()
-            normed_choices = [choice.lower() for choice in normed_choices]
+            if PY2:
+                lower = str.lower
```
sirosen: Using casefold below seems correct.
Should this actually be
```
if PY2:
    casefold = lamba x: x.upper().lower()
else:
    casefold = str.casefold
```

instead? `.upper().lower()` is kind of a "poor man's casefold" which will work for most things.

sirosen (commented):
I'm sorry that this took me so long to circle back to and review. I got pulled into other projects (after we spent good effort discussing this behavior, that feels like a waste).

I think this is the right thing to do. Even after a while away from it, reading it makes sense and is pretty reasonable.

I'd like to merge, but before I do, I had one comment about `lower` vs `casefold`. Let me know what you think about that.

click/types.py
altendky: I can't say I really know my way around this.  What's an example of a thing that `.upper().lower()` covers in py2 that `.lower()` doesn't?  Since I don't know I don't care either way but figure I may as well learn.  `:]`

sirosen: Sorry, I've misled you here. I'm not as clever as I think I am 😉 .

A case for casefold that I was thinking of is `ß`.
Uppercase `ß` should be `SS` -- I mean that in terms of how human languages, e.g. German, work -- so that `ß` and `ss` and `SS` all are treated as equivalent. But it seems that under py2, that's not how `upper()` on a unicode string behaves.

I don't think that case-folding on python2 is that reasonable to implement -- at least, not so trivial that it should simply be folded in here.

sirosen:
Just to keep the history a little bit clean, I rebased this before merging. (It's my own fault for taking months to review it.)

click/types.py
altendky: I tried it with ß and saw py2 not doing anything at all with that regardless.  But, someone in `#python` found the example below to vary between `.lower()` and `.upper().lower()` when using Unicode.  The tests don't look to be using Unicode for choice though and the docstring doesn't even mention the expected type for the `choices` parameter so I don't know.  I'm fine with whatever (and it's already merged).  I suppose it could even be argued that my addition of using `str.casefold` at all is an unrelated change.  `:]`

```python2
Python 2.7.15 (default, Dec 11 2018, 00:01:04) 
[GCC 6.3.0 20170516] on linux2
>>> x = u'µıſςϐϑϕϖϰϱϵẛι'
>>> x
u'\xb5\u0131\u017f\u03c2\u03d0\u03d1\u03d5\u03d6\u03f0\u03f1\u03f5\u1e9b\u1fbe'
>>> x.lower()
u'\xb5\u0131\u017f\u03c2\u03d0\u03d1\u03d5\u03d6\u03f0\u03f1\u03f5\u1e9b\u1fbe'
>>> x.upper().lower()
u'\u03bcis\u03c3\u03b2\u03b8\u03c6\u03c0\u03ba\u03c1\u03b5\u1e61\u03b9'
>>> x = 'µıſςϐϑϕϖϰϱϵẛι'
>>> x
'\xc2\xb5\xc4\xb1\xc5\xbf\xcf\x82\xcf\x90\xcf\x91\xcf\x95\xcf\x96\xcf\xb0\xcf\xb1\xcf\xb5\xe1\xba\x9b\xe1\xbe\xbe'
>>> x.lower()
'\xc2\xb5\xc4\xb1\xc5\xbf\xcf\x82\xcf\x90\xcf\x91\xcf\x95\xcf\x96\xcf\xb0\xcf\xb1\xcf\xb5\xe1\xba\x9b\xe1\xbe\xbe'
>>> x.upper().lower()
'\xc2\xb5\xc4\xb1\xc5\xbf\xcf\x82\xcf\x90\xcf\x91\xcf\x95\xcf\x96\xcf\xb0\xcf\xb1\xcf\xb5\xe1\xba\x9b\xe1\xbe\xbe'
```

PR #1320, opened 2019-05-23
Remove one unused import and one redundant import

AlexTereshenkov:
Here is the fix to two minor issues:

* `io` module is already imported at https://github.com/pallets/click/blob/master/click/_compat.py#L2 and do not need to be re-imported
* `PY2` is not used in the `_termui_impl.py` module and can be safely removed

The tests were run locally and have passed on 2.7 and 3.6.

These issues (among a few others) were flagged up by LGTM.com: https://lgtm.com/projects/g/pallets/click/alerts/?mode=list. Some of the other `pallets` repositories such as `flask` have been analyzed there as well: https://lgtm.com/projects/g/pallets/flask/.

If you like, you can use LGTM for automatically reviewing code in pull requests. Here's an example of how Google's AMPHTML use that to flag up security vulnerabilities in their code base: [ampproject/amphtml#13060](https://github.com/ampproject/amphtml/pull/13060)

There is also a shield badge you could put into the `readme` file, 
image 
which would show excellent quality of Python code in the `click` project.

(full disclosure: I'm a `click` user, and also part of the team that runs [LGTM.com](https://lgtm.com/))

### DIFF
--- a/click/_compat.py
+++ b/click/_compat.py
@@ -258,7 +258,6 @@ if PY2:
             value = value.decode(get_filesystem_encoding(), 'replace')
         return value
 else:
-    import io
     text_type = str
     raw_input = input
     string_types = (str,)
--- a/click/_termui_impl.py
+++ b/click/_termui_impl.py
@@ -16,7 +16,7 @@ import sys
 import time
 import math
 import contextlib
-from ._compat import _default_text_stdout, range_type, PY2, isatty, \
+from ._compat import _default_text_stdout, range_type, isatty, \
      open_stream, strip_ansi, term_len, get_best_encoding, WIN, int_types, \
      CYGWIN
 from .utils import echo


davidism:
Thanks! It's an interesting site, although it looks like a lot of the results are false positives due to compatibility fixes.

I'm moving towards using a standard set of pre-commit hooks, including flake8 and black, so those should catch these types of issues eventually. The thing holding us back right now is that applying those makes it more difficult to merge existing PRs (58 open at time of writing).

AlexTereshenkov:
@davidism thanks for posting back! I've noticed that some alerts (such as when one has to import `collections` / `abc` differently in Python 2 and Python 3) are due to compatibility hacks and have registered them internally so we could mitigate those. The migration to Python 3 is becoming of importance to many Python developers now due to the deprecation of Python 2 so we may look into addressing common issues with Python 2/3 compatibility to not raise false positives in as many cases as possible.

I have noticed that this [PR AppVeyor build for Python 2.7 32-bit and Python 2.7 64-bit](https://ci.appveyor.com/project/pallets/click/builds/24766605/job/uliv5vpwrh7aj8q2) failed with issue that looks to be connected to the `subprocess`? Wanted to let you know should you have missed this.

PR #1328, opened 2019-05-31
renamed context to environment to reduce confusion

### LINKED ISSUE #716: complex example - misleading name for context
The name `Context` and `pass_context` are misleading in the complex example, since the `Context` defined in the example shares a name with the click `Context`. Maybe a different name such as "Environment" or "Options" would be more appropriate.

blackmagic02881:
closes #716


blackmagic02881:
this is for click issues #716

PR #1329, opened 2019-05-31
Adding support to the context to distinguish the source of a command line parameter

### LINKED ISSUE #1264: Differentiate defaults from explicitly passed values
Unless I've missed something it seems there is currently no way to differentiate between an explicitly passed value and a default value.

This is problematic for establishing value precedence when dealing with other sources of parameter values (a config file in our case).
We're currently using the default values to detect if a given option was passed on the command line but this is imperfect and leads to the following problem:

- `@option('--some-option', default=1)`
- Some other way of specifying option values (e.g. a config file) sets `some-option = 2`
- User calls cli with `--some-option 1`

Under the above workaround this makes it appear as if the config file value (`2`) should have precedence even though the user explicitly requested value `1`.

joshuastorck:
Fixes #1264

### DIFF
--- a/click/core.py
+++ b/click/core.py
@@ -369,7 +369,7 @@ class Context(object):
 
         self._close_callbacks = []
         self._depth = 0
-        self._parameter_sources = {}
+        self._sources_by_paramname = {}
         
     def __enter__(self):
         self._depth += 1
@@ -614,7 +614,7 @@ class Context(object):
                        should be a valid ParameterSource value
         """
         ParameterSource.validate(source)
-        self._parameter_sources[name] = source
+        self._sources_by_paramname[name] = source
 
     def get_parameter_source(self, name):
         """Get the source of a parameter.
@@ -630,7 +630,7 @@ class Context(object):
         :returns: the source
         :rtype: ParameterSource
         """
-        return self._parameter_sources[name]
+        return self._sources_by_paramname[name]
 
 
 class BaseCommand(object):
--- a/click/core.py
+++ b/click/core.py
@@ -130,7 +130,7 @@ def iter_params_for_processing(invocation_order, declaration_order):
     return sorted(declaration_order, key=sort_key)
 
 class ParameterSource(object):
-    """This is an enum that indicates the source of a command line option.
+    """This is an enum that indicates the source of a command line parameter.
 
     The enum has one of the following values: COMMANDLINE,
     ENVIRONMENT, DEFAULT, DEFAULT_MAP.  The DEFAULT indicates that the
@@ -143,7 +143,7 @@ class ParameterSource(object):
     DEFAULT = "DEFAULT"
     DEFAULT_MAP = "DEFAULT_MAP"
     
-    VALUES = [COMMANDLINE, ENVIRONMENT, DEFAULT, DEFAULT_MAP]
+    VALUES = {COMMANDLINE, ENVIRONMENT, DEFAULT, DEFAULT_MAP}
     
     @classmethod
     def validate(clz, value):
@@ -157,7 +157,7 @@ class ParameterSource(object):
         """
         if value not in clz.VALUES:
             raise ValueError("Invalid ParameterSource value: '{}'. Valid "
-                             "values are: {}".format(",".join(VALUES)))
+                             "values are: {}".format(value, ",".join(clz.VALUES)))
 
 
 class Context(object):
@@ -617,7 +617,7 @@ class Context(object):
         self._parameter_sources[name] = source
 
     def get_parameter_source(self, name):
-        """Get the `source` of a parameter.
+        """Get the source of a parameter.
 
         This indicates the location from which the value of the
         parameter was obtained.  This can be useful for determining
@@ -630,7 +630,7 @@ class Context(object):
         :returns: the source
         :rtype: ParameterSource
         """
-        return self._parameter_sources.get(name)
+        return self._parameter_sources[name]
 
 
 class BaseCommand(object):
--- a/click/core.py
+++ b/click/core.py
@@ -130,13 +130,14 @@ def iter_params_for_processing(invocation_order, declaration_order):
     return sorted(declaration_order, key=sort_key)
 
 class ParameterSource(object):
+    """This is an enum that indicates the source of a command line option.
+
+    The enum has one of the following values: COMMANDLINE,
+    ENVIRONMENT, DEFAULT, DEFAULT_MAP.  The DEFAULT indicates that the
+    default value in the decorator was used.  This class should be
+    converted to an enum when Python 2 support is dropped.
     """
-    This is an enum that indicates the source of a command line option,
-    which is one of the following: COMMANDLINE, ENVIRONMENT, DEFAULT, DEFAULT_MAP.
-    The DEFAULT indicates that the default value in the decorator was used.
-    This class should be converted to an enum when Python 2 support is
-    dropped.
-    """
+
     COMMANDLINE = "COMMANDLINE"
     ENVIRONMENT = "ENVIRONMENT"
     DEFAULT = "DEFAULT"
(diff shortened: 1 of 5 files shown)


jcrotts:
Could you add an entry to the changelog for this, and it might be good to add a note to the  docs about the change.

joshuastorck:
What section would you recommend that I add the docs to?

jcrotts:
Good question, I think it makes sense to add a section to https://click.palletsprojects.com/en/7.x/parameters/ since your change applies to everything that inherits from BaseCommand.

joshuastorck:
Maybe it makes more sense to put it in the advanced.rst file? The parameters.rst file is pretty small and the concepts around environment variables and default values are in the options.rst and arguments.rst files, which come after the parameters.rst file.

jcrotts:
That sounds good to me, it is a little to specific to go into the parameters section.

joshuastorck:
@jcrotts, docs have been added and tests are passing. Please let me know if there are any other changes you want.

click/core.py
```diff
@@ -129,6 +129,27 @@ def sort_key(item):
 
     return sorted(declaration_order, key=sort_key)
 
+class ParameterSource(object):
```
jab: nit: Should this be named `OptionSource`? Positional arguments can't be set via the environment, have default values, etc. so this is only relevant for options. I think this was the intention since the docstring says "indicates the source of a command line option" already. https://click.palletsprojects.com/en/7.x/parameters/#differences

click/core.py
```diff
@@ -129,6 +129,27 @@ def sort_key(item):
...
+    DEFAULT = "DEFAULT"
+    DEFAULT_MAP = "DEFAULT_MAP"
+    
+    VALUES = [COMMANDLINE, ENVIRONMENT, DEFAULT, DEFAULT_MAP]
+    
+    @classmethod
+    def validate(clz, value):
+        if value not in clz.VALUES:
+            raise ValueError("Invalid ParameterSource value: '{}'. Valid "
+                             "values are: {}".format(",".join(VALUES)))
```
jab:
```python
def validate(cls, value):
    if value not in cls.VALUES:
        raise ValueError("Invalid value: %r not in %r" % (value, cls.VALUES))
```

click/core.py
```diff
@@ -129,6 +129,27 @@ def sort_key(item):
...
+    The DEFAULT indicates that the default value in the decorator was used.
+    This class should be converted to an enum when Python 2 support is
+    dropped.
+    """
+    COMMANDLINE = "COMMANDLINE"
+    ENVIRONMENT = "ENVIRONMENT"
+    DEFAULT = "DEFAULT"
+    DEFAULT_MAP = "DEFAULT_MAP"
+    
+    VALUES = [COMMANDLINE, ENVIRONMENT, DEFAULT, DEFAULT_MAP]
```
jab: `    VALUES = {COMMANDLINE, ENVIRONMENT, DEFAULT, DEFAULT_MAP}`

click/core.py
```diff
@@ -339,7 +360,8 @@ def __init__(self, command, parent=None, info_name=None, obj=None,
 
         self._close_callbacks = []
         self._depth = 0
-
+        self._parameter_sources = {}
```
jab: `        self._source_by_optname = {}`

click/core.py
```diff
@@ -572,6 +594,32 @@ def forward(*args, **kwargs):
 
         return self.invoke(cmd, **kwargs)
 
+    def set_parameter_source(self, name, source):
```
jab: `    def set_option_source(self, name, source):`

click/core.py
```diff
@@ -572,6 +594,32 @@ def forward(*args, **kwargs):
...
+        """Sets the `source` of a parameter. This indicates the 
+        location from which the value of the parameter was obtained.
+
+        :param name: the name of the command line parameter
```
jab: s/parameter/option

click/core.py
```diff
@@ -572,6 +594,32 @@ def forward(*args, **kwargs):
...
+        :param source: the source of the the command line parameter,
+                       which should be a valid ParameterSource value
+        """
+        ParameterSource.validate(source)
+        self._parameter_sources[name] = source
+
+    def get_parameter_source(self, name):
+        """Get the `source` of a parameter. This indicates the
```
jab: nit: docstring fixes as per above

click/core.py
```diff
@@ -572,6 +594,32 @@ def forward(*args, **kwargs):
...
+        an option on the command line that is the same as the default.
+        In that case, the source would be ParameterSource.COMMANDLINE,
+        even though the value of the parameter was equivalent to the
+        default.
+
+        :param name: the name of the command line parameter
+        :returns: the source
+        :rtype: ParameterSource
+        """
+        return self._parameter_sources.get(name)
```
jab: `        return self._source_by_optname[name]`

(raise `KeyError` if invalid name given rather than returning `None`)

jab (changes requested):
Awesome @joshuastorck, I've been wanting this – thanks for submitting! A few suggestions – what do you think? Psyched to get this landed!

click/core.py
joshuastorck: https://click.palletsprojects.com/en/7.x/arguments/#environment-variables states that arguments can be set via the environment, will change the docs to parameter from option.

click/core.py
joshuastorck: I'll add a unit test.

click/core.py
```diff
@@ -572,6 +593,32 @@ def forward(*args, **kwargs):
 
         return self.invoke(cmd, **kwargs)
 
+    def set_parameter_source(self, name, source):
+        """Sets the `source` of a parameter. This indicates the 
+        location from which the value of the parameter was obtained.
+
+        :param name: the name of the command line parameter
+        :param source: the source of the the command line parameter,
```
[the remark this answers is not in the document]
joshuastorck: Fixed

click/core.py
```diff
@@ -572,6 +593,32 @@ def forward(*args, **kwargs):
 
         return self.invoke(cmd, **kwargs)
 
+    def set_parameter_source(self, name, source):
+        """Sets the `source` of a parameter. This indicates the
```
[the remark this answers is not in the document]
joshuastorck: Fixed

click/core.py
```diff
@@ -129,6 +129,26 @@ def sort_key(item):
 
     return sorted(declaration_order, key=sort_key)
 
+class ParameterSource(object):
+    """
```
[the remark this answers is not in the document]
joshuastorck: Fixed

click/core.py
joshuastorck: Fixed

joshuastorck:
@jab, I think I've addressed all of your comments. Please take another look.

jab (changes requested):
Thanks @joshuastorck, looking great, just a couple last things and then I think this should be ready to merge!

tests/test_context.py
```diff
@@ -256,3 +256,62 @@ def cli(ctx):
...
+    def cli(ctx, option):
+        assert ctx.get_parameter_source("option") == click.ParameterSource.ENVIRONMENT
+        sys.exit(1)
+        
+    assert runner.invoke(cli, [], prog_name="test_parameter_source_environment", env={"NAME": "1"}).exit_code == 1
+
+
+def test_validate_parameter_source():
+    with pytest.raises(ValueError):
+        click.ParameterSource.validate("DEFAUL")
```
jab: maybe s/DEFAUL/NOT_A_VALID_PARAMATER_SOURCE?

click/core.py
```diff
@@ -129,6 +129,36 @@ def sort_key(item):
...
+    VALUES = {COMMANDLINE, ENVIRONMENT, DEFAULT, DEFAULT_MAP}
+    
+    @classmethod
+    def validate(clz, value):
+        """Validate that the specified value is a valid enum.
+
+        This method will raise a ValueError if the value is
+        not a valid enum.
+
+        :param clz: ParameterSource.class
```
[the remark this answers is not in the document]
joshuastorck: right, I was switching language context and forgot.

tests/test_context.py
joshuastorck: See https://github.com/pallets/click/blob/master/click/core.py#L1809 in regards to defaults for arguments.

jab (approved):
💯

davidism:
Looks good, thanks!

PR #1130, opened 2018-10-02
docs/testing.rst: Example with unit testing

dendrondal:
Proposed fix for Issue #1119. Tested with pytest 3.0.5 and Python 3.6.0 on Ubuntu 16.04, all 4 tests passed. Should provide an example that's extensible for having a test suite and core CLI in different parts of a python module.

Issue #1119, opened 2018-09-20
Need example with unit testing


mig82:
The examples on testing seem to suggest that the function being defined, say `foo` should be nested within it's corresponding unit test, say `test_foo`. This seems weird. I would have expected unit tests to be written either in a separate file or in a separate function, but nesting the definition within the test feels off. Can you please add a basic unit test to one of the examples for more clarity?

davidism:
All that's required is moving the command definition above the test definition.

Marking this as beginner friendly. Please feel free to submit a PR!

mig82:
Right, and then how do you execute all your defined tests in one go?

davidism:
However the test framework you're using (unittest, pytest, etc.) does.

dposada:
I'm assuming that the examples in question are these:

https://click.palletsprojects.com/en/7.x/testing/

davidism:
@dposada looks like #1130 already addressed this issue

dposada:
I went ahead and created #1330. Based on the original poster's confusion, I think it makes sense to give examples that show the source code in a separate module from the test code. Let me know what you think.

PR #1330, opened 2019-05-31
Separates source from test code in test examples

dposada:
Addresses https://github.com/pallets/click/issues/1119

This is different from #1130, which still shows the source and test code right next to each other.

Example of this PR:
image


dposada:
Tested with pytest 4.5.0 and Python 3.6.8 on Ubuntu 16.04, all four test examples in the doc passed.

PR #1332, opened 2019-05-31
Add an optional current_item parameter to ProgressBar.update()

### LINKED ISSUE #1226: Add `current_item` parameter to `ProgressBar.update()`
When using a ProgressBar in manual update mode, we can call the `ProgressBar.update(n_steps)` method to update the progress bar.  In this usage pattern, the `current_item` property doesn't get set, which means the `item_show_func` feature can't work as it should.  I propose adding an optional `current_item` parameter to update, so update would look like this:

```python
    def update(self, n_steps, current_item=None):
        self.make_step(n_steps)
        if current_item is not None:
            self.current_item = current_item
        self.render_progress()
```

This would allow a usage pattern like this (adapted from click documentation):

```python
with click.progressbar(length=total_size,
                       label='Unzipping archive',
                       item_show_func=lambda a: a.filename) as bar:
    for archive in zip_file:
        archive.extract()
        bar.update(archive.size, archive)
```

I can submit a PR for this if you don't see anything wrong with the idea.

thecav:
Fix #1226   Add an optional `current_item` parameter to `ProgressBar.update()` to allow for `item_show_func` to continue to display when using `update()`.  A usage example for this was added to `ProgressBar` and some new tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,11 +8,16 @@ Unreleased
 
 -   Adds a repr to Command, showing the command name for friendlier debugging. (`#1267`_, `#1295`_)
 -   Add support for distinguishing the source of a command line parameter. (`#1264`_, `#1329`_)
+-   Add an optional parameter to ``ProgressBar.update`` to set the
+    ``current_item``. (`#1226`_, `#1332`_)
 
 .. _#1267: https://github.com/pallets/click/issues/1267
 .. _#1295: https://github.com/pallets/click/pull/1295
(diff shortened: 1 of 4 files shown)


jcrotts:
Could you add an entry to the changelog, and a version added marker to the new docs you added. Thanks for the PR.

thecav:
Sure, would this go under the unreleased 7.1 section?

jcrotts:
David may say otherwise but since this is a new feature it will probably go under 8.0 unreleased. 7.1 is for bugs and docs changes.

PR #1336, opened 2019-06-10
[docs] clarify Parameter behavior when nargs=-1

mattsb42:
The API docs don't actually say how to do this, though there are examples that show it.


jab:
Thank you for the contribution.

PR #1354, opened 2019-07-13
Fix ProgressBar item_show_func so it displays the current item

### LINKED ISSUE #1353: ProgressBar item_show_func displays the previous item
I've noticed a bug with the `item_show_func` option for `click.progressbar` where the progress bar ends up showing the previous item instead of the current one.

The issue seems to be that `ProgressBar.generator()` will only render the progress bar _after_ the current item has been processed. As a result, the previous item is visible on screen while the current one is in processing.

I think this change to [lines 285-288](https://github.com/pallets/click/blob/7.x/click/_termui_impl.py#L285) should fix the issue:
```python
for rv in self.iter:
    self.current_item = rv
    self.render_progress()
    yield rv
    self.make_step(1)
```

JoelEager:
Per @davidism's request this PR contains my suggested fix for `ProgressBar.generator()`. In short, this change makes the `item_show_func` show the item currently being processed as opposed to the item previously processed (like it currently does). I've explained the bug more fully in #1353.

You can use this example to see the difference when run against current master vs this PR.
```python
from time import sleep
from click import progressbar


def format_item(item):
    if item:
        return "processing {}".format(item)


with progressbar(range(5), item_show_func=format_item) as pb:
    for item in pb:
        print(" actually on", item)
        sleep(1)
```

Thanks in advance for taking the time to look into this.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -178,6 +178,8 @@ Unreleased
     is not a TTY. :issue:`1138`
 -   Progress bar output is shown even if execution time is less than 0.5
     seconds. :issue:`1648`
+-   Progress bar ``item_show_func`` shows the current item, not the
+    previous item. :issue:`1353`
 
 
 Version 7.1.2
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -302,14 +302,13 @@ class ProgressBar:
             Only render when the number of steps meets the
             ``update_min_steps`` threshold.
         """
+        if current_item is not None:
+            self.current_item = current_item
+
         self._completed_intervals += n_steps
 
         if self._completed_intervals >= self.update_min_steps:
             self.make_step(self._completed_intervals)
-
-            if current_item is not None:
-                self.current_item = current_item
-
             self.render_progress()
             self._completed_intervals = 0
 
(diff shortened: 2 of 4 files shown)


davidism:
This seems incompatible the changes from #1698 which added an update interval. That PR relies on `update()`, but this one bypasses that and unconditionally calls `make_step()` and `render_progress()`. However, moving `update()` before `yield` means that the ETA calculation is wrong. Incidentally, that PR is when I decided that the progress bar was getting too complex without enough design oversight and started recommending tqdm.

davidism:
#1803 to stop hiding fast progress bars seems to have inadvertently mostly fixed this too. "processing" matches "actual" now, although it's not displayed for the first item (0).

davidism:
Regarding the ETA issue with this fix, this fix doesn't match the docs:

> Note that progressbar() updates the bar after each iteration of the loop. So code like this will render correctly:
> 
> ```python
> import time
> 
> with click.progressbar([1, 2, 3]) as bar:
>     for x in bar:
>         print('sleep({})...'.format(x))
>         time.sleep(x)
> ```

davidism:
In order to reconcile the update interval behavior, I changed this back to using `self.update(1)` after yielding. *Before* yielding, the progress bar is rendered with the new current item, but only at the start of the interval.

PR #1355, opened 2019-07-13
Suggest renaming option from `name` to `--name`

jtrakk:
Usually when I get this error message, it's because I wrote `@click.option('foo')` when I meant `@click.option('--foo')`. So this PR updates the error message to suggest the user fix that.


davidism:
Reworded the message a bit, updated test, added changelog.

PR #1368, opened 2019-08-14
[WIP/RFC] open_stream: pass through "errors" without "encoding"

blueyed:


blueyed:
My use case (why I was looking into it) was to specify "errors" only, but use the default for encoding (with a non-binary file).
Not sure if it's worth it?

davidism:
Closed by #1298

Issue #1370, opened 2019-08-15
click.confirm fails when piping input


holmescharles:
I string a lot of my programs together through piping, but some of them work best with confirmation prompting. However, if I pipe program A's output into program B and B asks for confirmation, click.confirm will not be able to get keyboard input since stdin is the pipe, not the terminal.

It seems that a work around would be to use click.getchar, which uses the terminal no matter what, but this would mean a few lines of code + an if statement to get the the same functionality.

Would it be possible to get click.confirm to always get input from the terminal? Or at least include an option for that?

jcrotts:
Feel free to make a PR for this if you have an implementation in mind.

In general saving a few lines in user code is not a strong reason for a feature request, but if the implementation is low impact it could get added.

aelbehe95:
I was also trying to implement the same functionality. For example 

```py
import click

def get_name(ctx, param, value):
    if not value and not click.get_text_stream('stdin').isatty():
        return click.get_text_stream('stdin').read().strip()
    else:
        return value


@click.command()
@click.argument('name', callback=get_name, required=False)
def say_hello(name):
    #click.confirm('Do you want to continue?', abort=True, default=True)
    click.echo('Hello {}'.format(name))



if __name__ == "__main__":
    say_hello()
```

#### Normal test case
```
λ python test_pipe.py click
Hello click
```
#### Missing argument
```
λ python test_pipe.py
Hello None
```

#### Pipelined argument
```
λ echo click | python test_pipe.py
Hello click
```

Now when confirmation logic added

```py
@click.command()
@click.argument('name', callback=get_name, required=False)
def say_hello(name):
    click.confirm('Do you want to continue?', abort=True, default=True)
    click.echo('Hello {}'.format(name))
```


#### Normal test case
```
λ python test_pipe.py click
Do you want to continue? [Y/n]: y
Hello click
```
#### Missing argument
```
λ python test_pipe.py
Do you want to continue? [Y/n]: y
Hello None
```

#### Pipelined argument
```
λ echo click | python test_pipe.py
Do you want to continue? [Y/n]: Aborted!
```

I have then added the following modifications

At `termui.py` i have added the following method

```py
def _get_user_confirmation():
    ret = ''
    while 1:
        c = getchar()
        if c == '\r':
            break
        elif c in ('\x08', '\x7f') and len(ret) != 0:
            echo("\b \b", nl=False)
            ret = ret[:-1]
        else:
            ret += c
            echo(c, nl=False)
[...]

PR #1372, opened 2019-08-20
make prompt work when stdin is a pipe

9ronoob:
A fix for https://github.com/pallets/click/issues/1370

### DIFF
--- a/click/termui.py
+++ b/click/termui.py
@@ -15,6 +15,7 @@ from .globals import resolve_color_default
 # The prompt functions to use.  The doc tools currently override these
 # functions to customize how they work.
 visible_prompt_func = raw_input
+console_open = open
 
 _ansi_colors = {
     'black': 30,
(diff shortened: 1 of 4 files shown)


click/termui.py
```diff
@@ -164,9 +165,15 @@ def confirm(text, default=False, abort=False, prompt_suffix=': ',
...
             # coloring through colorama on Windows
             echo(prompt, nl=False, err=err)
-            value = visible_prompt_func('').lower().strip()
+            old_stdin = sys.stdin
+            if WIN:
+                sys.stdin = console_open("CON:")
+            else:
+                sys.stdin = console_open("/dev/tty")
+            value = visible_prompt_func().lower().strip()
+            sys.stdin = old_stdin
```
davidism: This overwrites everything else Click is doing to set up a stream correctly.

aelbehe95:
This overwrites everything else Click is doing to set up a stream correctly.

Thanks for pointing out.

Correct me if i am wrong; this only overwrites the current behavior if WIN
 and PY2 conditions are met right ?

rougeth:
Any chance of having this approved and merged? @aelbehe95 can you update the PR with the recent changes?

PR #1377, opened 2019-08-25
Fix simple typo: aliase -> alias

timgates42:


examples/aliases/aliases.py
```diff
@@ -41,7 +41,7 @@ def get_command(self, ctx, cmd_name):
         # will create the config object is missing.
         cfg = ctx.ensure_object(Config)
 
-        # Step three: lookup an explicit command aliase in the config
+        # Step three: lookup an explicit command alias in the config
```
jab: Can you please change “lookup” to “look up” while you’re fixing this?

jab (changes requested):

examples/aliases/aliases.py
timgates42: Sure thing

timgates42: Thanks for spotting that - fixed

jab:
Thank you for the contribution!

Issue #320, opened 2015-04-05
Use python-atomicwrites for atomic file writes?


untitaker:
https://github.com/untitaker/python-atomicwrites

Armin thinks that it should be an optional dependency, just for Windows support (like colorama).

untitaker:
Decided to postpone this, Click's current thing supports Windows. The only advantage `python-atomicwrites` has is the better advanced API and the support for asserting that the target files don't exist.

davidism:
Going to close this. atomicwrites is cool, but we're unlikely to add an external dependency, although a click-atomicwrites extension could add a type that used it instead.

PR #1382, opened 2019-08-29
WIP: Set permissions for atomic open_file()

### LINKED ISSUE #1376: open_file(…, 'w', atomic=True) sets or resets permissions to 600, ignoring umask
Files are normally created with permission `666 & umask` as expected:

```pycon
>>> click.open_file('/tmp/test1', 'w').close()
>>> oct(os.stat('/tmp/test1').st_mode)
'0o100644'
```

However, opening a new or existing file with `atomic=True` sets or resets its permissions to `600` unconditionally:

```python
>>> click.open_file('/tmp/test2', 'w', atomic=True).close()
>>> oct(os.stat('/tmp/test2').st_mode)
'0o100600'
>>> click.open_file('/tmp/test1', 'w', atomic=True).close()
>>> oct(os.stat('/tmp/test1').st_mode)
'0o100600'
```

I’d expect a new file created with `atomic=True` to be created with the usual permission `666 & umask`, and an existing file opened with `atomic=True` to have the previous permissions preserved. (There’s probably no way to preserve permissions atomically; taking the previous permissions at the time of the initial open is fine.)

msmolens:
Previously, when `open_file()` was called with `atomic=True`, the target
file's permissions were always set to 0600, i.e. readable and writable
by the current user. These permissions come from the temporary file
created by [`tempfile.mkstemp()`](https://docs.python.org/3.7/library/tempfile.html#tempfile.mkstemp).

This commit changes an atomic `open_file()` call to set the permissions
of the target file. If the target file already exists, then its current
permissions are retained. Otherwise, the permissions respect the current
umask.

Fixes #1376

### DIFF
--- a/click/_compat.py
+++ b/click/_compat.py
@@ -1,6 +1,7 @@
 import re
 import io
 import os
+import stat
 import sys
 import codecs
 from weakref import WeakKeyDictionary
@@ -32,6 +33,12 @@ def _make_text_stream(stream, encoding, errors,
                                     line_buffering=True,
                                     force_readable=force_readable,
                                     force_writable=force_writable)
+
+def _get_current_umask():
+    """Get current umask."""
+    umask = os.umask(0)
+    os.umask(umask)
+    return umask
 
 
 def is_ascii_encoding(encoding):
@@ -512,7 +519,19 @@ def open_stream(filename, mode='r', encoding=None, errors='strict',
     else:
         f = os.fdopen(fd, mode)
 
-    return _AtomicFile(f, tmp_filename, os.path.realpath(filename)), True
+    real_filename = os.path.realpath(filename)
+
+    # Get permissions to set on the target file. If the target file already
+    # exists, retain its current permissions. If the target file will be
+    # created, respect the permissions from the current umask.
+    permissions = None
+    if not WIN:
+        try:
+            permissions = stat.S_IMODE(os.stat(real_filename).st_mode)
+        except OSError:
+            permissions = 0o666 & ~_get_current_umask()
+
+    return _AtomicFile(f, tmp_filename, real_filename, permissions), True
 
 
 # Used in a destructor call, needs extra protection from interpreter cleanup.
@@ -526,10 +545,11 @@ else:
 
 class _AtomicFile(object):
 
-    def __init__(self, f, tmp_filename, real_filename):
+    def __init__(self, f, tmp_filename, real_filename, permissions):
         self._f = f
         self._tmp_filename = tmp_filename
         self._real_filename = real_filename
+        self._permissions = permissions
         self.closed = False
 
     @property
@@ -546,6 +566,8 @@ class _AtomicFile(object):
             except OSError:
                 pass
         _replace(self._tmp_filename, self._real_filename)
+        if self._permissions is not None:
+            os.chmod(self._real_filename, self._permissions)
         self.closed = True
 
     def __getattr__(self, name):
--- a/click/_compat.py
+++ b/click/_compat.py
@@ -528,7 +528,7 @@ def open_stream(filename, mode='r', encoding=None, errors='strict',
     if not WIN:
         try:
             permissions = stat.S_IMODE(os.stat(real_filename).st_mode)
-        except OSError:
+        except (OSError, IOError):
             permissions = 0o666 & ~_get_current_umask()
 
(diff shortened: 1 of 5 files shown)


click/_compat.py
```diff
@@ -33,6 +34,12 @@ def _make_text_stream(stream, encoding, errors,
                                     force_readable=force_readable,
                                     force_writable=force_writable)
 
+def _get_current_umask():
+    """Get current umask."""
+    umask = os.umask(0)
```
andersk: This is, unfortunately, a potential security problem in the presence of multiple threads. (See https://bugs.python.org/issue21082.)

I’m not sure what to suggest instead, other than avoiding `tempfile` and writing the loop to safely create a fresh filename manually.

click/_compat.py
```diff
@@ -512,7 +519,19 @@ def open_stream(filename, mode='r', encoding=None, errors='strict',
...
+
+    # Get permissions to set on the target file. If the target file already
+    # exists, retain its current permissions. If the target file will be
+    # created, respect the permissions from the current umask.
+    permissions = None
+    if not WIN:
+        try:
+            permissions = stat.S_IMODE(os.stat(real_filename).st_mode)
+        except OSError:
+            permissions = 0o666 & ~_get_current_umask()
```
atugushev: IMO for the fallback it's better to set here a default permission, like 0o666 or even more paranoidal 0o600. As @andersk right said it's not thread-safe.

atugushev: Or even better, just `permissions = None`.

click/_compat.py
```diff
@@ -512,7 +519,19 @@ def open_stream(filename, mode='r', encoding=None, errors='strict',
...
+    real_filename = os.path.realpath(filename)
+
+    # Get permissions to set on the target file. If the target file already
+    # exists, retain its current permissions. If the target file will be
+    # created, respect the permissions from the current umask.
+    permissions = None
+    if not WIN:
+        try:
+            permissions = stat.S_IMODE(os.stat(real_filename).st_mode)
+        except OSError:
```
atugushev:
```suggestion
        except (OSError, IOError):
```

Handle `IOError` for py2 compatibility.

click/_compat.py
andersk: The default permissions if you didn’t specify `atomic=True` are implicitly `0o666 & ~umask`, because that’s the default behavior of `open()` (no explicit umask computation is required). So it’s unexpected for `atomic=True` to result in different permissions. `permissions = 0o666` is definitely too permissive, as `chmod` is *not* implicitly masked by `umask`, while `permissions = 0o600` (or `pass`, which has the same effect) is too restrictive.

The underlying problem is that `tempfile` simply has the wrong permissions behavior for this use case. It’s designed for temporary files that will later be deleted, not for creating files with which to atomically replace other files.

atugushev: You are right, anyways at least on save an existing file there will be correct permissions.

click/_compat.py
msmolens: baded409cfef144e3acd0e8d8f3c7c7e197723e7

davidism:
I really appreciate the work you're putting into this, but this is *way* more complexity than I'm willing to add and support. Vendoring two copies of Python's tempfile module is not worth it. Does python-atomicwrites suggested in #320 have this issue? cc @untitaker

msmolens:
@davidism Thanks for the review. I agree that vendoring the modules is heavy-handed and not a great way forward. I'll close this pull request for now.

Note that python-atomicwrites also changes the permissions: https://github.com/untitaker/python-atomicwrites/issues/42

untitaker:
Atomicwrites does NOT intentionally change permissions. Different permissions come from a result of how the file write is done.

andersk:
I took a more minimalist stab at this myself in #1400. @msmolens I copied your tests from here; hope that’s fine.

@untitaker You can certainly argue that it “does not change permissions” because it internally works by creating a new file with different permissions and renaming it over the old file. But that argument is of little interest to a developer who wanted the permissions to stay the same.

untitaker:
I believe #1400 is a correct fix for this, and I am surprised that one exists. I would be happy to have this in atomicwrites. After all its stated motto is to solve this rabbit hole once and for all.

We could keep the impl in click simple and have atomicwrites as optional dep for "better" behavior.

PR #1385, opened 2019-09-02
Documentation about Shared options

edusantana:
This way we can see a nice workaround.

Extracted from #108


davidism:
Thanks, but as stated in the issue this is not a pattern we want to promote.

Issue #1065, opened 2018-07-10
Incorrect OSError raised when redirecting to NUL on Windows


uranusjr:
Reproducible for both Python 3.6 and 3.7. I did not test against older versions.

Tested on the Click 6.x series. Windows 10, Build 17134.

Steps to reproduce:

```python
# t.py
import click

@click.command()
def main():
    click.echo('run')

if __name__ == '__main__':
    main()
```

```
> python t.py
run

> python t.py >NUL
Traceback (most recent call last):
  File "t.py", line 10, in <module>
    main()
  File ".venv\lib\site-packages\click\core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File ".venv\lib\site-packages\click\core.py", line 697, in main
    rv = self.invoke(ctx)
  File ".venv\lib\site-packages\click\core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File ".venv\lib\site-packages\click\core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "t.py", line 6, in main
    click.echo('run')
  File ".venv\lib\site-packages\click\utils.py", line 259, in echo
    file.write(message)
  File ".venv\lib\site-packages\click\_winconsole.py", line 180, in write
    return self._text_stream.write(x)
  File ".venv\lib\site-packages\click\_winconsole.py", line 164, in write
    raise OSError(self._get_error_message(GetLastError()))
OSError: Windows error 1
```

I believe the problem is that Click tries to be helpful, and raises an error when there are not bytes written when there should be. This is, however, exactly what is expected to happen when you redirect things to `NUL` on Windows.

Curiously, however, Windows seems to act a little differently for stderr. `2>NUL` works as expected.

uranusjr:
I added an additional check based on [MSDN’s suggestion](https://docs.microsoft.com/en-us/windows/console/writeconsole#remarks):

```diff
--- a/click/_winconsole.py
+++ b/click/_winconsole.py
@@ -33,6 +33,8 @@ kernel32 = windll.kernel32
 GetStdHandle = kernel32.GetStdHandle
 ReadConsoleW = kernel32.ReadConsoleW
 WriteConsoleW = kernel32.WriteConsoleW
+WriteFile = kernel32.WriteFile
+GetConsoleMode = kernel32.GetConsoleMode
 GetLastError = kernel32.GetLastError
 GetCommandLineW = WINFUNCTYPE(LPWSTR)(
[...]

PR #1393, opened 2019-09-13
Fix OSError when running on MinGW64 with MSYS2

### LINKED ISSUE #1338: OSError on MinGW64 with MSYS2 on Windows 8.1
A most basic script containing:

    import click
    click.echo('run')

Produces the following stacktrace:

      File "t.py", line 3, in <module>
        click.echo('run')
      File "C:/msys64/mingw64/lib/python3.7/site-packages\click\utils.py", line 260, in echo
        file.write(message)
      File "C:/msys64/mingw64/lib/python3.7/site-packages\click\_compat.py", line 630, in _safe_write
        return _write(s)
      File "C:/msys64/mingw64/lib/python3.7/site-packages\colorama\ansitowin32.py", line 41, in write
        self.__convertor.write(text)
      File "C:/msys64/mingw64/lib/python3.7/site-packages\colorama\ansitowin32.py", line 164, in write
        self.wrapped.write(text)
      File "C:/msys64/mingw64/lib/python3.7/site-packages\click\_winconsole.py", line 180, in write
        return self._text_stream.write(x)
      File "C:/msys64/mingw64/lib/python3.7/site-packages\click\_winconsole.py", line 164, in write
        raise OSError(self._get_error_message(GetLastError()))
    OSError: Windows error 1

I am running under a MinGW64 bash shell installed through MSYS2 on a Windows 8.1 machine. 
This is with Python 3.7, colorama 0.4.1, and Click 7.0, but the error goes back to 6.0, whereas it does not occur on 5.1. Python 2 does not exhibit the error.

danyeaw:
Fixes #1338.

In version 6.0, echo and prompt functions were modified to work with full unicode in the windows console by emulating an output stream. Since MSYS2 uses a bash console it was creating an OSError to apply these modifications when running in this environment.

This PR changes these modifications to only apply them if not running in a MSYS2 environment.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,7 @@ Unreleased
 -   Fix output on Windows Python 2.7 built with MSVC 14. :pr:`1342`
 -   Always return of of the passed choices for ``click.Choice``
     :issue:`1277`, :pr:`1318`
+-   Fix ``OSError`` when running in MSYS2. :issue:`1338`
 
 
 Version 7.0
--- a/click/_compat.py
+++ b/click/_compat.py
@@ -8,10 +8,11 @@ from weakref import WeakKeyDictionary
 
 PY2 = sys.version_info[0] == 2
 CYGWIN = sys.platform.startswith('cygwin')
+MSYS2 = sys.platform.startswith('win') and ('GCC' in sys.version)
 # Determine local App Engine environment, per Google's own suggestion
 APP_ENGINE = ('APPENGINE_RUNTIME' in os.environ and
               'Development/' in os.environ['SERVER_SOFTWARE'])
-WIN = sys.platform.startswith('win') and not APP_ENGINE
+WIN = sys.platform.startswith('win') and not APP_ENGINE and not MSYS2
 DEFAULT_COLUMNS = 80
 
 


danyeaw:
Thanks for the awesome library! :heart: Any interest in merging these changes or will we need to fix this downstream?

davidism:
This check seems like it could cause false positives. Is there no better way to detect that MinGW is used?

danyeaw:
@davidism I am not sure how you could have GCC on a Windows platform without using MINGW. Was there a false positive you had in mind?

Downstream, MINGW patches sysconfig, so we could do something like:
```
try:
     from sysconfig import _POSIX_BUILD
     MSYS2 = _POSIX_BUILD
except ImportError:
     MSYS2 = False

This seems kind of messy compared to the solution I used though. :smile:

davidism:
I'm not super familiar with these types of tools, but what about others like Cygwin?

danyeaw:
@davidism I ended up installing Cygwin to test this out. You are correct in that Python installed in Cygwin is compiled with GCC as well. However, `sys.platform.startswith('win')` returns False, since `sys.platform` returns `cygwin`. I don't have access to WSL, but I understand that `sys.platform` returns `linux`. Click in Cygwin works fine currently with master and this branch.

I think my changes might be a little more clear if I did:
`MSYS2 = sys.platform.startwith('win') and ('GCC' in sys.version)`
and then:
`WIN = sys.platform.startswith('win') and not APP_ENGINE and not MSYS2`

and then remove my changes from the `if` statement. Do you agree? If so I can update my PR.

ThiefMaster:
Yes, WSL is Linux as far as userspace is concerned (and `sys.platform` indeed returns `'linux'` there)

davidism:
I like the suggested more specific version.

danyeaw:
@davidism I have updated the PR with the discussed changes :+1:

danyeaw:
Thanks @davidism

davidism:
Thanks for the fix!

Right after I merged this, I noticed that it's very similar to issue #1065, although the fix in #1135 addresses the Windows API rather than detecting MSYS2. Any thoughts on their relation?

davidism:
It appears that this fixed #1065, so I need to figure out if I should merge #1135 too. cc @segevfiner

PR #1395, opened 2019-09-15
Assure choices are strings

### LINKED ISSUE #591: Support data types other than string for click.Choice values
Several times I have used click.Choice where I want my set of values to be integers.  However this class requires that the values be strings, and the resulting value is also a string (requiring conversion in my code before and after).  Is it not possible to allow other types?  Certainly it would be reasonable to raise an exception if the set of choices contained a variety of types, but as long as they were consistently one type this seems a useful improvement.  I'd be happy to work on such a feature, but wanted to check if this seems reasonable before I expend the effort.  Thanks!  (I LOVE the library, by the way!)

ssbarnea:
Fixes problem where the iterator may return different objects, making
choice fail when trying to use string specific methods.

Example:
```
    return '[%s]' % '|'.join(self.choices)
TypeError: sequence item 0: expected str instance, XXX found
```
Fixes: #591

### DIFF
--- a/click/types.py
+++ b/click/types.py
@@ -142,7 +142,10 @@ class Choice(ParamType):
     name = 'choice'
 
     def __init__(self, choices, case_sensitive=True):
-        self.choices = choices
+        # assures we have a list of strings or integers to avoid join failure
+        self.choices = [
+            str(s) if not isinstance(s, (int, str)) else s
+            for s in choices]
         self.case_sensitive = case_sensitive
 
     def get_metavar(self, param):


magiskboy (approved):

click/types.py
```diff
@@ -142,7 +142,10 @@ class Choice(ParamType):
     name = 'choice'
 
     def __init__(self, choices, case_sensitive=True):
-        self.choices = choices
+        # assures we have a list of strings or integers to avoid join failure
```
hg-pyun: Good comment.

jcrotts:
Thanks for the PR! I'm not sure if we should convert to strings here. I think as a user I would prefer the error message over the implicit conversion.

davidism:
The point about raising an error stands, but I'm also not sure I want to add more runtime type checking. Now that Pallets is Python 3.6+, typing should probably be handled by adding annotations throughout the library.

davidism:
Thanks for looking at this, but at this point other Pallets projects are adding annotations, which I want to prefer over runtime type checking. The docs already call out that all choices must be strings. On that point, this PR allows ints, which will still fail.

PR #1400, opened 2019-09-19
Preserve expected permissions in atomic open_file

### LINKED ISSUE #1376: open_file(…, 'w', atomic=True) sets or resets permissions to 600, ignoring umask

andersk:
Fixes #1376. Replacement for closed PR #1382.

### DIFF
--- a/click/_compat.py
+++ b/click/_compat.py
@@ -504,9 +504,38 @@ def open_stream(filename, mode='r', encoding=None, errors='strict',
     # as a proxy in the same folder and then using the fdopen
     # functionality to wrap it in a Python file.  Then we wrap it in an
     # atomic file that moves the file over on close.
-    import tempfile
-    fd, tmp_filename = tempfile.mkstemp(dir=os.path.dirname(filename),
-                                        prefix='.__atomic-write')
+    import errno
+    import random
+
+    try:
+        perm = os.stat(filename).st_mode
+    except OSError:
+        perm = None
+
+    flags = os.O_RDWR | os.O_CREAT | os.O_EXCL
+    if 'b' in mode:
+        flags |= getattr(os, 'O_BINARY', 0)
+
+    while True:
+        tmp_filename = os.path.join(
+            os.path.dirname(filename),
+            '.__atomic-write%08x' % (random.randrange(1 << 32),),
+        )
+        try:
+            fd = os.open(tmp_filename, flags, 0o666 if perm is None else perm)
+            break
+        except OSError as e:
+            if e.errno == errno.EEXIST or (
+                os.name == "nt"
+                and e.errno == errno.EACCES
+                and os.path.isdir(dir)
+                and os.access(dir, os.W_OK)
+            ):
+                continue
+            raise
+
+    if perm is not None:
+        os.chmod(tmp_filename, perm)  # in case perm includes bits in umask
 
     if encoding is not None:
         f = io.open(fd, mode, encoding=encoding, errors=errors)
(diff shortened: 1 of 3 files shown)


click/_compat.py
```diff
@@ -502,9 +502,33 @@ def open_stream(filename, mode='r', encoding=None, errors='strict',
...
+    while True:
+        tmp_filename = os.path.join(
+            os.path.dirname(filename),
+            '.__atomic-write%08x' % (random.randrange(1 << 32),),
+        )
+        try:
+            fd = os.open(tmp_filename, flags, 0o666 if perm is None else perm)
+            break
+        except OSError as e:
+            if e.errno == errno.EEXIST:
```
untitaker: The errorhandling in the implementation of `mkstemp` seems to be more sophisticated, and I believe it would be good if it were ported over as well.

untitaker: See https://github.com/python/cpython/blob/c47c8ba2969c9faf1c036b3dc4b0a5cf0d3aa0cc/Lib/tempfile.py#L248

davidism:
I'm fine with this in 7.1, but that means you need to rebase this onto the 7.x branch.

untitaker:
Nvm it still stats the old file. Still racy

andersk:
> Nvm it still stats the old file. Still racy

@untitaker This is an atomic write API, not a transaction API. As you know, if you use the file contents (say) to maintain a counter *n*, and one process increments it by 2 while a concurrent process increments it by 3, you will get one of *n* + 2, *n* + 3, *n* + 5. In other words, we guarantee that at least one of the updates will be applied consistently, but we can’t guarantee that the other update won’t be lost.

The OS’s `chmod` API is also an atomic write API. If two processes attempt to update the permissions concurrently, say to add independent bits, at least one of the updates will be applied consistently, but the other might be lost.

This PR allows these guarantees to be combined, treating the contents and permissions as a single atomic unit. If one process updates the file contents while a concurrent process updates the permissions, you might get the new contents with the old permissions, or the new contents with the new permissions. At least one of the updates will be applied consistently, but the other might be lost.

What this PR doesn’t provide is the ability to treat the contents and permissions as _independent_ atomic units. I’m not aware of any OS interface short of special transactional filesystems that would allow that to be implemented.

But that’s okay because people don’t want to use permission bits to store concurrently updated mutable state. They just want the existing permissions to be preserved. If an application doing atomic writes needs to be stopped to reliably reconfigure its permissions, that’s fine. It’s certainly better than making no effort to preserve permissions at all.

click/_compat.py
andersk: Sure, done.

andersk:
@davidism Rebased onto 7.x.

untitaker:
From my understanding the content and permissions are not necessarily part of the same atomic unit, because the file content and the permissions might be read independently (i.e. the permissions fetched via `stat()` might be from a different file than the one that is being read for its contents). In that sense the inconsistent state is created when reading the old one, not when writing.

I'm not implying that your patch is bad (I believe there can't be a better one), but I believe there is an argument for not trying to solve this problem at all such that the user has to think about it explicitly.

PR #1402, opened 2019-09-30
Add repeating behavior when click.confirm() is called with default=None.

### LINKED ISSUE #1381: Option to reask on no input in `click.confirm`
It would be nice if similarly to `click.prompt`, `click.confirm` would allow to reask the user if no input is given.

yurihs:
Hello!

This fixes #1381. With these changes, when `confirm()` is called with  `default=None`, the prompt will repeat until valid input or an abort. Otherwise, the previous behavior is preserved (`default=False`).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -164,6 +164,8 @@ Unreleased
 -   Control the location of the temporary directory created by
     ``CLIRunner.isolated_filesystem`` by passing ``temp_dir``. A custom
     directory will not be removed automatically. :issue:`395`
+-   ``click.confirm()`` will prompt until input is given if called with
+    ``default=None``. :issue:`#1381`
 
 
 Version 7.1.2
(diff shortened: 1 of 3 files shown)


chaoflow:
@yurihs This seems to implement ``default=None`` for ``click.confirm``, which is great and I'd like to see this merged. Only your current commit message and changelog entry are misleading.

click/termui.py
```diff
@@ -171,7 +175,7 @@ def confirm(text, default=False, abort=False, prompt_suffix=': ',
             rv = True
         elif value in ('n', 'no'):
             rv = False
-        elif value == '':
+        elif default is not None and value == '':
```
chaoflow: @yurihs Changing 164/165 above to the following might make sense:
```python
    prompt = click.termui._build_prompt(
        text, prompt_suffix, show_default,
        'y/n' if default is None else default and 'Y/n' or 'y/N',
    )
```

yurihs: Thanks for reviewing! Yeah, I hadn't noticed that part. I've implemented your change.

I totally mixed up the functions in the changelog entry 😅

I think everything is OK now.

PR #1404, opened 2019-10-03
Begin adding a loading bar utility

nvander1:
This similar to `progressbar` in that it shows an animation while a task is being performed. However, this is intended to show more of cycling animation for a background task (think the beachball on macOS or the hourglass on Windows).

It is configurable with the character used in the bar, the length of the bar, how fast the animation moves, and what message to display alongside the bar.

For example,

```python
import click
import time

print "foo"
with click.loadingbar(msg='Doing work', fill_char='o', interval=0.1, width=8):
    print "baz"
    time.sleep(3)

print "bar"
```

Will produce a prompt like this:

```
foo
Doing work [ o      ]
bar
```

Where the `o` is moving on character in either direction every 0.1 seconds. The "foo" is printed before we create the loadingbar. "baz" inside the loading bar is suppressed so we don't mess up the loadingbar. And then we print "bar" after we exit the loadingbar "with-statement" block.

### DIFF
--- a/click/__init__.py
+++ b/click/__init__.py
@@ -35,7 +35,7 @@ from .utils import echo, get_binary_stream, get_text_stream, open_file, \
 # Terminal functions
 from .termui import prompt, confirm, get_terminal_size, echo_via_pager, \
      progressbar, clear, style, unstyle, secho, edit, launch, getchar, \
-     pause
+     pause, loadingbar
 
 # Exceptions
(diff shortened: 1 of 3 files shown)


davidism:
Thanks for your interest in this. I don't want to merge this because I want to avoid adding more complexity to Click's progress bar support at this point. This might be good as an extension package. Or check out [tqdm](https://github.com/tqdm/tqdm) for a library dedicated to progress bars.

PR #1406, opened 2019-10-09
Make option format replace newlines with spaces.

jcrotts:
#834 
Remade to point 7.x


hg-pyun (commented):
Resolve conflict.

davidism:
Turns out the underlying `wrap_text` function already takes a `preserve_paragraphs` option, but it's not enabled for option help text right now. This enables the behavior in the command help text where `\n\n` is preserved and each paragraph is wrapped separately.

I'll submit a new PR that rolls this change back and uses `preserve_paragraphs` instead.

PR #1417, opened 2019-10-28
Undefined name: import sys for lines 300 and 311

cclauss:


davidism:
Thanks for pointing this out. The tests passed by accident since the import error also caused `exit_code == 1`. It looks like using `sys.exit()` instead of `ctx.exit()` was a typo, and the exit code isn't even needed, so I'll rewrite these tests.

PR #1418, opened 2019-10-30
ZSH completions escape

timka:
Completion in ZSH usually escapes spaces and other special characters in file names etc. I think this is the expected behavior.

As I see `-Q` has been introduced in https://github.com/stopthatcow/click/commit/13b84c5c4982f35988b812484cf9b76c8eba4ae6 which fixed sorting. However this options seems unnecessary and unrelated to sort order.

Also, I haven't found a straightforward way to test this. In case this breaks something else I'd figure out a way to override the default completion function.


davidism:
I'm unclear why you want to remove it. I had a hard time finding documentation on the `-Q` option to `_describe`, but it looks like that's what quotes values, which you said was expected.

PR #1425, opened 2019-11-15
Fix a small bug on the document about testing

wy193777:


davidism:
`@cli.command` already registered the `sync` command with the `cli` group.

PR #1429, opened 2019-11-22
click.style: add support for 256/RGB colors

lmjohns3:
This adds support for true-color text (using the \033[38;2;RRR;GGG;BBB escape sequence) and 256-color palette (using the \033[38;5;NNN escape sequence). It's backwards compatible with the current string-based colors, but adds a type-checking helper to handle integers or tuples/lists of 3 integers.


Zethson:
Very nice!

PR #1445, opened 2019-12-19
Fix SyntaxWarning in test over literal comparison.

tirkarthi:
The test generates a `SyntaxWarning` in Python 3.8 over using is to compare literals.

```
python -Wall tests/test_basic.py
tests/test_basic.py:51: SyntaxWarning: "is" with a literal. Did you mean "=="?
  assert rv is 42
```


jab:
Thank you for the contribution!

tirkarthi:
Thanks @jab :)

PR #1468, opened 2020-02-10
fix: always use parameter UTF-8 fallback

muggenhor:
Even when the filesystem encoding and sys.argv encoding are the same we
should still fall back to trying UTF-8 for decoding command line
arguments.

### DIFF
--- a/click/types.py
+++ b/click/types.py
@@ -119,6 +119,8 @@ class StringParamType(ParamType):
                         value = value.decode(fs_enc)
                     except UnicodeError:
                         value = value.decode('utf-8', 'replace')
+                else:
+                    value = value.decode('utf-8', 'replace')
             return value
         return value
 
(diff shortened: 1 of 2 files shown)


davidism:
This *seems* ok, but I can't tell what it might affect. Can you add a test case?

click/types.py
```diff
@@ -119,6 +119,8 @@ def convert(self, value, param, ctx):
                         value = value.decode(fs_enc)
                     except UnicodeError:
                         value = value.decode('utf-8', 'replace')
+                else:
+                    value = value.decode('utf-8', 'replace')
```
hg-pyun: Please, Add TC :)

muggenhor: Working on it. I'll aim to push a test case next week.

muggenhor:
I've made the test case commit come first. That way it's easier to verify that it fails without the fix commit.

davidism:
Just leaving a note for myself in case I need to revisit this later. It seems like `.decode("utf-8", "replace")` is already used as a fallback, it just wasn't applied consistently. Since this seems to improve consistency, I'm merging it.

Thanks for working on this!

PR #1492, opened 2020-03-09
deprecate old param callback style

davidism:
This has been emitting a warning since 2.0. Turn it into a deprecation warning to be removed in 8.0.


davidism:
Picked to 7.x branch: 18d6551b01fae57046520d8cc2a019ac0102e074

PR #1497, opened 2020-03-11
CliRunner isolation stdin has name and mode

### LINKED ISSUE #1064: stdin click.File has `.name` attribute when running from CLI, but doesn't when using `CliRunner`
### Expected Result

`@click.argument('infile', type=click.File('r'))` has attribute `name` in all invokation scenarios

### Actual result

`@click.argument('infile', type=click.File('r'))` doesn't have attribute `name` when using `click.testing.CliRunner.invoke()`

### Steps to reproduce

    @click.command()
    @click.argument('infile', type=click.File('r'))
    def test(infile):
        print(infile)
        print(infile.name)
    test()

run as

    python test.py -

returns

    <_io.TextIOWrapper name='<stdin>' mode='r' encoding='UTF-8'>
    <stdin>

whereas

    @click.command()
    @click.argument('infile', type=click.File('r'))
    def test(infile):
        print(infile)
        print(infile.name)  # Raises AttributeError

    runner = CliRunner()
    result = runner.invoke(test, ['-'], input="")
    print(result.output)
    print(result)

returns

    <_io.TextIOWrapper encoding='utf-8'>

    <Result AttributeError("'_io.BytesIO' object has no attribute 'name'",)>

jyscao:
* create subclass of io.TextIOWrapper, NamedTextIOWrapper, for PY3
* use NamedTextIOWrapper instance for sys.stdin & sys.stdout inside
  of CliRunner's isolation contextmanager

fixes #1064 

A few more notes:
* `'name'` & `'mode'` are assigned to `self.__dict__` as kv-pairs, and not as actual attributes to the `self` object, because b/c the `io.TextIOWrapper` object is not writable
* then `__getattribute__` is used to get the above 2 pseudo-attributes from `self.__dict__`
* the bug behavior described #1064 also exists in PY2, but this PR does not attempt to address that (since Python 2 support will eventually be dropped)
* `NamedTextIOWrapper` is not applied to `sys.stderr`, since click does not readily make it available as a `click.File` parameter type

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -79,6 +79,9 @@ Unreleased
     the parent context. :issue:`1565`
 -   ``click.style()`` can output 256 and RGB color codes. Most modern
     terminals support these codes. :pr:`1429`
+-   When using ``CliRunner.invoke()``, the replaced ``stdin`` file has
+    ``name`` and ``mode`` attributes. This lets ``File`` options with
+    the ``-`` value match non-testing behavior. :issue:`1064`
 
 
 Version 7.1.2
--- a/src/click/testing.py
+++ b/src/click/testing.py
@@ -40,6 +40,21 @@ class EchoingStdin:
         return repr(self._input)
 
 
+class _NamedTextIOWrapper(io.TextIOWrapper):
+    def __init__(self, buffer, name=None, mode=None, **kwargs):
+        super().__init__(buffer, **kwargs)
+        self._name = name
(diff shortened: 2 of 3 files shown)


kdeldycke:
Can confirm the issue and the fix! 👍

davidism:
Added changelog and test. Rewrote this to use properties instead of messing with `__dict__` and `__getattr__`. Applied it to all three streams.

Issue #1502, opened 2020-03-14
wrap_text() produces premature line break


andy-maier:
---

### Expected Behavior

`click.wrap_text(..., preserve_paragraphs=False)` should not produce the unmotivated premature line break shown below, but instead break as specified by the width.

```
$ python -c "import click; print(repr(click.wrap_text('line A1\nline A2\n', width=12)))"
'line A1\nline A2'
```

### Actual Behavior

```
$ python -c "import click; print(repr(click.wrap_text('line A1\nline A2\n', width=12)))"
'line A1\nline\nA2'
```
The second line got broken as if the character counting to determine the wrap position had ignored the line end after the first line. If the width gets varied, the second line does not get broken up to 11, and starting with 12 the extra break gets inserted.

### Environment

* Python version: 2.7 on MacOS, but happens also on Python 3
* Click version: master as of today (7.1.1+)

andy-maier:
PR #1503 adds unit testcases for click.wrap_text() that detail the current behavior, including the case reported in this issue. That PR can be merged independent of this issue. It can be used as a baseline, and if and when this issue is addressed, the corresponding testcase can be adjusted.

davidism:
Note that something similar happens with the Python built-in `textwrap.wrap`:

```pycon
>>> import textwrap
>>> textwrap.wrap("line A1\nline A2\n", width=12)                                                                                                                                                             
['line A1 line', 'A2']
```

I'm not particularly inclined to mess with text wrapping more. Is there a specific reason you need this change?

andy-maier:
On use cases:
1. We do not invoke `click.wrap_text()` directly in our projects.
2. `click.wrap_text()` is documented in the Click API, and the behaviour reported in this issue is unexpected and not described in the API.
3. The `click.HelpFormatter.write_usage()` method invokes `click.wrap_text()` without specifying the `preserve_paragraphs` argument, so it defaults to `False` and the behavior reported in this issue is visible to users of programs using Click in their usage help. Is it really intentional to use False in this case? Maybe the issue can be addressed by changing this to True?

davidism:
It appears usage is intended to be one line, I'm unclear how this is affecting that. Please update the first post with an minimal example that reproduces the issues this is causing for you.

andy-maier:
I experimented with it, and it turns out the usage line is properly wrapped when it does not fit into the terminal width.

So I do not have a case where it causes an issue for us, given that we do not use `click.wrap_text()` directly.

PR #1503, opened 2020-03-14
Added unit tests for click.wrap_text()

andy-maier:
Details:

* Added unit tests for click.wrap_text() that detail the current behavior,
  for both preserve_paragraphs=True/False.

* Two of the tests reproduce the behavior reported in issue #1502, marked
  with TODO and the issue number.

### DIFF
--- a/tests/test_formatting.py
+++ b/tests/test_formatting.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import pytest
 import click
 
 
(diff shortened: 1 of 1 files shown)


davidism:
This mostly tests the behavior of `textwrap.TextWrapper`, not the behavior added by Click for `wrap_text`. I think the only thing that needs to be tested is `preserve_paragraphs` and the no-wrap marker.

davidism:
Thanks for working on this. However, it's very hard for me to read these, and the things being tested don't seem to warrant this level of detail. Many of the tests are related to `textwrap`, not to Click's paragraph handling. I'm going to close this, if there's a specific behavior that needs to be tested, let's work on one specific test.

PR #1504, opened 2020-03-15
Include --help option in completion

tiangolo:
Include `--help` option in completion.

When executing completion (in Bash, Zsh, etc) the `--help` option is currently omitted, even though it's available.

This allows the `--help` option to show up in completion if it was enabled.

This is the same way it's done in `click-completion`: https://github.com/click-contrib/click-completion/blob/master/click_completion/core.py#L92

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,10 @@
 .. currentmodule:: click
 
+Version 7.2
+-----------
+
+-   Include ``--help`` option in completion. :pr:`1504`
+
 Version 7.1.1
 -------------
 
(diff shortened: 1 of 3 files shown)


tiangolo:
I'm not sure where should I add the change in `CHANGES.rst`, above:

```
Version 7.1.1
-------------
```

without a version title?

Or should I create a dummy unreleased version title that you change later?

ThiefMaster:
could you please omit the emoji in the commit messages? while they look pretty on GitHub, they are very noisy when reading the shortlog using the git CLI.

davidism:
For the changelog, just start a new 7.2 section.

tiangolo:
> could you please omit the emoji in the commit messages? while they look pretty on GitHub, they are very noisy when reading the shortlog using the git CLI.

@ThiefMaster sure thing. Just removed the hook. Changing the previous would mean force-pushing, so I won't do it, but I assume you squash and merge when/if you accept it so they won't end up in the history.

> For the changelog, just start a new 7.2 section.

@davidism perfect, done.

tiangolo:
Thanks! :rocket: :tada:

PR #1508, opened 2020-03-21
Implement flag options

### LINKED ISSUE #549: Support click.option with default values / optional arguments

### LINKED ISSUE #764: Flags with optional = argument

KangOl:
Allow to have flags with optional value.

Closes #549
Closes #548
Closes #764

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -339,6 +339,32 @@ And on the command line:
     invoke(info, args=['--lower'])
     invoke(info)
 
+Flag Options
+------------
+
+Alternatively, the flag options can be user provided by setting
+the ``flag_option=True`` parameter. The default value can be set via the
+``flag_value`` parameter.
+
+.. click:example::
+
+    @click.command()
+    @click.option('-S', '--gpg-sign',
+                  flag_option=True, flag_value="default GPG key",
+                  help="Sign the commit with GPG.", metavar='keyid')
+    def commit(gpg_sign):
+        if gpg_sign:
+            click.echo("sign with " + gpg_sign)
+        else:
+            click.echo("do not sign commit")
+
+.. click:run::
+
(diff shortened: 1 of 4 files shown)


libbkmz:
Any posibility that this PR will be accepted? 
I suppose this should fix #549

davidism:
Thanks for working on this. We're working on a hopefully more general solution to options that can omit their value in #1618.

PR #1517, opened 2020-03-30
With prompt, if no input is specified, the return value is not converted to the requested type

stgraham2000:
If the passed in default is a string but the type parameter is not a string then click should do its best to convert the default to the requested type.

This becomes an issue when using mypy with click.  MyPy will complain if you pass anything but a string for the default.  One option is to do a type ignore but that starts to pollute the code quickly if you have a lot of prompts that expect integer values.  Fixing this also will prevent some accidental bugs where you use different types with your `type` option and your `default`.  Click should do its best to honor the requested type.


davidism:
Looks like there was already a special check for the `Path` type, but it doesn't seem like there was any reason to restrict it to that type, so I simplified it.

PR #1518, opened 2020-03-30
Minor typo fix in bash completion template

bostrt:
Currently, the bash completion generated looks like:
```
_mycomm_completionetup;
```
^^ Missing the "s" in setup due to the Python template using the one "s" present.

### DIFF
--- a/src/click/_bashcomplete.py
+++ b/src/click/_bashcomplete.py
@@ -26,7 +26,7 @@ COMPLETION_SCRIPT_BASH = """
     return 0
 }
 
-%(complete_func)setup() {
+%(complete_func)s_setup() {
     local COMPLETION_OPTIONS=""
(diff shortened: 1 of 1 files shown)


src/click/_bashcomplete.py
```diff
@@ -26,7 +26,7 @@
     return 0
 }
 
-%(complete_func)setup() {
+%(complete_func)ssetup() {
```
davidism: _setup would match the zsh script

bostrt: I'm not sure I follow, are you looking for me to change this to `%(complete_func)s_setup`?

davidism: Yes

bostrt:
@davidism completion func now matches zsh per our conversation.

PR #1526, opened 2020-04-16
Improve instructions for fish completion

shields-fn:


davidism:
Thanks for looking at this. We ended up completely rewriting the completion system, including the docs, so I don't think this is needed anymore. https://click.palletsprojects.com/en/master/shell-completion/#enabling-completion

PR #1532, opened 2020-04-20
Drop Python 2

davidism:
Unlike the other projects, `_compat` is still here for Windows / str / bytes handling. Some of that can probably be cleaned up, but that's left for later.


davidism:
Looks like Azure is having issues. I'll try again in the morning.

PR #1534, opened 2020-04-20
Take out chain from core.py and _bashcomplete.py

jcrotts:
https://github.com/pallets/click/issues/1323


davidism:
Based on some discussions after this was submitted, we concluded that removing chain is probably not necessary to address the issues with it, so going to close this.

PR #1535, opened 2020-04-20
Make ParameterSource a proper Python 3 Enum subclass.

### LINKED ISSUE #1530: Make ParameterSource an Enum
`core.ParameterSource` was implemented as fake enum for compatibility with Python 2. Replace this with a simple `enum.Enum` now that we support Python >= 3.6. Can remove the `ParameterSource.validate` method too, it was just for ensuring the enum values were used.

alanbato:
This implements the change proposed in  #1530.

I tried to make as few changes as possible, that's why I went and implemented a custom `__eq__()` to support the current use case inside the library (mostly inside tests)

Any comments and feedback are welcome :)

fixes #1530

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1,3 +1,4 @@
+import enum
 import errno
 import inspect
 import os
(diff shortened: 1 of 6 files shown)


tests/test_context.py
```diff
@@ -323,5 +323,5 @@ def cli(ctx, option):
 
 
 def test_validate_parameter_source():
-    with pytest.raises(ValueError):
-        ParameterSource.validate("NOT_A_VALID_PARAMETER_SOURCE")
+    with pytest.raises(KeyError):
+        ParameterSource["NOT_A_VALID_PARAMETER_SOURCE"]
```
ThiefMaster, line 327: Probably this test can go away. Testing basic stdlib functionality doesn't make much sense IMO

alanbato: I agree, I wasn't sure if I should remove the test but I'm happy to do so after your suggestion

src/click/core.py
```diff
@@ -126,36 +127,27 @@ def sort_key(item):
...
     """This is an enum that indicates the source of a command line parameter.
 
     The enum has one of the following values: COMMANDLINE,
     ENVIRONMENT, DEFAULT, DEFAULT_MAP.  The DEFAULT indicates that the
-    default value in the decorator was used.  This class should be
-    converted to an enum when Python 2 support is dropped.
+    default value in the decorator was used.
+
+    .. versionchanged:: 8
+        This class is now a proper Python3 Enum
```
jab, line 138:
```suggestion
    .. versionchanged:: 8.0
        Made this an :class:`~enum.Enum`
        and dropped the now-superfluous ``validate`` method. 
```

jab: (Important to note when public API is removed)

jab:
> I tried to make as few changes as possible, that's why I went and implemented a custom __eq__() to support the current use case inside the library

A reasonable first cut, for sure. But @davidism, not sure if that's what you had in mind. It seems to me that changing the usage sites to do direct enum comparison (rather than having to keep the custom `__eq__` implementation here) would be more in the spirit of this change. WDYT?

PR #1537, opened 2020-04-21
Additional example for Quickstart for nested commands across files

grbd:
Hi,
I've added an example to the Quickstart docs
This lists the use of the add_command() function which I noticed was mentioned on the github readme but not on the main docs page https://click.palletsprojects.com/en/7.x/

This is just an example for splitting commands / sub commands across different files.
Which I think would be useful

https://stackoverflow.com/questions/34643620/how-can-i-split-my-click-commands-each-with-a-set-of-sub-commands-into-multipl/61353240#61353240

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -226,90 +226,23 @@ other invocations::
     if __name__ == '__main__':
         cli()
 
-Splitting Nested Commands Between Files
---------------------------------
+Registering commands with add_command
+-------------------------------------
 
-If we want to have nested commands which are split between files
-then we can use the function add_command()
-
-To give an example.
-Lets create a file called root.py to act as a main script
+Commands can be defined with @click.command() and registered with the group later
+by using group.add_command(). This could be used to split commands into multiple Python modules.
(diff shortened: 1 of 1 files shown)


davidism:
Thanks, however this doesn't seem right for the quickstart, it's a large section that's mostly boilerplate descriptions of modules and imports. I think the most I'd want for this is a few lines at the end of the previous section. Something like:

> Commands can be defined with `@click.command()` and registered with the group later by using `group.add_command()`. This could be used to split commands into multiple Python modules.
>
> ```python
> @click.group()
> def group():
>     pass
>  
> @click.command()
> def greet():
>    click.echo("Hello, World!")
>
> group.add_command(greet)
> ```

grbd:
Okay I've shortened the example for the Quickstart
let me know if any other changes are needed.

davidism:
Unfortunately, since you made this PR from a repo owned by a different user/org, I can't push to it to rebase or make final adjustments. I've opened a new PR with your amended commit: #1645.

Issue #1124, opened 2018-09-26
allow for prompting in invoked command


ltalirz:
When [invoking](https://click.palletsprojects.com/en/7.x/api/#click.Context.invoke) one click command from another command, options that are not specified are replaced by default values (or `None`), even if the option has `prompt=True`.

Is there currently a way to have an invoked command prompt for its options?
If not, where would I need to inject code in order to enable this?

Here a slightly modified invoke example [from the documentation](https://click.palletsprojects.com/en/5.x/advanced/#invoking-other-commands)

```python
#!/usr/bin/env python
# file: test.py
import click
cli = click.Group()

@cli.command()
@click.option('--other', prompt=True)
@click.option('--count', default=1)
def test(other, count):
    click.echo('Count: %d' % count)
    click.echo('Other: {}'.format(other))

@cli.command()
@click.option('--count', default=1)
@click.pass_context
def dist(ctx, count):
    ctx.invoke(test, count=42)

if __name__ == '__main__':
    cli()
```

```
 ./test.py  test  # prompts
Other: 
```

```
 ./test.py  dist  # does not prompt
Count: 42
Other: None
```

davidism:
This one took a while to pin down, but I've concluded that this is the expected behavior. Thanks to @amy-lei for working through this with me. The documentation is a bit ambiguous when it talks about what `invoke` and `forward` do, so I'll try to explain.

Click is divided into a few phases. The main CLI is executed by calling `command()` (or `command.main()`), the command line is parsed with `command.parse_args()`, then processed into a context with `command.make_context()`. That is when things like environment variables and prompts are handled. Then the command callback is invoked, passing the processed values to it as keyword arguments. At that point, command line processing is done, you're in user space.

`invoke` and `forward` take *Python* arguments, already processed into their correct names and types. With `invoke`, you explicitly pass the arguments you want to pass to the command. `forward` is a shortcut that passes the processed values for the current command to `invoke` as default values, but it still passes the *processed* values. Therefore, if you want to prompt for values, you need to do that in the command calling `invoke` or `forward`, either with options added to that command or further prompting and processing done in the decorated function. The command that's actually called from the command line is what's processed, anything else happens after that's done.

PR #1540, opened 2020-04-25
Call parameter callbacks when invoking command with `Context.invoke`

### LINKED ISSUE #950: Context.invoke doesn't run callback on the other command
When invoking a command using `ctx.invoke(other_cmd, ...)`, I expected that the parameters passed would be validated per `other_cmd`'s `callback` methods as declared on the arguments but that seems to be not the case.

Is this a known thing? Is there a workaround?

sphuber:
Fixes #950 

The `Context.invoke` method was not calling any callbacks that might
have been defined for the parameters of the command that is being
invoked.

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -603,6 +603,10 @@ class Context(object):
             for param in other_cmd.params:
                 if param.name not in kwargs and param.expose_value:
                     kwargs[param.name] = param.get_default(ctx)
+                    if param.callback is not None:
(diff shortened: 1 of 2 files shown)


davidism:
Thanks for working on this. However, the issue it's fixing is closed with the following reason:

This is intentional behavior. `invoke` occurs in user code, after the command line processing is over, so it should be passed valid Python values that do not need to be processed. See #1124 for a detailed explanation.

PR #1549, opened 2020-05-09
show_default looks at default_map

### LINKED ISSUE #1548: show_default and default_map don't play nice together
```python
#!/usr/bin/env python
import click

@click.command(context_settings={
    "show_default": True, "default_map": {"a": "b"}
})
@click.option("-a", default="c")
def func(a):
    print(a)

if __name__ == "__main__":
    func()
```
```console
$ # This is expected
$ ./cli.py
b
$ # This is not (default should be b)
$ ./cli.py --help
Usage: cli.py [OPTIONS]

Options:
  -a TEXT  [default: c]
  --help   Show this message and exit.  [default: False]
```

### Expected Behavior

I expect the default shown to be identical to the one being used.

### Actual Behavior

The default shown was the value provided as `default`, without taking the `default_map` into account

### Environment

* Python version: 3.7
* Click version: ``click==7.1.2``

ewjoachim:
Closes #1548 

It's my first PR on this project, so I may have missed things.

- Not sure if a ``versionchanged`` marker is needed to fix bugs.
- If a callable is passed as default, it will not be evaluated, but if it's passed as default map, it will. Is it a problem ?
- I've included a single-character typo fix in a related entry of the changelog. Is it ok ?

Also, I know how feedback is always nice so:
- Contributing doc was 👌 
- I have already worked with pytest-click, so nothing new here
- A bit confusing whether the fix was supposed to go to "commands", "options" or "defaults", so I made a choice :)
- Overall, very easy to contribute. Thanks and congrats.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -134,6 +134,8 @@ Unreleased
 -   Handle empty env vars as though the option were not passed. This
     extends the change introduced in 7.1 to be consistent in more cases.
     :issue:`1285`
+-   ``show_default`` considers ``default_map`` to display the
+    proper default value. :issue:`1548`
 
 
 Version 7.1.2
@@ -182,7 +184,7 @@ Released 2020-03-09
     :issue:`1277`, :pr:`1318`
 -   Add ``no_args_is_help`` option to ``click.Command``, defaults to
     False :pr:`1167`
--   Add ``show_defaults`` parameter to ``Context`` to enable showing
+-   Add ``show_default`` parameter to ``Context`` to enable showing
     defaults globally. :issue:`1018`
 -   Handle ``env MYPATH=''`` as though the option were not passed.
     :issue:`1196`
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -134,8 +134,9 @@ Unreleased
 -   Handle empty env vars as though the option were not passed. This
     extends the change introduced in 7.1 to be consistent in more cases.
     :issue:`1285`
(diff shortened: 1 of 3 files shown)


ewjoachim:
I there something I can do to ease reviewing ?

ewjoachim:
Hey there :) Anything I can do ?

davidism:
Thanks for submitting this. This looks like the right solution, but I'm currently focused on some other things so I haven't had a chance to review it yet. No need to do anything for now, this will get reviewed and merged when a maintainer has time.

davidism:
Was working on cleaning this up, and realized that it probably points to a design issue that `get_default` doesn't already call `ctx.lookup_default`. Based on the docstring it seems like that was the intended behavior.

But when I addressed that, it somehow made some unrelated tests fail. Opened #1649 to address that first.

ewjoachim:
Ok, thanks for the update. Let me know if there's anything I can do to help :)

davidism:
Finally worked through #1649, so this is ready to go in. While working on the other processing issues, I realized that this could be generalized to check `Context.lookup_default()` in `Parameter.get_default()`, so that it is applied consistently in help text, prompts, and `Context.invoke()`. In order to support that, I added a `call` parameter to allow turning off calling for help text.

ewjoachim:
Yay thanks :) I'm not sure there's much of my original work in the final PR but thank you very much for taking care of this :heart:

PR #1558, opened 2020-05-23
make_pass_decorator option to pass merged context

nixjdm:
Currently `make_pass_decorator` will pass a custom context object. If you need this and the normal context, you'd typically use both `@click.pass_context` and the decorator you make with `make_pass_decorator`. There are tests that do this. This can be cumbersome if you often want features in the normal context and custom features from your own, especially in a large cli with many commands.

This adds an option to have `make_pass_decorator` pass the normal context merged with the provided context object, preferring the custom in conflicts since that's the one the user wrote. This allows you to pass a single context object into the functions that has the built-in features and custom features, rather than two. This can simplify mental overhead in tracking which to use for what, and provides a single place to look for what you need.

This is opt-in - defaulted to preserve current behavior.

### DIFF
--- a/src/click/decorators.py
+++ b/src/click/decorators.py
@@ -33,7 +33,19 @@ def pass_obj(f):
     return update_wrapper(new_func, f)
 
 
-def make_pass_decorator(object_type, ensure=False):
+def merge_contexts(ctx, obj):
+    """Return the given context with the attributes of another object
+    merged into it. This is useful to attach a separate application
+    context to a Click context. Dunders and double prefixed attrs
+    are excluded from the merge.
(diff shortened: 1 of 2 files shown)


davidism:
After reviewing the decorators, I don't think this makes sense as a general behavior to be included in Click. The problem is that the provided `merge` function isn't general enough to cover common types of context objects. The most obvious is that the context object is often a `dict`. Or an object might have complex data, where overwriting the entire attribute isn't correct. Or than object might have computed properties that aren't assignable. Overall it's probably better to leave it up to a project to decide how their context objects should work.

PR #1579, opened 2020-06-09
Use asteriks for password input.

### LINKED ISSUE #1424: Password / hidden prompts should have option to display asterisks
For password / hidden prompts it would be great if instead of displaying nothing one could display one (or more) asterisk per character entered.

What do you think about this feature? I'd be happy to implement it.

teymour-aldridge:
This pull request fixes #1424. 
Co-authored-by: Mwiza Simbeye <mwiza@agripredict.com>

### DIFF
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -46,9 +46,9 @@ _ansi_reset_all = "\033[0m"
 
 
 def hidden_prompt_func(prompt):
-    import getpass
+    import stdiomask
 
-    return getpass.getpass(prompt)
+    return stdiomask.getpass(prompt=prompt, mask="X")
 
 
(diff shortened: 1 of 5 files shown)


src/click/termui.py
```diff
@@ -46,9 +46,9 @@
 
 
 def hidden_prompt_func(prompt):
-    import getpass
+    import stdiomask
```
davidism, line 49: Unfortunately, we can't add external dependencies.

Note that if you did add a dependency, you'd want to add it to `install_requires` as well.

davidism:
Closing, we discussed this in chat and I added an explanation to the linked issue.

Issue #1488, opened 2020-03-06
Don't use pkg_resources for --version


davidism:
The `--version` option uses pkg_resources to get the version information from the distribution metadata. Use [importlib_metadata](https://pypi.org/project/importlib-metadata/) instead, but don't make it a dependency. When creating the decorator, issue a warning if a `version` is not given and importlib_metadata is not installed.

PR #1531, opened 2020-04-20
Fix version detection when invoked with `python -m <package>`

cjolowicz:
click.version_option attempts to determine the version using `pkg_resources` when no version is passed. This fails when the program was invoked with `python -m <package>`. 

<details>
<summary>Traceback</summary>

```python
  Traceback (most recent call last):
    File ".../3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
      "__main__", mod_spec)
    File ".../3.7/lib/python3.7/runpy.py", line 85, in _run_code
      exec(code, run_globals)
    File ".../foobar/__main__.py", line 11, in <module>
      main()
    File ".../venv/lib/python3.7/site-packages/click/core.py", line 891, in __call__
      return self.main(*args, **kwargs)
    File ".../venv/lib/python3.7/site-packages/click/core.py", line 843, in main
      with self.make_context(prog_name, args, **extra) as ctx:
    File ".../venv/lib/python3.7/site-packages/click/core.py", line 762, in make_context
      self.parse_args(ctx, args)
    File ".../venv/lib/python3.7/site-packages/click/core.py", line 1115, in parse_args
      value, args = param.handle_parse_result(ctx, opts, args)
    File ".../venv/lib/python3.7/site-packages/click/core.py", line 1705, in handle_parse_result
      value = invoke_param_callback(self.callback, ctx, self, value)
    File ".../venv/lib/python3.7/site-packages/click/core.py", line 123, in invoke_param_callback
      return callback(ctx, param, value)
    File ".../venv/lib/python3.7/site-packages/click/decorators.py", line 295, in callback
      raise RuntimeError("Could not determine version")
  RuntimeError: Could not determine version
```
</details>

The reason version detection fails is that `__name__` in the calling module is `"__main__"`, while `entry_point.module_name` is `"<package>.__main__"`:

https://github.com/pallets/click/blob/19fdc8509a1b946c088512e0e5e388a8d012f0ce/src/click/decorators.py#L265-L269

https://github.com/pallets/click/blob/19fdc8509a1b946c088512e0e5e388a8d012f0ce/src/click/decorators.py#L288-L293

This PR fixes the issue by prefixing the module name with `__package__` from the calling module, if it was `"__main__"` to begin with.

---

**Minimal reproducible example**:

```
.
├── foobar
│   ├── __init__.py
│   └── __main__.py
├── setup.cfg
└── setup.py

1 directory, 4 files
```

```python
# setup.py
import setuptools
setuptools.setup()
```

[...]

PR #1582, opened 2020-06-10
changed version_option default package from pkg_resources to importli…

0x1za:
I believe this should close issue #1488

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -19,6 +19,7 @@ filelock==3.0.12          # via tox, virtualenv
 identify==1.4.16          # via pre-commit
 idna==2.9                 # via requests
 imagesize==1.2.0          # via sphinx
+importlib-metadata==1.6.1  # via -r requirements/tests.in
 jinja2==2.11.2            # via sphinx
 markupsafe==1.1.1         # via jinja2
 more-itertools==8.3.0     # via pytest
@@ -51,6 +52,7 @@ tox==3.15.2               # via -r requirements/dev.in
 urllib3==1.25.9           # via requests
 virtualenv==20.0.21       # via pre-commit, tox
 wcwidth==0.1.9            # via pytest
+zipp==3.1.0               # via importlib-metadata
 
 # The following packages are considered to be unsafe in a requirements file:
 # pip
--- a/src/click/decorators.py
+++ b/src/click/decorators.py
@@ -1,5 +1,4 @@
 import inspect
-import sys
 from functools import update_wrapper
 
 from .core import Argument
@@ -260,11 +259,6 @@ def version_option(version=None, *param_decls, **attrs):
                     (``'%(prog)s, version %(version)s'``)
     :param others: everything else is forwarded to :func:`option`.
     """
-    if version is None:
-        if hasattr(sys, "_getframe"):
-            module = sys._getframe(1).f_globals.get("__name__")
-        else:
-            module = ""
 
     def decorator(f):
         prog_name = attrs.pop("prog_name", None)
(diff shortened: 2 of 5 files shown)


mwizasimbeye11:
I removed a couple of no longer needed variables and imports.

src/click/decorators.py
```diff
@@ -279,16 +273,11 @@ def callback(ctx, param, value):
             ver = version
             if ver is None:
                 try:
-                    import pkg_resources
+                    import importlib_metadata
```
davidism, line 276: This is built in to Python 3.8: https://docs.python.org/3/library/importlib.metadata.html. Try importing the built in first, then try the backport.

requirements/dev.txt
```diff
@@ -19,6 +19,8 @@ filelock==3.0.12          # via tox, virtualenv
 identify==1.4.16          # via pre-commit
 idna==2.9                 # via requests
 imagesize==1.2.0          # via sphinx
+importlib-metadata==1.6.1  # via importlib-resources, pallets-sphinx-themes, pluggy, pre-commit, pytest, tox, virtualenv
```
davidism, line 22: It looks like you added `importlib-resources` to `dev.in` but forgot to commit that. We don't need importlib-resources, just importlib-metadata, and it should be added to `tests.in`, then the relevant `.txts` should be recompiled with `pip-compile requirements/test.txt` and `pip-compile requirements/dev.txt`.

davidism: Oops, my bad, I meant `pip-compile requirements/tests.in`, not `.txt` :sweat:

src/click/decorators.py
```diff
@@ -279,16 +273,11 @@ def callback(ctx, param, value):
             ver = version
             if ver is None:
                 try:
-                    import pkg_resources
+                    import importlib.metadata
```
davidism, line 276: Still need to fall back to `importlib_metadata` if this fails, since we support 3.6+.

```python
try:
    from importlib.metadata import version as get_version
except ImportError:
    try:
        from importlib_metadata import version as get_version
    except ImportError:
        get_version = None

if get_version is not None:
    ver = get_version(prog)
```

requirements/dev.txt
mwizasimbeye11: On it 😄

src/click/decorators.py
mwizasimbeye11: Yes definitely, forgot to backport.

0x1za (commented):
LGTM

davidism:
Squashed the original commits, added a changelog, added a change to `requirements/tests.in` that didn't get committed before. Then, while reviewing this, I realized that the entire function could use a refactor and new docs, so I did that in a separate commit.

* `option()` returns a decorator, so there's no point in wrapping it in another decorator internally.
* Version detection relies on being able to detect the module defining the CLI by examining the stack frames. Added the `package_name` parameter to skip or override detection if it fails. Also make `package` available in the message format, since it can be different than `prog`.
* Use `inspect.currentframe()` for stack inspection instead of a private method. Gets the top-level package name instead of the full dotted `__name__`.
* Use Python 3's `nonlocal` keyword for modifying args from the callback.
* Only try to detect the version if a package name was given or detected.
* Show a specific error if `importlib_metadata` needs to be installed.
* Show an error if `metadata.version()` can't find the package. This can happen if the package name doesn't match the PyPI name.
* Show the detected package name in the error message when the version wasn't detected, to make it clearer what Click was trying.

Also noticed that the change you made changed how the version was being detected, from looking at entry points to looking at the package name directly. It looks like the values returned by `importlib.metadata.entry_points()` aren't linked to the distributions that own them, so there's not really a nice way to use the previous implementation. Writing some notes below so the difference is clear later.

---

The previous implementation tried to find the program name among the `console_scripts` entry points, then if the module that entry pointed at matched the stack frame detected module, it took the version of the distribution that created the entry. The new implementation skips looking at entry points and assumes the package name is the same as the installed name. While that's not strictly true, it's definitely a good and common practice. The previous implementation wasn't strictly correct either, as there was no guarantee that the module that created the version option was also the one pointed to by the entry point.

davidism:
Incorporated #1531 for better version detection when using `python -m package`.

PR #1583, opened 2020-06-11
Fix #1460: hide prompt input if hide_input is enabled

### LINKED ISSUE #1460: click.prompt(type=int, hide_input=True) outputs rejected input
Despite asking `click.prompt()` to hide input it will still outputs the input if it rejects it.  I get the same behavior for `7.0` from PyPI, `7.x` from Git, and `master` from Git.

```bash
altendky@p1:~$ venv/bin/python -c 'import click; click.prompt("prompt", type=int, hide_input=True)'
prompt: 
Error: wait...  i get to see this? is not a valid integer
prompt:
```

If you specify a type (let's say you are inputting a numeric pin) then the same can happen with `click.password_option()`.

```python3
import click


@click.command()
@click.password_option(type=int)
def cli(password):
    pass


cli()
```

```bash
altendky@p1:~$ venv/bin/python x.py
Password: 
Error: lkjsaf is not a valid integer
Password:
```

The workaround for `click.prompt()` anyways I guess is to not specify a type and to implement the rejection manually.

If there is agreement that this is an issue that should be fixed then I can try to put together a PR for it.

csr:
This PR aims to fix #1460 (prompt input would be displayed in the error message even if `hide_input` was enabled).

These scripts now work just fine and display `Error: the value you entered was invalid` if the input isn't what Click expects. Script 1:
```
cesaredecal$ python  -c 'import click; click.prompt("prompt", type=int, hide_input=True)'
prompt: 
Error: the value you entered was invalid
```

Script 2 (that asks for a numeric pin):
```
import click


@click.command()
@click.password_option(type=int)
def cli(password):
    pass


cli()
```

```
cesaredecal$ python hello.py
Password: 
Error: the value you entered was invalid
```

All tests with `tox` pass. As far as the implementation goes, it would have probably been a little cleaner to handle the checking inside the `UsageError` class (or similar), but we don't have visibility on `hide_input` there. Feedback very welcome as I'm pretty new with Python.

### DIFF
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -153,7 +153,10 @@ def prompt(
         try:
             result = value_proc(value)
         except UsageError as e:
-            echo(f"Error: {e.message}", err=err)  # noqa: B306
+            if hide_input:
+                echo("Error: the value you entered was invalid", err=err)
+            else:
+                echo(f"Error: {e.message}", err=err)  # noqa: B306
             continue
         if not confirmation_prompt:
             return result
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,11 +12,6 @@ Unreleased
     parameter. :issue:`1264`, :pr:`1329`
 -   Add an optional parameter to ``ProgressBar.update`` to set the
     ``current_item``. :issue:`1226`, :pr:`1332`
-
-
-Version 7.2
------------
-
 -   Include ``--help`` option in completion. :pr:`1504`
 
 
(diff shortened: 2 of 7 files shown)


src/click/termui.py
```diff
@@ -153,7 +153,10 @@ def prompt_func(text):
         try:
             result = value_proc(value)
         except UsageError as e:
-            echo(f"Error: {e.message}", err=err)  # noqa: B306
+            if hide_input:
```
csr, line 156: I could have probably used `.format` to use conditional checking on `hide_input` to `echo` on a single line, but I thought that wouldn't have been very readble for its length. Feedback welcome.

PR #1586, opened 2020-06-15
Added IntRange min and max preview to --help output

### LINKED ISSUE #1525: Show range values for IntRange option --help
### Expected Behavior

Similar to the `show_defaults=True` parameter, options that have type IntRange should print the expected range values in the help text

```python
@click.option('-c', '--count', required=True, type=click.IntRange(1, 100, show_range=True))
def set_count(count)
   pass
```
```
>>> mycli set-count --help
Usage: mycli set-count [OPTIONS]

  Set the target count

Options:
  -c, --count INTEGER RANGE  [1-100 inclusive, required]
```

### Actual Behavior

```
   -c, --count INTEGER RANGE  [required]
```

### Environment

* Python version: 3.7.3
* Click version: 7.0

IamCathal:
Issue https://github.com/pallets/click/issues/1525  requested to have the following output for using the `--help` flag on options of type IntRange:
```
Usage: test.py [OPTIONS]

  Simple program that greets NAME for a total of COUNT times.

Options:
  --count INTEGER RANGE  Number of greetings.  [default: 1-32 inclusive]
  --help                 Show this message and exit.
```

This can be used like as a regular click user would expect it to:
````python
import click

@click.command()
@click.option("--count",type=click.IntRange(1, 32), show_default=True, default=1, help="Number of greetings.")
def hello(count):
    """Simple program that greets for a total of COUNT times."""
    for _ in range(count):
        click.echo("Hello!")

if __name__ == '__main__':
    hello()
````

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1936,6 +1936,10 @@ class Option(Parameter):
                 default_string = self.default
             extra.append(f"default: {default_string}")
 
+        if isinstance(self.type, IntRange):
+            if self.type.min is not None and self.type.max is not None:
+                extra.append(f"{self.type.min}-{self.type.max} inclusive")
+
         if self.required:
             extra.append("required")
         if extra:
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -216,6 +216,16 @@ def test_dynamic_default_help_text(runner):
     assert "(current user)" in result.output
 
 
+def test_intrange_default_help_text(runner):
+    @click.command()
+    @click.option("--count", type=click.IntRange(1, 32), show_default=True, default=1)
+    def cmd(arg):
+        click.echo(arg)
+
+    result = runner.invoke(cmd, ["--help"])
+    assert "1-32 inclusive" in result.output
+
+
 def test_toupper_envvar_prefix(runner):
     @click.command()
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: tests/test_options.py [hunk 1]
```diff
@@ -216,6 +216,16 @@ def cmd(username):
     assert "(current user)" in result.output
 
 
+def test_intrange_default_help_text(runner):
```
IamCathal, line 219: Simple test case added to check for correct output from IntRange type parameters

davidism: Looks good :+1:

src/click/core.py
```diff
@@ -1935,6 +1935,9 @@ def _write_opts(opts):
             else:
                 default_string = self.default
             extra.append(f"default: {default_string}")
+            if isinstance(self.type, IntRange):
```
davidism, line 1938: Still needs to be un-indented, it became part of the previous block when you removed the `if`.

IamCathal: Took a bit more time to make the change this time, hope there aren't any small bugs. In general for a simple feature is on test where it tries to send good data (like not a test case trying to crash it) good enough? Thanks

davidism (approved):

PR #1594, opened 2020-06-18
raise ValueError for duplicate flags

### LINKED ISSUE #1465: Duplicate Boolean flag options "--foo/--foo" are not detected (defaults to False)
I ran into an issue where I had the code:

    @click.option( "--print-verbose-info/--print-verbose-info", help="...", default=False )

...and running `cli.py --print-verbose-info` was apparently defaulting over to the `False` value.  Totally my mistake (it should have been `--no-print-verbose-info`), but it'd be nice to get a warning/error for boolean values which match exactly (ie: cannot be set to true).

It was kindof tricky to diagnose b/c I assumed that the `--print-verbose-info` that I was passing on the CLI was triggering the `True` path, not the `False` path, and usually Click option parsing is pretty bulletproof so it took a while to find that copy/paste problem.

This would be a `good-first-issue`, I'm submitting it now more as a reminder to maybe follow-up on it and figure out the appropriate P.R.

0x1za:
Specifying the same flag name twice will raise a `RuntimeError`.

```python
@click.option( "--foo/--foo", default=False)
```

fixes #1465

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,8 @@ Unreleased
     is not shown in the error message. :issue:`1460`
 -   An ``IntRange`` option shows the accepted range in its help text.
     :issue:`1525`
+-   An option defined with duplicate flag names (``"--foo/--foo"``)
+    raises a ``ValueError``. :issue:`1465`
 
 
 Version 7.1.2
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1840,6 +1840,11 @@ class Option(Parameter):
                     second = second.lstrip()
                     if second:
                         secondary_opts.append(second.lstrip())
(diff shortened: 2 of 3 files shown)


mwizasimbeye11:
#### Description
Given a flag option with duplicate flags eg: 

`@click.option( "--foo/--foo", help="some help about foo", default=False )`

Click does not issue a warning/error for duplicate flags.
#### Issue
This pull request is a fix for #1465

#### Testing
Tested in PR, `test/test_options.py::test_duplicate_flags`

davidism:
Squashed, rebased, added changelog, reverted changes to `requirements/tests.txt`, raised `ValueError` instead of `TypeError`.

PR #1604, opened 2020-06-25
nargs=-1 works with envvar 

### LINKED ISSUE #729: Can't mix nargs=-1 with envvar
`@click.argument(name, nargs=2, envvar=NAME)`

Works as expected. If you set `NAME="a b"` and run your command `name` in the called function will be a tuple of `('a', 'b')`. 

`@click.argument(name, nargs=-1, envvar=NAME)`

on the other hand doesn't work as expected. If you call it in the same exact way, `name` will be set to the empty tuple.

rch-liu:
This PR aims to fix two bugs:

1. Initially combining `nargs=-1` and `envvar` would yield an empty tuple where the expected behaviour of `nargs=-1` is to take on an unlimited number of arguments passed in `envvar`. @amy-lei and I added a fix so that now combining `nargs=-1` with `envvar` should output all the expected values from envvar rather than an empty tuple.

2. We also added error catching for nargs > len(envvar) where we would raise: `Got unexpected extra arguments` and when nargs < len(envvar) we would raise: `argument [argname] takes [nargs] values`.

The changelog is also updated with our new changes. Let us know what you think @davidism!

fixes #729

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,8 @@ Unreleased
 -   The ``BOOL`` type accepts the values "on" and "off". :issue:`1629`
 -   A ``Group`` with ``invoke_without_command=True`` will always invoke
     its result callback. :issue:`1178`
+-   ``nargs == -1`` and ``nargs > 1`` is parsed and validated for
+    values from environment variables and defaults. :issue:`729`
 
 
 Version 7.1.2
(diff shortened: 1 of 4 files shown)


amy-lei:
Fixes #729 

We discovered that `nargs` was also not enforced when the values came from `default`, so we addressed/added tests for it as well.

davidism:
Refactored the tests to use `parametrize`. Simplified the nargs validation so it uses the same message for any combination, rather than reusing the argument message, which didn't really fit.

PR #1609, opened 2020-06-29
detect program name with python -m

### LINKED ISSUE #1603: detection of program name when run with `python -m`
The program name (referred to as `prog_name` or `info_name`) is the name Click uses in usage help text. If it's not specified, Click tries to detect it by looking at `sys.argv[0]`.

https://github.com/pallets/click/blob/dfb842d95a7c146d8fe8140c8100065410d95c4d/src/click/core.py#L797-L800

This works when the CLI is run as a script, either directly, or indirectly as an entry point, or as a module when the program is a single file, but fails when the program is a package and run with `python -m`. In that case, Python converts the `-m name` to the Python file within the package that would be executed. (The use of `__file__` there is also incorrect, as that will always point to `click/core.py`. There should never be a case where `sys.argv` is empty. `make_str` is probably not needed either.)

This leads to the following names:

* `python example.py`, executing a script directly - `example.py`
* `python -m example`, the single file module `example.py` - `example.py`
* `python -m example`, a package like `example/__main__.py` - `__main__.py`
* `python -m example.cli`, a submodule like `example/cli.py` - `cli.py`
* `example`, an entry point, with any of the layouts above - `example`

What's usually expected is that executing a script will print the Python file name (`example.py`), executing an entry point will print the entry point name (`example`), and executing a module or package will print the Python command (`python -m example`, `python -m example.cli`)

This leads to projects that expect to be run as either an entry point or `-m` writing their own wrapper around the CLI to control what name gets used. For example, Flask uses a wrapper that the entry point will call with no args, while `__main__.py` or `__name__ == "__main__"` will pass `as_module=True`.

```python
def main(as_module=False):
    cli.main(prog_name="python -m flask" if as_module else None)
```

Unfortunately, detecting whether `python -m` was used is really, really complicated. Luckily, I did the work already in Werkzeug's reloader: https://github.com/pallets/werkzeug/blob/102bcda52176db448d383a9ef1ccf7e406a379eb/src/werkzeug/_reloader.py#L59. It won't be quite the same in Click because we're not concerned with the exact path to the `python` command or script file, or with arguments, but that should be a good start. Write a function `detect_program_name()` to detect this and call it in place of the existing code.

gmelodie:
Fixes #1603

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -33,6 +33,8 @@ Unreleased
     its result callback. :issue:`1178`
 -   ``nargs == -1`` and ``nargs > 1`` is parsed and validated for
     values from environment variables and defaults. :issue:`729`
+-   Detect the program name when executing a module or package with
+    ``python -m name``. :issue:`1603`
 
 
 Version 7.1.2
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -25,6 +25,7 @@ from .termui import style
 from .types import BOOL
 from .types import convert_type
 from .types import IntRange
+from .utils import _detect_program_name
 from .utils import echo
 from .utils import make_default_short_help
 from .utils import make_str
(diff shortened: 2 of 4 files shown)


davidism:
Rewrote the docstring. Simplified the function a bit more, we only care about getting a representative value for the help text. (Werkzeug needed the exact executable and file paths.) Files and entry points return only the file name. Modules and packages return `python -m name`. Made the function private (`_detect_program_name`) for now since I don't really see it being useful as a public API function.

PR #1610, opened 2020-06-29
Fix bug with envvars and invoked commands

### LINKED ISSUE #1079: Values from envvars are ignored when invoking and forwarding
`foo.py`:
```
import click

@click.group()
def cli():
    pass

@cli.command()
@click.option('--value')
def a(value):
    print(value)

@cli.command()
@click.pass_context
def invoke_a(ctx):
    ctx.invoke(a)
    ctx.parent.invoke(a)

@cli.command()
@click.pass_context
def forward_a(ctx):
    ctx.forward(a)
    ctx.parent.forward(a)

if __name__ == '__main__':
    cli(auto_envvar_prefix='FOO')
```
```
$ FOO_A_VALUE=1 python3 foo.py a
1
$ FOO_A_VALUE=1 python3 foo.py invoke_a
None
None
$ FOO_A_VALUE=1 python3 foo.py forward_a
None
None
```
When invoking and forwarding other commands, I think environment variables (and default maps) should be processed as if the command was directly called, which for both is currently not the case (see also #988 ).

amy-lei:
Commands invoked via `Context.invoke` or `Context.forward` only checked defaults if no values were passed in. Fixed this to also check envvars with the prefix matching the path taken through the contexts. 

fixes #1079

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,8 @@ Unreleased
     raises a ``ValueError``. :issue:`1465`
 -   ``echo()`` will not fail when using pytest's ``capsys`` fixture on
     Windows. :issue:`1590`
+-   Include check for ``envvar`` values in commands invoked with
+    ``Context.invoke`` and ``Context.forward``. :issue:`1079`
 
 
 Version 7.1.2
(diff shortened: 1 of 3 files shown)


davidism:
See #1124 for a detailed explanation for why we concluded this behavior was intentional and shouldn't be changed. Thanks for working on this and helping figure it all out!

PR #1617, opened 2020-07-01
Show flag name instead of True/False with Boolean flags

### LINKED ISSUE #1538: Improvement for show_default showing in case of a flag the name instead of True/False
First to say: It's not a bug ... it's (in my opinion) a reasonable improvement.

### Expected Behavior

That's how the boolean flag is configured:

```python
@click.option('--cache/--no-cache', default=False, show_default=True,
              help="Enable/Diable the cache.")
```
Using the --help you see following (just the line of interest):

```
--cache / --no-cache        Enable/Diable the cache.  [default: False]
```

but preferable woud be:

```
--cache / --no-cache        Enable/Diable the cache.  [default: no-cache]
```

It's clear to me that this works for --cache/--no-cache like flags only.

### Actual Behavior

The output is as documented by the API the default **value** (*not the name*) which is for a boolean True or False.

### Environment

* Python version: 3.8.1
* Click version: 7.0

ovezovs:
Fixes #1538

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1948,6 +1948,12 @@ class Option(Parameter):
                 default_string = ", ".join(str(d) for d in self.default)
             elif inspect.isfunction(self.default):
                 default_string = "(dynamic)"
+            elif self.is_bool_flag and self.secondary_opts:
+                # For boolean flags that have distinct True/False opts,
+                # use the opt without prefix instead of the value.
+                default_string = split_opt(
+                    (self.opts if self.default else self.secondary_opts)[0]
+                )[1]
             else:
                 default_string = self.default
             extra.append(f"default: {default_string}")
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -37,6 +37,8 @@ Unreleased
     ``python -m name``. :issue:`1603`
 -   Include required parent arguments in help synopsis of subcommands.
     :issue:`1475`
+-   Help for boolean flags with ``show_default=True`` shows the flag
+    name instead of ``True`` or ``False``. :issue:`1538`
 
 
 Version 7.1.2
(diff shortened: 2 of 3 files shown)


ovezovs:
@davidism, I have one major question regarding this fix: Should we show the flag names only for boolean flags or any? For example, [test_global_show_default](https://github.com/pallets/click/blob/fb1ca91ab00a2ce12da84addc459e1983aac2fcf/tests/test_formatting.py#L303) expects default on the help flag line to be false. For now, I did a workaround to not display the default flag name when it is help. I am not sure if that's a bad thing.

src/click/core.py
```diff
@@ -1939,6 +1939,10 @@ def _write_opts(opts):
                 default_string = "(dynamic)"
             else:
                 default_string = self.default
+                if self.is_bool_flag and self.name != "help":
+                    default_string = (
+                        split_opt(rv[0])[1] if self.default else split_opt(rv[1])[1]
```
juped, line 1944: Maybe this line could use a short comment

davidism:
Unless the test is particularly complex, I usually just roll it into the code commit. Same for multiple code and cleanup commits into a single commit.

davidism:
Changed the implementation so that it checks that the boolean flag actually has separate flags for `True` and `False`. It selects them from `self.opts` and `self.secondary_opts` instead of `rv`, since `rv` will have `,` joined alternative names together. That would cause the following to output `[default: no-cache, --nc]`.

```python
@click.option("--cache/--no-cache", "--c/--nc", show_default=True)
```

Updated the test to check when `default` is `True` and `False`, and added a second test for when there's only one name.

ovezovs:
> Updated the implementation so that it checks that the boolean flag actually has separate flags for `True` and `False`. It selects them from `self.opts` and `self.secondary_opts` instead of `rv`, since `rv` will have `,` joined alternative names together. That would cause the following:
> 
> ```python
> @click.option("--cache/--no-cache", "--c/--nc", show_default=True)
> ```
> 
> ```
> [default: no-cache, --nc]
> ```
> 
> Now it outputs only `no-cache`.
> 
> Updated the test to check when `default` is `True` and `False`, and added a second test for when there's only one name.

Thank you so much for the feedback!

PR #1618, opened 2020-07-01
option without value can prompt or use default

### LINKED ISSUE #549: Support click.option with default values / optional arguments

### LINKED ISSUE #736: Option that is not required and prompts if specified without an arg
Hi - 

I am working on a ticket for mycli - the gist is that the team wants the '-p' flag to behave like th e mysql client - i.e. prompt if you don't pass an arg, but it is not required so do NOT prompt if the -p is not included.  Today if I use standard click->prompt then scenario 2 gives an error.  Below is the desired behavior:

1)  $>  ./hello
                HELLO WORLD
2)  $>  ./hello -p
                Please enter your password:
3)  $>  ./hello -p s3cr3t
                HELLO WORLD

It appears looking at the code there is no way to do this, but it seems desirable - before working on a PR I wanted to validate that it can't be done and get input on the usefulness of the feature.

Thanks for your guidance
John

### LINKED ISSUE #764: Flags with optional = argument

### LINKED ISSUE #921: Conditionally disable prompts
First of all, thanks a lot for this great package! CLI made easy!

I'm using `prompt` parameter in my options so CLI can be used interactively and well as scripted.
Also applying `default` parameters to each required option so users have to type less.

**Problem:** Defaults do not apply to options which have `prompt` parameter, and I cant find an elegant way to control that behaviour.

So when using CLI in non-interactive manner I have to specify all command options even the ones that have defaults.
Ideally I'm looking for an option to disable prompts globally, this way when using CLI in non-interactive way, I have to only specify options that have no defaults and are required.

But so far all my explorations are unsuccessful.

Any help or hints on how to approach this is appreciated. Thanks!

### LINKED ISSUE #1015: Creating an option with prompt always prompts even if the option isn't added to command call
Hi, I'm having an issue when I add a prompt or password prompt option to my command. It will prompt even if the command doesn't have that option specified when calling the command in terminal.

I used this cookiecutter template to generate the boilerplate code:
https://github.com/nvie/cookiecutter-python-cli

And the default configs generate a working command that i used to test this:

1. I added an password_prompt decorator in the main command at `my_tool/cli.py`:
```python
import click


@click.command()
@click.option('--as-cowboy', '-c', is_flag=True, help='Greet as a cowboy.')
@click.password_option('--password', '-p', help='Input a password')
@click.argument('name', default='world', required=False)
def main(name, as_cowboy):
    """My Tool does one thing, and one thing well."""
    greet = 'Howdy' if as_cowboy else 'Hello'
    click.echo('{0}, {1}.'.format(greet, name))
```
2. Installed in the project virtualenv with pip
```
pip install --editable .
```

3. Calling the my-tool command in virtualenv makes click prompt for the password:
```
$ my-tool
Password:
```

As it is an option I thought that it wouldn't prompt it it isn't passed when calling the command.

There is something that I'm missing?

amy-lei:
Resolves #549 
Resolves #736
Resolves #764 
Resolves #921 
Resolves #1015 

Add `prompt_required` attribute to `Option` so that when `prompt_required=False`, the user will only be prompted for an input if the option is specified without a value. The attribute defaults to `True` to maintain the current behavior of option prompts.

`is_flag=False` now means that providing the value to the option is optional. If the flag is provided without the value, it can be prompted for or use `flag_value`.

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -18,6 +18,7 @@ from .formatting import HelpFormatter
 from .formatting import join_options
 from .globals import pop_context
 from .globals import push_context
+from .parser import _flag_needs_value
 from .parser import OptionParser
 from .parser import split_opt
 from .termui import confirm
@@ -2148,6 +2149,9 @@ class Option(Parameter):
                    option name capitalized.
     :param confirmation_prompt: if set then the value will need to be confirmed
                                 if it was prompted for.
+    :param prompt_required: If set to ``False``, the user will be
+        prompted for input only when the option was specified as a flag
+        without a value.
     :param hide_input: if this is `True` then the input on the prompt will be
                        hidden from the user.  This is useful for password
                        input.
@@ -2177,6 +2181,7 @@ class Option(Parameter):
         show_default=False,
         prompt=False,
         confirmation_prompt=False,
+        prompt_required=True,
         hide_input=False,
         is_flag=None,
         flag_value=None,
@@ -2201,21 +2206,38 @@ class Option(Parameter):
             prompt_text = prompt
         self.prompt = prompt_text
         self.confirmation_prompt = confirmation_prompt
+        self.prompt_required = prompt_required
         self.hide_input = hide_input
         self.hidden = hidden
 
-        # Flags
+        # If prompt is enabled but not required, then the option can be
+        # used as a flag to indicate using prompt or flag_value.
+        self._flag_needs_value = self.prompt is not None and not self.prompt_required
+
         if is_flag is None:
             if flag_value is not None:
+                # Implicitly a flag because flag_value was set.
                 is_flag = True
+            elif self._flag_needs_value:
+                # Not a flag, but when used as a flag it shows a prompt.
+                is_flag = False
             else:
+                # Implicitly a flag because flag options were given.
                 is_flag = bool(self.secondary_opts)
+        elif is_flag is False and not self._flag_needs_value:
+            # Not a flag, and prompt is not enabled, can be used as a
+            # flag if flag_value is set.
+            self._flag_needs_value = flag_value is not None
+
         if is_flag and default_is_missing:
             self.default = False
+
         if flag_value is None:
             flag_value = not self.default
+
         self.is_flag = is_flag
         self.flag_value = flag_value
+
         if self.is_flag and isinstance(self.flag_value, bool) and type in [None, bool]:
             self.type = BOOL
             self.is_bool_flag = True
@@ -2486,11 +2508,23 @@ class Option(Parameter):
     def consume_value(self, ctx, opts):
         value, source = super().consume_value(ctx, opts)
 
+        # The parser will emit a sentinel value if the option can be
+        # given as a flag without a value. This is different from None
+        # to distinguish from the flag not being given at all.
+        if value is _flag_needs_value:
+            if self.prompt is not None and not ctx.resilient_parsing:
+                value = self.prompt_for_value(ctx)
+                source = ParameterSource.PROMPT
+            else:
+                value = self.flag_value
+                source = ParameterSource.COMMANDLINE
+
         # The value wasn't set, or used the param's default, prompt if
         # prompting is enabled.
-        if (
+        elif (
             source in {None, ParameterSource.DEFAULT}
             and self.prompt is not None
+            and (self.required or self.prompt_required)
             and not ctx.resilient_parsing
         ):
             value = self.prompt_for_value(ctx)
--- a/src/click/parser.py
+++ b/src/click/parser.py
@@ -29,6 +29,11 @@ from .exceptions import BadOptionUsage
 from .exceptions import NoSuchOption
 from .exceptions import UsageError
 
+# Sentinel value that indicates an option was passed as a flag without a
+# value but is not a flag option. Option.consume_value uses this to
+# prompt or use the flag_value.
+_flag_needs_value = object()
+
 
 def _unpack_args(args, nargs_spec):
     """Given an iterable of arguments and an iterable of nargs specifications,
@@ -79,12 +84,6 @@ def _unpack_args(args, nargs_spec):
         rv[spos + 1 :] = reversed(rv[spos + 1 :])
 
     return tuple(rv), list(args)
-
-
-def _error_opt_args(nargs, opt):
-    if nargs == 1:
-        raise BadOptionUsage(opt, f"{opt} option requires an argument")
-    raise BadOptionUsage(opt, f"{opt} option requires {nargs} arguments")
 
 
 def split_opt(opt):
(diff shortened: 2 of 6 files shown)


src/click/parser.py
```diff
@@ -337,11 +337,27 @@ def _match_long_opt(self, opt, explicit_value, state):
             if explicit_value is not None:
                 state.rargs.insert(0, explicit_value)
 
+            # if the option is a prompt and `prompt_required` is false,
+            # then the current rarg will serve as a flag, meaning the
+            # user will be prompted for an input if no argument is passed.
             nargs = option.nargs
             if len(state.rargs) < nargs:
-                _error_opt_args(nargs, opt)
```
amy-lei, line 342: I've noticed this block of code in `_match_long_opt` is identical to the one in `_match_short_opt`. So when making changes, I kept it this way and added the changes twice. Should this be factored out or is it not too big of a deal?

src/click/parser.py
```diff
@@ -337,11 +337,27 @@ def _match_long_opt(self, opt, explicit_value, state):
...
+                if option.obj.prompt is not None and not option.obj.prompt_required:
+                    value = option.obj.prompt_for_value(self.ctx)
+                else:
+                    _error_opt_args(nargs, opt)
             elif nargs == 1:
-                value = state.rargs.pop(0)
+                next_rarg = state.rargs[0]
```
amy-lei, line 350: I've noticed in the current codebase that this code block in `_match_long_opts` is identical in `_match_short_opts`. I kept it this way when making changes and just added the changes in both places. Should this be factored out, or is it not that significant? 

https://github.com/pallets/click/blob/637ea206c348100582beda3c76a8c51e9d7437ab/src/click/parser.py#L341-L347

davidism:
I'm wondering if it's appropriate to have `prompt` get triggered from the parser, it seems like the parser should only be concerned with the command line, and prompting should happen during later processing.

amy-lei:
> I'm wondering if it's appropriate to have `prompt` get triggered from the parser, it seems like the parser should only be concerned with the command line, and prompting should happen during later processing.

I moved it into `full_process_value` since that's where it originally prompts for a value. But to do so, I had to add an optional param to let it know there was a flag present

davidism:
I think there's another issue about options where the value is optional, maybe this solution will help with that too.

davidism:
I was thinking of #549.

amy-lei:
> I was thinking of #549.

Just found it and was about to comment it too haha. Should I try to fit that into this PR too?

davidism:
It seems like it should be a natural fit with this, since we now track if only the flag was passed. If it's only a flag, and there is a default, use the default instead of prompting. I don't *think* that should collide with any existing combination of settings.

src/click/core.py
```diff
@@ -1661,7 +1661,7 @@ def handle_parse_result(self, ctx, opts, args):
         with augment_usage_errors(ctx, param=self):
             value = self.consume_value(ctx, opts)
             try:
-                value = self.full_process_value(ctx, value)
+                value = self.full_process_value(ctx, value, is_flag=self.name in opts)
```
davidism, line 1664: Options already have an `is_flag` attribute. Can we take advantage of that? This probably also ties in with #549.

amy-lei: An idea I have is to set `is_flag=True` for prompt options that have `prompt_required=False`

amy-lei:
@davidism I've added the ability to use options with optional args via the `_flag_needs_value` attribute we talked about. I currently have it so that setting `prompt_required=False` implies `_flag_needs_value=True`, and this simplifies the checks in the parser. 

However, this did not get rid of the additional parameter we added to `full_process_value` because the only way to distinguish  between the value being `None` due to the parser setting it that way and it being `None` due to the user not passing it in all can't be made without the `opts` dict.

An idea in mind is to instead of having `prompt_required=False` imply `_flag_needs_value=True` in the constructor, maybe we could set it to True in the parser? And then set it back after we've prompted? This way we wouldn't need the additional parameter anymore. But I'm also not too fond about this since we're changing attribute of the Option instance midway.

davidism:
Instead of the parser producing `None`, it could produce a token that processing later on can check for.

```python
# define the sentinel value at the top of the module
flag_needs_value = object()

# return it when parsing
def match(...):
    return _flag_needs_value

# check it when processing
def process(...):
    if value is _flag_needs_value:
        ...
    elif value is None:
        ...
```

davidism:
I updated the PR description with the issues related to using a value instead of prompting. Also closed some existing PRs that will be handled by this. Let's make sure any tests they had are reflected here if applicable.

amy-lei:
Sounds good :+1: I'll be sure to add those tests and update the docs.

Some clarifying questions:
1) If the user sets `is_flag=False` but does not pass in a `flag_value`, should an error be raised during initialization? during parsing? Or should it be fine to just let the value default to something like `None` instead?
2) From the description, does this mean that in order to allow the user to prompt when no value is passed, `is_flag=False` must be explicitly passed too? Is is just passing `prompt_required=False` sufficient?

davidism:
1. For now it's better to require that `flag_value` be set. While we could always remove that restriction later, it would be harder to add it in.
2. For this, I think if `is_flag` is left as `None`, `prompt_required=False` can set it to `False`.

amy-lei:
I've used the sentinel value idea, although was not too sure about the name or if it was appropriate to name it the same name as the attribute. 

I also noticed #1230 enforced that the option cannot be `multiple` or have `nargs!=1`. I'm wondering if that should be the case here too or if it should be allowed instead. I did not add tests for it nor explicitly mention it in the docs yet.

davidism:
Will get to this after #1649. They're all related, might affect my answer to the question about `multiple` and `nargs`.

brendanator:
You may have already addressed this but I was attempting to implement something similar to the color option from git diff:
```man
       --color[=<when>]
           Show colored diff.  --color (i.e. without =<when>) is the same as
           --color=always.  <when> can be one of always, never, or auto. It can
           be changed by the color.ui and color.diff configuration settings.
```

The color option has a flag value of `always` and defaults to `never` so it will be covered by this functionality. 

The main difference I see is that the value must be separated from the option by `=`.  I think this is probably required to avoid the option gobbling up a parameter. e.g. `git diff --color path/to/file` should not set the color option to `path/to/file`

The example in the docs don't seem to handle this.

davidism:
Working on rebasing this after #1649.

PR #1621, opened 2020-07-06
non-chained group invoked without subcommand invokes result callback

### LINKED ISSUE #1178: resultcallback is not being called when invoke_without_command=True
I have a group-command (called `groupcmd`) which I would like the user to be able to invoke without having to specify a subcommand. I also have a resultcallback (via `@groupcmd.resultcallback`) which I would like to call after the group-command finishes for processing errors and logging them to file. 

If I set `invoke_without_command=True` inside of the `@click.group` definition, then the resultcallback does not get called. 

I am using click 6.6 on python 2.7.12

IamCathal:
As mentioned in https://github.com/pallets/click/issues/1178 this behavior is actually desired. I believe that the changes to the comments seem to do the job but I am not 100%. Apart from that the simple code fix I made I'm not too sure if more changes are needed else where where result callbacks are referenced.

Added a simple test to show the functionality with my addition.

fixes #1178

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1170,7 +1170,7 @@ class MultiCommand(Command):
         self.format_commands(ctx, formatter)
 
     def resultcallback(self, replace=False):
-        """Adds a result callback to the chain command.  By default if a
+        """Adds a result callback to the command.  By default if a
         result callback is already registered this will chain them but
         this can be disabled with the `replace` parameter.  The result
         callback is invoked with the return value of the subcommand
@@ -1189,10 +1189,10 @@ class MultiCommand(Command):
             def process_result(result, input):
                 return result + input
 
-        .. versionadded:: 3.0
-
         :param replace: if set to `True` an already existing result
                         callback will be removed.
+
+        .. versionadded:: 3.0
         """
 
         def decorator(f):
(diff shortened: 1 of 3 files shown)


src/click/core.py
```diff
@@ -1258,15 +1258,10 @@ def _process_result(value):
...
         if not ctx.protected_args:
-            # If we are invoked without command the chain flag controls
-            # how this happens.  If we are not in chain mode, the return
-            # value here is the return value of the command.
-            # If however we are in chain mode, the return value is the
-            # return value of the result processor invoked with an empty
-            # list (which means that no subcommand actually was executed).
+            # The return value is the return value of the result processor
+            # invoked with an empty list (which means that no subcommand
+            # actually was executed).
```
teymour-aldridge, line 1263:
```suggestion
            # was actually executed).
```

IamCathal: Sounds much better, thank you

IamCathal:
Rebased commits and added changelog entry

davidism:
Changed so regular groups invoke the result callback with `None`, while chained groups invoke it with `[]`. Parametrized the test to test both cases.

PR #1630, opened 2020-07-17
Make ``BOOL`` accept the strings "on" and "off"

### LINKED ISSUE #1629: BOOL should accept "on" and "off"
The `BOOL` parameter type should accept & convert the strings "on" and "off".  With this change, `BOOL` would support all of the boolean strings that `configparser` supports.

jwodder:
Closes #1629.

### DIFF
--- a/docs/parameters.rst
+++ b/docs/parameters.rst
@@ -48,9 +48,9 @@ different behavior and some are supported out of the box:
 
 ``bool`` / :data:`click.BOOL`:
     A parameter that accepts boolean values.  This is automatically used
-    for boolean flags.  If used with string values ``1``, ``yes``, ``y``, ``t``
-    and ``true`` convert to `True` and ``0``, ``no``, ``n``, ``f`` and ``false``
-    convert to `False`.
+    for boolean flags.  If used with string values, ``1``, ``yes``, ``y``,
+    ``t``, ``true``, and ``on`` convert to `True`, and ``0``, ``no``, ``n``,
+    ``f``, ``false``, and ``off`` convert to `False`.
 
 :data:`click.UUID`:
     A parameter that accepts UUID values.  This is not automatically
--- a/docs/parameters.rst
+++ b/docs/parameters.rst
@@ -47,10 +47,10 @@ different behavior and some are supported out of the box:
     A parameter that only accepts floating point values.
 
 ``bool`` / :data:`click.BOOL`:
-    A parameter that accepts boolean values.  This is automatically used
-    for boolean flags.  If used with string values, ``1``, ``yes``, ``y``,
-    ``t``, ``true``, and ``on`` convert to `True`, and ``0``, ``no``, ``n``,
-    ``f``, ``false``, and ``off`` convert to `False`.
+    A parameter that accepts boolean values. This is automatically used
+    for boolean flags. The string values "1", "true", "t", "yes", "y",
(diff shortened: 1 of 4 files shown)


docs/parameters.rst
```diff
@@ -48,9 +48,9 @@ different behavior and some are supported out of the box:
 
 ``bool`` / :data:`click.BOOL`:
     A parameter that accepts boolean values.  This is automatically used
-    for boolean flags.  If used with string values ``1``, ``yes``, ``y``, ``t``
-    and ``true`` convert to `True` and ``0``, ``no``, ``n``, ``f`` and ``false``
-    convert to `False`.
+    for boolean flags.  If used with string values, ``1``, ``yes``, ``y``,
```
teymour-aldridge, line 51:
```suggestion
    for boolean flags.  If used with the string values, ``1``, ``yes``, ``y``,
```

docs/parameters.rst
```diff
@@ -48,9 +48,9 @@ different behavior and some are supported out of the box:
...
+    ``t``, ``true``, and ``on`` convert to `True`, and ``0``, ``no``, ``n``,
+    ``f``, ``false``, and ``off`` convert to `False`.
```
teymour-aldridge, line 53:
```suggestion
    ``f``, ``false``, and ``off`` the value is `False`.
```

docs/parameters.rst
```diff
@@ -48,9 +48,9 @@ different behavior and some are supported out of the box:
 
 ``bool`` / :data:`click.BOOL`:
     A parameter that accepts boolean values.  This is automatically used
-    for boolean flags.  If used with string values ``1``, ``yes``, ``y``, ``t``
-    and ``true`` convert to `True` and ``0``, ``no``, ``n``, ``f`` and ``false``
-    convert to `False`.
+    for boolean flags.  If used with string values, ``1``, ``yes``, ``y``,
+    ``t``, ``true``, and ``on`` convert to `True`, and ``0``, ``no``, ``n``,
```
teymour-aldridge, line 52:
```suggestion
    ``t``, ``true``, and ``on`` the value is `True`, and ``0``, ``no``, ``n``,
```

docs/parameters.rst
jwodder: That's bad grammar.

docs/parameters.rst
jwodder: That's bad grammar.

docs/parameters.rst
jwodder: All of your sugggestions combined are insufficient to produce a grammatical sentence.  It would have to be "If used with the string values *[no comma]* `1` ... and `on` *[insert comma]* the value is `True`, and *if used with the string values* `0` ... and `off` *[insert comma]* the value is `False`."  Moreover, the current sentence is perfectly grammatical, and I fail to see any improvements in what you're suggesting.

ThiefMaster: What about going for something much simpler:

```rst
The following string values are converted to their respective boolean counterparts:
``1``, ``yes``, ``y``, ``t``, ``true``, ``0``, ``no``, ``n``, ``f`` and ``false``
```

I think it's completely obvious which ones become `true` and which ones `false`...

docs/parameters.rst
davidism: Seems correct to me, but I went with a different rewrite.

PR #1666, opened 2020-09-09
Use in MSYS2 w/o termios available

ustsv:
MSYS2 and Python 3.8.5 not have termios module. Use WIN implementation instead of this.


davidism:
MSYS2 is not Windows. If your Python on MSYS2 doesn't have the termios module, you need to fix your Python.

PR #1686, opened 2020-10-11
Revert "Add python 3.9 to CI."

davidism:
Reverts pallets/click#1685

GitHub doesn't have pre-made env with 3.9, so it has to install it, which makes Windows and Mac builds too slow.


ThiefMaster:
Skipping those versions for win/mac builds wouldn't work?

askpatrickw:
https://github.com/actions/virtual-environments/issues/1740#issuecomment-718886755
Github Actions is live with Python 3.9 now

PR #1693, opened 2020-10-17
Case insensitive shell completions for Choice(..., case_sensitive=False)

### LINKED ISSUE #1692: shell completion is case sensitive, even with click.Choice(case_sensitive=False)
### Expected Behavior

Since I set case_sensitive to False, I would expect shell completions to search without case sensitivity.

```python
import click
from click._bashcomplete import get_choices

options = ['Paul', 'Simon', 'Art', 'Garfunkel']

@click.command()
@click.option('--opt', type=click.Choice(options))
def demo():
    pass

completions = list(get_choices(demo, 'dummy', ['--opt'], 'gar'))
assert len(completions) == 1, "expected to find 'Garfunkel' as a completion item"
```

### Actual Behavior

No completions are found. However, if I change the `incomplete` parameter to `"Gar"`, one completion is found as expected.

### Environment

* Python version: 3.7.7
* Click version: 7.0 (but I've reviewed the code on master and it doesn't seem to have a fix)

felixnorden:
This PR adds the logic and a test for checking case insensitive completions for `Choice` instances which have the argument `case_sensitive=False`. 

What changed was having the `Choice` class' `shell_complete`-method take `self.case_sensitive` into account when matching on the provided instance choices by using temporary string transforms on the `incomplete` argument and `self.choices` into all lower case when `self.case_sensitive=False`.

- Fixes #1692 

Checklist:

- [x] Run `pytest` and `tox`, no tests failed. (**NOTE**: No test results changed with this contribution, there was 1 failing test though)

### DIFF
--- a/tests/test_shell_completion.py
+++ b/tests/test_shell_completion.py
@@ -304,3 +304,13 @@ def test_context_settings(runner):
         env={"COMP_WORDS": "", "COMP_CWORD": "0", "_CLI_COMPLETE": "bash_complete"},
     )
     assert result.output == "plain,a\nplain,b\n"
+
+
+@pytest.mark.parametrize(("value", "expect"), [(False, ["Au", "al"]), (True, ["al"])])
+def test_choice_case_sensitive(value, expect):
+    cli = Command(
+        "cli",
+        params=[Option(["-a"], type=Choice(["Au", "al", "Bc"], case_sensitive=value))],
+    )
+    completions = _get_words(cli, ["-a"], "a")
+    assert completions == expect
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -151,6 +151,8 @@ Unreleased
     ending of a phrase or double linebreak. :issue:`1082`
 -   Skip progress bar render steps for efficiency with very fast
     iterators by setting ``update_min_steps``. :issue:`676`
+-   Respect ``case_sensitive=False`` when doing shell completion for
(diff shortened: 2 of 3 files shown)


tests/test_shell_completion.py
```diff
@@ -304,3 +304,14 @@ def complete(ctx, param, incomplete):
...
+
+def test_case_insensitive_completions():
+    options = ["Foo", "Bar", "Baz"]
+
+    opts = [Option(["--opt"], type=Choice(options, case_sensitive=False))]
+    cli = Command("cli", params=opts)
+
+    completions = _get_completions(cli, ["--opt"], "f")
+    assert len(completions) == 1
+    assert completions[0].value == "Foo", "expected to find 'Foo' as a completion item"
```
bgschiller, line 317: Consider adding a test with `case_sensitive=True`

bgschiller (commented):
Thanks for the quick PR! It might be worth adding a test for the opposite behavior: that case sensitive choice _don't_ autocomplete when the case doesn't match.

felixnorden:
Sorry for the delay, I have added a test case for this as well and the tests are passing as expected :)

tests/test_shell_completion.py
felixnorden: Added a case right below this one :)

davidism:
Refactored to avoid using an inner function. Refactored test to use `paremtrize` and test choices both cases.

PR #1698, opened 2020-10-26
Add update interval for progress bar

### LINKED ISSUE #676: Avoid rendering on highly frequent updates in progressbar
I am using a progressbar in a loop with short iterations, which negatively affects performance. It would be good to supply a chunk size such that the heavy part of an update is only performed every nth time. 

Currently I am using a small wrapper around it:

```
class ChunkProgressbar:
    def __init__(self, iterator, chunk_size=100, enabled=True, **kwargs):
        self.iterator = iterator
        self.chunk_size = chunk_size
        self.bar = progressbar(iterator, **kwargs)
        self.missed_updates = 0

    def __enter__(self):
        if self.bar is not None:
            self.bar.__enter__()
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        self.bar.update(self.missed_updates)
        self.bar.__exit__(exc_type, exc_val, exc_tb)

    def __iter__(self):
        for index, elem in enumerate(self.iterator):
            yield elem
            self.missed_updates += 1
            if index % self.chunk_size == self.chunk_size - 1:
                self.bar.update(self.missed_updates)
                self.missed_updates = 0
```

One could also throttle the rendering based on timing information (like the eta calculation), but that might be costly again due to timing calls or mess with heterogeneous iteration times.

srafi1:
Adds an optional `update_interval` for progress bars that allows the user to skip unnecessary renders for frequently updated iterators.

- Fixes #676

Checklist:

### DIFF
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -62,6 +62,7 @@ class ProgressBar:
         label=None,
         file=None,
         color=None,
+        update_min_steps=1,
         width=30,
     ):
         self.fill_char = fill_char
@@ -77,6 +78,8 @@ class ProgressBar:
             file = _default_text_stdout()
         self.file = file
         self.color = color
+        self.update_min_steps = update_min_steps
+        self._completed_intervals = 0
         self.width = width
         self.autowidth = width == 0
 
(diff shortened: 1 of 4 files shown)


davidism:
This is a good PR, but it's making me realize we're just slowly reimplementing [tqdm](https://github.com/tqdm/tqdm#monitoring-thread-intervals-and-miniters) without the design and focus of that library. I'll merge this, but I think I'm going to close any other progress bar feature requests.

src/click/_termui_impl.py
```diff
@@ -62,6 +62,7 @@ def __init__(
         label=None,
         file=None,
         color=None,
+        update_interval=1,
```
davidism, line 65: Maybe call this `update_min_steps` to make it clearer what the interval is and how it's checked.

src/click/termui.py
```diff
@@ -397,6 +402,10 @@ def progressbar(
                   default is autodetection.  This is only needed if ANSI
                   codes are included anywhere in the progress bar output
                   which is not the case by default.
+
```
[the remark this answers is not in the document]
srafi1: Done in d3c5e05

src/click/_termui_impl.py
srafi1: Done in d3c5e05. I named it before jumping into the implementation, but I agree that that's a better name.

PR #1709, opened 2020-11-10
Shell completion escape handling

### LINKED ISSUE #1708: Shell completion does't parse escaped slashes or spaces correctly
When a value with an escape string, for instance `my\ file`, is passed to a `click.ParamType`, the `incomplete` value which is passed to `click.ParamType.shell_complete` will only be `"my\ "` instead of `my file`.

### Expected Behavior

```python
import click

@click.command()
@click.argument("name", type=click.Choice(["my file", "my folder", "my\\ file", "other"]))
def cli(name):
    click.echo(name)

if __name__ == '__main__':
    cli()
```

Completing `my\ f` should be interpreted as the Python string `"my f"` and match the first two values.

```
$ example my\ f<TAB>
my\ file
my\ folder
```

### Actual Behavior

Completing `my\ f` is interpreted as the string `"my\\ f"` and matches the third value.

```shell
$ example my\ f<TAB>
my\\\ file
```

### Environment

* Python version: 3.8.6
* Click version: 8.0.0.dev0

SamSchott:
This PR fixes #1708 by using Python's `shlex` module to parse and split argument strings for shell completion. Incomplete arguments where splitting with `shlex.split` fails with a `ValueError` are handled by inspecting the state of the parser after the failure. There are only two such failures: 

1. A missing closing quote for an incomplete argument. This is handled by appending the missing quote. Therefore, shell completion suggestions for an argument `'path to` will be generated with `incomplete = "path to"`.
2. An incomplete escape sequence. This is handled by removing the incomplete sequence. Therefore, shell completion suggestions for an argument `'path\` will be generated with `incomplete = "path"`. This mirrors the behaviour of zsh and possibly other shells.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -107,6 +107,8 @@ Unreleased
         The shell now comes first, such as ``{shell}_source`` rather
         than ``source_{shell}``, and is always required.
 
+-   Completion correctly parses command line strings with incomplete
+    quoting or escape sequences. :issue:`1708`
 -   Extra context settings (``obj=...``, etc.) are passed on to the
     completion system. :issue:`942`
 -   Include ``--help`` option in completion. :pr:`1504`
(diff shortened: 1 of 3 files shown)


davidism:
`lex.token` contains the incomplete token after an error, so that can be used instead of modifying and re-splitting the string. Also, used `parametrize` and added a test of standard unquoted behavior.

PR #1715, opened 2020-11-23
Allow custom temp directories with CLIRunner.isolated_filesystem

### LINKED ISSUE #395: Allow custom retainable temporary directories in CLIRunner.isolated_filesystem
I would like to propose two changes to `CLIRunner.isolated_filesystem`, which I believe would significantly improve the experience of debugging tests using it:
1. Add an argument that allows users to specify the temporary directory.
2. Add an argument indicating that the temporary directory shouldn't be deleted (if an exception occurs).

The current behavior of `isolated_filesystem` of creating a random temporary directory without allowing specifying a specific directory or even a (partial) name and deleting it no matter what, makes it impossible to inspect the directory in case of test failures to gather more information. Even if the directory wouldn't be deleted, it's difficult to find.

By allowing the user to define a temporary directory, it's possible to take advantage of mechanisms such as [py.test's `tmpdir` fixture](http://pytest.org/latest/tmpdir.html) or other systems to separate setup code from the actual test and to make the directory known to the user for inspection.

Diaga:
CLIRunner.isolated_filesystem accepts a custom argument tempdir. If user specifies custom tempdir, we do not delete it afterwards assuming user will clean up afterwards.

- Fixes #395

Checklist:

### DIFF
--- a/src/click/testing.py
+++ b/src/click/testing.py
@@ -380,18 +380,30 @@ class CliRunner:
         )
 
     @contextlib.contextmanager
-    def isolated_filesystem(self):
-        """A context manager that creates a temporary folder and changes
-        the current working directory to it for isolated filesystem tests.
+    def isolated_filesystem(self, temp_dir=None):
+        """A context manager that creates a temporary directory and
+        changes the current working directory to it. This isolates tests
+        that affect the contents of the CWD to prevent them from
+        interfering with each other.
+
+        :param temp_dir: Create the temporary directory under this
+            directory. If given, the created directory is not removed
+            when exiting.
+
+        .. versionchanged:: 8.0
+            Added the ``temp_dir`` parameter.
         """
         cwd = os.getcwd()
-        t = tempfile.mkdtemp()
+        t = tempfile.mkdtemp(dir=temp_dir)
         os.chdir(t)
+
         try:
             yield t
         finally:
             os.chdir(cwd)
-            try:
-                shutil.rmtree(t)
-            except OSError:  # noqa: B014
-                pass
+
+            if temp_dir is None:
+                try:
+                    shutil.rmtree(t)
+                except OSError:  # noqa: B014
+                    pass
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -319,3 +319,18 @@ def test_file_stdin_attrs(runner):
 
     result = runner.invoke(cli, ["-"])
     assert result.output == "<stdin>\nr"
+
+
+def test_isolated_runner(runner):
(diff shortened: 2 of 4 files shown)


src/click/testing.py
```diff
@@ -380,18 +380,25 @@ def invoke(
         )
 
     @contextlib.contextmanager
-    def isolated_filesystem(self):
+    def isolated_filesystem(self, tempdir=None):
```
flosincapite, line 383: So presumably, you can add a test [here](https://github.com/pallets/click/blob/master/tests/test_testing.py) to exercise this new behavior. 

Actually, since this function is not tested directly (at present), you might want to add two tests: one where you call `isolated_filesystem` with an arg, and one without.

In each test, you can just use Python's `os` library to inspect the temporary directory. You should be able to call `os.path.exists` on the yielded directory both during and after the lifetime of the context manager. Something like

```
lingering_directory = None
with CLIRunner.isolated_filesystem() as directory:
    lingering_directory = directory
    self.assertTrue(os.path.exists(directory))
self.assertFalse(os.path.exists(directory))
```

I leave it to you to determine the testing logic for when `isolated_filesystem` takes an argument.

tests/test_testing.py
```diff
@@ -319,3 +321,22 @@ def cli(f):
...
+
+
+def test_isolated_runner_custom_tempdir(runner):
+    lingering_directory = None
+    with runner.isolated_filesystem(tempdir=tempfile.gettempdir()) as directory:
+        lingering_directory = directory
+        assert os.path.exists(directory)
+
+    assert os.path.exists(lingering_directory)
+    shutil.rmtree(lingering_directory)
```
flosincapite, line 342: Looking good!

One thing to notice: `shutil.rmtree` won't be called if any of the code above raises an exception, meaning that this code might or might not leave an artifact lying around. A context manager could fix this, as could a `try`/`finally` block.

However, in this case, it's not a big deal since `tempfile.gettempdir` will use the OS's temporary directory by default.

flosincapite (approved):

flosincapite:
Looks good to me!

src/click/testing.py
```diff
@@ -380,18 +380,25 @@ def invoke(
...
         """A context manager that creates a temporary folder and changes
         the current working directory to it for isolated filesystem tests.
+
+        :param tempdir: the path where the temporary directory should be created
+
+        .. versionchanged:: 7.1.2
         """
+        tempfile.tempdir = tempdir
```
davidism, line 391: This modifies global state and doesn't reset it. It's also not needed, `mkdtemp` takes a `dir` parameter.

PR #1741, opened 2020-12-27
Update documention for Custom Multi Commands

Saphyel:
I updated the documentation for custom multi commands to use the python 3 way. 

- [x] Add comment types
- [x] Replace the os.path for pathlib way


davidism:
Thanks for looking at this. However, these changes aren't really relevant to the example, so I don't think I'll merge them right now.

PR #1754, opened 2021-01-14
Replaced redundant mktemp() with mkstemp() 

### LINKED ISSUE #1752: Use of deprecated & insecure tempfile.mktemp()
### Expected Behavior

_tempfilepager() in _termui_impl.py uses a depreciated function that is considered insecure by the core [Python team](https://docs.python.org/3/library/tempfile.html#tempfile.mktemp). It is recommended to use mkstemp() instead. 

```python
import tempfile
filename = tempfile.mkstemp()  # Suggested replacement
```

### Actual Behavior

Python's warning: 

> Use of this function may introduce a security hole in your program. By the time you get around to doing anything with the file name it returns, someone else may have beaten you to the punch.

```python
import tempfile
filename = tempfile.mktemp()  # Depreciated and insecure
```

### Environment

* Python version: 3.8.5
* Click version: 7.1.2

MIS-Alex:
Replaced redundant function with the recommended alternative.  

- Fixes #1752

Checklist:

### DIFF
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -424,7 +424,7 @@ def _tempfilepager(generator, cmd, color):
     """Page through text by invoking a program on a temporary file."""
     import tempfile
 
-    filename = tempfile.mktemp()
+    filename = tempfile.mkstemp()
     # TODO: This never terminates if the passed generator never terminates.
     text = "".join(generator)
     if not color:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -155,6 +155,7 @@ Unreleased
     iterators by setting ``update_min_steps``. :issue:`676`
 -   Respect ``case_sensitive=False`` when doing shell completion for
     ``Choice`` :issue:`1692`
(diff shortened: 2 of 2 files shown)


src/click/_termui_impl.py
```diff
@@ -424,7 +424,7 @@ def _tempfilepager(generator, cmd, color):
     """Page through text by invoking a program on a temporary file."""
     import tempfile
 
-    filename = tempfile.mktemp()
+    filename = tempfile.mkstemp()
```
ThiefMaster, line 427: This doesn't look correct. `mkstemp` returns a tuple, but below we use `filename` as if it was a string filename...

ThiefMaster: Also further below, shouldn't we use `subprocess.call()` instead of `os.system()`? I don't think you can put anything nasty in the temp filename, but I think it'd also fail in other cases like `cmd` containing a space..

davidism: Treating the return as a tuple was fixed later. Using `subprocess.call` is an open issue #1476, I need to test things like this before I merge it.

PR #1755, opened 2021-01-15
show_defaults now accepts string value as suggested by the docs

### LINKED ISSUE #1732: Custom default label ignored if default is None
I found out that custom default label is not shown when `default` is `None`. In some cases `default = None` is intentional, but custom default label would be still helpful.

An example with `--limit` option, with `None` as `default` meaning no limit:
```python
import click

@click.command()
@click.option('--limit', type=int, show_default='unlimited')
def main(limit: int):
    print('Limit: {}'.format(limit))

if __name__ == '__main__':
    main()
```

### Expected Behavior
Tell us what should happen.

```sh
$ python3 /tmp/foo.py --help
Usage: foo.py [OPTIONS]

Options:
  --limit INTEGER  [default: unlimited]
```

### Actual Behavior
Tell us what happens instead.

```sh
$ python3 /tmp/foo.py --help
Usage: foo.py [OPTIONS]

Options:
  --limit INTEGER
```
Default is not shown, making the description less useful.

### Environment

* Python version: 3.9.1
* Click version: 7.1.2

b3n4kh:
Fixes #1732 by removing the mandatory "default" flag for using show_defaults with a string value.

- Fixes #1732 

Checklist:

- [docs already up2date] Add or update relevant docs, in the docs folder and in code. 

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -157,6 +157,8 @@ Unreleased
     ``Choice`` :issue:`1692`
 -   Use ``mkstemp()`` instead of ``mktemp()`` in pager implementation.
     :issue:`1752`
+-   If ``Option.show_default`` is a string, it is displayed even if
+    ``default`` is ``None``. :issue:`1732`
 
 
 Version 7.1.2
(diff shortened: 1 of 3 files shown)


davidism:
If `Option.show_default` or `Context.show_default` is `True`, this changes causes it to show `None` instead of hiding the value. Needs some extra checks to only show when it's a string.

PR #1772, opened 2021-02-02
Enable mypy checking and add minimal hints

### LINKED ISSUE #1764: python type hints and well defined parameters for decorators
I would like to have well defined types and type hints for click so that click is easier and safer to use.

Having this would make it possible to detect some classes of click usage errors with a python type checker like mypy. This problem could be solved by wrapping Click, but I would prefer to just add the type checking in click itself.

I am working on a patch to add type hints, I'm opening this issue so I can gauge what would be acceptable for you so I don't do something that you won't accept.

My plan more or less is:
 1. enable mypy checking from your pre-commit hook for all files in src with this config:
  ```ini
[mypy]                                                                                   
python_version = 3.6                                                                     
warn_unused_configs = true                                                               
ignore_missing_imports = true                                                            
disallow_untyped_defs = true                                                             
warn_redundant_casts = true                                                              
warn_unused_ignores = true                                                               
warn_no_return = false                                                                   
  ```
 2. Add type hinting so mypy completes without errors
 3. Add well defined parameters for decorators (I will likely use type overload so that you can still pass arbitrary parameters if you use custom classes) so that auto completion and type checking works properly for decorators

aucampia:
This pull request aims to:
 - enable mypy typing with same settings as in [werkzeug](https://github.com/pallets/werkzeug/)
 - add minimal hinting so that mypy completes without errors

- Relates to #1764

Checklist:

  - Updated tox.ini to run mypy
  - I checked docs for other things flake8, Black, did not seem there is something more worth mentioning, as tox is already mentioned
  - Don't think this makes sense for this pull request
  - Not applicable for this change


aucampia:
I made a pull request with minimal typing and [werkzeug](https://github.com/pallets/werkzeug/) config. I can add more to this patch but I think it would be good to merge this regardless of any further changes.

davidism:
Thanks for working on this. Right off the bat, I can say that `t.Any` should not be used unless absolutely necessary. Either type it accurately or don't type it.

src/click/core.py
```diff
@@ -1638,7 +1638,8 @@ class Group(MultiCommand):
     #: custom groups.
     #:
     #: .. versionadded:: 8.0
-    # TODO: when l.Literal can be used `type` should be replaced `t.Literal[type]`
+    # TODO: when support for python versions below 3.8 is dropped
+    #   `type` should be replaced with `t.Literal[type]`
     group_class: t.Optional[t.Union[t.Type["Group"], type]] = None
```
davidism, line 1643: I don't understand the `Union[Type[Group], type]]`, why is `type` needed? Also, if you want to use `Literal`, you can use it in quotes right now.

aucampia: AFAIU from the documentation and code valid values for group_class is subclasses of `Group` and the literal `type`:

[here](https://github.com/aucampia/pallets-click/blob/3b01ca3a52862435897e6118e5da82c0c04a61e7/src/click/core.py#L1635)
```python
    #: If set to the special value :class:`type` (literally
    #: ``group_class = type``), this group's class will be used as the
    #: default class. This makes a custom group class continue to make
    #: custom groups.
```

[here](https://github.com/aucampia/pallets-click/blob/iwana-typing-000/src/click/core.py#L1705)
```python
            if self.group_class is type:
                kwargs["cls"] = type(self)
            else:
                kwargs["cls"] = self.group_class
```

type in itself is a class defined in mypy as `class type(object):` so, in this case I guess it is the most appropriate type for a variable that stores the actual type `type`. Maybe there is a better option but other than literal I don't think there is.

`typing.Literal` is only available from python 3.8. I tried it in quotes but it does not help.

I tried replacing

```python
group_class: t.Optional[t.Union[t.Type["Group"], type]] = None
```

with each of these options:

```python
group_class: t.Optional[t.Union[t.Type["Group"], "t.Literal[type]"]] = None
group_class: t.Optional[t.Union[t.Type["Group"], "typing.Literal[type]"]] = None
group_class: t.Optional[t.Union[t.Type["Group"], "Literal[type]"]] = None
```

It results in these errors:

```
src/click/core.py:1643: error: Name 't.Literal' is not defined
src/click/core.py:1644: error: Name 'typing.Literal' is not defined
src/click/core.py:1643: error: Name 'Literal' is not defined
```

mypy docs suggests using [`typing_extentions`](https://mypy.readthedocs.io/en/stable/literal_types.html#literal-types)


I also tried this:

```python
if t.TYPE_CHECKING:
    from typing import Literal
```

But this resulted in errors (python 3.7):

```
src/click/core.py:39: error: Module 'typing' has no attribute 'Literal'
```

davidism: Oh right, I wrote that code, forgot that `type` was a special case.

Weird about quoting `t.Literal`, I definitely tried that in Werkzeug and it worked at the time, but I can reproduce the error you get now. That said, I don't particularly care about supporting old versions, if you're on 3.7 you should probably consider upgrading to 3.9 anyway. ;-) You can continue leaving todo notes, or do `if t.TYPE_CHECKING: import typing_extensions as te` if you want.

aucampia: As far as I can tell click is intended to support back to python 3.6 and I think for supporting older versions it is best to rather go for `typing_extensions`. This way the typing will be useful for those versions.

I checked werkzeug, and indeed it fails if I use python versions below 3.8, e.g.

```
$ pipx run --python python3.7 tox -e typing 
...
src/werkzeug/datastructures.pyi:14: error: Module 'typing' has no attribute 'Literal'
src/werkzeug/http.py:377: error: Name 't.Literal' is not defined
src/werkzeug/http.py:383: error: Overloaded function signature 2 will never be matched: signature 1's parameter type(s) are the same or broader
src/werkzeug/http.py:384: error: Name 't.Literal' is not defined
...
```

I did however check what happens if I try and use this from python 3.6 vs 3.8, and as far as I can tell mypy just treats literals as a `typing.Any`, which is probably not that bad.

test code `test_case.py`

```python
import werkzeug

from werkzeug.datastructures import MultiDict

print("test: ", MultiDict([("foo", "bar"), ("foo", "hehe")]).to_dict("this should not be accepted"))
```

setup

```bash
python3.6 -m venv env-3.6
env-3.6/bin/pip install mypy git+https://github.com/pallets/werkzeug.git@master
python3.9 -m venv env-3.9
env-3.9/bin/pip install mypy git+https://github.com/pallets/werkzeug.git@master
```

test:

```bash
env-3.6/bin/mypy --python-version 3.6 test_case.py
env-3.9/bin/mypy --python-version 3.9 test_case.py
```

output:

```
$ env-3.6/bin/mypy --python-version 3.6 test_case.py
Success: no issues found in 1 source file

$ env-3.9/bin/mypy --python-version 3.9 test_case.py
test_case.py:5: error: No overload variant of "to_dict" of "MultiDict" matches argument type "str"
test_case.py:5: note: Possible overload variant:
test_case.py:5: note:     def to_dict(self, flat: Literal[False]) -> Dict[str, List[str]]
test_case.py:5: note:     <1 more non-matching overload not shown>
```

I think if you want to still support python 3.6 the best option is `if t.TYPE_CHECKING: import typing_extensions as te` and I will change to that then.

davidism: Using `typing_extensions` is fine, thanks for checking!

aucampia: Actually `Literal[type]` does not work, because Literal only supports strings, bools, Enum values and None[1]. I changed the typing to `group_class: t.Optional[t.Union[t.Type["Group"], t.Type[type]]] = None` - it is pretty restrictive this way because I don't think there are many things that are subclasses of type (if there are any).

I added a note about this also.

[1]: https://www.python.org/dev/peps/pep-0586/#legal-parameters-for-literal-at-type-check-time

PR #1782, opened 2021-02-16
skip code tests when only docs change

davidism:
Skip running the code tests when only docs pages are changed.

This *should* work, but I can't test it in this PR, since it changes the workflow file, which is not a docs file. But there's an easy docs issue that I'll use to test this after merging.


davidism:
See #1783, only docs were changed and the tests workflow didn't run.

PR #1786, opened 2021-02-18
confirmation prompt text can now be customized

### LINKED ISSUE #723: Configurable "Repeat for confirmation:" text
It would be nice to be able to set this prompt confirmation text.

Saif807380:
The user can now customize the confirmation prompt text. If `confirmation_prompt` is set to `True`, the default prompt text will be used. If a non-empty string is provided then that will be used.

- fixes #723 

Checklist:

### DIFF
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -85,21 +85,14 @@ def prompt(
     If the user aborts the input by sending a interrupt signal, this
     function will catch it and raise a :exc:`Abort` exception.
 
-    .. versionadded:: 7.0
-       Added the show_choices parameter.
-
-    .. versionadded:: 6.0
-       Added unicode support for cmd.exe on Windows.
-
-    .. versionadded:: 4.0
-       Added the `err` parameter.
-
     :param text: the text to show for the prompt.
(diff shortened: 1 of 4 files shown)


src/click/termui.py
```diff
@@ -137,6 +141,18 @@ def prompt_func(text):
         text, prompt_suffix, show_default, default, show_choices, type
     )
 
+    if confirmation_prompt:
+        confirmation_prompt = _build_prompt(
```
davidism, line 145: `_build_prompt` is for adding information about the type and default value, which isn't relevant when confirming.

src/click/termui.py
```diff
@@ -137,6 +141,18 @@ def prompt_func(text):
...
+            "Repeat for confirmation"
+            if confirmation_prompt is True
+            else confirmation_prompt,
+            prompt_suffix,
+            show_choices=False,
+        )
+
+    elif confirmation_prompt is False:
+        confirmation_prompt = None
```
davidism, line 154: This isn't needed, it doesn't change the behavior further down.

PR #1805, opened 2021-03-05
Better error message for bad parameter default

### LINKED ISSUE #1749: Traceback when CLI option has bad default could be more helpful
### Expected Behavior

In the following command, I add an option `--multiple` with the metadata `multiple=True`, but the default `False` instead of `[False]`.

Ideally, the traceback should indicate that the `--multiple` option has a bad default.

```python
#!/usr/bin/env python
import click

@click.command()
@click.option("--multiple", multiple=True, default=False, show_default=True)
def cli(debug):
    pass

if __name__ == '__main__':
    cli()
```

### Actual Behavior

Tell us what happens instead.

```pytb
  File "./test_click.py", line 12, in <module>
    cli()
  File "/home/user/anaconda3/envs/pybids-pre/lib/python3.8/site-packages/click/core.py", line 1025, in __call__
  File "/home/user/anaconda3/envs/pybids-pre/lib/python3.8/site-packages/click/core.py", line 954, in main
  File "/home/user/anaconda3/envs/pybids-pre/lib/python3.8/site-packages/click/core.py", line 852, in make_context
    self.parse_args(ctx, args)
  File "/home/user/anaconda3/envs/pybids-pre/lib/python3.8/site-packages/click/core.py", line 1261, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "/home/user/anaconda3/envs/pybids-pre/lib/python3.8/site-packages/click/core.py", line 2078, in handle_parse_result
    value, source = self.consume_value(ctx, opts)
  File "/home/user/anaconda3/envs/pybids-pre/lib/python3.8/site-packages/click/core.py", line 2509, in consume_value
    value, source = super().consume_value(ctx, opts)
  File "/home/user/anaconda3/envs/pybids-pre/lib/python3.8/site-packages/click/core.py", line 1975, in consume_value
    value = self.get_default(ctx)
  File "/home/user/anaconda3/envs/pybids-pre/lib/python3.8/site-packages/click/core.py", line 2453, in get_default
    return super().get_default(ctx, call=call)
  File "/home/user/anaconda3/envs/pybids-pre/lib/python3.8/site-packages/click/core.py", line 1957, in get_default
    return self.type_cast_value(ctx, value)
  File "/home/user/anaconda3/envs/pybids-pre/lib/python3.8/site-packages/click/core.py", line 2007, in type_cast_value
    return _convert(value, (self.nargs != 1) + bool(self.multiple))
  File "/home/user/anaconda3/envs/pybids-pre/lib/python3.8/site-packages/click/core.py", line 2005, in _convert
    return tuple(_convert(x, level - 1) for x in value)
TypeError: 'bool' object is not iterable
```

I discovered this while running tests on pre-released dependencies, picking up click 8.0.0a1, which no longer accepts the above. The offending case was in a subcommand with many options, so it was not as obvious as it would be with a good error message.

### Environment

* Python version: 3.8
* Click version: 8.0.0a1

Saif807380:
Added a `try` block to check whether the default value provided to an option with `multiple=True` is a valid iterable. If the value is not a valid iterable, a `TypeError` with a better error message is raised.

- fixes #1749 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -182,6 +182,8 @@ Unreleased
     previous item. :issue:`1353`
 -   The ``Path`` param type can be passed ``path_type=pathlib.Path`` to
     return a path object instead of a string. :issue:`405`
+-   TypeError is raised when parameter with ``multiple=True`` or
+    ``nargs > 1`` has non-iterable default. :issue:`1749`
 
 
 Version 7.1.2
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -182,7 +182,7 @@ Unreleased
     previous item. :issue:`1353`
 -   The ``Path`` param type can be passed ``path_type=pathlib.Path`` to
     return a path object instead of a string. :issue:`405`
--   TypeError is raised when parameter with ``multiple=True`` or
+-   ``TypeError`` is raised when parameter with ``multiple=True`` or
     ``nargs > 1`` has non-iterable default. :issue:`1749`
 
 
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2015,7 +2015,7 @@ class Parameter:
                 iter_value = iter(value)
             except TypeError:
                 raise TypeError(
-                    "Default for parameter with multiple = True or nargs > 1"
+                    "Value for parameter with multiple = True or nargs > 1"
                     " should be an iterable."
                 )
 
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2011,7 +2011,15 @@ class Parameter:
             if level == 0:
                 return self.type(value, self, ctx)
 
-            return tuple(_convert(x, level - 1) for x in value)
+            try:
+                iter_value = iter(value)
+            except TypeError:
+                raise TypeError(
+                    "Default for parameter with multiple = True or nargs > 1"
+                    " should be an iterable."
+                )
+
+            return tuple(_convert(x, level - 1) for x in iter_value)
 
         return _convert(value, (self.nargs != 1) + bool(self.multiple))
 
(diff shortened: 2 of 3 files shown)


src/click/core.py
```diff
@@ -2011,7 +2011,15 @@ def _convert(value, level):
             if level == 0:
                 return self.type(value, self, ctx)
 
-            return tuple(_convert(x, level - 1) for x in value)
+            try:
+                iter_value = iter(value)
+            except TypeError:
+                raise TypeError(
+                    "Default for parameter with multiple = True or nargs > 1"
```
davidism, line 2018: This method is called for all values, not only default values.

Saif807380: I'll change the error message

CHANGES.rst
```diff
@@ -182,6 +182,8 @@ Unreleased
     previous item. :issue:`1353`
 -   The ``Path`` param type can be passed ``path_type=pathlib.Path`` to
     return a path object instead of a string. :issue:`405`
+-   TypeError is raised when parameter with ``multiple=True`` or
```
davidism, line 185: `TypeError` is a class name, it should be code formatted with double backticks.

Saif807380: Missed that one. Will fix it.

effigies:
Ah, I wish I'd followed this. Would it be possible to identify *which* argument generated the `TypeError`? It looks like there's a `self.name` that could be used.

davidism:
Both the errors raised in `type_cast_value` should be in `__init__` instead. The composite error should be removed, and both places that `value` is iterated in `type_cast_value` should raise a `BadParameter` error instead of `TypeError`.

PR #1807, opened 2021-03-05
add pass_meta_key decorator

### LINKED ISSUE #1739: @pass_meta_key decorator
### Idea Proposal:

As a click's plugin developer, we use the `click.Context.meta` dictionary to store any state. So IMHO, click could provide another decorator: `@click.pass_metavar(key="foo")` similar to `@click.pass_obj`.

**Example implementation:**

```python
def pass_metavar(key: str):
    """Similar to :func:`click.pass_obj`,
    this passes the ctx.meta[key] variable instead as the first argument.
    """

    def decorator(f):
            @click.pass_context
            def new_func(ctx: click.Context, *args, **kwargs):
                var = ctx.meta[key]
                return ctx.invoke(f, var, *args, **kwargs)

        return update_wrapper(new_func, f)
    
    return decorator
```

**Example usage:**

```python
@click.pass_metavar(key="foo")
def bar(var):
    click.echo(var)
```

I think this would be beneficial for click's plugin developers and naturally anyone building a complex CLI app which requires using the `click.Context.meta` to store state. 

If you like the idea, I will be glad to raise a PR for it asap.

davidism:
Extensions using `Context.meta` can use `pass_meta_key` to create decorators for passing their data to callbacks. The `doc_description` parameter can be used to add a specific description of the object, so that the decorator's docs can be included in the extension's docs.

```python
# in the extension
from click.decorators import pass_meta_key
pass_auth = pass_meta_key("click-auth.auth", doc_description="the :class:`ClickAuth` object")
```

```
# in the extension's docs
pass_auth.__doc__
Decorator that passes the :class:`ClickAuth` object as the first argument to the decorated function.
```

```python
# in the user code
@pass_auth
def list_prs(auth):
    ...
```

Also changed docs config so typing information is shown next to parameter descriptions instead of in the signature. https://pallets-click--1807.org.readthedocs.build/en/1807/api/#click.decorators.pass_meta_key

- fixes #1739

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -184,6 +184,9 @@ Unreleased
     return a path object instead of a string. :issue:`405`
 -   ``TypeError`` is raised when parameter with ``multiple=True`` or
     ``nargs > 1`` has non-iterable default. :issue:`1749`
+-   Add a ``pass_meta_key`` decorator for passing a key from
+    ``Context.meta``. This is useful for extensions using ``meta`` to
+    store information. :issue:`1739`
 
 
 Version 7.1.2
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -31,6 +31,9 @@ Decorators
 
 .. autofunction:: make_pass_decorator
 
+.. autofunction:: click.decorators.pass_meta_key
+
+
 Utilities
 ---------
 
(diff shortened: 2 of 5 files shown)


Eshaan7:
Really glad that you liked my idea. :D

PR #1809, opened 2021-03-06
Default value iterable checks are done in __init__()

### LINKED ISSUE #1806: Move iterable checks from type_cast_value to __init__
Both the errors raised in `type_cast_value` should be in `__init__` instead. The composite error should be removed from `type_cast_value`, it makes no sense at runtime. Both places that `value` is iterated in `type_cast_value` should raise a `BadParameter` error instead of `TypeError`.

_Originally posted by @davidism in https://github.com/pallets/click/issues/1805#issuecomment-791522850_

Saif807380:
I have moved both checks to `Parameter.__init__()`. I've added a new parameter `multiple` in the `__init__` method of `Parameter`, which can be passed by both `Option` and `Argument` via `super.__init__()`, so that the checks can be performed there for both types of parameters. The check raises `BadParameter` with the name of the parameter that has the bad default. I've also updated the test.

fixes #1806 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -187,6 +187,11 @@ Unreleased
 -   Add a ``pass_meta_key`` decorator for passing a key from
     ``Context.meta``. This is useful for extensions using ``meta`` to
     store information. :issue:`1739`
+-   Param ``multiple`` is moved from ``Option`` to ``Parameter`` to support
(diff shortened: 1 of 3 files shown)


davidism:
Closing as there is more complexity than I originally anticipated. I'm working on a refactor.

PR #1815, opened 2021-03-17
Fix: Deprecation notice now appears at the top of the help text

### LINKED ISSUE #1791: Deprecated notice in help text can be hard to see
The current `(DEPRECATED)` deprecation notice added to help text can be hard to see plainly since it can't consider the existing format of the help text.

For example:  I use sphinx style formatting in my help text and if this help text ends with a Note block like below:

```python
def encrypt(ctx, **kwargs):
    """Encrypt archives.

    .. Note::

        Here is a notice.
    """
```

Then a deprecation notice would append to the Note block, making it look like part of the Note and not that the command itself is deprecated.

```shell
Usage: cmd encrypt [OPTIONS]

  Encrypt archives.

  .. Note::

      Here is a notice. (DEPRECATED)
```

I would have expected the deprecation notice to be at the top of my help text making it clear to the user that this command is deprecated.  Or at the least a couple a new lines added to differentiate the notice from any existing help text.

Environment:

- Python version: Python 3.7.9
- Click version: click==7.1.2

BALaka-18:
This PR aims to fix the Issue of the deprecation notice getting printed alongside the help text. This PR makes the deprecation notice appear at the top of the help text.

- fixes #1791

Checklist:

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1224,7 +1224,7 @@ class Command(BaseCommand):
             with formatter.indentation():
                 help_text = self.help
                 if self.deprecated:
-                    help_text += DEPRECATED_HELP_NOTICE
+                    help_text = DEPRECATED_HELP_NOTICE + "\n" + help_text
                 formatter.write_text(help_text)
         elif self.deprecated:
             formatter.write_paragraph()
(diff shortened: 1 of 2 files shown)


src/click/core.py
```diff
@@ -2011,15 +2011,7 @@ def _convert(value, level):
...
-            try:
-                iter_value = iter(value)
-            except TypeError:
-                raise TypeError(
-                    "Value for parameter with multiple = True or nargs > 1"
-                    " should be an iterable."
-                )
-
-            return tuple(_convert(x, level - 1) for x in iter_value)
+            return tuple(_convert(x, level - 1) for x in value)
```
davidism, line 2014: Looks like this change isn't related.

BALaka-18: No, I think I went wrong somewhere while fetching the files, I have to see.

PR #1816, opened 2021-03-17
Deprecation notice now appears at the top of the help text

### LINKED ISSUE #1791: Deprecated notice in help text can be hard to see

BALaka-18:
This PR makes the deprecation notice appear at the beginning of the help text.

- fixes #1791

Checklist:

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -40,14 +40,6 @@ _missing = object()
 SUBCOMMAND_METAVAR = "COMMAND [ARGS]..."
 SUBCOMMANDS_METAVAR = "COMMAND1 [ARGS]... [COMMAND2 [ARGS]...]..."
 
-DEPRECATED_HELP_NOTICE = " (DEPRECATED)"
-DEPRECATED_INVOKE_NOTICE = "DeprecationWarning: The command {name} is deprecated."
-
-
-def _maybe_show_deprecated_notice(cmd):
-    if cmd.deprecated:
-        echo(style(DEPRECATED_INVOKE_NOTICE.format(name=cmd.name), fg="red"), err=True)
-
 
 def _fast_exit(code):
     """Low-level exit that skips Python's cleanup but speeds up exit by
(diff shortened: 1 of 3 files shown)


src/click/core.py
```diff
@@ -1224,7 +1224,7 @@ def format_help_text(self, ctx, formatter):
             with formatter.indentation():
                 help_text = self.help
                 if self.deprecated:
-                    help_text += DEPRECATED_HELP_NOTICE
+                    help_text = DEPRECATED_HELP_NOTICE + "\n" + help_text
```
davidism, line 1227: After thinking about the formatting, I think putting the notice at the front of the line instead of on its own line looks better. Also, the current notice has a space at the front, which can be removed since it's being prepended instead of appended now.

davidism:
This should update `get_short_help_str` as well.

src/click/core.py
```diff
@@ -1193,8 +1193,10 @@ def get_short_help_str(self, limit=45):
         """Gets short help for the command or makes it by shortening the
         long help string.
         """
+        short_help_text = DEPRECATED_HELP_NOTICE + self.short_help
```
davidism, line 1196: Should only be appended if the command is deprecated.

BALaka-18: Yes I overlooked that, updated the code.

davidism:
Ended up doing some more refactoring to avoid some repetition. Added the deprecation notice when generating short help from help as well. Inlined the deprecation messages and made the notice title case.

PR #1820, opened 2021-03-22
Fix issues with CliRunner's echo_stdin

### LINKED ISSUE #1101: CliRunner prompt input echos twice on Python 2
```python
import click

@click.command()
def cli():
    prompt("value")
    click.echo("done")

runner = CliRunner(echo_stdin=True)
result = runner.invoke(cli, input="test")
print(result.output)
```

Python 3 output (correct):

```
value: test
done
```

Python 2 output:

```
value: testtest
done
```

I went back and this appears to be the case in every version since 1.x. The issue seems to be with the `visible_input` function in `CliRunner.isolation`. It uses `input` after that has been wrapped in `EchoingStdin`. It appears to work in Python 3 due to some weird lookup behavior, but that appears to be an accident.  There are no tests for `echo_stdin` and `prompt` at the same time.

It didn't show up in the docs because the docs runner overrode `isolation` with different code that had `visible_input` pull from the original input and only replaced `sys.stdin` with the EchoingStdin wrapper, rather than both with the wrapper. While porting the docs runner to pallets-sphinx-themes, I removed that code in favor of simpler subclassing, and noticed that the docs had double input, which led me to discover the full issue.

saebischer:
Fixes #1101 and some related issues with `CliRunner(echo_stdin=True)`.

As raised in #1101, with Python 2 any prompted input appears twice in the output recorded by `CliRunner` when using `echo_stdin=True`. It appears to work with Python 3 because `EchoingStdin` does not always echo. For example, using `input` instead of `click.prompt`:

```python
@click.command()
def test():
    foo = input("Foo: ")
    click.echo("foo={}".format(foo))

runner = CliRunner(echo_stdin=True)
result = runner.invoke(test, input="bar\n")
print(result.output)
```

**Expected output:**

    Foo: bar
    foo=bar

**Actual output:**

    Foo: foo=bar

**Environment:** Click 7.1.2, Python 3.7.7

Here the problem was that `TextIOWrapper` tries to call `read1` on the underlying input stream, which wasn't being overridden in `EchoingStdin`. After adding an echoing `read1`, both Python 2 and Python 3 double-echo with `click.prompt`.

I've modified `visible_input` not to echo when `echo_stdin=True`. To get consistent behaviour for hidden prompts (which should not echo even if `echo_stdin=True`) I also added a way to temporarily turn off echoing inside `hidden_input`. This is also used in `_getchar` if it gets passed `echo=False`.

Finally, there was one more subtle problem in Python 3. `TextIOWrapper` provides a buffered stream that reads a large chunk from the input before parsing out individual lines. When the underlying input stream is `EchoingStdin`, this causes all input (up to the chunk size) to be echoed at once. Here's another example:

```python
@click.command()
def test():
    click.prompt("a")
    click.prompt("b")

runner = CliRunner(echo_stdin=True)
result = runner.invoke(test, input="foo\nbar\n")
print(result.output)
```

**Expected output:**

    a: foo
    b: bar

**Actual output:**

    a: foo
    bar
    b: 

**Environment**: commit `21cde2f` (after the previously mentioned issues are fixed), Python 3.7.7

The only solution I found for this was to set the `_CHUNK_SIZE` attribute of `TextIOWrapper`, which is not documented (I found it [here](https://github.com/python/cpython/blob/7cb033c423b65def1632d6c3c747111543b342a2/Lib/_pyio.py#L1995)) and I guess that means it might not be compatible with all Python implementations. It would be great if someone knows a cleaner approach, otherwise I hope it's okay to contribute something like this.

---

Checklist:

- [x] Add or update relevant docs, in the docs folder and in code. (n/a)
- [x] Add `.. versionchanged::` entries in any relevant code docs. (n/a)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -208,6 +208,8 @@ Unreleased
     :issue:`1568`
 -   The ``MultiCommand.resultcallback`` decorator is renamed to
     ``result_callback``. The old name is deprecated. :issue:`1160`
+-   Fix issues with ``CliRunner`` output when using ``echo_stdin=True``.
+    :issue:`1101`
 
 
 Version 7.1.2
--- a/src/click/testing.py
+++ b/src/click/testing.py
@@ -16,16 +16,21 @@ class EchoingStdin:
     def __init__(self, input, output):
         self._input = input
         self._output = output
+        self.paused = False
 
     def __getattr__(self, x):
         return getattr(self._input, x)
 
     def _echo(self, rv):
-        self._output.write(rv)
+        if not self.paused:
+            self._output.write(rv)
         return rv
 
     def read(self, n=-1):
         return self._echo(self._input.read(n))
+
+    def read1(self, n=-1):
+        return self._echo(self._input.read1(n))
 
     def readline(self, n=-1):
         return self._echo(self._input.readline(n))
@@ -204,10 +209,16 @@ class CliRunner:
 
         if self.echo_stdin:
             input = EchoingStdin(input, bytes_output)
+            echo_input = input
 
         sys.stdin = input = _NamedTextIOWrapper(
             input, encoding=self.charset, name="<stdin>", mode="r"
         )
+        if self.echo_stdin:
+            # Force unbuffered reads, otherwise the underlying EchoingStdin
+            # stream will echo a big chunk of input on the first read.
+            input._CHUNK_SIZE = 1
+
         sys.stdout = _NamedTextIOWrapper(
             bytes_output, encoding=self.charset, name="<stdout>", mode="w"
         )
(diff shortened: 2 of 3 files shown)


saebischer:
```
src/click/utils.py:129:9: B014 Redundant exception types in `except (IOError, OSError) as e:`.  Write `except OSError as e:`, which catches exactly the same exceptions.
```

Not sure about this one as I haven't made any changes to `src/click/utils.py`. I can see on `master` that it has been changed to just `OSError` as suggested. Should I do the same here even though it's unrelated to the main issue? New to this so guidance appreciated!

davidism:
You should target this pr at master, not 7.x.

saebischer:
Thanks @davidism, I've updated the base branch. I thought since this is a bugfix the contributing guide said to target the latest maintenance branch.

davidism:
You need to rebase, in addition to changing it in github.

davidism:
#1836 changed how the prompt was echoed to work around a readline issue. Splitting echoing the prompt from the space after it seems to mess up the tests you wrote. I changed to unconditionally disabling `echo_stdin` behavior during the `visible_input`, `hidden_input`, and `_getchar` functions, and this seems to have fixed it. Used a context decorator instead of managing the flag manually.

PR #1825, opened 2021-03-29
Fix: resolve_path logic is now cross-platform compatible

### LINKED ISSUE #1813: resolve_path differs on Windows depending on Python version
Hi. There was an issue filed under [Typer](https://github.com/tiangolo/typer/issues/244#issuecomment-792455309) that gives a full explanation, but basically the use of `os.path.realpath` in the `resolve_path` logic for click.Path differs between Python 3.7 and 3.8 on Windows. Prior to 3.8, `os.path.realpath` did not resolve symlinks. Therefore Click users on Windows using Python 3.7 or lower are getting the wrong results for resolve_path.

More info: https://docs.python.org/3/library/os.path.html#os.path.realpath

BALaka-18:
The resolve_path logic, as mentioned in the Issue, is now cross-platform compatible.

- fixes #1813 

Checklist:

### DIFF
--- a/src/click/types.py
+++ b/src/click/types.py
@@ -734,6 +734,10 @@ class Path(ParamType):
 
         if not is_dash:
             if self.resolve_path:
+                # realpath on Windows Python < 3.8 doesn't resolve symlinks
+                if os.path.islink(rv):
+                    rv = os.readlink(rv)
+
                 rv = os.path.realpath(rv)
 
             try:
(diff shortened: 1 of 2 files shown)


src/click/types.py
```diff
@@ -733,8 +733,8 @@ def convert(self, value, param, ctx):
         is_dash = self.file_okay and self.allow_dash and rv in (b"-", "-")
 
         if not is_dash:
-            if self.resolve_path:
-                rv = os.path.realpath(rv)
+            if self.resolve_path and os.path.islink(rv):
```
davidism, line 736: `islink` and `readlink` needs to be a separate step within the `if` block before `realpath`.

PR #1840, opened 2021-04-09
Fixed inconsistency in forwarding params in Context.forward

### LINKED ISSUE #1568: context.forward inconsistently removes undeclared options
### Expected Behavior

Three commands. The first takes a subset of the options that the last two take. Each prints the arguments they were called with. All but the first call the one before it via `context.forward`.

```python
import click
from pprint import pprint

@click.group()
def main():
    pass

@main.command()
@click.option('-a')
def first(**kwargs):
    print('first')
    pprint(kwargs)

@main.command()
@click.option('-a')
@click.option('-b')
@click.pass_context
def second(context, **kwargs):
    print('second')
    pprint(kwargs)
    context.forward(first)

@main.command()
@click.option('-a')
@click.option('-b')
@click.pass_context
def third(context, **kwargs):
    print('third')
    pprint(kwargs)
    context.forward(second)

main()
```

Here's the behavior I expect from each command. Notably, `first` is called with the same arguments, its declared options, each time.

```
$ python main.py first
first
{'a': None}
$ python main.py second
second
{'a': None, 'b': None}
first
{'a': None}
$ python main.py third
third
{'a': None, 'b': None}
second
{'a': None, 'b': None}
first
{'a': None}
```

### Actual Behavior

```
$ python main.py first
first
{'a': None}
$ python main.py second
second
{'a': None, 'b': None}
first
{'a': None, 'b': None}     # 'first' is called with 'b' here
$ python main.py third
third
{'a': None, 'b': None}
second
{'a': None, 'b': None}
first
{'a': None}                # but not here
```

### Environment

* Python version: 3.7.5
* Click version: 7.1.2

Saif807380:
`Context.forward` uses `Context.params` to fill in the options and passes the `kwargs` to `Context.invoke` which creates a new `Context` for the sub-command but doesn't pass the params to the new `Context` nor does it remove unused `kwargs`.

From the example mentioned in the [issue](https://github.com/pallets/click/issues/1568), now a union of params from the context and sub-context will be forwarded. 

For e.g. - If `command_1` has options `a and b` and `command_2` has options `a and c`, command_1 forwards to command_2, then `command_2` will receive `a, b and c` as `kwargs`. 

I was unsure whether or not to remove the unused options, so I've kept them for now. 

- fixes #1568

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -203,6 +203,9 @@ Unreleased
     does not raise a ``UnicodeEncodeError``. :issue:`848`
 -   Fix an issue where ``readline`` would clear the entire ``prompt()``
     line instead of only the input when pressing backspace. :issue:`665`
+-   Add all kwargs passed to ``Context.invoke()`` to ``ctx.params``.
+    Fixes an inconsistency when nesting ``Context.forward()`` calls.
+    :issue:`1568`
 
 
 Version 7.1.2
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -678,6 +678,10 @@ class Context:
         in against the intention of this code and no context was created.  For
         more information about this change and why it was done in a bugfix
         release see :ref:`upgrade-to-3.2`.
+
+        .. versionchanged:: 8.0
+            All ``kwargs`` are tracked in :attr:`params` so they will be
+            passed if :meth:`forward` is called at multiple levels.
         """
         self, callback = args[:2]
         ctx = self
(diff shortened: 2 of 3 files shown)


davidism:
It seems a little weird to treat all kwargs as params, since `params` is supposed to hold the parsed values from `@argument` and @option`. Another option is to separate into `ctx.params` and `ctx.kwargs` to track the decorated parameters vs everything the command was called with. I think it's fine as is for now, I'm not entirely sure the `kwargs` idea is any more "correct" in terms of the intentions of `invoke`, `forward`, and `params`.

PR #1843, opened 2021-04-11
Use `col_max` for `format_commands`

### LINKED ISSUE #1842: Use `col_max` for `format_commands`
Actually use `HelpFormatter.write_dl`'s `col_max` when generating usage. I have a lot of commands with long names, and I don't want the description of these to come on the following line, when simply adding some characters to the column with of the command name would make everything look very nice.

Before would be like:
```
short-command                      Docstring for short command
very-very-enormously-long-command-name
                                   Docstring for long command
```

And after bumping `col_max` a bit above 30:
```
short-command                           Docstring for short command
very-very-enormously-long-command-name  Docstring for long command
```

I am opening a pull request for this as well.

MarcelRobitaille:
- fixes #1842

Checklist:

I don't think all parameters for `Context` are documented (such as `terminal_width`, and this is more niche than that).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -203,6 +203,10 @@ Unreleased
     does not raise a ``UnicodeEncodeError``. :issue:`848`
 -   Fix an issue where ``readline`` would clear the entire ``prompt()``
     line instead of only the input when pressing backspace. :issue:`665`
+-   Added the ``col_max`` parameter to ``Context`` which gets passed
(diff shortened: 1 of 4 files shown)


MarcelRobitaille:
There are probably some other functions, not just `format_commands`, that should pipe this along too.

davidism:
Thanks for working on this, however it's better served by solving #561. For now with #938, you can write a custom formatter subclass and set `Context.formatter_class` to use it.

PR #1845, opened 2021-04-11
Organized test_basic.py (partial)

### LINKED ISSUE #1828: Organize tests
This is a meta issue to describe ongoing attention that should be devoted to tests. I don't want anyone to do this in a single commit or even a single PR, it's something that anyone can pick up one file, or even one part of a file, at a time. The later tasks about reorganizing will be easier once the earlier tasks are done, but will still require a good amount of familiarity and planning.

The tests in Pallets projects slowly evolved from using unittest to using pytest, and were also contributed by many users over time, so the style and organization are often inconsistent. The goal is to eventually have all tests organized into smaller, specific files to make it easier for contributors to identify where to add a new test.

- [ ] Replace for loops with `pytest.mark.parametrize`.
- [ ] Replace repeated groups of asserts in one test with `parametrize`.
- [ ] Refactor separate, similar test functions into a single parametrized function. Don't go overboard, sometimes it's easier to read when functions are kept separate.
- [ ] Remove class-based tests. If the class was used to group tests for a distinct function, consider moving all the tests to a separate file.
- [ ] Make sure test names follow the pattern `test_{function}_{behavior}`.
- [ ] Split up large test files into multiple, more specific files.
- [ ] Rearrange tests to group similar functions / behaviors together or move into specific files.
- [ ] Run `coverage run -m pytest` to identify missing coverage and write tests (or add parametrize cases to existing tests).

Anyone can help contribute to this effort. Pick a file or function to work on, and make a PR for it. Again, don't do all bullet points in a single commit, and don't do too much in one PR, otherwise it will be harder to review.

BALaka-18:
Organized ```test_basic.py```, awaiting feedback to be able to continue further.

- fixes #1828 

Checklist:

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1,5 +1,4 @@
 import os
-import uuid
 from itertools import chain
 
 import pytest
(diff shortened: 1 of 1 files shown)


tests/test_basic.py
```diff
@@ -104,92 +104,44 @@ def sub():
...
+    [
+        ("no value", "FOO:[no value]"),
+        ("42", "FOO:[42]"),
+        (None, "Option '--foo' requires an argument."),
+        ("", "FOO:[]"),
+        ("\N{SNOWMAN}", "FOO:[\N{SNOWMAN}]"),
+    ],
+)
+def test_basic_option(runner, default, expect):
+    param = click.Option(["--foo"], default=default)
```
davidism, line 118: This doesn't match the test it's replacing. This is testing default values, while the original was testing passed arguments.

tests/test_basic.py
```diff
@@ -104,92 +104,44 @@ def sub():
...
-    result = runner.invoke(cli, ["--foo=23"])
-    assert not result.exception
-    assert "FOO:[46]" in result.output
-
-    result = runner.invoke(cli, ["--foo=bar"])
-    assert result.exception
-    assert "Invalid value for '--foo': 'bar' is not a valid integer." in result.output
-
-
-def test_uuid_option(runner):
```
davidism, line 153: This shouldn't be completely removed.

tests/test_basic.py
```diff
@@ -104,92 +104,44 @@ def sub():
...
+)
+def test_basic_option(runner, default, expect):
+    param = click.Option(["--foo"], default=default)
 
-    result = runner.invoke(cli, ["--u=bar"])
-    assert result.exception
-    assert "Invalid value for '--u': 'bar' is not a valid UUID." in result.output
+    cli = click.Command("cli", params=[param], callback=lambda foo: foo)
+    result = runner.invoke(cli, standalone_mode=False)
+    # result = runner.invoke(cli, [])
```
davidism, line 122: Remove comment.

tests/test_basic.py
```diff
@@ -104,92 +104,44 @@ def sub():
...
 
+    assert result.exception is None
+    assert result.return_value == expect
 
-def test_float_option(runner):
```
davidism, line 175: This shouldn't be completely removed.

tests/test_basic.py
```diff
@@ -104,50 +104,46 @@ def sub():
...
-    result = runner.invoke(cli, ["--foo="])
-    assert not result.exception
-    assert "FOO:[]" in result.output
+    result = runner.invoke(cli, [values])
 
-    result = runner.invoke(cli, ["--foo=\N{SNOWMAN}"])
-    assert not result.exception
-    assert "FOO:[\N{SNOWMAN}]" in result.output
+    assert result.exception is None
+    assert result.return_value == expect
```
davidism, line 126: `return_value` is only set when running in `standalone_mode=False`, you can test `result.output` in this case.

PR #1864, opened 2021-05-03
fix: allow 'autocompletion' function to return any kind of iterable 

muggenhor:
This is related to #1856 (0103c9570650daa59560baf42ad0a27e57b3157f) and addresses the same issue as in python/typeshed#5265.

Specifically the type annotation introduced there forces the `autocompletion` function to have `List[str]` as return type even though the implementation never required that. It only needs `Iterable[str]`. This causes some [functional autocompletion functions](https://github.com/tomtom-international/hopic/blob/release/1/hopic/cli/autocomplete.py#L91-L100) to get rejected by `mypy` even though there's (ostensibly) no reason for that.

I'm not sure whether an issue needs to be created for this as this problem isn't present in any released code (yet).

Checklist:

no relevant docs:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -101,3 +101,6 @@ ignore_missing_imports = True
 
 [mypy-importlib_metadata.*]
 ignore_missing_imports = True
+
+[mypy-pytest.*]
+ignore_missing_imports = True
--- a/setup.cfg
+++ b/setup.cfg
@@ -76,7 +76,7 @@ per-file-ignores =
     src/click/__init__.py: F401
 
 [mypy]
-files = src/click
+files = src/click, tests/test_shell_completion.py
 python_version = 3.6
 disallow_subclassing_any = True
 disallow_untyped_calls = True
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2027,7 +2027,7 @@ class Parameter:
         ] = None,
         autocompletion: t.Optional[
             t.Callable[
-                [Context, t.List[str], str], t.List[t.Union[t.Tuple[str, str], str]]
+                [Context, t.List[str], str], t.Iterable[t.Union[t.Tuple[str, str], str]]
             ]
         ] = None,
     ) -> None:
--- a/tests/test_shell_completion.py
+++ b/tests/test_shell_completion.py
@@ -1,11 +1,16 @@
+# mypy: disallow_untyped_defs = False
+# mypy: disallow_untyped_calls = False
 import sys
+import typing as t
 
 import pytest
 
 from click.core import Argument
 from click.core import Command
+from click.core import Context
 from click.core import Group
 from click.core import Option
+from click.core import Parameter
 from click.shell_completion import CompletionItem
 from click.shell_completion import ShellComplete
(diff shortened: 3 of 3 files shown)


davidism:
Please fix the annotation to match the code.

setup.cfg
```diff
@@ -76,7 +76,7 @@ per-file-ignores =
     src/click/__init__.py: F401
 
 [mypy]
-files = src/click
+files = src/click, tests/test_shell_completion.py
```
davidism, line 79: We're not checking tests.

setup.cfg
```diff
@@ -101,3 +101,6 @@ ignore_missing_imports = True
 
 [mypy-importlib_metadata.*]
 ignore_missing_imports = True
+
+[mypy-pytest.*]
+ignore_missing_imports = True
```
davidism, line 106: If we were checking tests (we're not), you would add pytest in requirements/typing.in, not ignore it.

src/click/core.py
```diff
@@ -2027,7 +2027,7 @@ def __init__(
         ] = None,
         autocompletion: t.Optional[
             t.Callable[
-                [Context, t.List[str], str], t.List[t.Union[t.Tuple[str, str], str]]
+                [Context, t.List[str], str], t.Iterable[t.Union[t.Tuple[str, str], str]]
```
davidism, line 2030: Don't change this from `List` to `Iterable`, I was deliberate about returning `List`.

tests/test_shell_completion.py
```diff
@@ -158,11 +163,20 @@ def custom(ctx, param, incomplete):
...
         assert isinstance(args, list)
-        return [("art", "x"), "bat", "cat"]
+        yield "art", "x"
+        yield "bat"
+        yield "cat"
 
     with pytest.deprecated_call():
+        if t.TYPE_CHECKING:
+            # Force type checking against base class' constructor signature
+            Parameter(autocompletion=custom)
```
davidism, line 179: The base class isn't instantiated, this doesn't need to be tested (and we're not checking tests either). Non-matching signatures should be picked up by mypy without specific tests.

davidism (changes requested):

davidism:
After reviewing, going to close this in favor of a new PR, since basically everything would change.

setup.cfg
muggenhor: that's the only way for me to achieve this part of the contribution checklist: "Add tests that demonstrate the correct behavior of the change. Tests should fail without the change"

src/click/core.py
muggenhor: What's the reason for requiring user code to adhere to the more strict `List` type with their return value instead of `Iterable`?

For the purpose of type checking this is (seemingly) not backwards compatible as code that previously passed now starts being flagged. (generator function used as `autocompletion` function). I understand that this is a new major release, so breaking backwards compatibility is an option. But simultaneously the entire purpose of this intermediate shape of `autocompletion` seems to be a migration path to allow people to move away from that towards `shell_complete`. So this confuses me...

PR #1870, opened 2021-05-04
repr is erasing ANSI escapes codes

### LINKED ISSUE #1869: Allow ANSI escapes codes in metavar hints
Hello,

little quirk when printing metavar hint : 
`
Error: Missing argument '\x1b[95mREGION\x1b[0m'.
`

region argument was declared like this : 
`
region_argument = click.argument('region', metavar=click.style('REGION', fg='bright_magenta'))
`

It should be printed with magenta color : 

`
Error: Missing argument 'REGION'.
`

Environment:

- Python version: 3.6
- Click version: > 8.0 / master

AdrienPensart:
This change will allow colors in argument hint.

What about str() vs f"{}" ?

Should I update : return " / ".join(repr(x) for x in hint_list) too ?

- fixes #1869

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2372,7 +2372,7 @@ class Parameter:
         indicate which param caused the error.
         """
         hint_list = self.opts or [self.human_readable_name]
-        return " / ".join(repr(x) for x in hint_list)
+        return " / ".join(f"'{x}'" for x in hint_list)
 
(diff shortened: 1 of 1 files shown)


davidism:
Thanks for catching this, I fixed the other repr you pointed out. What's happening is that repr is escaping the non-printable control characters so that they're displayed as their string repr instead of controlling the terminal. repr was used as a shortcut for putting single quotes around things, so this might be something to change in other places as well.

PR #1872, opened 2021-05-07
Fix HelpFormatter.write_text() to use the full line width

### LINKED ISSUE #1871: `HelpFormatter.write_text()` is not using all the available line width
`HelpFormatter.write_text()` uses the function `wrap_text(text, width, initial_indent, ...)` internally. This function expects `width` to be the line width **including** the eventual indentation. `HelpFormatter.write_text()` gets this wrong and passes `self.width - self.current_indent` instead of just `self.width`.

janluke:
Fixes #1871.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -215,6 +215,8 @@ Unreleased
 -   When defining a parameter, ``default`` is validated with
     ``multiple`` and ``nargs``. More validation is done for values being
     processed as well. :issue:`1806`
+-   ``HelpFormatter.write_text`` uses the full line width when wrapping
+    text. :issue:`1871`
 
 
 Version 7.1.2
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -215,7 +215,8 @@ Unreleased
 -   When defining a parameter, ``default`` is validated with
     ``multiple`` and ``nargs``. More validation is done for values being
     processed as well. :issue:`1806`
-
+-   Fix :meth:`HelpFormatter.write_text`, which didn't use the full line
+    width. :issue:`1871`
```
davidism, line 220: Two spaces before headers is intentional formatting, please keep it as-is.

tests/test_formatting.py
```diff
@@ -335,3 +336,14 @@ def test_formatting_with_options_metavar_empty(runner):
     cli = click.Command("cli", options_metavar="", params=[click.Argument(["var"])])
     result = runner.invoke(cli, ["--help"])
     assert "Usage: cli VAR\n" in result.output
+
+
+def test_help_formatter_write_text():
+    # Failed because of https://github.com/pallets/click/issues/1871
```
davidism, line 342: In general we don't leave comments pointing to issues. No need for a comment here, but if you want to write one just describe what's being tested succinctly and directly.

janLuke: I removed the comment.

PR #1896, opened 2021-05-14
Add support for optional decorator parentheses

### LINKED ISSUE #1359: raise a more descriptive exception when @command decorator isn't called
Click 7.0, Python 3.7.4

Having done `pip install -e .` on my package to set up the CLI command, this error is raised because I decorated the main CLI function with `click.command` instead of `click.command()`:

```
  File "/Users/zev/.local/share/virtualenvs/myvirtualenvname-CI6hzx9c/bin/capture", line 11, in <module>
    load_entry_point('myvirtualenvname', 'console_scripts', 'my_module_name')()
TypeError: decorator() missing 1 required positional argument: 'f'
```

Would it be possible to raise a more informative exception referring to the decorator not having been called?

angus-lherrou:
- fixes #1359

Checklist:

### DIFF
--- a/src/click/decorators.py
+++ b/src/click/decorators.py
@@ -153,11 +153,29 @@ def _make_command(
     )
 
 
+@t.overload
 def command(
     name: t.Optional[str] = None,
     cls: t.Optional[t.Type[Command]] = None,
     **attrs: t.Any,
 ) -> t.Callable[[F], Command]:
+    ...
+
+
+@t.overload
+def command(
+    name: t.Callable,
+    cls: t.Optional[t.Type[Command]] = None,
+    **attrs: t.Any,
+) -> Command:
+    ...
+
+
+def command(
+    name: t.Union[str, t.Callable, None] = None,
+    cls: t.Optional[t.Type[Command]] = None,
+    **attrs: t.Any,
+) -> t.Union[Command, t.Callable[[F], Command]]:
     r"""Creates a new :class:`Command` and uses the decorated function as
     callback.  This will also automatically attach all decorated
     :func:`option`\s and :func:`argument`\s as parameters to the command.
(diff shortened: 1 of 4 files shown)


angus-lherrou:
Maintainers, I'm not sure if this change needs any docs updates?

ThiefMaster:
Wouldn't it be better if the decorators were smart enough to work even without parentheses, ie if there's exactly one positional argument and that argument is a callable, assume it's the decorated function...

davidism:
I go back and forth on that pattern. Right now I'm on the side of "there should be one way to do things". Even if we can detect it, better to tell everyone to do something consistently. I'd have to look at the mypy docs, but I have a feeling the "decorator or factory" pattern makes typing more complicated too.

ThiefMaster:
I know pytest does it for example, but no idea if they use type hints..

ThiefMaster:
https://github.com/pytest-dev/pytest/blob/main/src/_pytest/fixtures.py#L1239-L1272

davidism:
@angus-lherrou if you're up for it, want to implement this dual option instead of showing an error? If not, we can merge this then make another issue about supporting plain decorators.

angus-lherrou:
My first thought was actually the same, but I spent some time trying to implement the optional parentheses and it just wasn't working the way I was hoping it would. I can take a look at how pytest implements it and take another crack at it.

angus-lherrou:
@davidism I figured it out! When I tried before I didn't realize that I needed to _call_ the inner function before returning it if the decorator isn't called at the invocation site.

angus-lherrou:
I suppose it would be good to add this new capability in the documentation somewhere

src/click/decorators.py
```diff
@@ -153,11 +153,29 @@ def _make_command(
...
 def command(
     name: t.Optional[str] = None,
     cls: t.Optional[t.Type[Command]] = None,
     **attrs: t.Any,
 ) -> t.Callable[[F], Command]:
+    ...
+
+
+@t.overload
+def command(  # noqa: F811
```
ThiefMaster, line 166: I'm surprised the `noqa: F811` is needed - https://github.com/PyCQA/pyflakes/pull/435 looks like flake8 should be able to detect `@overload` usage and not trigger this error..

angus-lherrou: @ThiefMaster ah, I just checked in on this and it looks like flake8 doesn't detect cases like here where `@overload` is accessed like:
```python
import typing as t
@t.overload
...
```

Whether we want to do away with the aliased import of `typing` here or just leave the `noqa` is up to y'all. Separately, I'm going to try to fix this for flake8.

ThiefMaster: I'd keep the `noqa` and report it as a bug in flake8/pyflakes

davidism: That's awesome that you fixed the pyflakes issue, but the way I've been working around this for now is to do both:

```
import typing
import typing as t

t.Any
@typing.overload
```

angus-lherrou: That's definitely more portable, since you only need to do it once per file and not scatter `noqa` everywhere. I'll change it

src/click/decorators.py
```diff
@@ -153,11 +153,29 @@ def _make_command(
     )
 
 
+@t.overload
```
davidism, line 156: Add `import typing` and use `@typing.overload` for this specific case, I've been doing the same elsewhere around the pallets code.

angus-lherrou: Done in newest commit

PR #1900, opened 2021-05-15
fix argument parsing when generating help text

### LINKED ISSUE #1899: Arguments may break `--help`
Using `click.Path` for an argument may prevent the `--help` from showing up. A simple example is the following:

```
@click.command()
@click.option(
    "-f",
    type=click.Path(exists=True, file_okay=True),
    default="this does not exist",
    help="Output file name",
    show_default=True,
)
def cli():
    pass
```

When trying to print the `--help` the default argument will be checked for validity and instead print an error that `-f` is an invalid value.

Environment:

- Python version: Python 3.8.9
- Click version: 8.0.0

terencehonles:
Using `click.Path` for an argument may prevent the `--help` from showing up. A simple example is the following:

```
@click.command()
@click.option(
    "-f",
    type=click.Path(exists=True, file_okay=True),
    default="this does not exist",
    help="Output file name",
    show_default=True,
)
def cli():
    pass
```

When trying to print the `--help` the default argument will be checked for validity and instead print an error that `-f` is an invalid value. This change allows the default value to be displayed even if parsing of the value failed (but it will still try to parse in case it can better represent the default).

Environment:

- Python version: Python 3.8.9
- Click version: 8.0.0

- fixes #1899

Checklist:

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2219,7 +2219,15 @@ class Parameter:
 
             value = value()
 
-        return self.type_cast_value(ctx, value)
+        try:
+            # allow the type casting to fail if ``call`` is False which is used for
+            # generating help.
+            return self.type_cast_value(ctx, value)
+        except BadParameter:
+            if not call:
+                return value
+
+            raise
 
     def add_to_parser(self, parser: OptionParser, ctx: Context) -> None:
         raise NotImplementedError()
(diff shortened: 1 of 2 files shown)


davidism:
Thanks for working on this, however I don't think it's the correct fix, `call=False` is used for more than generating help, ignoring errors in that case has wider potential effects.

PR #1904, opened 2021-05-17
Add Python3 classifiers to setup.cfg

maxnoe:
This adds python 3 / python 3 only classifiers to the setup.cfg, which is e.g. used for searching packages on PyPI. It is only metadata, not used for any actual checks (this is what `python_requires` is for). 

Checklist:

Not relevant for this change


davidism:
I decided it wasn't worth keeping this list maintained, so now we support "Python".

maxnoe:
fair enough.

PR #1906, opened 2021-05-18
don't ignore value when nargs=-1 and envvar set

### LINKED ISSUE #1903: argument with nargs=-1 and envvar ignores values from command line
Since click 8.0, the value(s) provided on the command line for a nargs=-1 argument are ignored if the argument has `envvar` set.

Example:

```python
#!/usr/bin/env python
import click

@click.command()
@click.argument("foo", nargs=-1, envvar="FOO")
def cli(foo):
    click.echo(foo)

if __name__ == "__main__":
    cli()
```

```
$ ./clickbug arg1 arg2
()
```

When arguments are passed on the command line, the `foo` value remains empty. Removing the `envvar` option makes it work again. Values from the environment variable are picked up, though. Reverting to `click<8` also fixes the problem.

The cause of the bug is possibly [here](https://github.com/pallets/click/commit/3d3ea9c64420fd8978a41ce8b4db3ac2d245849c#diff-948f9e0cc733616a6b18aa152ac6703e819fec61dd752d1efa40e5736c3bdce7): The incoming `value` (which at this point still contains the correct arguments) is overwritten with `None` if `nargs` is -1 and the argument has an envvar set.

Environment:

- Python version: 3.8.5
- Click version: 8.0.0

mschmitzer:
A change in 3d3ea9c6 broke variadic arguments with an `envvar`
setting: the `process` method replaced `value` with `None`
uncoditionally, thus ignoring the values passed on the command line.
The intention of this was probably to replace the empty tuple (when no
arguments were passed) with `None` so that the value from the
environment would be picked later on. Adjust the condition accordingly
and add a test that covers this.

- fixes #1903

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,7 @@ Unreleased
 -   Annotate ``Context.obj`` as ``Any`` so type checking allows all
     operations on the arbitrary object. :issue:`1885`
 -   Fix some types that weren't available in Python 3.6.0. :issue:`1882`
+-   Fix arguments with ``nargs=-1`` and ``envvar`` set. :issue:`1903`
 
 
 Version 8.0.0
--- a/src/click/parser.py
+++ b/src/click/parser.py
@@ -233,7 +233,9 @@ class Argument:
                     )
                 )
 
-        if self.nargs == -1 and self.obj.envvar is not None:
+        if self.nargs == -1 and self.obj.envvar is not None and value == ():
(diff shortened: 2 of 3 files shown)


davidism:
It looks like you've made this PR from an org account instead of your user account. This means the "allow changes from maintainers" option doesn't work, so we can't prepare your PR for merge. Be sure to make PRs from your own account in the future. I'll resubmit your branch with cleanup in a new PR.

Issue #1096, opened 2018-08-24
Expand globs in arguments on Windows?


njsmith:
I'm using a variadic `Path` argument:

```python
@click.command()
@click.argument(
    "paths", nargs=-1, type=click.Path(exists=True, dir_okay=False)
)
def whatever(paths)
   ...
```

When I run `myprogram.py *.txt`, then on unix of course it works, because the shell expands on the glob. But on Windows, glob expansion is the responsibility of the program being invoked, and click doesn't appear to do it – even in a case like this where it has all the information needed to tell that glob expansion is appropriate. And click kind of has to do it, because click wants to validate the paths, and this has to happen *after* glob expansion – if I pass `*.txt`, it says `Error: Invalid value for "paths": Path "*.txt" does not exist.`, and my code doesn't even get invoked, so I can't even fix up the globs myself, except by disabling all of click's validation and then doing it all myself.

davidism:
* Should it be Click's (built-in) job to make up for cmd not having globs?
* How can we safely tell that a name is a glob?  There are other patterns besides `*`.
* What if the name comes from Bash and `*.txt` was actually the name of the file after expansion?
* Do we also want to expand `~`?
* If we add glob expansion to parameters, do we also need to do it for env vars on Unix too? For example, if a parameter has an env var `NAME="*.txt"`, no expansion is done, Click currently expects you to handle that.

My first reaction is that this shouldn't be done by Click, although it should be possible to subclass `Argument.process_value()` or `Path` to work with globs.

ThiefMaster:
> What if the name comes from Bash and *.txt was actually the name of the file after expansion?

IMHO such behavior would have to be completely disabled on non-Windows systems.

Maybe the least strange/error-prone solution would be to have a `windows_expand_globs` argument that can be set to True if you want glob expansion on Windows.

njsmith:
> Should it be Click's (built-in) job to make up for cmd not having globs?

The Windows convention that it is every program's job to make up for cmd not having globs, yes. "Annoying thing that every cli app has to do" seems like it falls within click's scope.

> How can we safely tell that a name is a glob? There are other patterns besides *.

I believe in traditional Windows globbing the metacharacters are `*`, `?` only. The python `glob` module also does `[]`. You probably can't exactly match what, like, MSVC would do, but that's ok, it's Windows, it's not your fault and no one expects everything to work consistently.

> What if the name comes from Bash and *.txt was actually the name of the file after expansion?

[...]

PR #1918, opened 2021-05-19
flag to control Windows pattern expansion

### LINKED ISSUE #1901: Allow disabling pattern expansion on Windows
## TL;DR

Essentially, make this optional: https://github.com/pallets/click/pull/1830

## Use case

We have a command that passes arbitrary arguments to be executed in other contexts such as Docker containers, servers over SSH, etc.

On Windows since Click 8.0 this breaks (depending on files in cwd, etc.) and unlike on Linux we can't just escape special characters

## Implementation

Setting could be called `disable_pattern_expansion`, e.g.:

```python
import click


@click.command(
    context_settings={'help_option_names': [], 'ignore_unknown_options': True, 'disable_pattern_expansion': True},
)
@click.argument('args', required=True, nargs=-1)
@click.pass_obj
def run(app, args):
    environment = app.get_environment()
    environment.send_command(args)
```

davidism:
Windows terminals don't do pattern expansion, so #1096 requested simulating it in Click. However, Windows also doesn't have a way to escape patterns so that the literal value is used. This adds a flag `cli(windows_expand_args=False)` to disable expansion.

- fixes #1901 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,6 +27,10 @@ Unreleased
     needed. :issue:`1895`
 -   If a default value is invalid, it does not prevent showing help
     text. :issue:`1889`
+-   Pass ``windows_expand_args=False`` when calling the main command to
+    disable pattern expansion on Windows. There is no way to escape
+    patterns in CMD, so if the program needs to pass them on as-is then
+    expansion must be disabled. :issue:`1901`
 
 
 Version 8.0.0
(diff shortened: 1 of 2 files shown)


ofek:
How is this used? I tried:

```python
import click

@click.group(windows_expand_args=False)
@click.pass_context
def cli(ctx):
    ...
```

and

```python
import click

@click.group(context_settings={'windows_expand_args': False})
@click.pass_context
def cli(ctx):
    ...
```

davidism:
It's an argument to the main program.


```
cli(windows_expanfd_args=False)
```

davidism:
```
def main():
    cli(...)
```

PR #1930, opened 2021-05-23
Fix shell completion for arguments with non-alphanumeric characters

### LINKED ISSUE #1929: Shell completion fails with for arguments starting with "/"
After enabling shell completion support, arguments which start with a leading forward slash "/" fail to complete.

After a bit of debugging, it appears that if the argument starts with a "/", the completions are generated for the command itself instead of for the argument. For example, `my_app open /fold⇥` would generate completions for `open` instead of `/fold` whereas `my_app open fold⇥` correctly generates completions for `fold`.

Edit: To clarify the above, in both cases either "/fold" or "fold" is correctly picked up as the incomplete string. But in the first case, `obj.shell_complete(ctx, "/fold")` is called with the "open" command as the `obj`.

Environment:

- Python version: 3.9
- Click version: 8.0.1
- Shell: zsh, bash

samschott:
This PR fixes shell completion for arguments with non-alphanumeric characters by implementing a less stringent test in `shell_completion._start_of_option()`.

- fixes #1929

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ Unreleased
 -   Bash version detection is locale independent. :issue:`1940`
 -   Empty ``default`` value is not shown for ``multiple=True``.
     :issue:`1969`
+-   Fix shell completion for arguments that start with a forward slash
+    such as absolute file paths. :issue:`1929`
 
 
 Version 8.0.1
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -450,7 +450,12 @@ def _is_incomplete_argument(ctx: Context, param: Parameter) -> bool:
 
 def _start_of_option(value: str) -> bool:
     """Check if the value looks like the start of an option."""
-    return not value[0].isalnum() if value else False
+    if not value:
+        return False
+
+    c = value[0]
+    # Allow "/" since that starts a path.
+    return not c.isalnum() and c != "/"
 
 
 def _is_incomplete_option(args: t.List[str], param: Parameter) -> bool:
--- a/tests/test_shell_completion.py
+++ b/tests/test_shell_completion.py
@@ -125,6 +125,14 @@ def test_path_types(type, expect):
     assert c.type == expect
 
 
+def test_absolute_path():
+    cli = Command("cli", params=[Option(["-f"], type=Path())])
+    out = _get_completions(cli, ["-f"], "/ab")
+    assert len(out) == 1
+    c = out[0]
+    assert c.value == "/ab"
+
+
 def test_option_flag():
     cli = Command(
         "cli",


davidism:
Please rebase to 8.0.x:

```
$ git rebase --onto origin/8.0.x origin/main
$ git push -f
```

CHANGES.rst
```diff
@@ -5,6 +5,8 @@ Version 8.0.2
 
 Unreleased
 
+-   Fix shell completion for arguments that start with non-alphanumeric
+    characters such as absolute file paths.
```
davidism, line 9: Keep two newlines between headers, add `:issue:` directive.

davidism: still missing `:issue:` directive

SamSchott: Yeah, working on it...

davidism:
Please use `git rebase -i origin/8.0.x` to squash these commits into one.

SamSchott:
All of them, or just the changelog entries?

davidism:
All of them.

SamSchott:
Any comments regarding the actual changes? Am I overlooking some cases by assuming that all options begin with a "-"? There might have been a good reason for the alphanumeric check.

davidism:
Options can start with any non-alpha character, not only `-`. Windows tools often use `/`. The other most common, shown in the docs, is `+`. Different options can have different prefixes, it doesn't have to be consistent across options or commands.

SamSchott:
Ah, yes, I should have red [https://click.palletsprojects.com/en/8.0.x/options/#other-prefix-characters](https://click.palletsprojects.com/en/8.0.x/options/#other-prefix-characters) 🤦. Apologies for that.

This makes things a bit more difficult. I can see following choices:

1. Keep the code as is and document that shell completion is only supported for values that start with alphanumeric characters. This would be a shame, especially since paths are such a common argument type for command line tools.
2. Collect all possible option prefixes for a specific command. If the incomplete string starts with one of those prefixes, recognise it as the start of an option. This would still prioritise the completion of option names over arguments or option values (as is the case currently), but only if there are actually any option names starting with the given character.

If you agree, I can try to implement the second.

davidism:
How about just don't support completion of Windows-style options, since that already wasn't supported.

SamSchott:
So we would hard-code a set of non-alphanumeric characters that are supported for the shell completion of options? For example `{"-", "+"}` or just `"-"` as in this PR?

One advantages of the 2. approach from my the previous comment: if a command has only arguments and *no* options, we could can offer completion for arguments that start with `-`.

The most rigorous solution probably would be: if the incomplete string starts with a non-alphanum character, we offer completions for both matching options and arguments. This will however be complex to implement, especially when offering completion for both a path and an option.

It's really up to you. As you say in the docs, any prefix other than "-" or "--" for options is discouraged. And the reverse, argument values that start with "-", are quite uncommon. The most probable use case is free string input and shell completion won't be very useful there.

davidism:
I was thinking `not value[0].isalnum() and value != "/"`

SamSchott:
Ok, done.

PR #1934, opened 2021-05-24
Fix `is_bool_flag` being set to True even when `is_flag` is False

### LINKED ISSUE #1925: Option.is_bool_flag is set to True even if the option is not a flag
This is relevant only for Click 8.0.1, since the following line was introduced in PR https://github.com/pallets/click/pull/1913:

https://github.com/pallets/click/blob/af0af571cbbd921d3974a0ff9cf58a4b26bb852b/src/click/core.py#L2531

The solution is as easy as:

```python
self.is_bool_flag = self.is_flag and isinstance(self.type, types.BoolParamType)
```
Available to open a PR.

## Demonstration code
```python
import click

@click.command()
@click.option('--switch', type=bool)
def f(switch):
    print(switch)


if __name__ == '__main__':
    assert not f.params[0].is_flag
    assert f.params[0].is_bool_flag
    f('--switch'.split())
```
```
Error: Option '--switch' requires an argument.
```

janluke:
Fixes #1925.

The test is probably overkill.

### DIFF
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -4,6 +4,7 @@ import re
 import pytest
 
 import click
+from click import Option
 
 
 def test_prefixes(runner):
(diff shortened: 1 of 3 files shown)


tests/test_options.py
```diff
@@ -4,6 +4,7 @@
 import pytest
 
 import click
+from click import Option
```
davidism, line 7: Use `click.Option` below.

janLuke: Okay, but note that using `click.Option` will require 2x the number of lines because of the line width limit.

davidism: Ah, never mind then, import is not a big deal.

PR #1942, opened 2021-06-01
bash version detection is locale independent

### LINKED ISSUE #1940: Bash version detection fails with certain locales
I have a machine where the system locale is set to `de_DE.UTF-8`. The output of `bash --version` here is:
~~~
GNU bash, Version 5.1.4(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2020 Free Software Foundation, Inc.
Lizenz GPLv3+: GNU GPL Version 3 oder jünger <http://gnu.org/licenses/gpl.html>

Dies ist freie Software. Sie darf verändert und verteilt werden.
Es wird keine Garantie gewährt, soweit das Gesetz es zulässt.
~~~

This causes clicks shell completion activation to die with an exception because of the capitalized `V` in `Version`:
~~~
  File "/home/losinski/.local/bin/template-test", line 8, in <module>
    sys.exit(main())
  File "/home/losinski/.local/lib/python3.9/site-packages/click/core.py", line 1137, in __call__
  File "/home/losinski/.local/lib/python3.9/site-packages/click/core.py", line 1057, in main
  File "/home/losinski/.local/lib/python3.9/site-packages/click/core.py", line 1132, in _main_shell_completion
    rv = shell_complete(self, ctx_args, prog_name, complete_var, instruction)
  File "/home/losinski/.local/lib/python3.9/site-packages/click/shell_completion.py", line 45, in shell_complete
    echo(comp.source())
  File "/home/losinski/.local/lib/python3.9/site-packages/click/shell_completion.py", line 324, in source
    self._check_version()
  File "/home/losinski/.local/lib/python3.9/site-packages/click/shell_completion.py", line 319, in _check_version
    raise RuntimeError(
RuntimeError: Couldn't detect Bash version, shell completion is not supported.
~~~

A quickfix would be to set `LANG=C` as environment on the `bash --version` call here: https://github.com/pallets/click/blob/main/src/click/shell_completion.py#L305. Another option would be
~~~
bash -c 'echo ${BASH_VERSION}'
5.1.4(1)-release
~~~

Environment:

- Python version: 3.9.2
- Click version: 8.0.1

janLo:
The `BASH_VERSION` variable holds the bash version number in a much 
better consumable and especially locale independent format than the 
`--version` flag.

This fixes #1940

Checklist:

### DIFF
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -302,8 +302,10 @@ class BashComplete(ShellComplete):
     def _check_version(self) -> None:
         import subprocess
 
-        output = subprocess.run(["bash", "--version"], stdout=subprocess.PIPE)
-        match = re.search(r"version (\d)\.(\d)\.\d", output.stdout.decode())
+        output = subprocess.run(
+            ["bash", "-c", "echo ${BASH_VERSION}"], stdout=subprocess.PIPE
+        )
+        match = re.search(r"^(\d+)\.(\d+)\.\d+", output.stdout.decode())
 
         if match is not None:
             major, minor = match.groups()
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Unreleased
 
 -   ``is_bool_flag`` is not set to ``True`` if ``is_flag`` is ``False``.
     :issue:`1925`
+-   Bash version detection is locale independent. :issue:`1940`
 
 
 Version 8.0.1


davidism:
Please rebase to 8.0.x and add a changelog entry.

janLo:
> Please rebase to 8.0.x and add a changelog entry.

Done.

src/click/shell_completion.py
```diff
@@ -302,8 +302,10 @@ class BashComplete(ShellComplete):
     def _check_version(self) -> None:
         import subprocess
 
-        output = subprocess.run(["bash", "--version"], stdout=subprocess.PIPE)
-        match = re.search(r"version (\d)\.(\d)\.\d", output.stdout.decode())
+        output = subprocess.run(
+            ["bash", "-c", "echo ${BASH_VERSION}"], stdout=subprocess.PIPE
```
igrr, line 306: Please note that this does not always return the version of `bash` where autocompletion has been invoked from.

Here's an example, taken from one macOS system:
```
$ which -a bash
/usr/local/bin/bash
/bin/bash
$ help | head -1
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
$ bash -c 'echo ${BASH_VERSION}'
4.4.23(1)-release
$ echo $BASH_VERSION
3.2.57(1)-release
```
Here, /bin/bash is the OS-provided bash 3.2.57, while /usr/local/bin/bash is bash 4.4.23 installed using Homebrew. The user profile is configured with /bin/bash as the login shell.

I don't see a good way to resolve this issue other than to tell the user that such setup is "broken". I'm mentioning it anyway in case someone else has an idea.

davidism: If `BASH_VERSION` is available in the env, then couldn't we look at that directly?

igrr: Unfortunately BASH_VERSION is not exported by default, so while
```bash
$ echo ${BASH_VERSION}
4.4.23(1)-release
```
works,
```bash
$ python -c 'import os; print(os.environ.get("BASH_VERSION"))'
None
```
doesn't. This can be worked around by doing `export BASH_VERSION` before invoking `click` for autocompletion, but I think this would be a breaking change.

sscherfke:
I would change it this for now:
```python
        output = subprocess.run(["bash", "--version"], stdout=subprocess.PIPE)
        match = re.search(r"(\d+)\.(\d+)\.\d+", output.stdout.decode())
```
This solves the locale dependent spelling of *version* ans also adds support for multi-digit versions (e.g. 5.12.23).  Since this still uses `bash --version`, it won’t introduce new issues.  

So while this may not be the perfect solution, it will certainly improve the current state. :)

src/click/shell_completion.py
janLo: But the current code has the very same problem, as you initially mentioned @igrr, as it also just calls the first bash in `$PATH`.

janLo:
Why not use the version variable. This would completely avoid future surprises that always happen when one tries to parse data out of human readable descriptions one does not control.

The variable is at least clearly defined:

> BASH_VERSION
> The version number of the current instance of Bash.

(From  https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html)

src/click/shell_completion.py
igrr: I agree with that and with your proposed solution @janLo, I only mentioned this (extra) issue in case anyone looking at this PR comes up with a way to deal it.

PR #1946, opened 2021-06-03
Stop using exception.message

hexagonrecursion:
https://bugzilla.mozilla.org/show_bug.cgi?id=1713825#c0

> In PEP 352, exception.message was deprecated (now, you can get the string from of an exception by simply doing str(exception). As of Python 3.0, exception.message was dropped, and attempting to access it causes an error.

### DIFF
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -177,7 +177,7 @@ def prompt(
             if hide_input:
                 echo(_("Error: The value you entered was invalid."), err=err)
             else:
-                echo(_("Error: {e.message}").format(e=e), err=err)  # noqa: B306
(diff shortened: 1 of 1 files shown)


davidism:
`ClickException` defines `self.message`. The `UsageError` where you made this change derives from that, so it's not an issue. Also, there are many more locations using `message`, usually as `self.message` in `click.exceptions`.

davidism:
I reviewed this more and concluded it's fine as-is. While you're correct that `Exception.message` is removed, from my reading of PEP 352 that means that `message` is not a universal API for all exceptions, not that subclasses can't define that attribute.

PR #1948, opened 2021-06-03
update link

Yourun-proger:
There is already a new stable version of the documentation.

### DIFF
--- a/examples/README
+++ b/examples/README
@@ -9,4 +9,4 @@ Click Examples
   through the wrong interpreter.
 
   For more information about this see the documentation:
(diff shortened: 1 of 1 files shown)


greyli:
If we need to update this link, it's better to use `https://click.palletsprojects.com/setuptools/` so that we don't need to update it in the future.

Yourun-proger:
@greyli , yes. This is realy better)

Yourun-proger:
Will you accept this or should I close this PR?

davidism:
Please don't ask when we'll merge things, well get to it as we have time.

PR #1949, opened 2021-06-06
update info about progress-bar

### LINKED ISSUE #1926: point to tqdm as an alternative progress bar in the docs
I just spent time changing my program to use `click.progressbar` and then [submitted a bug about it](https://github.com/pallets/click/issues/1924). It was closed with the message "We're no longer working on the progress bar implementation. Use a dedicated library like tqdm to render progress bars."

I wish I knew that in advance. I suggest that the section about `progressbar` in the docs would say that it's deprecated, and to use `tqdm` instead.

Yourun-proger:
fixes #1926

### DIFF
--- a/docs/utils.rst
+++ b/docs/utils.rst
@@ -346,6 +346,9 @@ but you want to quickly show the user some progress about how long that
 will take.  Click supports simple progress bar rendering for that through
 the :func:`progressbar` function.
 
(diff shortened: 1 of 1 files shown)


davidism:
Thanks, however this is not accurate. I'll have to update this with exactly what I intend.

Yourun-proger:
Okey)

PR #1962, opened 2021-06-17
add an ability to check that a Path has an executable bit set

### LINKED ISSUE #1961: Allow for the ability to check whether a Path has an executable bit set
It would be nice to be able to use click in the scenario where the user might pass in a path to an executable. In the case that a binary is produced in a folder of the same name, this would catch the scenario where the user passes the directory containing the executable instead of the correct `/some/long/path/programname/programname`

This is very straightforward to add: https://github.com/sielicki/click/commit/9cd5150da149fa2bff21208811deac6938acfece

sielicki:
- fixes #1961 

Checklist:

tox failed but as far as I can tell it's unrelated to my code. I'm also not sure where pre-commit hooks are documented, but as far as I can tell this is formatted fine.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,7 +3,8 @@
 Version 8.1.0
 -------------
 
-Unreleased
+-   An an argument to ``Path`` which mirrors writable/readable, but for
+    executable bits. :issue:`1961`
 
 -   Drop support for Python 3.6. :pr:`2129`
 -   Remove previously deprecated code. :pr:`2130`
--- a/src/click/types.py
+++ b/src/click/types.py
@@ -753,8 +753,9 @@ class Path(ParamType):
         exist, then all further checks are silently skipped.
     :param file_okay: Allow a file as a value.
     :param dir_okay: Allow a directory as a value.
-    :param writable: The file or directory must be writable.
-    :param readable: The file or directory must be readable.
+    :param readable: if true, a readable check is performed.
+    :param writable: if true, a writable check is performed.
+    :param executable: if true, an executable check is performed.
     :param resolve_path: Make the value absolute and resolve any
         symlinks. A ``~`` is not expanded, as this is supposed to be
         done by the shell only.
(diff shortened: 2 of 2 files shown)


CHANGES.rst
```diff
@@ -9,8 +9,8 @@ Unreleased
 Version 8.0.2
 -------------
 
-Unreleased
-
+-   An an argument to ``Path`` which mirrors writable/readable, but for
```
davidism, line 12: This needs to go in the 8.1 section.

src/click/types.py
```diff
@@ -768,6 +769,9 @@ class Path(ParamType):
         ``None``, keep Python's default, which is ``str``. Useful to
         convert to :class:`pathlib.Path`.
 
+    .. versionchanged:: 8.0.2
+        Added the `executable` parameter.
```
davidism, line 773: Use double backticks to format code.

sielicki: Fixed, sorry for the lack of care with that.

CHANGES.rst
```diff
@@ -3,15 +3,14 @@
 Version 8.1.0
 -------------
 
-Unreleased
-
+-   An an argument to ``Path`` which mirrors writable/readable, but for
+    executable bits. :issue:`1961`
```
davidism, line 7: Keep two lines before subsequent headers, this format is used throughout the docs.

CHANGES.rst
sielicki: Thanks, wasn't sure whether this could be minor or not.

CHANGES.rst
```diff
@@ -3,15 +3,14 @@
...
 
 Version 8.0.2
 -------------
 
 Unreleased
 
-
```
davidism, line 14: Don't remove intentional formatting.

sielicki: I'm sloppy. I *think* I got it right this time. ;)

sielicki:
@davidism is there anything I can help with to make sure this lands in the next milestone? If this PR needs to be redone to point at a pallets:next branch, let me know.

PR #1964, opened 2021-06-18
Give `Option.show_default` higher priority wrt `Context.show_default`

### LINKED ISSUE #1963: Give `Option.show_default` higher priority wrt `Context.show_default`
Currently, `Context.show_default` (a global setting) overrides `Option.show_default` (a local setting). This is counterintuitive and limit expressiveness: once `Context.show_default` is set to `True`, it's impossible to disable `show_default` for a specific option.

janluke:
Fixes #1963. 

After this, a non-null `show_default` passed to an option overrides the corresponding context setting. 

### Implementation notes
- This also fixes the previous type hint of `Option.show_default`, adding `str` to the `Union`. 
- I didn't include `str` in the type of `Context.show_default` because, as far as I understand, passing a `str` makes sense for dynamic options (not for a global default).
- Added a static constant `Option.DEFAULT_SHOW_DEFAULT = False`. If you change your mind about the default, the tests I added don't require any change.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,8 @@ Version 8.1.0
     without parentheses. :issue:`1359`
 -   The ``Path`` type can check whether the target is executable.
     :issue:`1961`
+-   ``Command.show_default`` overrides ``Context.show_default``, instead
+    of the other way around. :issue:`1963`
 
 
 Version 8.0.4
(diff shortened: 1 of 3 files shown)


janluke:
@davidism I think `Option.DEFAULT_SHOW_DEFAULT` is redundant. We could just set `Context.show_default = False` rather than to `None`. Let me know what you think.

janluke:
@davidism Give me a shout when it's time to release 8.1.0 and I'll resolve conflicts.

PR #1965, opened 2021-06-18
Across the codebase, handle the case a keyword argument is provided but set to `None`

### LINKED ISSUE #1959: Param decorators assume `attrs[key] is not None` just because `key in attrs`
The following lines cause a `TypeError` if one provide `help=None`:

https://github.com/pallets/click/blob/8b48450d5d63c747600e069d4c3e2274f41c8360/src/click/decorators.py#L244-L245

By the way, why is `cleandoc` called by the decorator and not by `Option` itself? The docstring says "all keyword arguments are forwarded unchanged (except ``cls``)".

The same problem appears when passing `cls=None` (also with `@argument`):

https://github.com/pallets/click/blob/8b48450d5d63c747600e069d4c3e2274f41c8360/src/click/decorators.py#L246-L247

I encountered this problem while writing a wrapper for the above decorators that explicitly list arguments for a better IDE support.

janluke:
In several points, Click relies _exclusively_ on the following check to verify if an argument `arg_name` has been provided: `arg_name in attrs`. Then it assumes that `attrs[arg_name] is not None`. This PR is about treating `None` values in `attrs` as "not provided" too.

- Fixes #1959 and similar issues in command decorators (see commit messages).

There are probably other points in the codebase that do the same thing. If you intend to merge this, I'll go on and search more in depth. I'll also add a test and a changelog entry.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,9 @@ Version 8.1.0
     :issue:`1961`
 -   ``Command.show_default`` overrides ``Context.show_default``, instead
     of the other way around. :issue:`1963`
+-   Parameter decorators and ``@group`` handles ``cls=None`` the same as
+    not passing ``cls``. ``@option`` handles ``help=None`` the same as
+    not passing ``help``. :issue:`#1959`
 
 
 Version 8.0.4
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1819,7 +1819,7 @@ class Group(MultiCommand):
         """
         from .decorators import command
 
-        if self.command_class is not None and "cls" not in kwargs:
+        if self.command_class and kwargs.get("cls") is None:
             kwargs["cls"] = self.command_class
 
         func: t.Optional[t.Callable] = None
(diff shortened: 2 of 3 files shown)


davidism:
This seems fine, if you want to go through the code more like you mentioned.

davidism:
That one seems to have broken a behavior.

Besides that change, were there any others? Let me know when this is ready for merge.

janluke:
My last commit broke a test. This is due to the fact that `default=None` is not equivalent to not providing `default` for binary flags:
- when you don't provide `default`, default is set to `False`
- when `default=None`, default is set to `None`.

I think this may be accidental behavior but it's not senseless: settings `default=None` you can force the prompt to require an explicit value to be provided (it'll be `[y/n]` rather than `[y/N]`). 

Nonetheless, I noticed that if you provide `default=None` to a flag without a secondary option, the prompt message is broken: it won't include "[y/n]" at the end.

@davidism I think it's better to handle this issue in another PR eventually. Thus, I'd revert my last commit here. What do you think?

davidism:
Sounds good to me, I'll merge this one.

janluke:
I reverted the commit, I didn't remove it. This is ready to be merged.

> Besides that change, were there any others? Let me know when this is ready for merge.

I searched for `"in attrs"` and `"in kwargs"`. There are a few other places where Click relies on these but they seem legit.

PR #1970, opened 2021-06-22
Empty default value in help output

### LINKED ISSUE #1969: Empty default value in --help output for option with multiple=True
## Minimal reproducible example that demonstrates the bug

```python
#!/usr/bin/env python3

import click

@click.command()
@click.option('--multioption', multiple=True, default=None, help='My help', show_default=True)
def info(_multioption):
    pass


if __name__ == '__main__':
    info()
```

Notice `multiple=True` which is crucial. If you replace it with `multiple=False` it will work just fine.

### What I get

```console
$ ./test.py --help
Usage: test.py [OPTIONS]

Options:
  --multioption TEXT  My help  [default: ]
  --help              Show this message and exit.  [default: False]
```

### Expected output

```console
$ ./test.py --help
Usage: test.py [OPTIONS]

Options:
  --multioption TEXT  My help
  --help              Show this message and exit.  [default: False]
```

### Python version

```console
$ python3 --version
Python 3.9.2
```

### Click version

```console
$ python -c "import click; print(click.__version__)"
8.0.1
```
(I checked that `8.0.2.dev0` is also affected)

pbeza:
This PR solves the problem with printing empty default values in `--help` output when `@click.option` has `multiple=True` set.

- fixes #1969

Checklist:

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2748,7 +2748,8 @@ class Option(Parameter):
             else:
                 default_string = str(default_value)
 
-            extra.append(_("default: {default}").format(default=default_string))
+            if default_string:
+                extra.append(_("default: {default}").format(default=default_string))
 
         if isinstance(self.type, types._NumberRangeBase):
             range_str = self.type._describe_range()
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2461,7 +2461,7 @@ class Option(Parameter):
     def __init__(
         self,
         param_decls: t.Optional[t.Sequence[str]] = None,
-        show_default: bool = False,
+        show_default: t.Union[bool, str] = False,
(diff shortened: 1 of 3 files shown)


src/click/core.py
```diff
@@ -2731,7 +2731,8 @@ def _write_opts(opts: t.Sequence[str]) -> str:
         show_default_is_str = isinstance(self.show_default, str)
 
         if show_default_is_str or (
-            default_value is not None and (self.show_default or ctx.show_default)
+            (default_value or isinstance(default_value, bool))
```
davidism, line 2734: This fails for a default int value of 0, and any other falsey values.

pbeza: Oh.. obviously... I repushed, please verify is it better now.

PR #1979, opened 2021-06-29
allow exception text override

### LINKED ISSUE #1917: Allow overriding exception text
Inheriting from `ClickException` has a small problem: It always appends `Error: ` to the front of the exception text.

https://github.com/pallets/click/blob/9da166957f5848b641231d485467f6140bca2bc0/src/click/exceptions.py#L43

Overriding that would be useful if you are trying to maintain a consistent style and/or color scheme with log messages and warnings/errors

---

Could easily make a somewhat backwards-compatible update to allow overriding the prepended string via the following:

```python
    def format_message(self) -> str:
        return f"Error: {self.message}"

    ...

    def show(self, file: t.Optional[t.IO] = None) -> None:
        if file is None:
            file = get_text_stderr()

        echo(_(self.format_message()), file=file)
```

cAtaman:
This is to allow overriding of the base Exception text: "Error: {message}"
Accomplished by moving the error-text definition and translation to format_message()

- fixes #1917

Checklist:

### DIFF
--- a/tests/test_custom_classes.py
+++ b/tests/test_custom_classes.py
@@ -1,3 +1,6 @@
+from abc import ABCMeta
+from typing import IO
+
 import click
 
 
@@ -110,3 +113,29 @@ def test_group_group_class_self(runner):
     group = CustomGroup()
     subgroup = group.group()(lambda: None)
     assert type(subgroup) is CustomGroup
+
+
+def test_custom_exception_class():
+    """Subclasses of click.ClickException should be able to override
+    the default message start "Error: {message}" """
+
+    class CustomException(click.ClickException):
+        def format_message(self) -> str:
+            return f"CustomError: {self.message}"
+
+    class MockStdIO(IO, metaclass=ABCMeta):
+        value = None
+
+        def write(self, *args, **kwargs):
+            self.value = args[0]
+
+    try:
+        raise CustomException("Bad things happened")
+    except CustomException as excinfo:
+        exc = excinfo
+
+    mock_file = MockStdIO()
+    exc.show(mock_file)
+
+    assert isinstance(exc, click.ClickException)
+    assert str(mock_file.value.strip()) == "CustomError: Bad things happened"
--- a/tests/test_custom_classes.py
+++ b/tests/test_custom_classes.py
@@ -1,5 +1,4 @@
-from abc import ABCMeta
-from typing import IO
+import io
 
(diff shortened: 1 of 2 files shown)


tests/test_custom_classes.py
```diff
@@ -110,3 +113,29 @@ class CustomGroup(click.Group):
...
+
+def test_custom_exception_class():
+    """Subclasses of click.ClickException should be able to override
+    the default message start "Error: {message}" """
+
+    class CustomException(click.ClickException):
+        def format_message(self) -> str:
+            return f"CustomError: {self.message}"
+
+    class MockStdIO(IO, metaclass=ABCMeta):
```
davidism, line 126: This isn't needed, can use `io.StringIO()`

tests/test_custom_classes.py
```diff
@@ -110,3 +113,29 @@ class CustomGroup(click.Group):
...
+        value = None
+
+        def write(self, *args, **kwargs):
+            self.value = args[0]
+
+    try:
+        raise CustomException("Bad things happened")
```
davidism, line 133: Don't need to raise, can call `CustomException("...").show(f)`

tests/test_custom_classes.py
```diff
@@ -110,3 +113,29 @@ class CustomGroup(click.Group):
...
+    except CustomException as excinfo:
+        exc = excinfo
+
+    mock_file = MockStdIO()
+    exc.show(mock_file)
+
+    assert isinstance(exc, click.ClickException)
```
davidism, line 140: This isn't needed, if this were false Python itself would be broken.

tests/test_custom_classes.py
cAtaman: Alright. I'll resolve

tests/test_custom_classes.py
cAtaman: haha true

davidism:
Actually, now that I look into how `format_message` is used, the original request doesn't seem correct. Other exceptions override this method, which means most errors will no longer have "Error: " at the front when output.

tests/test_custom_classes.py
cAtaman: I knew there'd be a better way. Lemme try that

davidism:
Closing as I don't think the original issue is valid anymore.

PR #1998, opened 2021-07-08
Hide single option boolean flags in "help" when the default is False

### LINKED ISSUE #1971: Consider not showing the default value of boolean flags without a secondary option (False)
Unless I'm missing something, the default value of such options should be obvious: `False`. As a consequence, `[default: False]` feels like superfluous noise. This is certainly true for the `--help` option, which should never show a default.

cAtaman:
Hides the default value in help texts for single option boolean flags when the default is False

- fixes #1971

Checklist:

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2416,25 +2416,26 @@ class Option(Parameter):
 
     All other parameters are passed onwards to the parameter constructor.
 
-    :param show_default: controls if the default value should be shown on the
-                         help page. Normally, defaults are not shown. If this
-                         value is a string, it shows the string instead of the
-                         value. This is particularly useful for dynamic options.
-    :param show_envvar: controls if an environment variable should be shown on
-                        the help page.  Normally, environment variables
-                        are not shown.
-    :param prompt: if set to `True` or a non empty string then the user will be
-                   prompted for input.  If set to `True` the prompt will be the
-                   option name capitalized.
+    :param show_default: Controls if the default value should be shown
+        on the help page. Normally, defaults are not shown. If this
+        value is a string, it shows that string in parentheses instead
+        of the actual value. This is particularly useful for dynamic
+        options. For single option boolean flags, the default remains
+        hidden if its value is ``False``.
+    :param show_envvar: Controls if an environment variable should be
+        shown on the help page. Normally, environment ariables are not
+        shown.
+    :param prompt: If set to ``True`` or a non empty string then the
+        user will be prompted for input. If set to ``True`` the prompt
+        will be the option name capitalized.
     :param confirmation_prompt: Prompt a second time to confirm the
         value if it was prompted for. Can be set to a string instead of
         ``True`` to customize the message.
     :param prompt_required: If set to ``False``, the user will be
         prompted for input only when the option was specified as a flag
         without a value.
-    :param hide_input: if this is `True` then the input on the prompt will be
-                       hidden from the user.  This is useful for password
-                       input.
+    :param hide_input: If this is ``True`` then the input on the prompt
+        will be hidden from the user. This is useful for password input.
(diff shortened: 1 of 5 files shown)


tests/test_formatting.py
```diff
@@ -316,6 +316,8 @@ def cmd1():
 
 
 def test_global_show_default(runner):
+    """Note: default to "--help" would not be shown"""
```
davidism, line 319: Add this as a comment, docstrings are meant to describe a test not add comments.

tests/test_options.py
```diff
@@ -700,16 +700,32 @@ def test_show_default_boolean_flag_name(runner, default, expect):
...
+        default=True,
+        help="Enable the cache.",
+    )
+    ctx = click.Context(click.Command("test"))
+    message = opt.get_help_record(ctx)[1]
+    assert "[default: True]" in message
+
+
+def test_hide_false_default_boolean_flag_value(runner):
+    """When a boolean flag only has one opt and its default is False,
```
davidism, line 720: The default is not `False` here, it's not set / `None`.

src/click/core.py
```diff
@@ -2454,6 +2456,10 @@ class Option(Parameter):
 
     .. versionchanged:: 8.0.1
         ``type`` is detected from ``flag_value`` if given.
+
+    .. versionchanged:: 8.1.0
```
davidism, line 2460: Order change entries for the newest version at the top.

tests/test_options.py
```diff
@@ -700,16 +700,32 @@ def test_show_default_boolean_flag_name(runner, default, expect):
...
+    it will not show the default
     """
     opt = click.Option(
         ("--cache",), is_flag=True, show_default=True, help="Enable the cache."
     )
     ctx = click.Context(click.Command("test"))
     message = opt.get_help_record(ctx)[1]
-    assert "[default: False]" in message
+    assert "[default: False]" not in message
```
davidism, line 728: If it tries to show some other default somehow, this test will still pass. Probably `'default' not in message` is safer.

src/click/core.py
```diff
@@ -2420,6 +2420,8 @@ class Option(Parameter):
                          help page. Normally, defaults are not shown. If this
                          value is a string, it shows the string instead of the
                          value. This is particularly useful for dynamic options.
+                         For single option boolean flags, the default remains hidden
```
davidism, line 2423: Looks like this line sticks out more. Docstrings and docs pages are wrapped at 72 characters (code is 88).

tests/test_options.py
```diff
@@ -700,16 +700,32 @@ def test_show_default_boolean_flag_name(runner, default, expect):
     assert f"[default: {expect}]" in message
 
 
-def test_show_default_boolean_flag_value(runner):
-    """When a boolean flag only has one opt, it will show the default
-    value, not the opt name.
+def test_show_true_default_boolean_flag_value(runner):
```
davidism, line 703: Seems like this behavior now contradicts the requested change.

tests/test_formatting.py
cAtaman: alright

tests/test_options.py
cAtaman: I'll rename the test and parametrize to handle both False and None

src/click/core.py
cAtaman: reordered

tests/test_options.py
cAtaman: I'm concerned the word "default" might feature in the actual help text. Perhaps I could use 
```assert "[default: " not in message```

tests/test_options.py
cAtaman: It seems prudent to alert the user when such a boolean flag has its default set to `True` as it is not the norm.

PR #2004, opened 2021-07-13
Fix default values of multiple options with optional values

### LINKED ISSUE #2001: Can't set the default value for multiple options with optional values
The following example combines [optional values](https://click.palletsprojects.com/en/8.0.x/options/#optional-value) with [multiple options](https://click.palletsprojects.com/en/8.0.x/options/#multiple-options):

```python
#!/usr/bin/env python3

import click


@click.command()
@click.option('--label', multiple=True, is_flag=False, flag_value='default')
def cli(label):
    print(label)
    print(f"click.core._flag_needs_value: {click.core._flag_needs_value}")


cli()
```

```
$ ./test.py --label --label test
('<object object at 0x7f5ef63ec250>', 'test')
click.core._flag_needs_value: <object object at 0x7f5ef63ec250>
```

Expected result: `('default', 'test')`

Looks like when `multiple=True`, the sentinel value (`click.core._flag_needs_value`) is never substituted with the default value specified by `flag_value`.

Environment:

- Python version: 3.7.3
- Click version: 8.0.1

ikapelyukhin:
Fixes default values for multiple options with optional values. The sentinel object (`click.core._flag_needs_value`) is now properly replaced by the value of `flag_value`.

- fixes #2001

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,8 @@ Unreleased
     :issue:`2072`
 -   Default values are not cast to the parameter type twice during
     processing. :issue:`2085`
+-   Options with ``multiple`` and ``flag_value`` use the flag value
+    instead of leaving an internal placeholder. :issue:`2001`
 
 
 Version 8.0.1
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2853,6 +2853,14 @@ class Option(Parameter):
                 value = self.flag_value
                 source = ParameterSource.COMMANDLINE
 
+        elif (
+            self.multiple
+            and value is not None
+            and any(v is _flag_needs_value for v in value)
+        ):
+            value = [self.flag_value if v is _flag_needs_value else v for v in value]
+            source = ParameterSource.COMMANDLINE
+
         # The value wasn't set, or used the param's default, prompt if
         # prompting is enabled.
         elif (
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -792,6 +792,29 @@ def test_option_with_optional_value(runner, args, expect):
     assert result.return_value == expect
 
 
+def test_multiple_option_with_optional_value(runner):
+    cli = click.Command(
+        "cli",
+        params=[
+            click.Option(["-f"], is_flag=False, flag_value="flag", multiple=True),
+            click.Option(["-a"]),
+            click.Argument(["b"], nargs=-1),
+        ],
+        callback=lambda **kwargs: kwargs,
+    )
+    result = runner.invoke(
+        cli,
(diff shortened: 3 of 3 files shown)


ikapelyukhin:
Things get a bit weird with `ParameterSource` and multiple options: you might get some of the option values from defaults or the prompt, others specified explicitly on the command line.

Also, default values populated from `flag_value` have `ParameterSource.COMMANDLINE` rather than `ParameterSource.DEFAULT`, which doesn't seem right.

davidism:
> Things get a bit weird with ParameterSource and multiple options: you might get some of the option values from defaults or the prompt, others specified explicitly on the command line.

That needs to be addressed, the value may only come from one source.

ikapelyukhin:
> > Things get a bit weird with ParameterSource and multiple options: you might get some of the option values from defaults or the prompt, others specified explicitly on the command line.
> 
> That needs to be addressed, the value may only come from one source.

@davidism OK, I'll need some more input from you then, there are two potential cases that I'm seeing:

#### Without a prompt
```
@click.option('--foo', multiple=True, is_flag=False, flag_value='default')
```

So if you do `./script.py --foo --foo test` you'd get `('default', 'test')` as a value for `foo` option and the source is `ParameterSource.COMMANDLINE`, which is accurate for both items in the list because of how `flag_value` source is determined:

https://github.com/pallets/click/blob/ce152bd7215038f762ee918e6e181a5309ad57c9/src/click/core.py#L2867-L2869

So everything seems fine in this case as is, assuming that `ParameterSource.COMMANDLINE` is correct and shouldn't be `ParameterSource.DEFAULT` instead for one of the items.

#### With a prompt
```
@click.option('--foo', multiple=True, is_flag=False, flag_value='default', prompt="Foo")`
```

1) How do you even specify multiple values with a prompt? I'm getting an `Error: Value must be an iterable.` for whatever input I provide. Is prompting even supposed to work for options with `multiple=True`?

2) If I do `./script.py --foo --foo test`, am I expected to provide values via the prompt for both of them, or just the first one that's missing a value?

3) Provided that it is possible to use prompts with `multiple=True` and specify just one of the values, if I do `./script.py --foo --foo test` and specify one of the values after being prompted what's the `ParameterSource` then? One part of the value is coming from `ParameterSource.PROMPT` and the other one is coming from `ParameterSource.COMMANDLINE`.

Seeing how it doesn't seem possible to set an iterable value via the prompt, there's no problem with `ParameterSource` as is here either :sweat_smile:

ikapelyukhin:
@davidism OK, I've reverted the changes related to prompting, as it's not really related to the issue at hand. So things should be pretty straightforward now and source is set to `ParameterSource.COMMANDLINE` whenever option's default value is being used.

PR #2009, opened 2021-07-15
Remove logically-dead code in decorators.py::version_option()

### LINKED ISSUE #1989: Coverity 370644 Logically dead code in def version_option(), line 345
Coverity scan reports the following. The `>>>` markers indicate Coverity's inline comments. 

```
1    if version is None and package_name is None:
342        frame = inspect.currentframe()
343        assert frame is not None
>>>   implied_types: frame.f_back implies that the type of frame cannot be any of undefined and null.
344        assert frame.f_back is not None
>>>   possible_types: At condition frame !== None, the type of frame cannot be any of undefined and null.
>>>   dead_error_condition: The condition frame !== None must be true.
>>>   CID 370644 (#1 of 1): Logically dead code (DEADCODE)dead_error_line: Execution cannot reach the expression None inside this statement: f_globals = ((frame !== Non....
345        f_globals = frame.f_back.f_globals if frame is not None else None
```

Full report is at https://scan4.coverity.com/reports.htm#v38579/p14953/fileInstanceId=103660997&defectInstanceId=14283947&mergedDefectId=370644&eventId=14283947-3  but that isn't publicly accessible :-(.

If I understand this correctly on line 345 we're guaranteed that the `frame` variable has a value different than `None` so there's no need to check for it again.

cAtaman:
Remove redundant null checks from decorators.py's version_option() function as highlighted by Coverity scans

- fixes #1989

Checklist:

### DIFF
--- a/src/click/decorators.py
+++ b/src/click/decorators.py
@@ -340,9 +340,8 @@ def version_option(
 
     if version is None and package_name is None:
         frame = inspect.currentframe()
-        assert frame is not None
-        assert frame.f_back is not None
-        f_globals = frame.f_back.f_globals if frame is not None else None
+        f_back = frame.f_back if frame is not None else None
+        f_globals = f_back.f_globals if f_back is not None else None
         # break reference cycle
         # https://docs.python.org/3/library/inspect.html#the-interpreter-stack
         del frame


src/click/decorators.py
```diff
@@ -342,19 +342,16 @@ def version_option(
         frame = inspect.currentframe()
         assert frame is not None
         assert frame.f_back is not None
-        f_globals = frame.f_back.f_globals if frame is not None else None
+        assert frame.f_back.f_globals is not None
```
davidism, line 345: I think the report in the issue means that all three asserts here are not needed.

src/click/decorators.py
```diff
@@ -342,19 +342,16 @@ def version_option(
...
-        if f_globals is not None:
-            package_name = f_globals.get("__name__")
-
-            if package_name == "__main__":
-                package_name = f_globals.get("__package__")
-
-            if package_name:
-                package_name = package_name.partition(".")[0]
+        package_name = f_globals.get("__name__")
+        if package_name == "__main__":
```
davidism, line 351: Keep empty lines formatting around blocks.

src/click/decorators.py
```diff
@@ -342,19 +342,16 @@ def version_option(
         frame = inspect.currentframe()
         assert frame is not None
         assert frame.f_back is not None
-        f_globals = frame.f_back.f_globals if frame is not None else None
+        assert frame.f_back.f_globals is not None
+        f_globals = frame.f_back.f_globals
         # break reference cycle
         # https://docs.python.org/3/library/inspect.html#the-interpreter-stack
         del frame
```
davidism, line 349: Can collapse this to `f_globals = inspect.currentframe().f_back.f_globals` and remove the `frame = `, `del`, and comment.

src/click/decorators.py
cAtaman: Well, there are two ways to go: 
 - Using the `asserts` and removing all null checks (as we'd be certain there are no null values)
 - removing the assertions and check for null values

src/click/decorators.py
cAtaman: done

src/click/decorators.py
cAtaman: alright

src/click/decorators.py
cAtaman: I collapsed everything as you mentioned below. This is surrounded by a try block that checks for attribute error. tThis would be thrown when accessing attributes of a NoneType object.

```
        try:
            f_globals = inspect.currentframe().frame.f_back.f_globals
        except AttributeError:
            assert False
```

src/click/decorators.py
```diff
@@ -339,22 +339,16 @@ def version_option(
...
-
-            if package_name == "__main__":
-                package_name = f_globals.get("__package__")
-
-            if package_name:
-                package_name = package_name.partition(".")[0]
+        try:
+            f_globals = inspect.currentframe().frame.f_back.f_globals
+        except AttributeError:
+            raise AssertionError()
```
davidism, line 345: This pattern doesn't make sense. If there's an attribute error raise, just let that be raised, no need to silence it and turn it into an empty assertion error.

src/click/decorators.py
```diff
@@ -339,22 +339,16 @@ def version_option(
...
+
+        package_name = f_globals.get("__name__")
+        if package_name == "__main__":
+            package_name = f_globals.get("__package__")
+        if package_name:
```
davidism, line 350: Please leave empty lines between blocks.

src/click/decorators.py
cAtaman: alright

src/click/decorators.py
cAtaman: oh yeah, forgot about that one

cAtaman:
```tox -e typing``` seems to complain unless I check every step of this assignment 
```
f_globals = inspect.currentframe().frame.f_back.f_globals
```

File: src/click/decorators.py [hunk 6]
```diff
@@ -339,22 +339,14 @@ def version_option(
...
-
-        if f_globals is not None:
-            package_name = f_globals.get("__name__")
-
-            if package_name == "__main__":
-                package_name = f_globals.get("__package__")
-
-            if package_name:
-                package_name = package_name.partition(".")[0]
+        f_globals = inspect.currentframe().f_back.f_globals
```
cAtaman, line 342: ```tox -e typing``` seems to complain unless I check every step of this assignment

cAtaman: It seems to break it up might be the solution
```python3
        frame = inspect.currentframe()

        if frame is not None:
            f_b = frame.f_back

        if f_b is not None:
            f_globals = f_b.f_globals

        del frame
```

cAtaman: or perhaps

```python3
        frame = inspect.currentframe()
        assert frame is not None
        assert frame.f_back is not None
        assert frame.f_back.f_globals is not None
        f_globals = frame.f_back.f_globals
        # break reference cycle
        # https://docs.python.org/3/library/inspect.html#the-interpreter-stack
        del frame
```

Coverity's issue was that there were null checks after asserting the value was not **```None```**

cAtaman: Which of these would be more acceptable?

cAtaman: I think this might be better
```python3
        frame = inspect.currentframe()
        f_back = frame.f_back if frame is not None else None
        f_globals = f_back.f_globals if f_back is not None else None
        # break reference cycle
        # https://docs.python.org/3/library/inspect.html#the-interpreter-stack
        del frame
```

davidism: I agree, that looks good.

davidism:
@atodorov Not really sure if we've addressed the findings or not. The original report was a bit hard to parse, and we don't have a way to run it again. Any chance you could run against this branch and see if those issues have been addressed?

atodorov:
Can't run this against a branch unfortunately. It's a proprietary service and it gets run against all of my site-packages/ directory.  Coverity Scan is free to use for public repositories on GitHub so you may want to configure that for click yourself to scan the latest main branch.

Issue #665, opened 2016-09-28
Click prompts do not cooperate with readline module


gumho:
When readline is imported and a user presses backspace during a Click prompt, the entire prompt message is erased and cursor goes to beginning of the line. This can be reproduced with the following code:

```
import click, readline
click.prompt("Input")
```

This is due to [Python issue 12833](https://bugs.python.org/issue12833). The official recommendation is that the prompt should be displayed with raw_input. I noticed that Click first displays the prompt message using echo() and then uses raw_input("") for the prompt. There is also a workaround by passing " " to raw_input, however, that may be ugly in the display.

hany:
Any feedback about this particular issue?

famecastle:
This becomes now a little problematic, since the great 'click-shell` package, which easily extends the `click` package by an interactive shell if no commands or parameters are provided, relies on readline.

What I've done in my (venv'ed) python project to circumvent this problem for now:

- copy the `click` directory from `venv/lib/site-packages/` to the main folder
- run `pip uninstall click`
- edit `click/termui.py` and go to the `f("")` line.
- Replace `f("")` with `f(" ")` (notice the space)
- In the line above, add an `rstrip(" ")` to remove ugly display: `echo(text.rstrip(" "), nl=False, err=err)`

I think the click team should fix this issue since readline is quite a common module included in the Python standard library. How about echoing the prompt as usual, but displaying the prefix in the input-method (like `f(prefix)`) ?

PR #2019, opened 2021-07-27
Fix click.prompt() printing a single space to stdout when err=True.

### LINKED ISSUE #2018: click.prompt(..., err=True) produces a single space on stdout
Using the following script

```python
import click

click.prompt("asd", err=True)
```

and running it like

```
python test.py 1> stdout 2> stderr
```

The file `stderr` should receive all output from the script, and `stdout` should remain empty. However we can see that the file `stdout` receives a single space for each round that `click.prompt()` does (i.e. when entering nothing, it keeps re-asking).

This is unexpected and should not happen.

I believe [this comment](https://github.com/pallets/click/blob/main/src/click/termui.py#L141) is related to this issue.

Environment:

- Python version: 3.7.10
- Click version: 8.0.1

nils-werner:
When using `click.prompt(..., err=True)`, the underlying code sends a single space to `stdout`, which may confuse programs who are trying to digest the output. This PR fixes that by sending the space to whatever output stream was chosen instead. For `err=True`, the space is sent to `sys.stdout`.

Note that I am not sure how to test this behaviour, that's why this PR is a WIP. Any suggestions are welcome.

- fixes #2018

Checklist:

### DIFF
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -67,11 +67,7 @@ def hidden_prompt_func(prompt: str, err: bool = False) -> str:
     """
     import getpass
 
-    try:
-        sys.stdout = sys.__stderr__ if err else sys.__stdout__
-        return getpass.getpass(prompt)
-    finally:
-        sys.stdout = sys.__stdout__
+    return getpass.getpass(prompt, stream=sys.stderr if err else sys.stdout)
 
 
 def _build_prompt(
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -53,10 +53,12 @@ def visible_prompt_func(prompt: str, err: bool = False) -> str:
 
     """
     try:
-        sys.stdout = sys.__stderr__ if err else sys.__stdout__
+        old_stdout = sys.stdout
+        if err:
+            sys.stdout = sys.stderr
         return input(prompt)
     finally:
-        sys.stdout = sys.__stdout__
+        sys.stdout = old_stdout
 
 
 def hidden_prompt_func(prompt: str, err: bool = False) -> str:
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -24,10 +24,6 @@ if t.TYPE_CHECKING:
 
 V = t.TypeVar("V")
 
-# The prompt functions to use.  The doc tools currently override these
-# functions to customize how they work.
-visible_prompt_func: t.Callable[[str], str] = input
-
 _ansi_colors = {
     "black": 30,
     "red": 31,
@@ -50,10 +46,32 @@ _ansi_colors = {
 _ansi_reset_all = "\033[0m"
 
 
-def hidden_prompt_func(prompt: str) -> str:
+def visible_prompt_func(prompt: str, err: bool = False) -> str:
+    """
+    .. versionadded:: 8.0.2
+        Added the `err` parameter.
+
+    """
+    try:
+        sys.stdout = sys.__stderr__ if err else sys.__stdout__
+        return input(prompt)
+    finally:
+        sys.stdout = sys.__stdout__
+
+
+def hidden_prompt_func(prompt: str, err: bool = False) -> str:
+    """
+    .. versionadded:: 8.0.2
+        Added the `err` parameter.
+
+    """
     import getpass
 
-    return getpass.getpass(prompt)
+    try:
+        sys.stdout = sys.__stderr__ if err else sys.__stdout__
+        return getpass.getpass(prompt)
+    finally:
+        sys.stdout = sys.__stdout__
 
 
 def _build_prompt(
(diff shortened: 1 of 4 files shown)


davidism:
The comment above the code you've changed calls out using `stdout` for the space specifically.

> Echo a space to stdout to work around an issue where readline causes backspace to clear the whole line.

This was from https://github.com/pallets/click/issues/665 and https://github.com/pallets/click/pull/1836. Please confirm that this does not cause the issue with readline again before I can consider merging this.

nils-werner:
I checked and can confirm that it does not regress issue #665.

I checked running

```python
import click, readline

click.prompt("Input")
```

and

```python
import click, readline

click.prompt("Input", err=True)
```

and hitting backspace. The prompt wasn't deleted in either case.

src/click/termui.py
```diff
@@ -50,10 +46,32 @@
...
 
-def hidden_prompt_func(prompt: str) -> str:
+def visible_prompt_func(prompt: str, err: bool = False) -> str:
+    """
+    .. versionadded:: 8.0.2
+        Added the `err` parameter.
+
+    """
+    try:
+        sys.stdout = sys.__stderr__ if err else sys.__stdout__
```
davidism, line 56: This is not correct. Click purposefully uses `sys.stdout`, not the original `__stdout__`, in order to allow overriding `stdout`, such as in tests.

src/click/termui.py
```diff
@@ -50,10 +46,32 @@
...
+    .. versionadded:: 8.0.2
+        Added the `err` parameter.
+
+    """
     import getpass
 
-    return getpass.getpass(prompt)
+    try:
+        sys.stdout = sys.__stderr__ if err else sys.__stdout__
+        return getpass.getpass(prompt)
```
davidism, line 72: This `try/finally` isn't needed here, `getpass` takes a `stream` parameter.

davidism:
It seems like while your issue report regarded printing a space to stdout, what you were really reporting was that Python's `input` and `getpass` always print to stdout. While your patching approach does work, it's something Click deliberately avoids doing outside of tests. I'm not comfortable with merging this type of change.

> may confuse programs who are trying to digest the output

It seems like it might be easier / safer to adjust how that program is understanding the output, such as ignoring empty space, rather than patching `sys.stdout` for `input`.

I do want to allow the developer to override what things like `prompt` do, so that might help with this use case eventually.

GaretJax:
> It seems like it might be easier / safer to adjust how that program is understanding the output, such as ignoring empty space, rather than patching sys.stdout for input.

@davidism that sadly not always possible, as the program digesting the output is not always under our control.

davidism:
@nils-werner were you planning to address my comments and review?

nils-werner:
I was under the impression that you had rejected my proposed changes.

If not, could you repeat what you wanted to see changed?

src/click/termui.py
nils-werner: I am not sure I understand. The purpose of `sys.__stdout__` et al is to be the [original value of `sys.stdout`, and the Python docs recommend to use it to reset them to that original value](https://docs.python.org/3/library/sys.html#sys.__stdin__).

davidism: Tools like pytest and Click's CliRunner already override `sys.stdout`. We would want to use those, not the originals.

nils-werner: OK, understood.

davidism:
OK, I think this change makes sense given the other existing capabilities, but we're going to need to document that using `prompt` with `stderr` is not thread (or coroutine) safe because it doesn't lock the change in global state. If you're looking to make a contribution to Python, maybe open an issue that `input` should take a stream to show output on, like `getpass`.

davidism:
I'm surprised that readline continues to work. I don't see anything about it knowing what stream to re-output on, it should always use stdout from what I can tell.

davidism:
I think I've got a different solution. Changing the approach in #1836:

```diff
-            echo(text.rstrip(" "), nl=False, err=err)
+            echo(text, nl=False, err=err)
             # Echo a space to stdout to work around an issue where
             # readline causes backspace to clear the whole line.
-            return f(" ")
+            return f(" \b")
```

Don't strip the space off the prompt, so it's written to the stream as in 7.x. However, instead of sending a space to `input`, send a space *and backspace*. This prints the entire prompt to stderr, still triggers whatever in readline preserves the prompt, and doesn't actually include the extra space or backspace character in either stream.

jab (changes requested):
Before we moved forward with an approach like this (that temporarily reassigns the shared global resource sys.stdout, which could maybe pull the carpet out from under some programs), first it seems worth having a discussion with the Python maintainers about this issue, and whether it should be fixed in Python itself (e.g. by adding support for `input(..., stream=sys.stderr)`, or whether they have some better recommendation than this).

In the meantime, certainly seems worth documenting the current behavior of the space being written to stdout. Would you like to submit that as a separate PR in the meantime, @nils-werner?

(While I was looking at this quickly, I also caught a possible issue with a couple of the type hints, though that's very minor.)

PR #2023, opened 2021-07-28
Add platform check to gate _fast_exit()

### LINKED ISSUE #2017: Resource tracker warning on fish completion
Issue:
I am having the same issue described in #1738 although I am using `fish`. Essentially, I get an error from `resource_tracker.py` related to "leaked semaphore objects," when I run the completion. The completion still works, but the text of the error is garbled along with the completion information before tabbing between options. This does not happen when I use Python 3.7, but does when I use Python 3.8.

The error I am getting ends up looking like this after pressing `tab` to get completion text (the third line starting with `❯` is where I pressed `tab`):

```
----
❯ poetry shell
Spawning shell within ----.p3d-sRVjZJIG-py3.8
source ----.p3d-sRVjZJIG-py3.8/bin/activate.fish
Welcome to fish, the friendly interactive shell

----
❯ source ----.p3d-sRVjZJIG-py3.8/bin/activate.fish

----
❯ p3d /Users/----/.asdf/installs/python/3.8.8/lib/python3.8/multiprocessing/resource_tracker.py:216: UserWarning: resource_tracker: There appear to be 32 leaked semaphore objects to clean up at shutdownz API calls.)  pipeline  (Run pipeline processes.)
  warnings.warn('resource_tracker: There appear to be %d '

```

Reproduce:
I don't have a cut down example that I can provide, but I am wondering if maybe the fact that I am using `poetry` and trying to run completion in a virtualenv is causing an issue? As noted above, this works for Python 3.7 but has an issue with 3.8.

I am also using `click-didyoumean` and I manage my Python installations using `asdf`.

Expected Behavior:
I would expect the same behavior (no error) across different supported versions of Python.

Environment:

- Python version: 3.8.8
- OS: macOS 10.15.7
- Click version: 8.0.1
- click-didyoumean version: 0.0.3
- asdf version: 0.8.1
- Poetry version: 1.1.7

taranlu-houzz:
- Platforms that do not support using fork for multiprocessing have issues freeing
    up resources correctly when using os._exit() as opposed to sys.exit()

- fixes #2017 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,8 @@ Unreleased
     :issue:`1969`
 -   Fix shell completion for arguments that start with a forward slash
     such as absolute file paths. :issue:`1929`
+-   Fix use of ``_fast_exit`` on non-Linux platforms.
+    :issue:`2017`
 
 
 Version 8.0.1
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1,6 +1,7 @@
 import enum
 import errno
 import os
+import platform
 import sys
 import typing
 import typing as t
@@ -1130,7 +1131,11 @@ class BaseCommand:
         from .shell_completion import shell_complete
 
         rv = shell_complete(self, ctx_args, prog_name, complete_var, instruction)
-        _fast_exit(rv)
+
+        if platform.system() == "Linux":
+            _fast_exit(rv)
+        else:
+            sys.exit(rv)
 
     def __call__(self, *args: t.Any, **kwargs: t.Any) -> t.Any:
         """Alias for :meth:`main`."""
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -53,6 +53,9 @@ def _fast_exit(code: int) -> "te.NoReturn":
     Fallback to sys.exit() on non-Linux platforms.
 
     :param code: Exit code.
+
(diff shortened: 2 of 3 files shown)


taranlu-houzz:
@davidism This is what I have in mind. I will run through the check list later when I have some time.

taranlu-houzz:
@davidism So, it seems that I need to add `platform` to the list of approved modules in order to pass `test_imports.py`. Is that alright, or is there a different way that you typically check the OS in `click`?

davidism:
Avoid top-level imports in Click that aren't already allowed. Instead, add the import where it's used (local instead of global).

taranlu-houzz:
@davidism Okay, I have run through the checklist and added a new test. I changed the platform check to be inside the `_fast_exit()` function itself. I think this is ready to be reviewed.

davidism:
The more I think about this, the more I think we should just drop `_fast_exit` as I suggested initially. Guarding this for "non Linux platforms" seems very heavy handed, and the fact that we're saying "skipping cleanup is safe on Linux" seems like we're just asking for trouble at some point in the future.

taranlu-houzz:
@davidism I created a new PR that removes `_fast_exit()` and uses `sys.exit()` instead: #2030

PR #2034, opened 2021-08-05
small refactorings in command parsing methods

dlax:
While reading the code, I noticed that:
- 'args' does not need to be passed to Command.handle_parse_result()
- the return value from Command.handle_parse_result() is not used by caller

I'm suggesting two changes which, IMO, make the code easier to follow.

Checklist:


davidism:
The default implementation doesn't use this, presumably it's like that to be an interface subclasses could use.

Issue #1905, opened 2021-05-17
PermissionError on Windows using echo_via_pager


thorstenkampe:
```python
import click
click.echo_via_pager('2')
2
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "F:\scoop\apps\miniconda3\current\envs\main\lib\site-packages\click\termui.py", line 302, in echo_via_pager
    return pager(itertools.chain(text_generator, "\n"), color)
  File "F:\scoop\apps\miniconda3\current\envs\main\lib\site-packages\click\_termui_impl.py", line 357, in pager
    return _tempfilepager(generator, "more <", color)
  File "F:\scoop\apps\miniconda3\current\envs\main\lib\site-packages\click\_termui_impl.py", line 440, in _tempfilepager
    os.unlink(filename)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'R:\\Temp\\tmpn3vjxjwz'
```

Click should use the system's pager to display the input object without error

Environment:

- Python version: 3.9.4
- Click version: 8.0.0 (the issue does NOT occur with Click 7.1.2!)
- OS: Windows 10 (the issue does not NOT occur on Linux or Cygwin!)

PR #2036, opened 2021-08-07
close temporary file used by echo_via_pager

jcrotts:
On Windows, `unlink` would fail because the file was still open.

- fixes #<https://github.com/pallets/click/issues/1905>

Checklist:


davidism:
We should probably always close the file descriptor, not only on Windows, right?

PR #2041, opened 2021-08-09
Make shell completion prioritize option values over new options

### LINKED ISSUE #2040: Shell completion fails for option values that start with symbols
Running into the same case as #1929, except specific to option values that may start with symbols such as '+', '~', and '!'. Shell completion treats these values as the start of new options, so doesn't complete them.

Reproduction script with the cli command referenced by setuptools:

```python
class CustomType(click.ParamType):
    name: str = "custom"

    def shell_complete(self, ctx: click.Context, param: click.Parameter, incomplete: str) -> t.List[CompletionItem]:
        values = ["+all", "+one", "+two", "~all", "~one", "~two", "!all", "!one", "!two"]
        return [CompletionItem(v) for v in values if v.startswith(incomplete)]


@click.command()
@click.option("-c", type=CustomType())
def cli(c: t.Optional[str]):
    click.echo(f"Value is {c}")
```

With shell completion enabled:

```sh
$ click-bug -c <TAB><TAB>
+all  +one  +two  ~all  ~one  ~two  !all  !one  !two
$ click-bug -c +<TAB><TAB>
# Nothing happens
```

I was looking at the discussion on #1930, and it makes sense that '+' may be considered the start of an option. I'm wondering if it would be valid for completion to prioritize option values over new option names. As in, within shell_completion.py's _resolve_incomplete function, moving the `_is_incomplete_option` check above the `if "--" not in args and _start_of_option(incomplete)` call allows the tab completion in this example to work as expected:

```sh
$ click-bug -c +<TAB><TAB>
+all  +one  +two
```

I can look into putting up a PR to do that if it seems valid.

Environment:

- Python version: 3.6.9
- Click version: 8.0.1

spanglerco:
Swap the order of checks in _resolve_incomplete so that we first check to see if we're completing an option value before detecting if we're completing a new option name. Allows option values that start with non-alphanumeric characters to complete rather than treating them like new options.

- fixes #2040
- related to #1930

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -50,6 +50,8 @@ Version 8.1.0
     value. :issue:`2124`
 -   ``to_info_dict`` will not fail if a ``ParamType`` doesn't define a
     ``name``. :issue:`2168`
+-   Shell completion prioritizes option values with option prefixes over
+    new options. :issue:`2040`
 
 
 Version 8.0.4
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -292,6 +292,8 @@ class Context:
         #: must be never propagated to another arguments.  This is used
         #: to implement nested parsing.
         self.protected_args: t.List[str] = []
+        #: the collected prefixes of the command's options.
+        self._opt_prefixes: t.Set[str] = set(parent._opt_prefixes) if parent else set()
 
         if obj is None and parent is not None:
             obj = parent.obj
@@ -1385,6 +1387,7 @@ class Command(BaseCommand):
             )
 
         ctx.args = args
+        ctx._opt_prefixes.update(parser._opt_prefixes)
         return args
 
     def invoke(self, ctx: Context) -> t.Any:
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -448,17 +448,16 @@ def _is_incomplete_argument(ctx: Context, param: Parameter) -> bool:
     )
 
 
-def _start_of_option(value: str) -> bool:
+def _start_of_option(ctx: Context, value: str) -> bool:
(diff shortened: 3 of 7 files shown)


davidism:
How does this affect options without values?

spanglerco:
> How does this affect options without values?

Flag options are unaffected (`_is_incomplete_option` [returns False early if is_flag](https://github.com/pallets/click/blob/8.0.x/src/click/shell_completion.py#L470)).

Count options do end up unintentionally taking over the completion for the next arg. That could be easily fixed by extending the previously mentioned `is_flag` check to `if param.is_flag or param.count`.

Are there other kinds of options without values? Those are the only two cases I've seen in the documentation.

spanglerco:
Ah, I guess there's also options with optional values. This change also makes claim the next arg for completion. If I understand correctly, the parser [collects all of the possible prefixes from each option](https://github.com/pallets/click/blob/8.0.x/src/click/parser.py#L310) and then decides whether [an arg is an optional value vs. another option based on those prefixes](https://github.com/pallets/click/blob/8.0.x/src/click/parser.py#L485).

Should the shell_completions do the same, using the collected prefixes from the parser?

spanglerco:
@davidism I pushed a rewrite of the fix to instead grab the opt_prefixes from the parser and store them into the context. Then the shell completion can use that information to more closely match what the parser will do.

> How does this affect options without values?

The new implementation should treat them the same as the parser: if the incomplete string starts with a symbol used by any of the options, it will complete as a new option. Otherwise, it will complete as the value of the current option. See the added tests in test_shell_completion.py.

spanglerco:
Hi @davidism is there any interest in this PR or anything else I could help answer about it?

PR #2064, opened 2021-09-11
Fix type annotation for `type` arg

### LINKED ISSUE #2062: mypy raises errors when using `click.prompt()` due to `type` argument having incompatible type
I have a project that uses the latest version of mypy (0.910) and click (8.0.1). I am attempting to use `click.prompt()` and pass in the `type` argument but this is always causing mypy (with default settings) to find errors.

Here is an example:
```python
import click
name = click.prompt('What is your name?', type=str)
age = click.prompt('What is your age?', type=int)
```

When I run mypy against this file, I get two errors:
```
main.py:2: error: Argument "type" to "prompt" has incompatible type "Type[str]"; expected "Optional[ParamType]"
main.py:3: error: Argument "type" to "prompt" has incompatible type "Type[int]"; expected "Optional[ParamType]"
```

To address these, I tried passing in some different values for `type` but to no avail (same issues all happen for `int` as well):
```python
name = click.prompt('What is your name?', type=Optional[str])
# Returns error: Argument "type" to "prompt" has incompatible type "object"; expected "Optional[ParamType]"

name = click.prompt('What is your name?', type=StringParamType)
# Returns error: Argument "type" to "prompt" has incompatible type "Type[StringParamType]"; expected "Optional[ParamType]"

name = click.prompt('What is your name?', type=Optional[StringParamType])
# Returns error: Argument "type" to "prompt" has incompatible type "object"; expected "Optional[ParamType]"
```

Is this a bug or should I be providing something different to `type`?

Environment:

- Python version: 3.7.10
- Click version: 8.0.1
- mypy version: 0.910

Yourun-proger:
- fixes #2062

Checklist:

- [x]  Run `pytest` and `tox`, no tests failed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,8 @@ Unreleased
     to be relative to the containing directory. :issue:`1921`
 -   Completion does not skip Python's resource cleanup when exiting,
     avoiding some unexpected warning output. :issue:`1738, 2017`
+-   Fix type annotation for ``type`` argument in ``prompt`` function.
+    :issue:`2062`
 
 
 Version 8.0.1
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -84,7 +84,7 @@ def prompt(
     default: t.Optional[t.Any] = None,
     hide_input: bool = False,
     confirmation_prompt: t.Union[bool, str] = False,
-    type: t.Optional[ParamType] = None,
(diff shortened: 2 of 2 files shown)


Yourun-proger:
True don't know how i can fix typing error. May be `# type:ignore` or no?

PR #2075, opened 2021-09-29
Update utils.py

ltbam:
On some usage 'file' object can be left at 'None' which in my case prevents Flask to start correctly.


davidism:
I don't understand this change. `file` can indeed be omitted from the call, in which case it defaults to `stdout`. This is not an issue.

PR #2138, opened 2021-11-21
Link code examples to reference documentation

felix-hilden:
Hi! This PR makes Click's code examples clickable (heh.) to take the user to reference documentation. The commit does a few things:

* introduce and configure [`sphinx-codeautolink`](https://sphinx-codeautolink.rtfd.io)
* skip the parsing of a single decorator example to avoid a syntax error
* update and pin docs requirements (on Python 3.9)

The result looks like this:

home page

So clicking in this case would take you to the reference entry of `click.option`. I think it improves navigation quite a bit when you're browsing examples.

I'm currently developing that extension and spent the weekend trying out a few big projects, Click being one of them. This time the setup was so easy that I couldn't resist just a cold PR 😄 If this isn't to your liking, feel free to shoot me down and close the PR right away! Or if any modifications are needed, I'd be happy to make them 👍 or even set up a demo RTD site if you'd like. I verified (to the best of my abilities) that the links are indeed everywhere they should be and that the build is completely clean.

Lemme know what you think!

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -23,10 +23,12 @@ extensions = [
     "pallets_sphinx_themes",
     "sphinx_issues",
     "sphinx_tabs.tabs",
+    "sphinx_codeautolink",
 ]
 autodoc_typehints = "description"
 intersphinx_mapping = {"python": ("https://docs.python.org/3/", None)}
 issues_github_path = "pallets/click"
+codeautolink_global_preface = "import click"
 
 # HTML -----------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


felix-hilden:
Seems like I don't have to create a RTD example since you have already enabled the PR builds! But for some reason the extension isn't online. Have I missed something in the workflow?

davidism:
This is cool, sort of like a "go to definition" feature in an IDE. However, I'm not really sure it's making the code examples more helpful. It adds links for all the built-in modules and functions too, and doesn't seem to be able to link to attributes or methods on Click objects, only the top-level functions, which means a lot of the interesting bits of the code examples go unlinked. I think for now I'm not going to add this, but I'll keep it in mind.

felix-hilden:
Thanks for having a look! And a fair assessment. It is because the extension doesn't execute code and we don't currently follow decorator return type hints that the links are only produced for top-level functions. Would you like for me to resubmit if we get that working?

PR #2139, opened 2021-11-23
Minor doc fix (o not?): Usage of `!!!!111` as a suffix for `shout` examples looks a typo

dcaba:
But maybe not, and there is a historical reason, joke, or something else I may be missing because of my ignorance... Already checked the commit history, and nothing made me think that this was intentional.

### DIFF
--- a/docs/options.rst
+++ b/docs/options.rst
@@ -265,7 +265,7 @@ Example:
     def info(shout):
         rv = sys.platform
         if shout:
-            rv = rv.upper() + '!!!!111'
+            rv = rv.upper() + '!!!!!!!'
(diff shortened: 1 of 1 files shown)


dcaba:
I guess my sense of humor or capacity to identify it is poor, and now I feel that this may be intentional / when people is shouting may make typos like this.. so closing, apologies for the noise

ThiefMaster:
I think this is intentional, considering it's an example and if you want to shout on purpose in text form without being *too* serious, it's kind of common to write "!!!!11" as if you are angry and hammering the keyboard without managing to keep SHIFT pressed all the time.

PR #2150, opened 2021-12-08
Store raw help string for commands

### LINKED ISSUE #2149: Access untruncated help string
There's an [open feature request against sphinx-click](https://github.com/click-contrib/sphinx-click/issues/56), asking that we stop ignoring truncation of help strings when generating the documentation. Unfortunately, a quick look suggests [this is not currently possible](https://github.com/click-contrib/sphinx-click/issues/56#issuecomment-988996878) because we don't store the help string before we truncate it.

This RFE ask that we provide some mechanism to access the raw help string. I suspect this will involve either (a) storing the "raw" help string in a different attribute or (b) storing this in the `help` attribute and truncating on load rather than on store.

stephenfin:
Some tools, such as [sphinx-click](https://github.com/click-contrib/sphinx-click/issues/56), may wish to access the full help string for a command including any text after the form feed character ([which indicates truncation](https://click.palletsprojects.com/en/latest/documentation/#truncating-help-texts)). Make this possible by storing this help string before truncation.

An alternative to this approach would be store the untruncated string in `help` and simply truncate on load. That's technically a breaking change. Another one is to do nothing, and I can close the RFE against sphinx-click.

- fixes #2149

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,8 @@ Unreleased
 
 -   Single options boolean flags with ``show_default=True`` only show
     the default if it is ``True``. :issue:`1971`
+-   Store untruncated help string in ``Command.raw_help`` attribute.
+    :issue:`2149`
 
 
 Version 8.0.4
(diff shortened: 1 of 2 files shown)


stephenfin:
https://github.com/pallets/click/issues/2149#issuecomment-989083217

PR #2151, opened 2021-12-09
Store raw help string for commands

### LINKED ISSUE #2149: Access untruncated help string

stephenfin:
Some tools, such as [sphinx-click](https://github.com/click-contrib/sphinx-click/issues/56), may wish to access the full help string for a command including any text after the form feed character ([which indicates truncation](https://click.palletsprojects.com/en/latest/documentation/#truncating-help-texts)). Make this possible by storing this help string before truncation. Truncation is achieved by the users of the attribute.

This is an alternative to #2150, based on feedback on the issue, #2149

- fixes #2149

Checklist:

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1156,6 +1156,10 @@ class Command(BaseCommand):
                              the command is deprecated.
 
     .. versionchanged:: 8.1
+        ``help`` is stored untruncated, the formatter applied
+        truncation later.
+
+    .. versionchanged:: 8.1
         Indentation in ``help`` is cleaned here instead of only in the
         ``@command`` decorator. ``epilog`` and ``short_help`` are also
         cleaned.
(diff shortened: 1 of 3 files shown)


src/click/core.py
```diff
@@ -1330,6 +1330,10 @@ def format_help_text(self, ctx: Context, formatter: HelpFormatter) -> None:
         """Writes the help text to the formatter if it exists."""
         text = self.help or ""
 
+        # truncate the help text to the content preceding the first form feed,
+        # if any
+        text = text.split("\f", 1)[0]
```
stephenfin, line 1335:
```python
>>> 'test'.split('\f', 1)[0]
'test'
>>> 'test\f'.split('\f', 1)[0]
'test'
```

davidism:
Added another commit to defer `inspect.cleandoc()` for `help`, `epilog`, and `short_help`.

PR #2155, opened 2021-12-17
Fix TypeVar for decorators that handle Commands

euresti:
The old type var would throw away the types of Callables.

```
import click

@click.option("--name", prompt="Your name", help="The person to greet.")
def hello(name: str) -> None:
    ...

reveal_type(hello)   #  note: Revealed type is 'def (*Any, **Any) -> Any'
```

With this small adjustment mypy now knows that it's a `'def (name: builtins.str)'`

Checklist:


davidism:
That's sort of better, but it's still incorrect, the return value of the `@command` and `@group` decorators is `Command` and `Group` instances, not the original function.

PR #2173, opened 2022-01-10
Fix typing issue in `click.secho`

vrslev:
[Pyright](https://github.com/microsoft/pyright) tells that there's type checking issue: IO descriptors are not specified.

This PR fixes it.

### DIFF
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -604,7 +604,7 @@ def unstyle(text: str) -> str:
 
 def secho(
     message: t.Optional[t.Any] = None,
-    file: t.Optional[t.IO] = None,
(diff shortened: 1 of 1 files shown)


src/click/termui.py
```diff
@@ -604,7 +604,7 @@ def unstyle(text: str) -> str:
 
 def secho(
     message: t.Optional[t.Any] = None,
-    file: t.Optional[t.IO] = None,
+    file: t.Optional[t.IO[t.Any]] = None,
```
Jonxslays, line 607: You could look at using `typing.AnyStr` here, rather than just a blanket `typing.Any`.

davidism:
Closing in favor of #2175

PR #2187, opened 2022-02-07
Fix type for `file` argument in `click.echo` function

### LINKED ISSUE #2185: click.echo is improperly typed
I'm getting a repeat of #2174 : although click.secho has been fixed, pyright is continuing to complain about the type annotation for click.echo 

#2175 only fixes the problem for click.secho, I think the same should be done for click.echo?

(Running with 1c588834)

Yourun-proger:
- fixes #2185

Checklist:


Yourun-proger:
Oops. Do I need `t.Any`?

Yourun-proger:
It seems that I fixed everything and the tests passed. Let me know if the typing needs further clarification.

PR #2212, opened 2022-03-11
Improve type hints

### LINKED ISSUE #2211: Incorrect type hint when @command uses a custom class
The `@command` decorator is annotated as always returning `Command`.

This is a problem when supplying custom class, e.g.:
```python
@click.command(cls=MyClass)
def foo():
     ...

reveal_type(foo)  # revealed type is `Command`
foo.my_property = 6  # error: `Command` does not have a member `my_property`
```

Environment:

- Python version: any
- Click version: 8.x

matejcik:
This PR improves the typpe annotation on `@command` decorator so that when `cls` is submitted, the return value is an instance of `cls`.

In addition, as a small tweak, I lowered the requirement on `choices` of `Choice` to `Collection` -- we only need iteration and length. This way `Choice(choices=somedict.keys())` is valid.

- fixes #2211

Checklist:

- [ ] (n/a) Add tests that demonstrate the correct behavior of the change. Tests should fail without the change.
- [ ] (n/a) Add or update relevant docs, in the docs folder and in code.
- [ ] (n/a) Add `.. versionchanged::` entries in any relevant code docs.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -43,6 +43,8 @@ Version 8.1.0
 -   It's possible to pass a list of ``params`` to ``@command``. Any
     params defined with decorators are appended to the passed params.
     :issue:`2131`.
+-   ``@command`` decorator is annotated as returning the correct type if
+    a ``cls`` argument is used. :issue:`2211`
 
 
 Version 8.0.4
--- a/src/click/decorators.py
+++ b/src/click/decorators.py
@@ -121,19 +121,38 @@ def pass_meta_key(
     return decorator
 
 
+CmdType = t.TypeVar("CmdType", bound=Command)
+
+
 @t.overload
 def command(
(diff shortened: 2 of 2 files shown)


src/click/types.py
```diff
@@ -234,7 +234,7 @@ class Choice(ParamType):
 
     name = "choice"
 
-    def __init__(self, choices: t.Sequence[str], case_sensitive: bool = True) -> None:
+    def __init__(self, choices: t.Collection[str], case_sensitive: bool = True) -> None:
```
davidism, line 237: Please revert this. Requiring a sequence was deliberate on my part, as the order may matter for display purposes.

matejcik: easy enough; force-pushed back to a0bbbf2

PR #2215, opened 2022-03-17
postpone referencing sys.modules["__main__"]

adriangb:
This currently breaks anyio (and presumably other things that do multiprocessing stuff) _just from an import_. 
To reproduce:

```python
from time import sleep

from anyio import to_process, run
import click  # this import breaks the program

async def main() -> None:
    await to_process.run_sync(sleep, 0)

if __name__ == "__main__":
    run(main)
```

cc @agronholm

### DIFF
--- a/src/click/utils.py
+++ b/src/click/utils.py
@@ -475,7 +475,7 @@ class PacifyFlushWrapper:
 
 
 def _detect_program_name(
-    path: t.Optional[str] = None, _main: ModuleType = sys.modules["__main__"]
+    path: t.Optional[str] = None, _main: t.Optional[ModuleType] = None
 ) -> str:
     """Determine the command used to run the program, for use in help
     text. If a file or entry point was executed, the file name is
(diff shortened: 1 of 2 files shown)


src/click/utils.py
```diff
@@ -474,8 +474,11 @@ def __getattr__(self, attr: str) -> t.Any:
         return getattr(self.wrapped, attr)
 
 
+_unset = ModuleType("_click_unset")
+
+
 def _detect_program_name(
-    path: t.Optional[str] = None, _main: ModuleType = sys.modules["__main__"]
+    path: t.Optional[str] = None, _main: ModuleType = _unset
```
adriangb, line 481: Looks like `_main=None` is being passed in by a test. But if I fix the type annotation, then I get an error for accessing an attribute of None. Without understanding the logic of the function, I get the feeling that either the test is wrong of the type annotation is wrong. If it's the former, then fixing it in this MR makes sense because the default can then be None (instead of a faux module). If it's the latter, I can do it in this PR if you need me to or punt it for a future PR.

PR #2218, opened 2022-03-19
`ParamType.to_info_dict()` works with no name

### LINKED ISSUE #2168: `ParamType` missing `name` errors on `to_info_dict`
When calling `to_info_dict` on a `ParamType` instance which doesn't have a `name` attribute, an exception is raised. While this is a bug in the downstream code, only `to_info_dict` is affected, so the bug often goes unnoticed.

```py3
class TestType(click.ParamType):
    pass

TestType().to_info_dict()
```

```
>       return {"param_type": param_type, "name": self.name}
E       AttributeError: 'TestType' object has no attribute 'name'

src/click/types.py:66: AttributeError
```

It would be nice to have some sort of reasonable fallback. I've put coded such possibility at https://github.com/pallets/click/compare/main...phy1729:fix-ParamType-no-name . I also fixed where the docs neglected to specify `name` in an example.

Environment:

- Python version: Python 3.9.9
- Click version: 8.0.3

davidism:
fixes #2168

### DIFF
--- a/docs/shell-completion.rst
+++ b/docs/shell-completion.rst
@@ -132,6 +132,8 @@ with the incomplete value.
 .. code-block:: python
 
     class EnvVarType(ParamType):
+        name = "envvar"
+
         def shell_complete(self, ctx, param, incomplete):
             return [
                 CompletionItem(name)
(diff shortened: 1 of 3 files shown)


davidism:
Remaking this since I couldn't push more changes.

davidism:
https://github.com/pallets/click/pull/2219

PR #2265, opened 2022-04-29
Use t.Mapping not t.Dict for parameters

### LINKED ISSUE #2264: Unable to properly inherit and hint Group due to use of Dict for parameters
I am working on updating to Click 8 in https://github.com/Chia-Network/chia-blockchain/pull/11361 and wanted to make `windows_expand_args=True` default for all my uses and have a single place to apply any other adjustments I need at that level.  I wrote a custom subclass of `Group` as below to achieve that, as well as a similar subclass of `Command`.  The issue is that when I hint that only my `_Command` subclass should be passed mypy complains due to the variance of `Dict`.  Modifying `click.Group.__init__()` to accept `Mapping[str, Command]` instead of `Dict[str, Command]` resolves the issue.  I will submit a PR immediately following this.

```
chia/util/click.py:22: error: Argument "commands" to "__init__" of "Group" has incompatible type "Union[Dict[str, _Command], Sequence[_Command], None]"; expected "Union[Dict[str, Command], Sequence[Command], None]"  [arg-type]
```

```python
class _Group(click.Group):
    def __init__(
        self,
        name: Optional[str] = None,
        commands: Optional[Union[Dict[str, _Command], Sequence[_Command]]] = None,
        **attrs: Any,
    ) -> None:
        # _commands = cast(Optional[Union[Dict[str, click.Command], Sequence[click.Command]]], commands)
        super().__init__(name=name, commands=commands, **attrs)
        self.command_class = _Command

    def main(
        self,
        args: Optional[Sequence[str]] = None,
        prog_name: Optional[str] = None,
        complete_var: Optional[str] = None,
        standalone_mode: bool = True,
        **extra: Any,
    ) -> Any:
        return super().main(
            args=args,
            prog_name=prog_name,
            complete_var=complete_var,
            standalone_mode=standalone_mode,
            windows_expand_args=True,
            **extra,
        )
```

Environment:

- Python version: 3.10.1
- Click version: 8.1.3

altendky:
https://docs.python.org/3/library/typing.html#typing.Dict
> To annotate arguments it is preferred to use an abstract collection type such as Mapping.

This PR does almost that to fix the issue below.  I instead use `MutableMapping` here since we do mutate in some cases.  I chose to use `MutableMapping` everywhere for consistency and leaving it open for more mutation in the future.  I understand that this may not be preferred and would be happy to use `Mapping` where possible instead.

- fixes #2264

It was suggested I could follow https://github.com/pallets/click/pull/2256 as a guide for how much of this checklist applies to hint-only changes.  As such some have been skipped.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Version 8.1.4
 
 Unreleased
 
+-   Replace all ``typing.Dict`` occurrences to ``typing.MutableMapping`` for
+    parameter hints.  :issue:`2255`
 -   Improve type hinting for decorators and give all generic types parameters.
     :issue:`2398`
 
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1804,7 +1804,9 @@ class Group(MultiCommand):
     def __init__(
         self,
         name: t.Optional[str] = None,
-        commands: t.Optional[t.Union[t.Mapping[str, Command], t.Sequence[Command]]] = None,
+        commands: t.Optional[
+            t.Union[t.Mapping[str, Command], t.Sequence[Command]]
+        ] = None,
         **attrs: t.Any,
     ) -> None:
         super().__init__(name, **attrs)
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -264,7 +264,7 @@ class Context:
         info_name: t.Optional[str] = None,
         obj: t.Optional[t.Any] = None,
         auto_envvar_prefix: t.Optional[str] = None,
-        default_map: t.Optional[t.Dict[str, t.Any]] = None,
+        default_map: t.Optional[t.Mapping[str, t.Any]] = None,
         terminal_width: t.Optional[int] = None,
         max_content_width: t.Optional[int] = None,
         resilient_parsing: bool = False,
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -5,6 +5,8 @@ Version 8.2.0
 
 Unreleased
 
+-   Switch from ``typing.Dict`` to ``typing.MutableMapping`` for
```
saroad2, line 8:
```suggestion
-   Replace all ``typing.Dict`` occurrences to ``typing.MutableMapping`` for
```

saroad2 (changes requested):
Hey @altendky , Thanks for the PR and sorry for the long delay!

This PR makes sense to me. I fixed some merging issues and we are ready to merge.
The only thing that I think worth fine tuning is the change note. I wrote a suggestion, but if you have better one - please feel free to paraphrase.

After this is fixed, we could merge!

altendky:
Your suggestion reads well to me.  Thanks for your interest and effort on this.

saroad2:
I'm now rethinking that this PR might should be merged into *8.1.x* instead of *main*.

@davidism , I would appriciate your input on it.

davidism:
Yes, this can go in 8.1.x.

Issue #2267, opened 2022-05-02
investigate and improve startup speed


davidism:
Use `python -X importtime` to analyze the start up time based on the time it takes to import modules. Let's see how much we can move into function-level imports to make the initial startup faster. Tools like `tuna` can analyze the `importtime` output, and there are also other profilers, including memray, that may show additional places for speedup.

paxnovem:
I can pick up this issue.

PR #2270, opened 2022-05-02
Delay evaluation of type hint imports

paxnovem:
This PR reduces the time at startup that is taken at import time. This is achieved by delaying the evaluation of type imports.

- https://github.com/pallets/click/issues/2267

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,9 @@ Unreleased
         unneeded. Parsing works directly without building a separate parser.
     -   ``split_arg_string`` is moved from ``parser`` to ``shell_completion``.
 
+-   Enable deferred evaluation of annotations with
+    ``from __future__ import annotations``. :pr:`2270`
+
 
 Version 8.1.7
 -------------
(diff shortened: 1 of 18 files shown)


saroad2:
Since this PR was written, the PSF has decided to go on a different direction than `from __future__ import annotations`. You can read about it [here](https://discuss.python.org/t/type-annotations-pep-649-and-pep-563/11363). Right now they consider moving towards [PEP 649](https://peps.python.org/pep-0649/) instead.

Unfortunately, I think we might should drop this PR. @davidism , would love to hear your opinion on it.

davidism:
`from __future__ import annotations` will continue to work as it does, and any change will still result in deferred evaluation of annotations.

davidism:
Updated to use modern typing syntax and features. Both MyPy and PyRight support using modern syntax with older Pythons when using `from __future__ import  annotations`.

* Remove `typing_extensions` where possible.
* Use shorthand union syntax `|` instead of `Union` and `Optional`.
* Use generic types from built-ins, `collections, `collections.abc` and `contextlib`.
* Quote `TypeVar` `bound` and `cast` types to avoid runtime evaluation of generics and unions.

PR #2275, opened 2022-05-03
Normalize dots in script names

### LINKED ISSUE #2166: Shell completion doesn't work on scripts with a dot in name
I found out, that a shell completion doesn't work on scripts with a dot in their name, mainly the scripts themselves (without an entry point linked) or entry points with a dot in their name (less common). The main problem is that the dot is copied into the name of the completion variable.

MWE:
Let's have a script `/tmp/example.py`
```python
import click


@click.command()
def main():
    pass


if __name__ == '__main__':
    main()
```

Completion on this script doesn't do anything using any of reasonable variable names:
```sh
_EXAMPLE_COMPLETE=bash_source python3 /tmp/example.py
_EXAMPLEPY_COMPLETE=bash_source python3 /tmp/example.py
_EXAMPLE_PY_COMPLETE=bash_source python3 /tmp/example.py
```

I found out, click requires the variable `_EXAMPLE.PY_COMPLETE` to be set, which is not possible (at least) in bash.

It would help if the dot was replaced by an underscore, the same way a dash is handled.

Environment:

- Python version: 3.9.9
- Click version: 8.0.3

nicksspirit:
- fixes #2166 

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,7 @@ Unreleased
 -   Fix return value and type signature of `shell_completion.add_completion_class`
     function. :pr:`2421`
 -   Bash version detection doesn't fail on Windows. :issue:`2461`
+-   Completion works if there is a dot (``.``) in the program name. :issue:`2166`
 
 
 Version 8.1.3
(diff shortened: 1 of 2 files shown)


hmaarrfk:
Who! Thank you for this. I was hitting the same issue. I hope that it gets reviewed soon!

PR #2286, opened 2022-05-13
flush when calling `clear`, use colorama on Windows

### LINKED ISSUE #2284: add the option to click.clear() to use "clear" instead of printing "\033[2J\033[1;1H"
Hi! I'm writing a small script and I'm using `click` for user input and console text formatting. In practice, I use the `click.getchar()` function to read a single char from stdin, and do some processing from there; I construct a string representing a visual representation of the output and I print it. But! Whenever I use `click.clear()` instead of `os.system('clear')`, I get an inconsistent behavior: the terminal doesn't get cleared at the beginning of the script,

Here is a minimal reproducible example (imports omitted for brevity): `script1.py`
```
def main():
    os.system('clear')
    c = click.getchar()

if __name__ == '__main__':
    main()
```
`python3 script1.py` -> the terminal gets cleared and there's only the blinking cursor; this is the expected behavior. Instead, here's `script2.py`:
```
def main():
    click.clear()
    c = click.getchar()

if __name__ == '__main__':
    main()
```
`python3 script2.py` -> see: image

The terminal doesn't get cleared, I can still see my username, the command, etc. Obviously this is trivial to circumvent; my implementation is:
```
def clearscreen(use_click=True):
    click.clear() if use_click else os.system('cls' if os.name == 'nt' else 'clear')
```
Something like this could be mainlined in `click.clear()`, for example something like (taking from the current implementation):
```
def clear(scroll: bool = True) -> None:
    if not isatty(sys.stdout):
        return
    if WIN:
        os.system("cls")
    else:
        sys.stdout.write("\033[2J\033[1;1H") if scroll else os.system("clear")
```
Obviously the default value would follow the current behavior for consistency. I hope I've been thorough enough. Thanks!
- Python version: 3.10.4
- Click version: 8.0.3

belliriccardo:
- fixes #2284

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@ Unreleased
 -   Bash version detection doesn't fail on Windows. :issue:`2461`
 -   Completion works if there is a dot (``.``) in the program name. :issue:`2166`
 -   Improve type annotations for pyright type checker. :issue:`2268`
+-   Improve responsiveness of ``click.clear()``. :issue:`2284`
 
 
 Version 8.1.3
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -1,14 +1,12 @@
 import inspect
 import io
 import itertools
-import os
 import sys
 import typing as t
 from gettext import gettext as _
 
 from ._compat import isatty
 from ._compat import strip_ansi
-from ._compat import WIN
 from .exceptions import Abort
 from .exceptions import UsageError
 from .globals import resolve_color_default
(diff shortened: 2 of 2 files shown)


saroad2 (dismissed):
Hey @belliriccardo , thanks for the PR and sorry for the long delay!

I had a single comment on the PR. Once you address this comment, this PR could be merged.

belliriccardo:
Hi! It's been a while but I havent forgotten! Hope the change is ok!

saroad2:
Hey @belliriccardo , thanks for the clarification! Now it is much clearer.

I'm sorry for the long delay. I'm now walking over older PRs and reviewing them, helping out the *pallet* team.

One more thing regarding this PR that I forgot to ask: would you mind to add a test or two for the new behavior?

PR #2335, opened 2022-08-05
Use `t.Collection` instead of `t.Sequence` in `types.Choice`

Viicos:
Using `typing.Collection` instead of `typing.Sequence` allows sized iterables to be used as choices. One example is when using keys of a dictionnary:

```py
@click.option("--choices", type=click.Choice(my_dict.keys()))
```

Right now, mypy is complaining about it:

```sh
error: Argument 1 to "Choice" has incompatible type "dict_keys[str, Union[str, bool]]"; expected "Sequence[str]"
```

Checklist:

- [x] Add tests that demonstrate the correct behavior of the change. Tests should fail without the change. (no changes)
- [x] Add or update relevant docs, in the docs folder and in code. (no changes)
- [x] Add `.. versionchanged::` entries in any relevant code docs. (no changes)

### DIFF
--- a/src/click/types.py
+++ b/src/click/types.py
@@ -241,7 +241,7 @@ class Choice(ParamType):
 
     name = "choice"
 
-    def __init__(self, choices: t.Sequence[str], case_sensitive: bool = True) -> None:
+    def __init__(self, choices: t.Collection[str], case_sensitive: bool = True) -> None:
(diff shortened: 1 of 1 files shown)


davidism:
I don't want to allow that though, the current type is intentional.

PR #2374, opened 2022-10-13
Suppress stderr

Freed-Wu:
Checklist:


davidism:
I'm not clear why this is needed. It seems important that any errors in our completion are shown so that users can report issues.

PR #2405, opened 2022-11-16
Default catch_exceptions in CliRunner to False

corey-nm:
Currently, exceptions are silently caught by CliRunner.invoke, and the only way to know about these is through the documentation. By defaulting catch_exceptions to False, exceptions will be re-raised automatically, and developers will have to explicitly opt-in to populate the .exception field of CliRunner

### DIFF
--- a/src/click/testing.py
+++ b/src/click/testing.py
@@ -352,7 +352,7 @@ class CliRunner:
         args: t.Optional[t.Union[str, t.Sequence[str]]] = None,
         input: t.Optional[t.Union[str, bytes, t.IO]] = None,
         env: t.Optional[t.Mapping[str, t.Optional[str]]] = None,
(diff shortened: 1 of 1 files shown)


davidism:
There are a few test patterns in Click:

1. Invoke a command, assert its return code and output.
2. Invoke a command, assert if an error was caught that would have resulted in a 1 return code.
3. Invoke a command, let the error raise in the test directly.

1 allows for testing a CLI the way it will be perceived by the user. 2 allows a finer grained check over _why_ it failed, along with 1. 3 is more like "normal" Python testing, where you want to fail directly.

All are valid, and I think 1 still makes more sense as the default since it mirrors what the output will actually be to a user. Regardless of which one anyone thinks is "right", it should be clear from the number of test failures here why we can't change this now.

PR #2420, opened 2022-12-24
Fix return value of `shell_completion.add_completion_class` function

alexreg:
Decorators should return a possibly modfiied version of the original type. Hence, `add_completion_class` should return `cls` rather than merely `None`. This conforms to Python conventions, and allows type checking to work properly.

- fixes #<issue number>

Checklist:

### DIFF
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -398,7 +398,7 @@ _available_shells: t.Dict[str, t.Type[ShellComplete]] = {
 
 def add_completion_class(
     cls: t.Type[ShellComplete], name: t.Optional[str] = None
-) -> None:
+) -> t.Type[ShellComplete]:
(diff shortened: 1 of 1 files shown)


davidism:
This isn't intended to be a decorator.

alexreg:
@davidism Then you need to fix the docs. See `docs/shell-completion.rst`.

davidism:
Happy to review a PR

davidism:
Never mind, I see what happened now. This needs a changelog, and should be targeted at the 8.1.x branch as a bugfix.

davidism:
Also, if this is intended to fix type checking as well, then this is not correct yet. You need to mark the argument and return value with a bound typevar.

alexreg:
@davidism Okay

alexreg:
I'll close this an open a bugfix PR against the 8.1.x branch.

PR #2421, opened 2022-12-24
Fix return value and type signature of `shell_completion.add_completion_class` function

alexreg:
Decorators should return a possibly modfiied version of the original type. Hence, add_completion_class should return cls rather than merely None. This conforms to Python conventions, and allows type checking to work properly.

Checklist:

### DIFF
--- a/tests/test_shell_completion.py
+++ b/tests/test_shell_completion.py
@@ -4,10 +4,10 @@ from click.core import Argument
 from click.core import Command
 from click.core import Group
 from click.core import Option
+from click.shell_completion import _available_shells
+from click.shell_completion import add_completion_class
 from click.shell_completion import CompletionItem
 from click.shell_completion import ShellComplete
-from click.shell_completion import add_completion_class
-from click.shell_completion import _available_shells
 from click.types import Choice
 from click.types import File
 from click.types import Path
--- a/tests/test_shell_completion.py
+++ b/tests/test_shell_completion.py
@@ -6,6 +6,8 @@ from click.core import Group
 from click.core import Option
 from click.shell_completion import CompletionItem
 from click.shell_completion import ShellComplete
+from click.shell_completion import add_completion_class
+from click.shell_completion import _available_shells
 from click.types import Choice
 from click.types import File
 from click.types import Path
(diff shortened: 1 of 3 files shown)


alexreg:
Pre-commit check failures seem to concern pre-existing issues.

alexreg:
Hi @davidism. Is this okay for you?

saroad2:
Hey @alexreg , thanks for the PR.

While this PR is pretty straight-forward, you didn't include any tests to verify it.
Can you add some tests to check that everything works smoothly?

alexreg:
@saroad2 Sure. I added a couple of tests for the return value.

tests/test_shell_completion.py
```diff
@@ -342,3 +344,25 @@ def test_choice_case_sensitive(value, expect):
     )
     completions = _get_words(cli, ["-a"], "a")
     assert completions == expect
+
+
+def test_add_completion_class():
+    _available_shells.clear()
```
saroad2, line 350: This is a risky move.

By clearing `_available_shells` you are clearing also all the pre-defined shells that's in there before the test.

A better way to go is to:
1. Save the previous content of `_available_shells` in a separted list
2. Clear `_available_shells`
3. Run the test
4. Insert back the shells you saved

In that way, your test will not effect other tests.
This is a perfect example for the use of [pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html). If you are femiliar with those, you can create a fixture that uses the keyword `yield` for before-test and after-test effects.

tests/test_shell_completion.py
```diff
@@ -342,3 +344,25 @@ def test_choice_case_sensitive(value, expect):
...
+        source_template = "dummy source"
+
+    assert add_completion_class(MyshComplete) is MyshComplete
+    assert _available_shells[MyshComplete.name] is MyshComplete
+
+
+def test_add_completion_class_decorator():
+    _available_shells.clear()
+
+    @add_completion_class
```
saroad2, line 363: Very important test. Thank you!

saroad2 (changes requested):
Looks good.

I had one comment about the tests regarding the usage of `_available_shells.clear()`. If you have any questions about my comment, please feel free to ask.

tests/test_shell_completion.py
alexreg: Thanks for your prompt review and for the suggestion. Indeed, clearing the dict of pre-defined shells gave me a bad feeling too... It was just a crude way to ensure these two tests didn't interfere with each other.

I didn't know about pytest fixtures before, but was glad to learn something! (And they seem very straightforward.) I've just pushed a new commit that uses a fixture, and hopefully that looks good to you as well.

saroad2:
Hey @alexreg , thank you for reposing my review. I'm glad I was able to introduce you to new concepts.

I've added some small tweaks to your PR, mainly a small refactoring for the fixture, elaboration on the steps of the tests you wrote, and add one more test about using `add_completion_class` with the `name` parameter.

Happily merging the PR and hope to see your contributions again in the future!

alexreg:
@saroad2 Looks good, re the tests. I'm not sure why I treated the dict like a value type in the fixture, but glad you fixed that too... Cheers.

PR #2432, opened 2023-01-12
document max wrapping width

### LINKED ISSUE #2253: Help text wrapping set to 80 irrespective of the terminal width
According to the [documentation](https://click.palletsprojects.com/en/8.1.x/documentation/#preventing-rewrapping) click should:

> The default behavior of Click is to rewrap text based on the width of the terminal

At least on a mac terminal, that's not the case.  For me, the output of `shutil.get_terminal_size()` is `os.terminal_size(columns=210, lines=44)`.  But click is wrapping all text to 80 characters and does not respond to resizes of the terminal before or during the program's execution.

This is a little jarring and it would be nice if this could either dynamically resize or offer a way to manually set it.

Environment:

- Python version: 3.9
- Click version: 8.1.2

couling:
The primary issue of #2253 was not so much the limit bit the discrepancy between documentation and behaviour.  This PR documents the 80 character limit and also incorporates the helpful response in the issue on how to change the limit.

- fixes #2253

### DIFF
--- a/docs/documentation.rst
+++ b/docs/documentation.rst
@@ -85,8 +85,9 @@ Preventing Rewrapping
 ---------------------
 
 The default behavior of Click is to rewrap text based on the width of the
-terminal.  In some circumstances, this can become a problem. The main issue
-is when showing code examples, where newlines are significant.
+terminal, to a maximum 80 characters. In some circumstances, this can become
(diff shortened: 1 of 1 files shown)


couling:
The build seems to have failed because of something I didn't change?  Not sure what the cause of that is.

PR #2467, opened 2023-03-13
Add support for MultiCommand method based sub-Commands

### LINKED ISSUE #601: use click.command in a class
hey, is it possible to do 

```
class Foo(object):
    @click.command()
    def cmd(self):
        pass
```

this code would fail because click will validate that a "argument self" is needed, is it possible to let click not check this in a easy way?

monkeyusage:
# MultiCommand Class Based methods

This PR enables `click` users to use class MultiCommand methods.
It fixes #601 where users request for a way to use commands that are also Python Object methods.

Such as:

```python

class Foo:
    @command()
    def bar(self):
        print("bar method / command")
```
Currently one might use staticmethods to create MultiCommands that use method based commands. However user might need to access object attributes in their commands.
We propose to enable users to create MultiCommands that also allow access to object attributes by setting the commands inside the __init__ method so the instance can have a reference to self while executing.

Here is example implementation of such a class:

```python
# in file example.py
class Foo(MultiCommand):
    def __init__(self, x: message):
        super().__init__()
        self.message = message
        self.commands = {"echo": click.command(self.echo)}

    # MultiCommand must implement get_command and list_commands
    def get_command(self, ctx: click.Context, cmd_name: str) -> Command:
        return self.commands.get(cmd_name)

    def list_commands(self, ctx: click.Context) -> list[str]:
        return list(self.commands.keys())

    # here is an example method  that we want to use as a click command
    # notice we have access to self and can use click.option and click.argument decorators
    @click.option("--upper", default=False)
    def echo(self, upper: bool) -> None:
        msg = self.message.upper() if upper else self.message
        click.echo(msg) 


# execute the click application
Foo("bar")

# test it using:
# python3 example.py echo --upper
```

- fixes #601

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,13 +1,5 @@
 .. currentmodule:: click
 
-Version 8.2.1
--------------
-
-Unreleased
-
--   Add support for method based commands
-    :issue:`601`
-
 
 Version 8.2.0
 -------------
@@ -25,6 +17,8 @@ Unreleased
 
 -   Improve type hinting for decorators and give all generic types parameters.
     :issue:`2398`
+-   Add support for method based commands
+    :issue:`601`
 
 
 Version 8.1.3
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,5 @@
 .. currentmodule:: click
 
-
 Version 8.2.0
 -------------
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,14 @@
 .. currentmodule:: click
 
+Version 8.2.1
+-------------
+
+Unreleased
+
+-   Add support for method based commands
+    :issue:`601`
+
+
 Version 8.2.0
 -------------
 
--- a/src/click/decorators.py
+++ b/src/click/decorators.py
@@ -226,12 +226,10 @@ def command(
         params = attr_params if attr_params is not None else []
 
         try:
-            decorator_params = f.__click_params__  # type: ignore
+            params.extend(reversed(f.__click_params__))  # type: ignore
+            del f.__click_params__  # type: ignore
         except AttributeError:
             pass
-        else:
-            del f.__click_params__  # type: ignore
-            params.extend(reversed(decorator_params))
 
         if attrs.get("help") is None:
             attrs["help"] = f.__doc__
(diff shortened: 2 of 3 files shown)


davidism:
Closing for the same reason as the linked issue:

> I don't think this is an intended use case. If you want to build a cli in a class, you should probably do that in `__init__` by calling `cli.add_command()` so that the methods are bound.

monkeyusage:
The problem is that regular methods are not bound, they must be static which is not ideal.

davidism:
That's not a problem, that's a design choice. Using instance methods as commands is not an intended design pattern. And there's already a way to use instance methods: call `self.add_command(self.method)` in `__init__`, no changes to or new releases of Click required.

monkeyusage:
We want to extend some unrelated class with click (basically a Mixin) wich is why I am using it this way. 
Thanks a lot for your fast response !

Sharing this since I could not find any examples online. A minimal reproduction (without bothering with Multiple Inheritence) would be:

```python
class Foo(Group):
    def __init__(self, msg: str):
        super().__init__()
        self.msg = msg
        self.commands = {}
        self.add_command(click.command(self.my_command))

    # you can use click.option and click.argument decorators here
    def my_command(self):
        click.echo(self.msg)
        ....
```

monkeyusage:
The problem is that `self.add_command(click.command(self.method))` will not work if the method is decorated with `click.option` or `click.argument` .

It results with an Attribute error telling the user that __click_params__ is not an attribute of the method. The decorators alter the method and the current code does not allow for the solution you propose.

Here is a reproduction of the problem:

```python
import click

class Foo(click.Group):
    def __init__(self, message):
        super().__init__()
        self.message = message
        self.commands = {}
        self.add_command(click.command(self.bar), name="bar")

    @click.option("--loud", default=False)
    def bar(self, loud: bool) -> None:
        msg = self.message if not loud else self.message.upper()
        click.echo(msg)

cli = Foo("methods with options and args cannot run") # raises AttributeError __click_params__
cli()
```

Hence my small modification in `decorators.py`.

davidism:
Again, that's not a problem with Click, because that's not an intended design pattern. The "problem" you're describing is a problem with your chosen code strategy. So if you want to code that way, _you_ need to solve it, not Click. In your example, you would build the commands and parameters in `__init__` (everything in Click can be constructed without decorators).

PR #2509, opened 2023-05-05
Hide default value when show_default is False

pfhayes:
Previously, when `show_default` was False on an `Option`, the value could still appear in the prompt. Remove this so that sensitive values can be hidden from prompts.

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,7 @@ Unreleased
 -   Show the ``types.ParamType.name`` for ``types.Choice`` options within
     ``--help`` message if ``show_choices=False`` is specified.
     :issue:`2356`
+-   Do not display values in prompts when `show_default=False` on Options
 
 
 Version 8.1.8
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,7 +34,7 @@ Unreleased
 -   Show the ``types.ParamType.name`` for ``types.Choice`` options within
     ``--help`` message if ``show_choices=False`` is specified.
     :issue:`2356`
--   Do not display values in prompts when `show_default=False` on Options
+-   Do not display default values in prompts when `show_default=False` on options
 
 
 Version 8.1.8
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,7 +34,8 @@ Unreleased
 -   Show the ``types.ParamType.name`` for ``types.Choice`` options within
     ``--help`` message if ``show_choices=False`` is specified.
     :issue:`2356`
--   Do not display default values in prompts when `show_default=False` on options
+-   Do not display default values in prompts when ``Option.show_default`` is
+    ``False``. :pr:`2509`
 
 
 Version 8.1.8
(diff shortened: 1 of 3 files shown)


src/click/core.py
```diff
@@ -2865,6 +2865,7 @@ def prompt_for_value(self, ctx: Context) -> t.Any:
             show_choices=self.show_choices,
             confirmation_prompt=self.confirmation_prompt,
             value_proc=lambda x: self.process_value(ctx, x),
+            show_default=(False if self.show_default is False else True),
```
saroad2, line 2868: Why not simply: `show_default=self.show_default`?

tests/test_termui.py
```diff
@@ -446,3 +446,13 @@ def cli(password):
 
     if prompt == "Confirm Password":
         assert "Confirm Password: " in result.output
+
+
+def test_show_default_false_hides_prompt(runner):
```
saroad2, line 451: This test name ican be clearer in order to explain what does this test check.

Mybe something like "test_false_show_default_cause_no_default_display_in_prompy".

tests/test_termui.py
```diff
@@ -446,3 +446,13 @@ def cli(password):
...
+    @click.command()
+    @click.option("--arg1", show_default=False, prompt=True, default="my-default-value")
+    def cmd(arg1):
+        pass
+
+    result = runner.invoke(cmd, input="my-input", standalone_mode=False)
```
saroad2, line 457: Maybe add a comment to explain what you are actually checking here.

That might help people in the future that will read your test

saroad2 (changes requested):
Hey, thanks for the contributaion!

This PR makes sense to me. I added some minor comments about the PR.
After fixing those, I could merge it.

src/click/core.py
pfhayes: In this class, `self.show_default` is a `Union[bool, str, None]`, but `termui.prompt` accepts only `bool`. Passing through show_default directly gives a type error, and also appears to do the wrong things for values of type `filenam`

tests/test_termui.py
pfhayes: Done

tests/test_termui.py
pfhayes: Done

src/click/core.py
saroad2: Got it. Thanks for the explanation.

Would you mind writing it down as a comment in the code? I think it would help others who read this line of code for the first time.

CHANGES.rst
```diff
@@ -7,6 +7,7 @@ Unreleased
 
 -   Use modern packaging metadata with ``pyproject.toml`` instead of ``setup.cfg``.
     :pr:`326`
+-   Do not display values in prompts when `show_default=False` on Options
```
saroad2, line 10:
```suggestion
-   Do not display default values in prompts when `show_default=False` on options
```

saroad2 (changes requested):
Thanks for minding my review.

I added two more small suggestions. After you respond to those, we could merge this PR.

saroad2:
Also, this PR doesn't pass pre-commit at the moment. You might want to run `pre-commit run -a` in order to make sure that all files are according to *Click*'s style guideline.

CHANGES.rst
pfhayes: Done.

pfhayes:
Issues fixed. Added comment and also restructured code to make behavior of `show_default` arg to `prompt` more explicit

PR #2510, opened 2023-05-07
Update contributing and support docs.

Rowlando13:
Summary:

- Update link from main icon to point to docs 
- Add basic contributing page 
- Add details about python version 
- Consolidate click-contrib page into python version support page.


saroad2:
I feel like this PR should get the attention of @davidism .

One thing that I can't wrap my head around is the mention of *Beeware*. *Beeware* has almost nothing to do with *click*.

Rowlando13:
@saroad2 I agree. Beeware has nothing to do with click, but they do have very extensive general information about contributing. Maybe my mentioning them gives the wrong idea or I should work it differently.  It is mentioned in on the docs channel in the discord as a good example of beginner docs. 
Does not seem worth to rewrite it since beeware already has nice materials. Since there was not contributing section to start with, I would rather start with the technical pieces that apply to the project.

PR #2515, opened 2023-05-14
Reduce click contrib

Rowlando13:
Reduce scope of click contrib.


Rowlando13:
@davidism. Please take a look.

PR #2526, opened 2023-05-30
Annotate choices for Choice as Collection

### LINKED ISSUE #2525: Argument `choices` of `Choice` constructor should be `Collection`, not `Sequence`
Implementation of `click.Choice` currently requires `choices` argument to match `typing.Sequence[str]` type, while a wider one can be used, namely: `typing.Collection[str]` . `Sequence` adds `__getitem__` which is never used on this argument, `Collection[str]` seems to be sufficient here.

Currently mypy will complain when `set` is used as an argument to `click.Choice`:

```python
@click.command()
@click.option("--select", type=click.Choice({"a", "b"})
def command(select):
    ...
```

error:

```text
error: Argument 1 to "Choice" has incompatible type "Set[str]"; expected "Sequence[str]"  [arg-type]
```

There would be no errors in runtime.

blazewicz:
This PR changes type annotation for `choices` argument on `Choice` from `typing.Sequence[str]` to `typing.Collection[str]`. Additional tests using `list`, `tuple` and `set` were also added.

- fixes #2525

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Unreleased
 
 -   Use modern packaging metadata with ``pyproject.toml`` instead of ``setup.cfg``.
     :pr:`326`
+-   Use ``typing.Collection`` to annotate ``choices`` argument to ``Choice``. :pr:`2526`
 
 
 Version 8.1.4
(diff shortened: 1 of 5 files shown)


davidism:
See discussion in linked issue.

PR #2551, opened 2023-07-04
Separate out virtualenv to its own sections

Rowlando13:
The virtualenv now has a comment that it is recommended in the quickstart and it broken out into its own section. 
Left most of the content unchanged. Moved it because most people won't need an explanation of virtualenv.


docs/virtualenv.rst
```diff
@@ -0,0 +1,45 @@
+.. _virtualenv-heading:
+
+Virtualenv
+=========================
+
+Why use virtualenv?
```
tachyondecay, line 6: These headings should be changed to Title Case for consistency with the rest of the Click docs.

docs/virtualenv.rst
```diff
@@ -0,0 +1,45 @@
...
+-------------------------
+
+You should use `Virtualenv <https://virtualenv.pypa.io/en/latest/>`_ because:
+
+*   It allows you to install multiple versions of the same dependency.
+
+*   If you have an operating system version of python, it prevents you from changing its dependencies and potentially messing up your os.
```
tachyondecay, line 13: Capitalize "Python," "OS" in all caps.

PR #2555, opened 2023-07-06
echo doesn't fail with no streams

### LINKED ISSUE #2415: Error in click.utils.echo() when console is unavailable
The `click.utils.echo()` function does not seem to account for the case when the console is not available on Windows, i.e., when running under `pythonw.exe` interpreter instead of the `python.exe` one.

Minimal example: 

```python
# program.py
import sys
import os

try:
    import click
    click.utils.echo("Hello world")
except Exception:
    import traceback
    error_file = os.path.join(os.path.dirname(__file__), "error_log.txt")
    with open(error_file, "w") as fp:
        traceback.print_exc(file=fp)
```

Running this program with `pythonw.exe program.py` produces the `error_log.txt`:
``` 
  File "C:\Users\Rok\Development\pyi-click\program.py", line 6, in <module>
    click.utils.echo("Hello world")
  File "C:\Users\Rok\Development\pyi-click\venv\lib\site-packages\click\utils.py", line 299, in echo
AttributeError: 'NoneType' object has no attribute 'write'
```

In contrast, standard `print()` function gracefully handles situations when console is unavailable and `sys.stdout` and `sys.stderr` are `None`.

This attempt at retrieving stdout/stderr:
https://github.com/pallets/click/blob/c65c6ad18471448c0fcc59ef53088787288c02cc/src/click/utils.py#L250-L254
should be followed by another `None` check, and if the stream is unavailable, the function should become a no-op (exit immediately).

Environment:

- OS: Windows
- Python version: any
- Click version: 8.1.3 (and earlier)

davidism:
When running with `pythonw` on Windows (GUI mode, no console), `sys.stdout` and `stderr` are `None`. This doesn't appear to be documented by Python. In this case, `echo` should return early since there's nowhere to write to. fixes #2415 

`progressbar` and `pager` also use these, they'll fall back to writing to a temporary `StringIO` since it wasn't clear how to make them exit early.

Presumably if you actually wanted to see output when using `pythonw` you would do something like `sys.stdout = open("output.txt", "w")` at the beginning of your program.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,6 +30,8 @@ Unreleased
     bytes with the replacement character (``�``). :issue:`2395`
 -   Remove unnecessary attempt to detect MSYS2 environment. :issue:`2355`
 -   Remove outdated and unnecessary detection of App Engine environment. :pr:`2554`
+-   ``echo()`` does not fail when no streams are attached, such as with ``pythonw`` on
+    Windows. :issue:`2415`
 
 
 Version 8.1.3
--- a/src/click/_compat.py
+++ b/src/click/_compat.py
@@ -576,12 +576,17 @@ def isatty(stream: t.IO[t.Any]) -> bool:
 
 
 def _make_cached_stream_func(
-    src_func: t.Callable[[], t.TextIO], wrapper_func: t.Callable[[], t.TextIO]
-) -> t.Callable[[], t.TextIO]:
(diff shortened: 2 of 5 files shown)


bwoodsend:
> This doesn't appear to be documented by Python

For future reference, it's [here](https://docs.python.org/3/library/sys.html?highlight=pythonw#sys.__stderr__). It would have been more helpful if that notice was under the regular `sys.std{out,in,err}`.

PR #2559, opened 2023-07-07
Fix decorator typing with _Decorator type alias

### LINKED ISSUE #2558: Typing changes in 8.1.4 cause failure of mypy check
Click's simple example https://github.com/pallets/click#a-simple-example doesn't pass mypy 1.4.1 since 8.1.4.  With click 8.1.3 the example did pass. This affects not only the simple example but also existing projects that use click. I ran mypy on one of my projects and everything passed, ran it 30 minutes later and saw failures out of the blue. Click 8.1.4 was published to PyPI between my checks.

To reproduce:

* Save the simple example as click.py
* `pip install mypy==1.4.1 click==8.1.3`
* `mypy click.py`

This is the output:

```
click.py:3: error: Argument 1 has incompatible type "Callable[[Any, Any], Any]"; expected <nothing>  [arg-type]
click.py:12: error: <nothing> not callable  [misc]
```

Environment:

- Python version: 3.11
- Click version: 8.1.4

sirosen:
`_Decorator[FC]` is identical to `t.Callable[[FC], FC]`. However, `mypy` currently reads the first type incorrectly and supports the second one fully.

This fix will make mypy checking work correctly on click decorators, and should have no negative impact on other type checkers because the types are effectively "the same".

- fixes #2558

I plan to open a bug report against `mypy` to ask about this issue. It may be a duplicate (https://github.com/python/mypy/issues/13250 ?), but we'll have a thread to pull on there to figure out why this isn't supported when there doesn't appear to be a meaningful difference between the two types.

Checklist:

I haven't added tests because any `mypy` testing will require some kind of new framework for testing types.
I think this would be valuable and not hard (e.g. [here's some prior art in marshmallow](https://github.com/marshmallow-code/marshmallow/blob/c5ba8012320f6c1d972d1b048f30728755c84d49/tox.ini#L13-L15) which I helped setup), but it's beyond the scope of a small typing fix.

Instead, I tested `mypy` v1.4.1 (latest) before and after this change on the following small sample:
```python
from click import option

def foo(a: int) -> str:
    return str(a)

reveal_type(option)
foo = option()(foo)
reveal_type(foo)
```

Before the change:
```
$ mypy click_example.py 
click_example.py:8: note: Revealed type is "def [FC <: Union[def (*Any, **Any) -> Any, click.core.Command]] (*param_decls: builtins.str, *, cls: Union[Type[click.core.Option], None] =, **attrs: Any) -> def (FC`-1) -> FC`-1"
click_example.py:9: error: Argument 1 has incompatible type "Callable[[int], str]"; expected <nothing>  [arg-type]
click_example.py:10: note: Revealed type is "<nothing>"
```

After the change:
```
$ mypy click_example.py
click_example.py:8: note: Revealed type is "def (*param_decls: builtins.str, *, cls: Union[Type[click.core.Option], None] =, **attrs: Any) -> def [FC <: Union[def (*Any, **Any) -> Any, click.core.Command]] (FC`-1) -> FC`-1"
click_example.py:10: note: Revealed type is "def (a: builtins.int) -> builtins.str"
Success: no issues found in 1 source file
```

- [x] (N/A) Add or update relevant docs, in the docs folder and in code.

I added a changelog section for 8.1.5 since this is the first change after 8.1.4. I'm happy to make tweaks and changes if this was not appropriate.

- [x] (N/A) Add `.. versionchanged::` entries in any relevant code docs.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,12 @@
 .. currentmodule:: click
 
+Version 8.1.5
+-------------
+
+Unreleased
+
+- Fix type annotations on the option and argument decorators. :issue:`2558`
+
 Version 8.1.4
 -------------
 
--- a/src/click/decorators.py
+++ b/src/click/decorators.py
@@ -21,7 +21,6 @@ if t.TYPE_CHECKING:
 R = t.TypeVar("R")
 T = t.TypeVar("T")
 _AnyCallable = t.Callable[..., t.Any]
-_Decorator: "te.TypeAlias" = t.Callable[[T], T]
 FC = t.TypeVar("FC", bound=t.Union[_AnyCallable, Command])
 
 
@@ -331,7 +330,7 @@ def _param_memo(f: t.Callable[..., t.Any], param: Parameter) -> None:
 
 def argument(
     *param_decls: str, cls: t.Optional[t.Type[Argument]] = None, **attrs: t.Any
-) -> _Decorator[FC]:
+) -> t.Callable[[FC], FC]:
     """Attaches an argument to the command.  All positional arguments are
     passed as parameter declarations to :class:`Argument`; all keyword
     arguments are forwarded unchanged (except ``cls``).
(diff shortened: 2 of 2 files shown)


sirosen:
No need to file a separate `mypy` bug. I'm pretty sure this is https://github.com/python/mypy/issues/13449 , and I've left a comment there to hopefully engage the maintainers there.

davidism:
Thanks for working on this and the good writeup. Looks like #2562 fixes the issue more broadly. Can you check that as well before I close this one, just in case.

PR #2562, opened 2023-07-09
Fix types, introduce type tests

### LINKED ISSUE #2558: Typing changes in 8.1.4 cause failure of mypy check

Tinche:
Fix signatures, introduce simple tests.

I followed the Flask model for typing tests. Initially I was going to replicate the attrs model (https://github.com/python-attrs/attrs/blob/main/tests/test_mypy.yml) but maybe this is good enough to start?

Also introduced Pyright tests, again copied from attrs.


src/click/decorators.py
```diff
@@ -150,16 +150,12 @@ def command(
...
-# However, mypy thinks this doesn't fit the overloaded function. Pyright does
-# accept that spelling, and the following work-around makes pyright issue a
-# warning that CmdType could be left unsolved, but mypy sees it as fine. *shrug*
+# variant: name omitted, cls _must_ be a keyword argument, @command(cls=CommandCls, ...)
 @t.overload
 def command(
     name: None = None,
-    cls: t.Type[CmdType] = ...,
+    *,
+    cls: t.Type[CmdType],
```
layday, line 158: I think the same change should be made to `group`'s third overload.

davidism:
We don't have CI for this yet, but check `pyright --verifytypes click` to make sure pyright can at least understand the types.

Should this be used instead of #2559?

Tinche:
> Should this be used instead of https://github.com/pallets/click/pull/2559?

I've added `help_option` to the typing tests, so this is now a strict superset. So pretty much yeah.

Tinche:
> We were also using pyright --verifytypes click, not sure what the exact difference is but might want to run both.

I've added it. Looks like it prints out a bunch of stuff and succeeds.

```
pyright: commands[0]> pyright --verifytypes click
Module name: "click"
Package directory: "/home/tin/pg/click/.tox/pyright/lib/python3.11/site-packages/click"
Module directory: "/home/tin/pg/click/.tox/pyright/lib/python3.11/site-packages/click"
Path of py.typed file: "/home/tin/pg/click/.tox/pyright/lib/python3.11/site-packages/click/py.typed"

Public modules: 12
   click
   click.core
   click.decorators
   click.exceptions
   click.formatting
   click.globals
   click.parser
   click.shell_completion
   click.termui
   click.testing
   click.types
   click.utils

Symbols used in public interface:

Symbols exported by "click": 646
  With known type: 646
  With ambiguous type: 0
  With unknown type: 0
  Functions without docstring: 98
  Functions without default param: 22
  Classes without docstring: 13

Other symbols referenced but not exported by "click": 247
  With known type: 247
  With ambiguous type: 0
  With unknown type: 0

Type completeness score: 100%

Completed in 1.389sec
```

src/click/decorators.py
Tinche: Sure, but I think it's going to be up to the community to fill out the tests and fix the fallout. This PR is mainly to set up the infrastructure for it.

ofek:
Is this ready to be merged?

davidism:
Yes, I just need to find some time, busy with regular job.

davidism:
I've switched from installing and using nodenv directly to installing the pyright package from PyPI. This allows us to pin the pyright version the same as mypy. I prefer to pin these tools because new versions regularly cause our tests to fail, which is confusing to contributors making PRs. (That said, the failures are usually nice things like being able to remove ignore comments.) Everything seems to work still, but let me know if there was a specific reason for using nodenv directly.

Tinche:
Cool, thanks for the tip. I've actually borrowed this code from attrs, I'm not a pyright expert or anything. Maybe @hynek has an opinion?

PR #2565, opened 2023-07-13
Fix `group` overload

layday:
Additional fixes for #2558 building on @tinche's PR :)

----

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Version 8.1.6
 
 Unreleased
 
+-   Fix an issue with type hints for ``@click.group()``. :issue:`2558`
+
 
 Version 8.1.5
 -------------
--- a/src/click/decorators.py
+++ b/src/click/decorators.py
@@ -271,15 +271,11 @@ def group(
 
 
 # variant: name omitted, cls _must_ be a keyword argument, @group(cmd=GroupCls, ...)
-# The _correct_ way to spell this overload is to use keyword-only argument syntax:
-# def group(*, cls: t.Type[GrpType], **attrs: t.Any) -> ...
-# However, mypy thinks this doesn't fit the overloaded function. Pyright does
-# accept that spelling, and the following work-around makes pyright issue a
-# warning that GrpType could be left unsolved, but mypy sees it as fine. *shrug*
 @t.overload
 def group(
     name: None = None,
-    cls: t.Type[GrpType] = ...,
+    *,
+    cls: t.Type[GrpType],
     **attrs: t.Any,
 ) -> t.Callable[[_AnyCallable], GrpType]:
     ...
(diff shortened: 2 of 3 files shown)


Tinche:
Out of curiosity, what does strict mode do for pyright?

layday:
It turns on [these](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#diagnostic-rule-defaults) additional checks.  In my experience Pyright won't error for type vars it's not able to solve in the default type checking mode.  The `group` overload error would've been caught by the assert, but there might be auxiliary errors we are not explicitly testing against that strict mode might be able to detect.

Tinche:
Cool, thanks. What's the default type checking mode? It's not `off` I hope

ofek:
While you are fixing things, does this look like a bug? https://github.com/pallets/click/issues/2558#issuecomment-1634555016

Tinche:
LGTM FWIW. Might want to add a changelog entry.

tests/typing/pyrightconfig.json
```diff
@@ -0,0 +1,3 @@
+{
```
davidism, line 1: Move this to `pyproject.toml`.

layday: As in, create a `pyproject.toml` in `tests/typing`? No `pyproject.toml` in the root of the 8.1.x branch this PR is targetting.

alex:
Will this resolve the typing of:

```py
@click.group()
def cli():
    pass


@cli.command()
@click.argument("version")
def release(version: str) -> None:
    # ...
```

or is this a distinct issue?

layday:
That doesn't produce an error for me. Have you tried with click 8.1.5?

tests/typing/pyrightconfig.json
layday: If you meant the latter, I think it might give people to idea that click can be type checked with Pyright (as opposed to its public types).  Also note that Pyright's TOML parser does not support v1 of the spec (https://github.com/microsoft/pyright/issues/4336), meaning that you won't be able to use v1 constructs in your `pyproject.toml`.

hauntsaninja:
@layday 

```
λ pip freeze | egrep 'click|mypy'
click==8.1.5
mypy==1.4.1
mypy-extensions==1.0.0

λ cat xx.py 
import click

@click.group()
def cli() -> None:
    pass


@cli.command()
@click.argument("version")
def release(version: str) -> None:
    pass

λ mypy --strict xx.py
xx.py:8: error: <nothing> has no attribute "command"  [attr-defined]
xx.py:8: error: Untyped decorator makes function "release" untyped  [misc]
Found 2 errors in 1 file (checked 1 source file)
```

tests/typing/pyrightconfig.json
davidism: That's fine, just leave it as is for this branch.

davidism:
Click 8.1.6 released.

PR #2576, opened 2023-07-29
Replace bash shell completion version error with warning

### LINKED ISSUE #2574: Don't fail with error when bash is not found or too old
As discussed in #2152 and #2200, `click` does not support bash shell autocompletion for bash versions older than version 4.4. I understand this and the rational behind this decision, and I am **not** requesting that this should be changed. However, the way that `click` makes developers aware of this version limitation has caused an issue with the ability for autocompletion generation to be scripted.

I have developed [a CLI tool](https://github.com/superatomic/tldr-man-client) using `click` and I have tried to package it for the [Homebrew](https://brew.sh) package manager. As a part of the installation script, autocompletion scripts for `bash`, `zsh`, and `fish` must be generated at installation time. [Homebrew isolates builds by removing `/usr/local/bin` and all user `PATH`s that are not essential for the build](https://docs.brew.sh/Formula-Cookbook#superenv-notes). This means that a newer version of `bash` (>4.4) will **not** be present at build time, **even if it is present on the system**, which causes `click` to fail to generate bash autocompletion, even if a user has a newer version of bash installed on their system:

```
  File "/Users/Olivia/Code/tldr-man/.venv/bin/tldr", line 6, in <module>
    sys.exit(cli())
  File "/Users/Olivia/Code/tldr-man/.venv/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
  File "/Users/Olivia/Code/tldr-man/.venv/lib/python3.10/site-packages/click/core.py", line 1050, in main
  File "/Users/Olivia/Code/tldr-man/.venv/lib/python3.10/site-packages/click/core.py", line 1125, in _main_shell_completion
    rv = shell_complete(self, ctx_args, prog_name, complete_var, instruction)
  File "/Users/Olivia/Code/tldr-man/.venv/lib/python3.10/site-packages/click/shell_completion.py", line 45, in shell_complete
    echo(comp.source())
  File "/Users/Olivia/Code/tldr-man/.venv/lib/python3.10/site-packages/click/shell_completion.py", line 326, in source
    self._check_version()
  File "/Users/Olivia/Code/tldr-man/.venv/lib/python3.10/site-packages/click/shell_completion.py", line 314, in _check_version
    raise RuntimeError(
RuntimeError: Shell completion is not supported for Bash versions older than 4.4.
```

This error is caused by the following function, which will  make bash autocompletion simply fail if it detects that an incompatible version is being used: https://github.com/pallets/click/blob/f45198a3bcaf41d72d31273f9e6a2e3dfeccd19b/src/click/shell_completion.py#L304-L325

It makes sense for bash autocompletion *generation* to always succeed, like it does for both `zsh` and `fish` (zsh and fish autocompletion generation will succeed no matter what, even if neither shell is present on the system). While it is important to notify the user if they are using a version of bash that is too old to be supported, it is a bug for this to cause autocompletion generation to fail entirely. I propose for these error messages to be changed into warning messages that are still displayed, but do not prevent the generation from succeeding.

I am willing to submit a pull request myself with the proposed changes if these changes are wanted.

Environment:

- Python version: 3.10.12
- Click version: 8.1.3

superatomic:
Replaces instances of `raise RuntimeError(...)` with `echo(..., err=True)` in `_check_version()`.

- fixes #2574

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@ Version 8.1.7
 Unreleased
 
 -   Fix issue with regex flags in shell completion. :issue:`2581`
+-   Bash version detection issues a warning instead of an error. :issue:`2574`
 
 
 Version 8.1.6
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -301,7 +301,8 @@ class BashComplete(ShellComplete):
     name = "bash"
     source_template = _SOURCE_BASH
 
-    def _check_version(self) -> None:
+    @staticmethod
+    def _check_version() -> None:
         import subprocess
 
         output = subprocess.run(
@@ -313,15 +314,17 @@ class BashComplete(ShellComplete):
             major, minor = match.groups()
 
             if major < "4" or major == "4" and minor < "4":
-                raise RuntimeError(
+                echo(
                     _(
                         "Shell completion is not supported for Bash"
                         " versions older than 4.4."
-                    )
+                    ),
+                    err=True,
                 )
         else:
-            raise RuntimeError(
-                _("Couldn't detect Bash version, shell completion is not supported.")
+            echo(
+                _("Couldn't detect Bash version, shell completion is not supported."),
+                err=True,
             )
 
     def source(self) -> str:


superatomic:
#2581 should be merged first before this PR is merged.

superatomic:
It might be worth removing the warning when Bash is not on the system at all, so the behavior is more in line with Zsh and Fish.

davidism:
I'm not planning to change it more. This is geared towards people using it directly, not towards other build systems.

superatomic:
Makes sense. Thank you for your review and your work! I appreciate it.

PR #2586, opened 2023-08-14
Break up arguments section

Rowlando13:
Break up the arguments section into arguments and handling files.


docs/arguments.rst
```diff
@@ -34,164 +38,82 @@ And what it looks like:
...
 
 .. _environment-variables:
 
 Environment Variables
 ---------------------
 
-Like options, arguments can also grab values from an environment variable.
-Unlike options, however, this is only supported for explicitly named
-environment variables.
+This feature is not recommended since it be confusing to users. Arguments can only pull environment variables from ? explicitly named environment variables. In that case, it can also be a list of different environment variables where the first one is picked. ?
```
Rowlando13, line 116: @AndreasBackx Do you know what the documentation could mean in this section? I have not a clue so providing clearer wording and a better example is difficult.

Rowlando13: @AndreasBackx I think actually only Armin knew. It was basically unchanged since the first commit. But I figured it out by playing around. Could you give this PR a once over?

docs/arguments.rst
```diff
@@ -34,166 +37,86 @@ And what it looks like:
...
-set to ``nargs=-1``, as it will eat up all arguments.
+.. admonition:: Note on Required Arguments
 
-Example:
+   It is possible to make an argument required by setting ``required=True``.  It is not recommended since we think command line tools should gracefully degrade into becoming no ops.  We think this because command line tools are often invoked with wildcard inputs and they should not error out if the wildcard is empty.
+
+Multiple Arguments
+-----------------------------------
+
+To set the number of argument use the ``nargs`` kwarg. It can be set to any positive integer and -1. Setting it to -1, makes the number of arguments arbitrary (which is called variadic) and can only be used once. The arguments are then packed as a tuple and passed to the function.
```
AndreasBackx, line 50: I think the previous explanation was more correct and clear. This confuses "multiple arguments" and "variadic arguments". See my previous comment.

docs/arguments.rst
```diff
@@ -34,166 +37,86 @@ And what it looks like:
...
 
 .. click:example::
 
     @click.command()
-    @click.argument('src', nargs=-1)
-    @click.argument('dst', nargs=1)
-    def copy(src, dst):
+    @click.argument('src', nargs=1)
+    @click.argument('dsts', nargs=-1)
+    def copy(src, dsts):
```
AndreasBackx, line 57:
```suggestion
    def copy(src: str, dsts: tuple[str, ...]):
```

docs/arguments.rst
```diff
@@ -34,166 +37,86 @@ And what it looks like:
...
-   ``nargs`` to ``+`` to indicate that at least one argument is required.
+    invoke(copy, args=['foo.txt', 'usr/david/foo.txt', 'usr/mitsuko/foo.txt'])
 
-   This is supported by setting ``required=True``.  However, this should
-   not be used if you can avoid it as we believe scripts should gracefully
-   degrade into becoming noops if a variadic argument is empty.  The
-   reason for this is that very often, scripts are invoked with wildcard
-   inputs from the command line and they should not error out if the
-   wildcard is empty.
+.. admonition:: Note on Handling Files.
```
AndreasBackx, line 68:
```suggestion
.. admonition:: Note on Handling Files
```

docs/arguments.rst
```diff
@@ -34,166 +37,86 @@ And what it looks like:
...
 
-.. _file-args:
+    This is not how you would handle files and files paths, since it passes them in as strings. For more see :ref:`handling-files`.
```
AndreasBackx, line 70:
```suggestion
    This is not how you should handle files and files paths. This merely used as a simple example. See :ref:`handling-files` to learn more about how to handle files in parameters.
```

docs/arguments.rst
```diff
@@ -34,166 +37,86 @@ And what it looks like:
...
-            chunk = input.read(1024)
-            if not chunk:
-                break
-            output.write(chunk)
-
-And what it does:
-
-.. click:run::
+    @click.argument('files', nargs=-1, type=click.Path())
+    def touch(files):
```
AndreasBackx, line 83:
```suggestion
    @click.argument('files', nargs=-1, type=click.Path(path_type=pathlib.Path))
    def touch(files: tuple[pathlib.Path, ...]):
```

I feel like we should make `path_type=pathlib.Path` the default. Until it's not, let's maybe make it a bit more clear they can get a `Path` type here as it otherwise defaults to `str`.

docs/arguments.rst
```diff
@@ -34,166 +37,86 @@ And what it looks like:
...
 
 .. _environment-variables:
 
 Environment Variables
 ---------------------
 
-Like options, arguments can also grab values from an environment variable.
-Unlike options, however, this is only supported for explicitly named
-environment variables.
+Arguments can use environment variables. To do so pass the name of the environment variable(s) into ``click.argument``.
```
AndreasBackx, line 117:
```suggestion
Arguments can use environment variables. To do so, pass the name(s) of the environment variable(s) via `envvar` in ``click.argument``.
```

docs/handling-files.rst
```diff
@@ -0,0 +1,86 @@
+.. _handling-files:
+
+Handling Files
+================
+
+.. currentmodule:: click
+
+Click has built in features to support file and file path handling. Since files and files paths are recommended to be passed in as arguments, the examples use arguments.
```
AndreasBackx, line 8:
```suggestion
Click has built in features to support file and file path handling. The examples use arguments but the same principle applies to options as well.
```

I would not say it is a recommendation to pass files as arguments solely. It's perfectly valid to use options. `-i` and `-o` for example is commonly used for input and output.

docs/handling-files.rst
```diff
@@ -0,0 +1,86 @@
...
+
+.. _file-args:
+
+File Arguments
+-----------------
+
+Click supports working with files with the :class:`File` type. Some notable features are:
+
+*   Support for   ``-`` to mean a special file that refers to stdin/stdout. This is common pattern for POSIX command line utilities, for example wget.
```
AndreasBackx, line 17:
```suggestion
*   Support for   ``-`` to mean a special file that refers to stdin when used for reading, and stdout when used for writing. This is a common pattern for POSIX command line utilities.
```

docs/handling-files.rst
```diff
@@ -0,0 +1,86 @@
...
+*   Deals with Unicode and bytes correctly for all versions of Python.
```
AndreasBackx, line 18:
```suggestion
*   Deals with ``str`` and ``bytes`` correctly for all versions of Python.
```

docs/handling-files.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Example:
+
+.. click:example::
+
+    @click.command()
+    @click.argument('input', type=click.File('rb'))
+    @click.argument('output', type=click.File('wb'))
+    def inout(input, output):
```
AndreasBackx, line 27:
```suggestion
    def inout(input: BinaryIO, output: BinaryIO):
```

docs/handling-files.rst
```diff
@@ -0,0 +1,86 @@
...
+        """Copy contents of INPUT to OUTPUT."""
+        while True:
+            chunk = input.read(1024)
+            if not chunk:
+                break
+            output.write(chunk)
+
+And what it does:
```
AndreasBackx, line 35: Can we use `And from the command line:` instead of `And what it does:`? I feel like the original is better.

docs/handling-files.rst
```diff
@@ -0,0 +1,86 @@
...
+
+    with isolated_filesystem():
+        invoke(inout, args=['-', 'hello.txt'], input=['hello'],
+               terminate_input=True)
+        invoke(inout, args=['hello.txt', '-'])
+
+File Path Arguments
+----------------------
+
+For handling paths, the :class:`Path` type is better than a string. Some notable features are:
```
AndreasBackx, line 47:
```suggestion
For handling paths, the :class:`Path` type is better than a ``str``. Some notable features are:
```

docs/handling-files.rst
```diff
@@ -0,0 +1,86 @@
...
+
+*   It can run checks such as does the path exist or is the file executable.
```
AndreasBackx, line 49:
```suggestion
*   The ``exists`` argument will verify whether the path exists.
*	``readable``, ``writable``, and ``executable`` can perform permission checks.
*	``file_okay`` and ``dir_okay`` allow specifying whether files/directories are accepted.
```

docs/handling-files.rst
```diff
@@ -0,0 +1,86 @@
...
+*   It also handles formatting error messages with :func:`format_filename` so any undecodable bytes will be printed nicely.
```
AndreasBackx, line 50:
```suggestion
*   Error messages are nicely formatted using :func:`format_filename` so any undecodable bytes will be printed nicely.
```

docs/handling-files.rst
```diff
@@ -0,0 +1,86 @@
...
+
```
AndreasBackx, line 51:
```suggestion

See :class:`Path` for all features.

```

docs/handling-files.rst
```diff
@@ -0,0 +1,86 @@
...
+            f.write('Hello World!\n')
+        invoke(touch, args=['hello.txt'])
+        println()
+        invoke(touch, args=['missing.txt'])
+
+
+File Opening Behaviors
+-----------------------------
+
+The :class:`File` type attempts to be "intelligent" about when to open a file. Stdin/stdout and files opened for reading will be opened immediately. This will give the user direct feedback when a file cannot be opened. Files opened for writing will only be open on the first IO operation. This is done by automatically wrapping the file in a special wrapper.
```
AndreasBackx, line 77: This removes the documentation on the benefits of lazily opening a file for writing: not emptying the file. Can we add this again here?

docs/handling-files.rst
```diff
@@ -0,0 +1,86 @@
...
+
+File open behavior can be controlled by the boolean kwarg ``lazy``. If a file is opened lazily:
+
+*   A failure at first IO operation will happen by raising an :exc:`FileError`.
+*   It can help minimize resource handling confusion.  If a file is opened in lazy mode, it will receive a ``close_intelligently`` method that can help figure out if the file needs closing or not. This is not needed for parameters, but is necessary for manually prompting. For manual prompts with the :func:`prompt` function you do not know if a stream like stdout was opened (which was already open before) or a real file was opened (that needs closing).
```
AndreasBackx, line 82: Do we have a page anywhere where `close_intelligently` is documented? Would be nice to link to it because this does not explain that enough imho.

AndreasBackx (changes requested):
What I really like about these changes is that some of this new documentation is so much more easily skimmable. Instead of rather large paragaphs, we have easy to understand lists. I am a big fan of this and we should to this elsewhere in the documentation as well!

Most of my comments are improvements towards:
1. Consistency in writing.
2. Adding typing which aids with what we're documenting here.
3. Some regressions I think we're making somewhere.

docs/arguments.rst
```diff
@@ -34,166 +37,86 @@ And what it looks like:
...
+An argument may be assigned a :ref:`parameter type <parameter-types>`. If no type is provided, the type of the default value is used. If no default value is provided, the type is assumed to be STRING.
 
-The value is then passed as a tuple.  Note that only one argument can be
-set to ``nargs=-1``, as it will eat up all arguments.
+.. admonition:: Note on Required Arguments
 
-Example:
+   It is possible to make an argument required by setting ``required=True``.  It is not recommended since we think command line tools should gracefully degrade into becoming no ops.  We think this because command line tools are often invoked with wildcard inputs and they should not error out if the wildcard is empty.
+
+Multiple Arguments
```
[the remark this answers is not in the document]
Rowlando13: Variadic refers to taking an indefinite number of arguments. ie the special case of nargs = -1

In mathematics and in computer programming, a variadic function is a function of indefinite arity, i.e., one which accepts a variable number of arguments

https://en.wikipedia.org/wiki/Variadic_function 

So I don't think its the right name for the heading. But I see how this is confusing, multiple arguments is used to mean something different in the options code. I will look for a different word than multiple

docs/arguments.rst
Rowlando13: In my view variadic is a special case of taking more than one argument, but that may not be the clearest way for documentation purposes. See above.

docs/handling-files.rst
Rowlando13: Yes. That was actually an unintentional to propagate it everywhere.

docs/handling-files.rst
Rowlando13: I agree it does not explain enough. It seems like it refers to specific parts of the code but I am not sure which.

docs/arguments.rst
```diff
@@ -34,166 +37,86 @@ And what it looks like:
...
-the ``nargs`` parameter.  If it is set to ``-1``, then an unlimited number
-of arguments is accepted.
+An argument may be assigned a :ref:`parameter type <parameter-types>`. If no type is provided, the type of the default value is used. If no default value is provided, the type is assumed to be STRING.
 
-The value is then passed as a tuple.  Note that only one argument can be
-set to ``nargs=-1``, as it will eat up all arguments.
+.. admonition:: Note on Required Arguments
 
-Example:
+   It is possible to make an argument required by setting ``required=True``.  It is not recommended since we think command line tools should gracefully degrade into becoming no ops.  We think this because command line tools are often invoked with wildcard inputs and they should not error out if the wildcard is empty.
```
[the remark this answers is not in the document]
Rowlando13: If you read the full docs, their are some opinionated pieces that look to be written by the original author and correspond to unix command line tool practices. This is one of them so I modified it slightly and left it in place. I was planning on leaving them in place on my first pass and then we could make a decision as a whole on how opinionated we wanted to be.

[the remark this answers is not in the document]
Rowlando13: If you ignore the heading on the note which is the almost the same as the paragraph it was in, then it seems the author was talking about required arguments in general.

docs/handling-files.rst
Rowlando13: I did not remove it. Just put it further down. 

Since files opened for writing will typically empty the file, the lazy mode should only be disabled if the developer is absolutely sure that this is intended behavior.

docs/arguments.rst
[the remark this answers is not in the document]
AndreasBackx: Makes sense. Could we rephrase "into becoming no ops" here? I think what is meant is that only using arguments makes a cli look like ancient magics, right?

Could we still document `nargs=-1` behaviour with `required=True` however? I agree that only using arguments is bad. Though there's a definite use case for having a variadic argument that requires at least one.

docs/handling-files.rst
AndreasBackx: It refers to `LazyFile.close_intelligently`. Could you add that reference? That should be good for now. `LazyFile` is in utils.

docs/arguments.rst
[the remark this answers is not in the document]
Rowlando13: I think it means that if you invoke the tool with no arguments then it should do nothing rather than failing with an error. I am going to leave in alone for now and open a meta issue on opinionated takes about things.

docs/arguments.rst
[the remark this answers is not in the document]
Rowlando13: I think those are more literally true but at the cost of being very long. I am going to leave this as Multiple Arguments for now. Feel free to add more examples if you like.

Rowlando13:
Was not able to include full type hints because of issues with custom sphinx directive we use to run the code.

PR #2630, opened 2023-11-07
Specialize `progressbar(length=...)` as `ProgressBar[int]`

### LINKED ISSUE #2626: Make ProgressBar type available in public interface
`mypy` considers function `click.progressbar()` as generic with type parameter `V`: https://github.com/pallets/click/blob/ca5e1c3d75e95cbc70fa6ed51ef263592e9ac0d0/src/click/termui.py#L287-L303

In most cases type parameter `V` will be inferred from `iterable` function parameter:

```python
from typing import reveal_type

import click

with click.progressbar([1, 2, 3]) as bar:
    reveal_type(bar) # Revealed type is "click._termui_impl.ProgressBar[builtins.int]"
```

However,  if `iterable` parameter is not specified (or if it's empty list), type parameter `V` cannot be inferred: 

```python
import click

with click.progressbar(length=1) as bar:
    pass
```

```
$ mypy mypy_click/__init__.py 
mypy_click/__init__.py:3: error: Need type annotation for "bar"  [var-annotated]
```

This issue can be solved with manually specified type parameter:

```python
import click

with click.progressbar[int](length=1) as bar:
    pass
```
Unfortunately, this doesn't work for functions:

```
$ mypy mypy_click/
mypy_click/__init__.py:3: error: Type application is only supported for generic classes  [misc]
```

Another way to fix this is specify type for variable `bar`:

```python
from typing import cast

import click
from click._termui_impl import ProgressBar

with cast(ProgressBar[int], click.progressbar(length=1)) as bar:
    pass

# or

with click.progressbar(length=1) as bar: # type: ProgressBar[int]
    pass
``` 

But in this case `ProgressBar` is imported as not a part of public interface.

So, I think `ProgressBar` *type* should be a part of the public interface. Maybe it should not
be exported directly like class, because users should use `click.progressbar()` function instead.
Public `ProgressBar` type can be defined as [protocol](https://peps.python.org/pep-0544/) and `click.progressbar()` just returns its implementation.

Environment:

- Python version: 3.11.5
- Click version: 8.1.7
- Mypy version: 1.6.1

asottile:
Checklist:

- [x] Add or update relevant docs, in the docs folder and in code. N/A

### DIFF
--- a/tests/typing/typing_progressbar.py
+++ b/tests/typing/typing_progressbar.py
@@ -1,0 +1,22 @@
+from typing_extensions import assert_type
+
+from click import progressbar
+from click._termui_impl import ProgressBar
+
+
+def test_length_is_int() -> None:
+    with progressbar(length=5) as bar:
+        assert_type(bar, ProgressBar[int])
+        for i in bar:
+            assert_type(i, int)
+
+
+def it() -> tuple[str, ...]:
+    return ("hello", "world")
+
+
+def test_generic_on_iterable() -> None:
+    with progressbar(it()) as bar:
+        assert_type(bar, ProgressBar[str])
+        for s in bar:
+            assert_type(s, str)
--- a/tests/typing/typing_progressbar.py
+++ b/tests/typing/typing_progressbar.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing_extensions import assert_type
 
 from click import progressbar
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,4 +1,4 @@
-2630.. currentmodule:: click
+.. currentmodule:: click
 
 Version 8.2.0
 -------------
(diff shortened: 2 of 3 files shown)


tests/typing/typing_progressbar.py
```diff
@@ -0,0 +1,22 @@
...
+
+def test_length_is_int() -> None:
+    with progressbar(length=5) as bar:
+        assert_type(bar, ProgressBar[int])
+        for i in bar:
+            assert_type(i, int)
+
+
+def it() -> tuple[str, ...]:
+    return ("hello", "world")
```
asottile, line 15: this separate function is entirely so pyright doesn't inference as `Literal[("hello", "world")]`

davidism (approved):

asottile:
I hate being that guy but I noticed this PR sticking around when I went to clean up some repos -- is there anything left here?

davidism:
It's good, just haven't had a chance to go through the click backlog in quite a while.

PR #2639, opened 2023-11-29
More robust check of bash version

### LINKED ISSUE #2638: Shell completion bash version check fail in git-bash on windows
I'm using windows with git-bash, and trying to set up bash shell completion following the readme.

I have added the following to my `.bashrc`

```sh
eval "$(_VREX_COMPLETE=bash_source vrex)"
```

But I get this error in new shell instances:

```
The file cannot be accessed by the system.
```

I have traced this to the subprocess execution of bash to check the version.

The problem is that the `bash` executable name is not working from my standard windows python installation.
Instead I need to explicitly add the full path to the git-bash executable, which is `C:\Program Files\Git\bin\bash.exe`

I have a working patch locally with using `shutil.which('bash')` to locate bash, which should work on all platforms. PR coming after this issue is submitted.

This is somewhat related to https://github.com/pallets/click/issues/2461, but that is a different issue, but both are specific to git-bash in windows.

Environment:

- Python version: 3.12.0
- Click version: 8.1.7

koliyo:
More robust check of bash version.

Specifically on windows with git-bash, where python is the normal windows python install

Also use `--norc` flag to avoid recursively loading .bashrc (which can happen when the click completion loading does not work properly)

Testing should not require any additional tests, but rather make sure to run the existing shell completion tests on multiple platforms, eg Linux, Mac, Windows. Or how do you otherwise recommend adding tests for this?

- fixes #2638

Checklist:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,7 +16,8 @@ Unreleased
     :issue:`2632`
 -   Fix ``click.echo(color=...)`` passing ``color`` to coloroma so it can be
     forced on Windows. :issue:`2606`.
-
+-   More robust bash version check, fixing problem on Windows with git-bash.
+    :issue:`2638`
 
 Version 8.1.7
 -------------
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -303,8 +303,8 @@ class BashComplete(ShellComplete):
 
     @staticmethod
     def _check_version() -> None:
-        import subprocess
         import shutil
+        import subprocess
 
         bash_exe = shutil.which("bash")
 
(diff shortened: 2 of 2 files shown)


koliyo:
Would appreciate some feedback if this is not considered completed. As mentioned, automated testing of this is very context specific, so not sure how, or if, this should be added.

Regarding tox test, tox actually fails locally on my mac, but in a completely unrelated section. NOTE: this also fails in `main`!

```
typing: commands[0]> mypy
src/click/utils.py:492: error: Statement is unreachable  [unreachable]
        return os.path.join(
        ^
src/click/_termui_impl.py:616: error: Statement is unreachable  [unreachable]
        try:
        ^
Found 2 errors in 2 files (checked 16 source files)
typing: exit 1 (19.01 seconds) /Users/nils/Work/misc/click> mypy pid=19222
typing: FAIL ✖ in 24.66 seconds
```

My branch has the same test result as main.

PR #2663, opened 2024-01-06
Make `click.Context` generic over `obj`

### LINKED ISSUE #2493: Make `click.Context` generic for `obj`
To add some type safety to the `obj` attribute of the `Context` class, I was thinking maybe it could be made generic with respect to `obj` attribute. It could then be used this way (with a `TypedDict` but it could be any user defined object):

```python
class ContextObj(TypedDict):
    attr: int

def subcommand(
    ctx: click.Context[ContextObj],
    ...
) -> None:
    reveal_type(ctx.obj["attr"])  # Revealed type is "int"
```

That might be a bit cumbersome, so I'd understand if this is rejected (but I'm open to alternatives). Otherwise, I will be happy to implement this

Viicos:
- fixes #2493 

This is backwards compatible, considering the type var will default to `Any` (as it is currently).

Checklist:

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -48,24 +48,25 @@ if t.TYPE_CHECKING:
 
     from .shell_completion import CompletionItem
 
-    ObjT = te.TypeVar("ObjT", default=t.Any)
+    ObjT = te.TypeVar("ObjT", default=None)
 else:
-    ObjT = t.TypeVar("ObjT")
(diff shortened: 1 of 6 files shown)


AndreasBackx:
I gave this a quick shot, see my commit. I don't think this is feasible to implement in a convenient manner. I feel like you might have come across similar issues. I'll close this for now. If anyone has a good implementation PoC, feel free to create a new PR.

Issue #1053, opened 2018-06-22
Problem with CliRunner and Python's logging module


iLoveTux:
If I have a `click.command` instance which calls `logging.basicConfig` with `stream=sys.stdout` and I use an instance of `click.CliRunner` to test the output that the command produces more than once it will fail after the first test.

```
import sys
import click
import logging
import unittest
from click.testing import CliRunner

@click.command()
@click.argument("name")
def greet(name):
    logging.basicConfig(stream=sys.stdout, level=10)
    logging.info('hello, {}!'.format(name))

class TestCli(unittest.TestCase):
    """Both tests are exactly the same"""
    def test_greet_1(self):
        runner = CliRunner()
        result = runner.invoke(greet, ["world"])
        self.assertEqual(result.exit_code, 0)
        self.assertIn('hello, world', result.output)

    def test_cli_2(self):
        runner = CliRunner()
        result = runner.invoke(greet, ["world"])
        self.assertEqual(result.exit_code, 0)
        self.assertIn('hello, world', result.output)

if __name__ == "__main__":
    unittest.main()

```

See the results:

```
$ python test_CliRunner.py
.F
======================================================================
FAIL: test_greet_1 (__main__.TestCli)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_CliRunner.py", line 18, in test_greet_1
    self.assertIn('hello, world', result.output)
AssertionError: 'hello, world' not found in ''

----------------------------------------------------------------------
Ran 2 tests in 0.010s

FAILED (failures=1)
```

The best I can figure out is that CliRunner intercepts `sys.stdout` and replaces it with a file-like object. A handler is then attached to the root logger by a call to `logging.basicConfig` within the command.

When the command is invoked the second time, a new file-like object is allocated for `sys.stdout` but the handler referencing the original file-like object persists rendering the call to `logging.basicConfig` a no-op. This leads to failures of any tests that are examining the output of the command.

I have successfully worked around this issue with the following code:

```
import sys
import click
import logging
import unittest
from click.testing import CliRunner

@click.command()
@click.argument("name")
def greet(name):
    logging.basicConfig(stream=sys.stdout, level=10)
    logging.info('hello, {}!'.format(name))

class TestCli(unittest.TestCase):
    def setUp(self):
        logging.getLogger("").handlers = [] 

    def test_greet_1(self):
        runner = CliRunner()
        result = runner.invoke(greet, ["world"])
        self.assertEqual(result.exit_code, 0)
[...]

PR #2680, opened 2024-02-23
Fix closing of callbacks on CLI exit

kdeldycke:
This PR forces the context to close itself on CLI exits.

This makes sure all callbacks registered by invoked options are properly called whenever a `ctx.exit()` action is triggered. Which has the effect of preventing state leaks.

These leaks cannot be witness in standard CLI operations, because on `ctx.exit()` call, the OS process that is running the Python CLI will be destroyed.

Still, in unittests, these state leaks will leads to flaky and non-deterministic tests, as the `pytest`, `tox` or any other test framework will keep using the same Python interpreter. To demonstrate this, I implemented a unit test using loggers as a fixture, in addition to a generic unit test.

Checklist:


kdeldycke:
Note that this might be related to: https://github.com/pallets/click/issues/1053

Issue #556, opened 2016-04-11
--help doesn't get forwarded when using click.UNPROCESSED


nickjj:
To forward command line options and flags, the Click documentation recommends that you do this:

```python
@click.command(context_settings=dict(
    ignore_unknown_options=True,
))
@click.option('-v', '--verbose', is_flag=True, help='Enables verbose mode')
@click.argument('timeit_args', nargs=-1, type=click.UNPROCESSED)
def cli(verbose, timeit_args):
    """A wrapper around Python's timeit."""
    cmdline = ['python', '-mtimeit'] + list(timeit_args)
    if verbose:
        click.echo('Invoking: %s' % ' '.join(cmdline))
    call(cmdline)
```
That can be found here: http://click.pocoo.org/6/advanced/#forwarding-unknown-options

---

**The problem is**, let's say you're wrapping another utility and you want to view the help menu for that utility. That's commonly accessed in 4 different ways:

1. Running the command with no options or aguments
2. `help`
3. `-help`
4. `--help`

The first 3 ways are supported by Click without issues.

However, the last one fails because instead of forwarding it to the utility, Click just reports back its own help for the command you've created.

I feel like this is a bug because you would expect to see the utility's help menu, not yours.

untitaker:
Click forwards unknown options... but `--help` is a known option! All the other examples you give are not.

I think you're supposed to use `add_help_option=False` to the keyword args of `click.command(..)`, as per http://click.pocoo.org/6/api/#click.Command

nickjj:
Your recommend works, thanks.

Is it worth adding that to the documentation's code snippet listed in the issue? I've used Click extensively but didn't even know that option existed, because this happened to be the first time I ever wanted to fully wrap a third party utility.

That option kind of goes hand in hand with wrapping a utility.

untitaker:
I'm not sure. After all the help output for the wrapper command is demoed further below, and how would you document the `-v` option of the wrapper if you forward `--help` as well?

Perhaps as a bullet point in the "list of things that are important to know", mention that `add_help_option` is a thing.

Anyway I'll close this, feel free to open a PR about the doc change if you care enough :))

Issue #489, opened 2015-12-14
Cannot ignore unknown options


asmodehn:
I m having a problem (on 6.2) that might be related to this issue https://github.com/mitsuhiko/click/issues/473. In my case, this http://click.pocoo.org/6/api/#click.Context.ignore_unknown_options doesn't seem to behave as documented.
I still get a parser error for unspecified options : 

``` python
import click
import rester
@click.command(context_settings=dict(
    ignore_unknown_options=True,
))
@click.option('--log', default='INFO', help='log help')
@click.option('--tc', help='tc help')
@click.option('--ts', help='ts help')
def roster(log, tc, ts):
    """A wrapper around Rester's cli runner that ignores unexpected arguments."""
    rester.apirunner.run()

if __name__ == '__main__':
    roster()
```

```
$ ./scripts/roster --tc rostests/testRosServices.json -sdlkfjs lksjdfsf
Usage: roster [OPTIONS]

Error: Got unexpected extra arguments (-sdlkfjs lksjdfsf)
```

wagnerluis1982:
According to documentation (http://click.pocoo.org/6/advanced/#forwarding-unknown-options) you have to pass `allow_extra_args=True` to `context_settings` and get the remaining args with `ctx.args`.

Or you have to add a `@click.argument('args')` to receive the args.

jcrotts:
Closed, @wagnerluis1982's answer is correct.

Issue #210, opened 2014-08-25
Command to act as a proxy to another command


alexcouper:
### Issue

I want to be able to accept all arguments on the command line and pass them on to another command.
### Use case

I have things like supervisor/postgres/nginx running using non-default configuration locations, and rather than having to type the config path each time I'd like to have it hardcoded in click. This would allow me to do the following:

``` python
@cli.command()
@cli.argument('everything', default='', greedy=True)
def supervisor(args):
    bash("supervisord -c supervisor_conf/supervisord.conf {0}".format(args))
```

The end result being that I could run something like:

``` bash

$ x supervisord -n
```

And supervisor would then be started with no daemon, using the correct configuration.

Is this something that `click` would like to support? Or do you see it as conflicting with the direction it is headed in some way?

PR #2686, opened 2024-03-02
add FORWARD parameter type and stop parsing arguments when FORWARD is…

Kamilcuk:
This change introduces a new `click.FORWARD` argument type, which stops the parser from further parsing arguments. Is it now possible to forward arguments without any parsing. 
For example:

```
import click


@click.group()
@click.option("-v", "--verbose", is_flag=True)
def docker(verbose):
    pass


@docker.command()
@click.option("-v", "--verbose", is_flag=True)
@click.option("-u", "--user")
@click.argument("image")
@click.argument("command", nargs=-1, type=click.FORWARD)
def run(verbose, user, image, command):
    cmdline = ["docker"] + ([f"-u{user}"] if user else []) + ["run", image] + list(command)
    print(cmdline)


docker()
```

Allows for the following:

```
$ python ./docker.py -v run -u kamil alpine sh --help -u -v
['docker', '-ukamil', 'run', 'alpine', 'sh', '--help', '-u', '-v']
```

Issues:

- https://github.com/pallets/click/issues/1183
- https://github.com/pallets/click/issues/556
- https://github.com/pallets/click/issues/360
- https://github.com/pallets/click/issues/210
- https://github.com/pallets/click/issues/489
- https://stackoverflow.com/questions/42457484/python-click-consume-options-as-arguments

Checklist:

### DIFF
--- a/src/click/__init__.py
+++ b/src/click/__init__.py
@@ -54,6 +54,7 @@ from .types import DateTime as DateTime
 from .types import File as File
 from .types import FLOAT as FLOAT
 from .types import FloatRange as FloatRange
+from .types import FORWARD as FORWARD
 from .types import INT as INT
 from .types import IntRange as IntRange
 from .types import ParamType as ParamType
@@ -61,7 +62,6 @@ from .types import Path as Path
 from .types import STRING as STRING
 from .types import Tuple as Tuple
 from .types import UNPROCESSED as UNPROCESSED
-from .types import FORWARD as FORWARD
 from .types import UUID as UUID
 from .utils import echo as echo
 from .utils import format_filename as format_filename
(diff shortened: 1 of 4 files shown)


davidism:
I'm not really clear what this does compared to adding a `nargs=-1` argument, or using `--`. Regardless, I'm currently rewriting the entire parser, so this code won't be correct anyway.

Kamilcuk:
> or using --

It doesn't require it. See the attached issues.

`docker run` does not require `--`, you can run `docker run alpine ls --help` and get help for `ls`, not for `docker`. In contrast, that is impossible with click.

> I'm currently rewriting the entire parser, so this code won't be correct anyway.

Ok, good luck! In that case, I will not fix the checklist.

I _really_ hope that the new parser will _not_ require `--` to stop parsing. If it does, please let me know, I would want to invest time that it does not.

In any case, I use python3.7. I will back port this and publish as I need it.

PR #2726, opened 2024-05-20
Doc Update Resolving Issue #2206

### LINKED ISSUE #2206: documentation about entry points is unclear
First, thanks for making an awesome tool! Click made knocking together a CLI for my masters work trivial. 

These documentation questions are an effort to clarify some of the language in the [setuptools](https://click.palletsprojects.com/en/8.0.x/setuptools/) and [shell completion](https://click.palletsprojects.com/en/8.0.x/shell-completion/?highlight=tab%20complete#enabling-completion) sections, to reduce friction for new users like me. 

1. Though **Shell Completion** depends on users installing through a setuptools entrypoint, and refers to the **Setuptools Integration** docs, the terminology is not aligned between the two sets of examples. **Scripts in Packages** refers to the package `yourpackage`, a subpackage `scripts`, a module `yourscript`, and a click command `cli`. Shell Completion refers to `PROG_NAME`, "the executable name in uppercase...". Does `PROG_NAME` map to the the name of the package `yourscript`, the entrypoint `yourscript`, the command `cli`, or something else entirely? Clarification of exactly what the "executable name" is here would be very helpful.
2. In the **Scripts in Packages** example, it is unclear that the `console_scripts` key is a "default" value, and not something package-specific. For users who have worked with entrypoints in setup.py before, this is probably obvious, but for newcomers there's no visual way to differentiate "standard" values from semantically meaningful ones in this example. A single-sentence note clarifying would help.

- Python version: 3.8
- Click version: 8

j7an:
- Update CONTRIBUTING.rst with latest Click version
- Add entry_points specification URL
- Clarify shell completion description

fixes #2206

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,12 @@
 .. currentmodule:: click
 
+Version 8.1.8
+-------------
+
+Unreleased
+
+-   Fix an issue with type hints for ``click.open_file()``. :issue:`2717`
+
 Version 8.1.7
 -------------
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -144,7 +144,7 @@ Start coding
     .. code-block:: text
 
         $ git fetch origin
-        $ git checkout -b your-branch-name origin/8.0.x
+        $ git checkout -b your-branch-name origin/8.1.x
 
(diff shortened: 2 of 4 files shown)


aenglander:
@dependabot rebase

aenglander (approved):

PR #2731, opened 2024-05-21
Add parametrized tests for `should_strip_ansi`

### LINKED ISSUE #2606: click.echo does not force colors on Windows with color=True
Using `click.echo(click.style("Foo", fg="red"), color=True)` on Windows will not print "Foo" in red if the output is not a terminal that supports colors, so it seems that `color=True` is ignored on Windows.

Here is an example code to reproduce the bug:
```
import click


def main():
    test_txt = "click_echo.txt"
    with open(test_txt, "w") as f:
        click.secho("Without setting color", file=f, fg="red")
        click.secho("With color=False", file=f, fg="red", color=False)
        click.secho("With color=True", file=f, fg="red", color=True)

    click.echo(f"Check the contents of {test_txt}")


if __name__ == "__main__":
    main()
```
After running it on Linux the third line of `click_echo.txt` will contain the ANSI color codes but on Windows all three lines will be the same (the behavior is the same if `echo` prints on the console and the script output is redirected to a file)

Environment:

- Python version: 3.9
- Click version: 8.1.7

Tyl13:
Issue #2606 described a inconsistency in how color=True works between Operating Systems. 
This is because ANSI styles can't always be applied to different terminals on windows. This did not cause issues when running it within the terminals because Colorama would change the ANSI style codes to escape codes for windows when it was required. However, if you were to pipe the `click.echo()` output to a file then the `should_strip_ansi` function would not work on Windows as the other operating systems. 

This is because the `auto_wrap_for_ansi` function was not being called with the color argument. After the `auto_wrap_for_ansi` function is called, it defaults `color=None` and will end up calling `should_strip_ansi` with `color=None` which is different from the original call to `should_strip_ansi` that was called before the wrapper function. 

By passing color to `auto_wrap_for_ansi`, then the behavior of `should_strip_ansi` becomes consistent.

An additional change applied within the PR is to the tests in `test_utils.py`. This is to increase test coverage for the different cases of `click.echo()` and removes the old version of the test. Instead of one test that mocks `isatty()`, there are now three different tests. One that mocks `isatty()=True` and `_is_jupyter_kernel_output()=False`, another that mocks `isatty()=False` and `_is_jupyter_kernel_output()=True` , and the last one mocks both `isatty()=False` and `_is_jupyter_kernel_output()=False`.

fixes #2606 

- [x] Add tests that demonstrate the correct behavior of the change. Tests

### DIFF
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -1,4 +1,10 @@
-from click._compat import should_strip_ansi
+from __future__ import annotations
+
+import sys
+
+import pytest
+
+import click
 
 
 def test_is_jupyter_kernel_output():
@@ -7,4 +13,43 @@ def test_is_jupyter_kernel_output():
 
     # implementation detail, aka cheapskate test
     JupyterKernelFakeStream.__module__ = "ipykernel.faked"
-    assert not should_strip_ansi(stream=JupyterKernelFakeStream())
+    assert click._compat._is_jupyter_kernel_output(stream=JupyterKernelFakeStream())
+
+
+@pytest.mark.parametrize(
+    "stream",
+    [None, sys.stdin, sys.stderr, sys.stdout],
+)
+@pytest.mark.parametrize(
+    ("color", "expected_override"),
+    [
+        (True, False),
+        (False, True),
+        (None, None),
+    ],
+)
+@pytest.mark.parametrize(
+    ("isatty", "is_jupyter", "expected"),
+    [
+        (True, False, False),
+        (False, True, False),
+        (False, False, True),
+    ],
+)
+def test_should_strip_ansi(
+    monkeypatch,
+    stream,
+    color: bool | None,
+    expected_override: bool | None,
+    isatty: bool,
+    is_jupyter: bool,
+    expected: bool,
+) -> None:
+    monkeypatch.setattr(click._compat, "isatty", lambda x: isatty)
+    monkeypatch.setattr(
+        click._compat, "_is_jupyter_kernel_output", lambda x: is_jupyter
+    )
+
+    if expected_override is not None:
+        expected = expected_override
+    assert click._compat.should_strip_ansi(stream=stream, color=color) == expected


Tyl13:
Rebased to fix merge conflicts more easily. However, something seems to sometimes cause the windows version to fail. Currently investigating that. I've had it fail locally and then a second rerun succeed. I'm unsure what is causing the issue atm.

Tyl13:
The issue is coming from my previous change to the runner. The issue is documented here #2732 with the PR to fix it here #2733.

AndreasBackx:
Fixed by #2607

davidism:
After rebasing this on latest stable that includes #2607 and #2733, the `test_echo_color_flag` that you change fails now. I have a feeling the test doesn't need to be changed if the code was changed. I'm not going to block 8.1.8 on this since it's only a change to tests. @Tyl13 if you still remember any discussion about this from the sprint, let me know, and if you can update the test to pass again that would be great.

Tyl13:
Yeah, I am also not sure why I changed the `isatty` to `should_strip_ansi`. That one doesn't need changed, maybe I was trying to make it closer to the new test in the `test_compat` but I don't have my previous laptop to check my reasoning anymore. Looking at the previous commit before I force-pushed the fix-2606 branch from [56d016e](https://github.com/pallets/click/commit/56d016ef912ad860fa38756003fa62efa0c709b5) to [8e34c62](https://github.com/pallets/click/commit/8e34c623a10ad2d26137971ec340eb84a04f6212) shows the tests being very different. I then moved those to test_compat on their own, and I'm guessing renamed the `isatty` to `should_strip_ansi` with a find and replace. I'm not sure how I was getting it to pass then though, unless I noticed it on my end and forgot to commit undoing it.

I can fix it by just removing those changes to that file,do you want it rebased on the most recent tag of 8.1.8?

kdeldycke:
The fix consisting in passing `color` to `auto_wrap_for_ansi` shipped in 8.2.0 via #2607. So this PR is only about adding color-related tests, which are worth reusing and maybe extend.

I will try to clean-up that PR.

AndreasBackx:
I cannot review my own PR, but the rebases look great. Thanks again!

kdeldycke:
> I cannot review my own PR, but the rebases look great. Thanks again!

Just added you as a review on this one if this help.

kdeldycke:
This PR has been cleanup of duplicate code and rebased on top of stable and ready to be merged upstream.

PR #2775, opened 2024-09-07
Improve echo_via_pager behaviour in face of errors

### LINKED ISSUE #2674: echo_via_pager with generators leaves terminal in broken state
# Steps to reproduce

Run the following test case:
```
import click

def output_generator():
    counter = 0
    while True:
        yield "this is a line of output\n"
        if counter == 1024:
            click.echo("kaboom", err=True)
            click.get_current_context().exit(0)
        counter += 1

@click.command
def kaboom():
    click.echo_via_pager(output_generator)

if __name__ == "__main__":
    kaboom()
```

Attempt to use the terminal afterwards

# Expected result

A working terminal.

# Actual behavior

Newlines and output are obscured.  Commands may be entered, but they are not displayed.

# Workaround

Run the `reset` command after the command terminates.

Environment:

Reproducible on OS X and Ubuntu (x86_64):

```
$ sw_vers
ProductName:            macOS
ProductVersion:         14.3.1
BuildVersion:           23D60
```
- Python version: 3.11.2
- Click version: 8.1.7

```
# cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.3 LTS"
```
- Python version: 3.10.12
- Click version: 8.1.7

stefreak:
fixes #2674 (echo_via_pager with generators leaves terminal in broken state).

- [x] Add tests that demonstrate the correct behavior of the change. Tests

### DIFF
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -427,26 +427,40 @@ def _pipepager(generator: cabc.Iterable[str], cmd: str, color: bool | None) -> N
                 text = strip_ansi(text)
 
             stdin.write(text.encode(encoding, "replace"))
-    except (OSError, KeyboardInterrupt):
+    except BrokenPipeError:
+        # In case the pager exited unexpectedly, ignore the broken pipe error.
         pass
-    else:
-        stdin.close()
-
-    # Less doesn't respect ^C, but catches it for its own UI purposes (aborting
-    # search or other commands inside less).
-    #
-    # That means when the user hits ^C, the parent process (click) terminates,
-    # but less is still alive, paging the output and messing up the terminal.
-    #
-    # If the user wants to make the pager exit on ^C, they should set
-    # `LESS='-K'`. It's not our decision to make.
-    while True:
+    except Exception as e:
+        # In case there is an exception we want to close the pager immediately
+        # and let the caller handle it.
+        # Otherwise the pager will keep running, and the user may not notice
+        # the error message, or worse yet it may leave the terminal in a broken state.
+        c.terminate()
+        raise e
+    finally:
+        # We must close stdin and wait for the pager to exit before we continue
         try:
-            c.wait()
-        except KeyboardInterrupt:
+            stdin.close()
+        # Close implies flush, so it might throw a BrokenPipeError if the pager
+        # process exited already.
+        except BrokenPipeError:
             pass
-        else:
-            break
+
+        # Less doesn't respect ^C, but catches it for its own UI purposes (aborting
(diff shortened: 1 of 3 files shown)


stefreak:
I've had a look at #350 and the new behaviour sounds like what @amjith describes in the comment https://github.com/pallets/click/issues/350#issuecomment-107271611 

The fix by @untitaker is still in place (https://github.com/pallets/click/commits/e71f0a040a81e03276b25ab1d1b0671822ab331b) – now in the finally block, which should be even better.

For me this works really well in all the situations I was able to come up with so far.

src/click/_termui_impl.py
```diff
@@ -424,26 +424,33 @@ def _pipepager(generator: cabc.Iterable[str], cmd: str, color: bool | None) -> N
                 text = strip_ansi(text)
 
             stdin.write(text.encode(encoding, "replace"))
-    except (OSError, KeyboardInterrupt):
```
stefreak, line 427: I'm not sure why OSError has been captured here; I think it's probably better to remove it and re-introduce it with a comment or a more specific solution if it causes a problem.

stefreak: `OSError` used to be `IOError` and it's in since the beginning (aef16506b5da26684103db2e47346dbf49a7e7a1).

If writing files for some reason, I'd like to see an error message, rather than swallowing any error that might occur – but I'm happy to revert this change if the reviewer disagrees or there is a reason I'm not aware of

src/click/_termui_impl.py
```diff
@@ -424,26 +424,33 @@ def _pipepager(generator: cabc.Iterable[str], cmd: str, color: bool | None) -> N
...
-        pass
-    else:
+            stdin.flush()
```
stefreak, line 427: Close implies flush, so this one is not necessary. https://stackoverflow.com/questions/2447143/does-close-imply-flush-in-python
```suggestion
```

src/click/_termui_impl.py
untitaker: most likely this is done for less noisy behavior in broken pipes or broken SSH sessions?

stefreak: Right, now I also ran into a case. Pushed a version now that ignores `BrokenPipeError` specifically

0xDEC0DE (commented):
I don't have rights to approve, but this looks like it will do the trick.

PR #2805, opened 2024-11-09
`Parameter` typing improvements

AndreasBackx:
This removes all of the `type: ignore` mentions in `Argument`, its subclasses, and the `exceptions` module. To do so, the following is done:

1. `Parameter` is now an abstract base class. This just clarifies it cannot be instantiated which was always the point.
2. `Parameter.name` can no longer be `None`. It only ever is `None` when `expose_value = False`. Instead of having to deal with `None` and `is None` everywhere, let's just use `""` instead of `None` when `expose_value` is `False`
3. `ctx` in `Parameter.get_error_hint` can be `None`. It always has been like this, but wasn't explicit in typing. Better to be explicit than implicit.

- ~~Add tests that demonstrate the correct behavior of the change. Tests
  should fail without the change.~~
	- `Parameter.name` is an implementation detail, it's not in the documentation. 
- ~~Add or update relevant docs, in the docs folder and in code.~~

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -7,6 +7,8 @@ import inspect
 import os
 import sys
 import typing as t
+from abc import ABC
+from abc import abstractmethod
 from collections import abc
 from collections import Counter
 from contextlib import AbstractContextManager
@@ -838,9 +840,7 @@ class Context:
                     # https://github.com/pallets/click/pull/3068
                     if default_value is UNSET:
                         default_value = None
-                    kwargs[param.name] = param.type_cast_value(  # type: ignore
-                        ctx, default_value
-                    )
+                    kwargs[param.name] = param.type_cast_value(ctx, default_value)
 
             # Track all kwargs as params, so that forward() will pass
             # them on in subsequent calls.
@@ -1320,7 +1320,7 @@ class Command:
                     or param.hidden
                     or (
                         not param.multiple
-                        and ctx.get_parameter_source(param.name)  # type: ignore
+                        and ctx.get_parameter_source(param.name)
                         is ParameterSource.COMMANDLINE
                     )
                 ):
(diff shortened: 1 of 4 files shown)


src/click/core.py
```diff
@@ -2343,17 +2345,17 @@ def handle_parse_result(
                 value = None
 
         if self.expose_value:
-            ctx.params[self.name] = value  # type: ignore
+            ctx.params[self.name] = value
 
         return value, args
 
     def get_help_record(self, ctx: Context) -> tuple[str, str] | None:
-        pass
```
AndreasBackx, line 2351: Linter complains about not having an implementation when it's an abstract base class. Using `return None` makes it not do that.

Rowlando13:
@AndreasBackx Would you mind rebasing this so it can go out with 8.3.0?

kdeldycke:
@AndreasBackx I just reviewed this PR following our discussion at https://github.com/pallets/click/discussions/3329 and just realized it is not that big of a change. So I took the liberty to rebase it on top of `stable` and fix all remaining issues. The only careful part was adapting it to the `UNSET`/sentinel but in the end it worked well.

I will program this PR for inclusion for the next 8.4.0 release cycle. Can you double-check everything looks OK to you?

AndreasBackx:
Wow ya, thank you so much. I reviewed, or still am, reviewing the other ones. It's been a while but it looks as expected I think. This PRO looks as how I'd expect it. I'm kind of fuzzy about changing one None -> "" here in a few places but I think that's a fine change for the edge case it might just be and them being internal implementations.

kdeldycke:
> Wow ya, thank you so much. I reviewed, or still am, reviewing the other ones. It's been a while but it looks as expected I think. This PRO looks as how I'd expect it. I'm kind of fuzzy about changing one None -> "" here in a few places but I think that's a fine change for the edge case it might just be and them being internal implementations.

Just refined some comparisons and removed an assert to clean up code.

So let's consider this PR ready to be merged upstream unless a review is raised.

Rowlando13:
I will fix changes then this is ready for 8.4.0.

kdeldycke:
> I will fix changes then this is ready for 8.4.0.

Just rebased on top of `main` with appropriate location of the changelog entry. Also solved a merge conflict due to https://github.com/pallets/click/pull/3371 merge.

This is ready to be merged into `main`.

davidism (approved):

kdeldycke:
Thanks @AndreasBackx for the merge and for removing the left-over branch!

PR #2808, opened 2024-11-17
Use SPDX license identifier

cdce8p:
[PEP 639](https://peps.python.org/pep-0639/) recommends the use of SPDX license identifiers.

At the moment the `license` key is ignored by `flit`, see https://github.com/pypa/flit/issues/525. However, it's still possible to replace it now. Once the support in flit is added, the value will be used automatically.

_At some point, the table will need be be changed to just a string for full PEP 639 compliance. That requires support for it first though. The change is also pretty simple._
```diff
-license = {text = "BSD-3-Clause"}
+license = "BSD-3-Clause"
```

--
Before the move to `pyproject.toml`, `setup.cfg` used `BSD-3-Clause` as well. https://github.com/pallets/click/blob/8.1.7/setup.cfg#L12

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "click"
 version = "8.2.0.dev"
 description = "Composable command line interface toolkit"
 readme = "README.md"
-license = {file = "LICENSE.txt"}
+license = {text = "BSD-3-Clause"}
 maintainers = [{name = "Pallets", email = "contact@palletsprojects.com"}]
 classifiers = [
     "Development Status :: 5 - Production/Stable",


Rowlando13:
Looks like license is a modified BSD-3 so I am not sure this is correct. @davidism thoughts?

cdce8p:
> Looks like license is a modified BSD-3 so I am not sure this is correct. @davidism thoughts?

Can you explain what specifically you mean? Github recognizes it correctly as the `BSD-3-Clause` license.
https://github.com/pallets/click/blob/main/LICENSE.txt
https://spdx.org/licenses/BSD-3-Clause.html

davidism:
It's not modified, should be standard BSD-3-Clause.

davidism:
I'm going to close this for now. Once Flit supports license = "string" and license_files = [...], you're welcome to create a PR for that.

PR #2811, opened 2024-11-24
Fix eagerness of help option generated by `help_option_names`

kdeldycke:
When a help option is generated by the use of `help_option_names` setting, it loose its eagerness. This issue has an effect on the order in which parameters (and their callbacks) are processed, and breaks the evaluation order described in the documentation: https://click.palletsprojects.com/en/stable/advanced/#callback-evaluation-order

That's because the utility method `iter_params_for_processing()` is comparing option objects, and the help option is generated on every  `get_help_option()` or `get_params(ctx)` call. This behavior is demonstrated by a unit-test in this PR.

My solution to fix this issue consist in caching locally the auto-generated help option. This solves the issue.

I used this oportunity to clarify some documentation, and unit-test the main method responsible for the ordering of parameters.

This is more or less a follow up of #2563, which has been merged in the upcoming v8.1.8 release.

Checklist:

- [x] Run pre-commit hooks and fix any issues.


src/click/core.py
```diff
@@ -1022,14 +1033,21 @@ def show_help(ctx: Context, param: Parameter, value: str) -> None:
...
+            self._help_option = Option(
+                help_options,
+                is_flag=True,
+                is_eager=True,
+                expose_value=False,
+                callback=show_help,
+                help=_("Show this message and exit."),
+            )
+
+        return self._help_option
```
AndreasBackx, line 1050: Can you instantiate `self._help_option: HelpOption | None = None` in the constructor so we can check for `None` here and have this typed?

Any reason we couldn't use the `HelpOption` type you introduced as well?

AndreasBackx (changes requested):

src/click/core.py
kdeldycke: Good idea on typing `_help_option`.

As for not reusing `HelpOption`, it's because I was targeting the `main` branch instead of the `stable` one. I just fixed that.

kdeldycke:
Just fixed all the remaining issues.

I was just wondering why `get_help_option()` couldn't be replaced by a `@cached_property` like that:
```python
    @cached_property
    def help_option(self) -> t.Optional["Option"]:
        ctx = self.get_current_context()
        help_options = self.get_help_option_names(ctx)

        if not help_options or not self.add_help_option:
            return None

        # Avoid circular import.
        from .decorators import HelpOption
        
        return HelpOption(help_options)

    def get_help_option(self, ctx: Context) -> t.Optional["Option"]:
        return self.help_option
```

But I guess there are some edge-cases regarding the context management. And I don't want to propose refactors that are too deep just as 8.1.8 is around the corner and the 8.2.0 is coming soon after.

Anyway, this PR is ready to be merged upstream for the 8.1.8 release.

AndreasBackx:
Thanks for the fix!

kdeldycke:
Thanks @AndreasBackx for the merge!

PR #2815, opened 2024-12-03
Add links to third-party projects enhancing Click

kdeldycke:
I propose to promote third-party projects which extends Click in major ways.

I referenced notable repositories that are actively maintained and brings non-trivial enhancements to Click. I sorted them by their popularity (i.e. number of GitHub stars).

CC: @ewels, @simonw, @janluke

### DIFF
--- a/docs/contrib.rst
+++ b/docs/contrib.rst
@@ -18,4 +18,41 @@ Please note that the quality and stability of those packages may be different
 than Click itself. While published under a common organization, they are still
 separate from Click and the Pallets maintainers.
 
+
+Third-party projects
+--------------------
+
+Other projects that extend Click's features are available outside of the
+click-contrib_ organization.
+
+Some of the most popular and actively maintained are listed below:
+
+==========================================================  ===========================================================================================  =================================================================================================  ======================================================================================================
+Project                                                     Description                                                                                  Popularity                                                                                         Activity
+==========================================================  ===========================================================================================  =================================================================================================  ======================================================================================================
+`Typer <https://github.com/fastapi/typer>`_                 Use Python type hints to create CLI apps.                                                    .. image:: https://img.shields.io/github/stars/fastapi/typer?label=%20&style=flat-square           .. image:: https://img.shields.io/github/last-commit/fastapi/typer?label=%20&style=flat-square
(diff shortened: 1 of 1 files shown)


Rowlando13:
Thanks for this! Can you change the PR to be against stable? Stable are the currently visible docs.

davidism:
A long time ago, we maintained a similar list for Flask extensions. It was impossible to maintain, and anything on the list immediately looked more "official". We couldn't keep track of projects over time, and many users would try to get their new or little used project on the list.

I do have a desire to create a site similar to https://djangopackages.org/ for Flask, Jinja, and Click eventually.

Rowlando13:
I agree with your sentiment @davidism. However, I think there is value to not always thinking super long term and seeing how it plays out. The additions are clear improvements over what we have. I only comment because you reopened the issue and I was planning on merging it.

Rowlando13:
Perhaps more clearly stated I am happy to take on the temporary burden. If it becomes to much I will simply, remove the section.

kdeldycke:
I had no clue [Typer](https://github.com/fastapi/typer) was built on top of Click. Just added it (CC @tiangolo).

My goal with that list is to increase Click usage. And also have a page to point out users to when they're requesting a feature that is out of Click's scope but available in third party tools.

For moderation and keep the signal to noise ratio high, I added a couple of simple editorial rules. They're easy to check and good enough to filter unimportant projects.

I made the list into a table, inspired by:
- https://docs.pytest.org/en/stable/reference/plugin_list.html
- https://kdeldycke.github.io/meta-package-manager/benchmark.html#activity

Now as @Rowlando13 said, we can just test this minimal arrangement and see how it is used or abused by the community.

kdeldycke:
And see how it renders at: https://github.com/pallets/click/blob/0966fa8bd68f0a5e1876568f49eb16caf5504561/docs/contrib.rst#third-party-projects

Screenshot 2024-12-07 at 10 35 09

Rowlando13:
Thanks for working this out. My goal is also to increase click usage and build community around its maintenance.

kdeldycke:
I just rebased this PR on top of `stable`. It is ready to be merged if you like it.

PR #2818, opened 2024-12-08
Add CliRunner default `catch_exceptions` parameter

### LINKED ISSUE #2817: Set the `catch_exceptions` value for the whole runner
I always pass `catch_exceptions=False` for all my test suites.
It would be convenient to configure this value once for all in `CLIRunner` for example.

For example, we could add a `CLIRunner.catch_exceptions` parameter, and make `invoke` use that value if not directly passed in parameter of `invoke`. Then I would just have to put the runner in a pytest fixture and that would do the thing for me.

If that is OK, I can open a PR for this.

What do you think?

azmeuk:
fixes #2817

Add a `catch_exceptions` parameter to `CliRunner`.
If the `catch_exceptions` parameters from `invoke` is not explicitly set, the value from `CliRunner` is used.

### DIFF
--- a/src/click/testing.py
+++ b/src/click/testing.py
@@ -227,6 +227,11 @@ class CliRunner:
                        to `<stdout>`.  This is useful for showing examples in
                        some circumstances.  Note that regular prompts
                        will automatically echo the input.
+    :param catch_exceptions: Whether to catch any other exceptions than
+                             ``SystemExit`` when running :meth:`~CliRunner.invoke`.
+
+    .. versionchanged:: 8.2
+        Added the ``catch_exceptions`` parameter.
 
     .. versionchanged:: 8.2
         ``mix_stderr`` parameter has been removed.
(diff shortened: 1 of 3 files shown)


src/click/testing.py
```diff
@@ -227,6 +227,11 @@ class CliRunner:
                        to `<stdout>`.  This is useful for showing examples in
                        some circumstances.  Note that regular prompts
                        will automatically echo the input.
+    :param catch_exceptions: Whether to catch any other exceptions than
+                             ``SystemExit`` when running :meth:`~CliRunner.invoke`.
```
cosmastech, line 231:
```suggestion
    :param catch_exceptions: Whether to catch any exceptions other than
                             ``SystemExit`` when running :meth:`~CliRunner.invoke`.
```

AndreasBackx:
Thank you for the issue and PR!

PR #2821, opened 2024-12-15
Rewrite help page section.

Rowlando13:
Rewrite help page generation section, formerly documenting script.

### DIFF
--- a/docs/documentation.rst
+++ b/docs/documentation.rst
@@ -1,18 +1,14 @@
-Documenting Scripts
+Help Pages
 ===================
 
 .. currentmodule:: click
 
-Click makes it very easy to document your command line tools.  First of
-all, it automatically generates help pages for you.  While these are
-currently not customizable in terms of their layout, all of the text
-can be changed.
+Click makes it very easy to document your command line tools. For most things Click automatically generates help pages for you. By design the text is customizable, but the layout is not.
 
 Help Texts
 ----------
 
-Commands and options accept help arguments.  In the case of commands, the
-docstring of the function is automatically used if provided.
+Commands and options accept help arguments. For commands, the docstring of the function is automatically used if provided.
 
 Simple example:
 
@@ -20,13 +16,16 @@ Simple example:
 
     @click.command()
     @click.option('--count', default=1, help='number of greetings')
-    @click.argument('name')
-    def hello(count, name):
-        """This script prints hello NAME COUNT times."""
+    @click.option('--name', help='a name')
+    def hello(count, name=False):
+        """This script prints hello and a name one or more times."""
         for x in range(count):
-            click.echo(f"Hello {name}!")
+            if name:
+                click.echo(f"Hello {name}!")
+            else:
+                click.echo("Hello!")
 
-And what it looks like:
+And from the command line:
 
 .. click:run::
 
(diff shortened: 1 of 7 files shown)


docs/help-pages.rst
```diff
@@ -0,0 +1,242 @@
...
+Commands and options accept help arguments. For commands, the docstring of the function is automatically used if provided.
+
+Simple example:
+
+.. click:example::
+
+    @click.command()
+    @click.option('--count', default=1, help='number of greetings')
+    @click.option('--name', help='a name')
+    def hello(count, name=False):
```
AndreasBackx, line 20:
```suggestion
    def hello(count: int, name: str):
```

`name=False` is incorrect, no?

Let's also add typing here while we're at it.

AndreasBackx: Seeing that this also changed from parameter, to option. I think we should not change this to an option. The goal here is to showcase how these have help pages. Highlighting parameters don't have help pages is important.

This is regardless of whether or not that's a good state right now. I think it should also get documentation, but that's outside of the scope of this PR.

AndreasBackx (changes requested):
This is pretty hard to review because of the following:

1. Git is not aware of the rename of `documentation.rst` -> `help-pages.rst`. This makes it hard to understand what has changed because it's all "new". See https://stackoverflow.com/questions/433111/how-to-make-git-mark-a-deleted-and-a-new-file-as-a-file-move.
2. The sections have been re-ordered.

Could we do the following to make reviewing easier:
1. Keep the order of the sections.
2. Tell git of the rename.

Then a reviewer will be able to see the changes under each section.

If you still want to re-order sections, could that be done in a separate PR?

AndreasBackx:
Could you also rebase this on main?

Rowlando13:
Can do both.

Rowlando13:
Do you want all docs based on main or just this branch?

Rowlando13:
@AndreasBackx Done.

AndreasBackx:
> Do you want all docs based on main or just this branch?

All docs, let's focus on just main for now.

> @AndreasBackx Done.

Thanks, this is much easier to review now.

docs/documentation.rst
```diff
@@ -189,65 +220,9 @@ Example:
...
-.. versionadded:: 2.0
-
-The help parameter is implemented in Click in a very special manner.
-Unlike regular parameters it's automatically added by Click for any
-command and it performs automatic conflict resolution.  By default it's
-called ``--help``, but this can be changed.  If a command itself implements
-a parameter with the same name, the default help parameter stops accepting
-it.  There is a context setting that can be used to override the names of
-the help parameters called :attr:`~Context.help_option_names`.
+The help parameter(s) is automatically added by Click for any command. The default is ``--help`` but can be override by the context setting :attr:`~Context.help_option_names`. Click also performs automatic conflict resolution on the default help parameter so if a command itself implements a parameter named ``help`` then the default help will no be run.
```
AndreasBackx, line 225:
```suggestion
Help parameters are automatically added by Click for any command. The default is ``--help`` but can be override by the context setting :attr:`~Context.help_option_names`. Click also performs automatic conflict resolution on the default help parameter so if a command itself implements a parameter named ``help`` then the default help will not be run.
```

AndreasBackx (approved):
I'll approve to get this out the door. Though please make the non-nit changes. I definitely like that this is easier to read and skim, less fluff.

docs/documentation.rst
```diff
@@ -1,32 +1,31 @@
...
+        """This script prints hello and a name one or more times."""
         for x in range(count):
-            click.echo(f"Hello {name}!")
+            if name:
+                click.echo(f"Hello {name}!")
+            else:
+                click.echo("Hello!")
 
-And what it looks like:
+And from the command line:
```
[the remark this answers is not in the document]
Rowlando13: Great idea.

docs/help-pages.rst
Rowlando13: Done

docs/documentation.rst
```diff
@@ -115,71 +159,58 @@ Example:
...
-Meta Variables
---------------
+Placeholder Variable
+-----------------------
 
-Options and parameters accept a ``metavar`` argument that can change the
-meta variable in the help page.  The default version is the parameter name
-in uppercase with underscores, but can be annotated differently if
-desired.  This can be customized at all levels:
+The default placeholder variable (`meta variable <https://en.wikipedia.org/wiki/Metasyntactic_variable#IETF_Requests_for_Comments>`_) in the help pages is the parameter name in uppercase with underscores. This can be changed for Commands and Parameters with the ``options_metavar`` and  ``metavar`` kwargs.
```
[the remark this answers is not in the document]
Rowlando13: Going to leave this alone.

PR #2829, opened 2024-12-22
Only try to set flag_value if is_flag is true

lpsinger:
This statement:

    flag_value = not self.default

requires the class to have a working `__bool__` method. However, there are some classes that explicitly disable this method:

- [`numpy.ndarray`](https://numpy.org/doc/stable/reference/arrays.ndarray.html) raises `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`
- [`astropy.units.Quantity`](https://docs.astropy.org/en/stable/api/astropy.units.Quantity.html) raises `ValueError: Quantity truthiness is ambiguous, especially for logarithmic units and temperatures. Use explicit  comparisons.`

Move this statement so that it is only executed if `is_flag` is true.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -91,6 +91,9 @@ Unreleased
 -   Add a ``catch_exceptions`` parameter to :class:`CliRunner`. If
     ``catch_exceptions`` is not passed to :meth:`CliRunner.invoke`,
     the value from :class:`CliRunner`. :issue:`2817` :pr:`2818`
+-   ``Option.flag_value`` will no longer have a default value set based on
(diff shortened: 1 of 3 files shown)


lpsinger:
See, for example, https://github.com/m4opt/m4opt/issues/220.

AndreasBackx:
This seems to make sense. Thank you! Here it would make sense to have some typing to clarify this requirement. Though that's out of scope for this PR.

lpsinger:
Thanks for the speedy review!

davidism:
Behavior further updated in #2930

PR #2830, opened 2024-12-23
Merge stable into main

AndreasBackx:
This merges stable into main but there's an openstanding problem that might needs fixing on stable before we merge into main.

Stable has moved from os.system to subprocess. As part of this, it also extended some testing for `echo_with_pager`. One test leverages `cat > PATH` which due to how the move to subprocess went, is no longer supported. We can decide to:
1. Create 8.1.9 with the fix bringing back support.
2. Only include the fix for support in 8.2.0.

I think to reduce the load and focus on future development, let's go with 2. The impact for users is the following:

- Those with `PAGER` setup including something that needs to be run in a shell, won't be used in 8.1.8. It will fallback to one of the pagers we try.

I think this is acceptable. Please let me know what you think.


kdeldycke:
I just fix all conflicts in #2873, from which Click 8.2.0 can be released.

kdeldycke:
Following the merge of #2873, this PR can be closed, and `stable-into-main` branch can be deleted.

PR #2840, opened 2025-01-09
Move `--help` option defaults from its class to its decorator

### LINKED ISSUE #2832: Regression: Help option subclassing no longer works with verion 8.1.8
Click 8.18 has a regression wrt. to help with Option subclasses.

This was found in this issue originally:
- https://github.com/aboutcode-org/scancode-toolkit/issues/4044

This was introduced in this PR by @kdeldycke :wave: :
- https://github.com/pallets/click/pull/2563

It breaks code that uses option subclasses. 

- This code works in 8.1.7 and  8.1.8 with no custom class
```Python
import click

@click.command()
@click.help_option("-h", "--help")
def scancode():
    """OK in pkg:pypi/click@8.1.8 and pkg:pypi/click@8.1.7"""
    pass


if __name__ == "__main__":
    scancode()
```
and then with 8.1.7 and 8.1.8:
```
$ python  works.py --help
Usage: works.py [OPTIONS]

  Regression in pkg:pypi/click@8.1.8

Options:
  -h, --help  Show this message and exit.
```

- This code works in 8.1.7 and fails in 8.1.8 with a custom class
```Python
import click


class PluggableCommandLineOption(click.Option):
    pass


@click.command()
@click.help_option("-h", "--help", cls=PluggableCommandLineOption)
def scancode():
    """Regression in pkg:pypi/click@8.1.8"""
    pass


if __name__ == "__main__":
    scancode()
```

and then with 8.1.7
```
 python  failing.py --help
Usage: failing.py [OPTIONS]

  Regression in pkg:pypi/click@8.1.8

Options:
  -h, --help  Show this message and exit.


$ python  failing.py -h
Usage: failing.py [OPTIONS]

  Regression in pkg:pypi/click@8.1.8

Options:
  -h, --help  Show this message and exit.

```

and then with 8.1.8
```
$ python  failing.py -h


$ python  failing.py --help
Error: Option '--help' requires an argument.


```

- This code works more or less in 8.1.7 and 8.1.8 with custom class and no "--help" option
```Python
import click


class PluggableCommandLineOption(click.Option):
    pass


@click.command()
@click.help_option("-h", cls=PluggableCommandLineOption)
def scancode():
    """Regression in pkg:pypi/click@8.1.8"""
    pass


if __name__ == "__main__":
    scancode()

```
and then with 8.1.7
```
$ python  works2.py -h
Usage: works2.py [OPTIONS]

  Regression in pkg:pypi/click@8.1.8

Options:
  -h      Show this message and exit.
```

and then with 8.1.7
```
$ python  works2.py --help
Usage: works2.py [OPTIONS]

  Regression in pkg:pypi/click@8.1.8

Options:
  -h      Show this message and exit.
  --help  Show this message and exit
```

and then with 8.1.8
```
$ python  works2.py -h

```
and then with 8.1.8 note the changes in `-h TEXT`
```
$ python  works2.py --help
Usage: works2.py [OPTIONS]

  Regression in pkg:pypi/click@8.1.8

Options:
  -h TEXT
  --help   Show this message and exit.

```

Environment:

- Python version: 3.9 to 3.11
- Click version: 8.1.8

kdeldycke:
This is a proposal to fix #2832, which points to a subtle change in behavior of the help option introduced in 8.1.8.

The root of the problem is that by deduplicating code in 8.1.8 (see: #2563), I introduced a `HelpOption` class, to centralize the setup necessary to make a standard `Option` behave like an help option.

In #2832, a user is relying on a custom class to feed the `@help_option` decorator. This bypass all the custom setup made by `HelpOption`, and produce the discrepancy in behavior between 8.1.7 and 8.1.8.

This PR:
- Reverts back to setting up all parameters of the help option in the decorator instead of the `HelpOption` class
- Aligns `@help_option` implementation with the way Click is doing for the others (like `@password_option`, `@version_option`, ...) 
- Removes the `click.decorators.HelpOption` class as a consequence
- Adds a complete unit test to check combination of custom class and option names leads to expected results

Checklist:

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1014,25 +1014,21 @@ class Command:
         return list(all_names)
 
     def get_help_option(self, ctx: Context) -> Option | None:
-        """Returns the help option object."""
-        help_options = self.get_help_option_names(ctx)
+        """Returns the help option object.
 
-        if not help_options or not self.add_help_option:
+        Skipped if :attr:`add_help_option` is ``False``.
+        """
+        help_option_names = self.get_help_option_names(ctx)
+
+        if not help_option_names or not self.add_help_option:
             return None
 
-        def show_help(ctx: Context, param: Parameter, value: str) -> None:
-            if value and not ctx.resilient_parsing:
-                echo(ctx.get_help(), color=ctx.color)
-                ctx.exit()
+        # Avoid circular import.
+        from .decorators import help_option
 
-        return Option(
-            help_options,
-            is_flag=True,
-            is_eager=True,
-            expose_value=False,
-            callback=show_help,
-            help=_("Show this message and exit."),
-        )
+        # apply help_option decorator and pop resulting option
+        help_option(*help_option_names)(self)
+        return self.params.pop() # type: ignore[return-value]
 
     def make_parser(self, ctx: Context) -> _OptionParser:
         """Creates the underlying option parser for this command."""
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -98,6 +98,8 @@ Unreleased
 -   ``Choice`` is now generic and supports any iterable value.
     This allows you to use enums and other non-``str`` values. :pr:`2796`
     :issue:`605`
+-   Fix setup of help option's defaults when using a custom class on its
+    decorator. Removes ``HelpOption``. :issue:`2832` :pr:`2840`
 
 Version 8.1.8
 -------------
--- a/src/click/__init__.py
+++ b/src/click/__init__.py
@@ -19,7 +19,6 @@ from .decorators import command as command
 from .decorators import confirmation_option as confirmation_option
 from .decorators import group as group
 from .decorators import help_option as help_option
-from .decorators import HelpOption as HelpOption
 from .decorators import make_pass_decorator as make_pass_decorator
 from .decorators import option as option
 from .decorators import pass_context as pass_context
--- a/src/click/decorators.py
+++ b/src/click/decorators.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import inspect
 import typing as t
-from collections import abc
 from functools import update_wrapper
 from gettext import gettext as _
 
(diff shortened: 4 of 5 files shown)


AndreasBackx:
Can you rebase this on stable then we can include this in 8.1.9 and release 8.2.0.

kdeldycke:
Yes! I can rebase it to 8.1.9. I was targeting 8.2.0 as I didn't know if a 8.1.9 would see the light of day given that 8.2.0 seems imminent.

Also, something I realized, when #2563 was merged back into `main` branch (i.e. the future 8.2.0 release), the new `HelpOption` class was properly merged, but the duplicate in-place `Option` creation in `get_help_option()` was not replaced with the new `HelpOption`. See: https://github.com/pallets/click/blob/8fcf9b0c5bbfcc9231298d76cde29ae8b0fc50d0/src/click/core.py#L1028-L1035

So I felt like this PR, by targeting `main` could solve both issues at the same time. What do you think?

src/click/core.py
```diff
@@ -1014,25 +1014,26 @@ def get_help_option_names(self, ctx: Context) -> list[str]:
...
+        from .decorators import help_option
+
+        def dummy_func() -> None:
+            pass
+
+        # Call @help_option decorator to produce an help option with proper
+        # defaults and attach it to the dummy function defined above.
+        help_option(*help_option_names)(dummy_func)
+
+        return dummy_func.__click_params__.pop()  # type: ignore[no-any-return, attr-defined]
```
davidism, line 1036: A throw-away function isn't needed here. The option decorator can be applied to the command `self` and popped from `self.params`.

```suggestion
        # apply help_option decorator and pop resulting option
        help_option(*help_option_names)(self)
        return self.params.pop() # type: ignore[return-value]
```

src/click/core.py
```diff
@@ -1014,25 +1014,26 @@ def get_help_option_names(self, ctx: Context) -> list[str]:
         return list(all_names)
 
     def get_help_option(self, ctx: Context) -> Option | None:
-        """Returns the help option object."""
-        help_options = self.get_help_option_names(ctx)
+        """Returns the help option object.
 
-        if not help_options or not self.add_help_option:
+        Unless ``add_help_option`` is ``False``.
```
davidism, line 1019:
```suggestion
        Skipped if :attr:`add_help_option` is ``False``.
```

kdeldycke:
I just merged the changes proposed by @davidism and rebase this PR on top of `main`.

All tests are passing but the `typing` workflow job, which is also failing for the same reason on the vanilla `main` branch. Same thing for the `pre-commit.yaml` workflow that is failing for reasons not related to this PR.

So unless you have other issues to point to, I think this PR is ready to be merged upstream.

PR #2846, opened 2025-01-17
Fix Zsh completions with colons

### LINKED ISSUE #2703: zsh completion still misbehaving in the presence of `:`
<!--
This issue tracker is a tool to address bugs in Click itself. Please use
Pallets Discord or Stack Overflow for questions about your own code.

Replace this comment with a clear outline of what the bug is.
-->

<!--
Describe how to replicate the bug.

Include a minimal reproducible example that demonstrates the bug.
Include the full traceback if there was an exception.
-->

Given `foo` containing:

```python
#!/usr/bin/env python3
from click.shell_completion import CompletionItem
import click

class Foo(click.ParamType):
        return value

    def shell_complete(self, ctx, param, incomplete):
        return [
            CompletionItem(value=value, help="adsf") for value in 
            ["baz:quux", "spam:eggs"]
        ]

@click.command()
@click.argument("foo", type=Foo())
def main(foo):
    print(foo)
```

and running shell completion generation, followed by attempting to complete the command (indicated via `<Tab>`, produces:

```
⊙  PATH=$PWD:/Users/Julian/.local/share/virtualenvs/bowtie/bin/:$PATH
⊙  eval "$(_FOO_COMPLETE=zsh_source ./foo)"                                                                                                                                                                                                                                                                         
⊙  foo <Tab>
baz   -- quux:adsf
spam  -- eggs:adsf
```

rather than the expected:

```
baz:quux   -- adsf
spam:eggs -- adsf
```

<!--
Describe the expected behavior that should have happened but didn't.
-->

This seems to be a regression on #1812.

The fix mentioned there (of "fake escaping" the colons via `.replace(":", "\\:")` seems to work.

Environment:

- Python version:

```
⊙  python --version                                                                                                                                                                                                                                                                                                 julian@Airm
Python 3.11.9
```

- Click version:

```
⊙  python -m pip show click                                                                                                                                                                                                                                                                                         julian@Airm
Name: click
Version: 8.1.7
Summary: Composable command line interface toolkit
Home-page: https://palletsprojects.com/p/click/
Author: 
Author-email: 
License: BSD-3-Clause
Location: /Users/julian/.dotfiles/.local/share/virtualenvs/bowtie/lib/python3.11/site-packages
Requires: 
Required-by: rich-click, trogon, virtue
```

the-13th-letter:
The Zsh completion script uses different primitives to load completion items into Zsh depending on whether a completion item has a description or not.

  - For items with descriptions, the Zsh `_describe` function is used, which expects item plus description pairs as values, separated by a colon.  In turn, any colon within the item must be escaped.
  - For items without descriptions, the Zsh `compadd` builtin is used, which expects literal items, without escaping.  Actually escaping colons in the item will corrupt the item.

`click`'s built-in Zsh completion class does not escape colons in the item.  So it only handles the second case correctly, not the first; see e.g. #2703.  This PR fixes this by choosing a suitable serialization, for each item separately, according to the abovementioned escaping requirements.

Since at its core this is an issue of two systems not agreeing on the format for exchanging data, and since we principally control both systems, we could fix this on either side.  However, because the Zsh completion script is typically already installed in the user's shell, which is both harder to detect within `click` and harder to change or influence, we opt to change the `click` side instead, even if it the serialization needs to be more irregular.

fixes #2703

---

  - “docs“ presumably not applicable in this case?
  - Would cause merge issues because `CHANGES.rst` has no entry yet for post-8.2.0 changes.
  - Not applicable in this case?

### DIFF
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -124,6 +124,8 @@ _SOURCE_BASH = """\
 %(complete_func)s_setup;
 """
 
+# See ZshComplete.format_completions below, and issue #2703, before
+# changing this script.
 _SOURCE_ZSH = """\
 #compdef %(prog_name)s
 
@@ -366,7 +368,23 @@ class ZshComplete(ShellComplete):
         return args, incomplete
 
     def format_completion(self, item: CompletionItem) -> str:
-        return f"{item.type}\n{item.value}\n{item.help if item.help else '_'}"
+        """"""
+        # See issue #1812 and issue #2703 for context.
+        #
+        # Items *with* help are registered with zsh's `_describe`, which
+        # splits off the help/description after the first colon.  So
+        # escape all colons in item.value so that they arrive intact.
+        #
+        # Items *without* help are registered with zsh's `compadd`,
+        # which does no colon handling.  So do *not* escape colons in
+        # item.value, lest the completions themselves ultimately get
+        # mangled.
+        #
+        # Finally, there is no functional difference between using an
+        # empty help and using "_" as help.
+        help_ = item.help or "_"
+        value = item.value.replace(":", r"\:") if help_ != "_" else item.value
+        return f"{item.type}\n{value}\n{help_}"
 
 
 class FishComplete(ShellComplete):
--- a/tests/test_shell_completion.py
+++ b/tests/test_shell_completion.py
@@ -1,3 +1,4 @@
+import textwrap
 import warnings
 
 import pytest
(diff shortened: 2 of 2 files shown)


Rowlando13:
Thanks for taking the time to submit this PR. We will review this.

the-13th-letter:
Rebased onto and retargeted to `main`, since `release-8.2.0` looks pretty dead to me. I used this opportunity to also add some TLDR-style inline commentary to the patch, and fix some other commentary typos.

AndreasBackx:
Apologies for the delay. This looks pretty good, but it's a bit hard to review. We'd ideally need some way to create tests for autocomplete. Does anyone know how we could possibly trigger autocomplete? If so, we might be able to use [scrut](https://github.com/facebookincubator/scrut) to create a test for it.

I think it should be fine with just a note in the changes and if that's done, we can merge this afaik.

src/click/shell_completion.py
```diff
@@ -366,7 +372,24 @@ def get_completion_args(self) -> tuple[list[str], str]:
...
+        # splits off the help/description after the first colon.  So
+        # escape all colons in item.value so that they arrive intact.
+        #
+        # Items *without* help are registered with zsh's `compadd`,
+        # which does no colon handling.  So do *not* escape colons in
+        # item.value, lest the completions themselves ultimately get
+        # mangled.
+        #
+        # Finally, there is no functional difference between using an
+        # empty help and using "_" as help.
```
AndreasBackx, line 389: Clean this up a tad perhaps? It'd be nice to have this explained in-line if possible.

AndreasBackx (approved):
See previous comment, a nitpick and otherwise just CHANGES that needs an addition.

AndreasBackx (changes requested):
Meant to click request changes, woopsie!

the-13th-letter:
Rebased onto `stable` as of today, and attempted to clean up the inline commentary as requested.

> Apologies for the delay. This looks pretty good, but it's a bit hard to review. We'd ideally need some way to create tests for autocomplete. Does anyone know how we could possibly trigger autocomplete? If so, we might be able to use [scrut](https://github.com/facebookincubator/scrut) to create a test for it.

Can't help you there, sorry. To test my submission, I did write a small < 50-line click application that completed filenames manually, sometimes adding help texts, sometimes even with colons, and then interactively tab-completed arguments in a `zsh` session, in a directory with specially-prepared filenames. This isn't very generalizable to other test uses, and the triggering of tab completion was still manual.

src/click/shell_completion.py
the-13th-letter: Commentary rewritten in the current iteration of this PR.

AndreasBackx:
I've rebased this on stable. Then I've removed the ZSH template check, I'm not sure why we're checking the output of the ZSH completion here. We only really care about the output, don't we? Let me know in case I missed something. Otherwise this seems good to land if CI is good.

AndreasBackx (approved):

PR #2854, opened 2025-02-12
Shell completion overrides are documented as taking keyword arguments

benblank:
The documentation indicates that shell completion override functions are "passed 3 keyword arguments", but are in fact [passed *zero* keyword arguments and three *positional* arguments](https://github.com/pallets/click/blob/5dd628854c0b61bbdc07f22004c5da8fa8ee9481/src/click/core.py#L2440). 🙂

### DIFF
--- a/docs/shell-completion.md
+++ b/docs/shell-completion.md
@@ -138,7 +138,7 @@ def cli(ev):
 ## Overriding Value Completion
 
 Value completions for a parameter can be customized without a custom type by providing a `shell_complete` function. The
-function is used instead of any completion provided by the type. It is passed 3 keyword arguments:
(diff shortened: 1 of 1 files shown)


Rowlando13:
Please rebase onto stable. No change log entry needed. Those are only needed for code changes. Also we changed the doc to myst instead of rst. So you will be looking for shell-completion.md

benblank:
Thanks for circling back on this; I've rebased on main, including the changes you mention.

Rowlando13:
Please rebase on stable. That is where docs fixes go so they are available immediately.

benblank:
Shoot, I completely missed that. 🤦

Fixed.

Issue #2706, opened 2024-04-16
Better i18n with gettext: use class-based API


carmenbianca:
Currently, Click implements gettext using the classic GNU gettext API. That looks like this:

```python
from gettext import _

print(_("Hello, world!"))
```

This API depends on a global state in the gettext module. By calling [`gettext.textdomain()`](https://docs.python.org/3/library/gettext.html#gettext.textdomain), the active translation domain is changed for _all_ Python modules that use the classic GNU gettext API.

This side effect is usually desirable, except when your module is imported by another module as a library. So you usually don't want to call `gettext.textdomain()` without putting it behind some sort of function call. With argparse, this is easy: put it in your `main` function before you even create the `ArgumentParser` object. With Click, I'm not sure this is possible:

- Your main group/command is decorated with all manner of Click magic and its contents may not actually be run (e.g. `--help`).
- There is no way to pre-hook a group/command with a function call that I am aware of.

So you end up having to call `gettext.textdomain()` on import of your module containing your Click groups/commands.

We can fix that by switching to the class-based API. Because Click will still need to support the old API as well for backwards compatibility, my proposal looks a little as follows. Create a module `click.i18n` with the following contents (simplified):

```python
import gettext as _gettext_module

TRANSLATIONS = None

def gettext(message):
    if TRANSLATIONS is None:
        return _gettext_module.gettext
    return TRANSLATIONS.gettext

# alias
_ = gettext
```

Now, elsewhere in Click, you replace all `from gettext import _` with `from .i18n import _`.

Subsequently, we can create a function `install_translations(translations)` in `i18n.py` that replaces the `TRANSLATIONS` global constant with an instantiated [`GNUTranslations`](https://docs.python.org/3/library/gettext.html#the-gnutranslations-class) object. This function would _still_ need to be called before the consumer's `main` function, but it wouldn't change the gettext global state—it would only change Click's. Which, as far perfectionism goes, is probably tolerable. It would be better still if there was a pre-hook, but this is fine.

Furthermore, the consumer could use different domains for Click's `TRANSLATIONS` object and their own, allowing them to separate their own translations from Click's, and hypothetically reuse the Click translations in other projects.

In fact, having done this plumbing, Click could even ship its own translation strings, getting rid of duplication efforts of translating the same Click strings. Click's own translations could then be activated using e.g. `install_click_translations()` without any arguments.

In summary, the problems solved by this:

- Importing the Click consumer's module no longer changes the global state of the gettext module.
[...]

PR #2864, opened 2025-03-01
Improve internationalization

v--:
While experimenting with internationalization of a personal project, I encountered two issues:
* I had to translate a bulk of messages coming from click.
* I had to configure gettext globally (or monkey-patch it) before loading click, which introduces many complications. This is discussed by another user, [carmenbianca](https://github.com/carmenbianca), in #2706.

I was initially thinking of creating a discussion about this, but I wanted to also share an implementation. I am myself inexperienced in translating software, so any criticism is welcome. I will list what I have done and what I am unsure about. Perhaps [carmenbianca](https://github.com/carmenbianca) can chime in?

* I have added some .po files with translations into Russian and Bulgarian. These are straightforward, but there are subtleties:
    * They must be compiled to .mo files. This is an issue because the setup becomes more difficult, especially on Windows where gettext tools must be installed manually.  I found no easy way to compile the .po files to .mo files. I have thus committed the .mo files, as well as a compilation script. gettext really is clunky.
    * The locale files are under the `click` domain. Client applications will have to use other gettext domains. I am unaware of best practices here.
    * In order for flit to package these files, they must be put in the source code directory.

* I have replaced the calls to gettext with a wrapper that stores a locale internally. This should not be a problem generally, but I want to hear if you have any precautions.

* I have added some unit tests that unfortunately depend on gettext. Hence, on gettext itself and the .mo files discussed above. Conditionally skipping tests in pytest is fortunately easy enough, if running them in CI (or on Windows) becomes a problem.


davidism:
Thank you for working on this. I will need to very carefully review this at some point, since it affects quite a bit of the code and introduces a new pattern that needs to be maintainable later. Additionally, based on experiments for docs translations, I want to manage translations using Weblate, so I'll have to figure out how to set that up for this project. This is unlikely to happen soon, so I just want to set expectations.

v--:
@davidism I appreciate your comment. I also just realized that I haven't thanked the team for developing click - it's one of the libraries I have been using consistently over the years.

Now, more to the point. I have no expectations for this pull request myself. As mentioned, the translations can easily be extracted into another package (or implemented differently in a later pull request). This would leave mostly mundane changes that should not induce maintenance overhead (replacing which implementations of `gettext` and `ngettext` get used).

Rowlando13:
@v--  I appreciate your candor. I don't have a super long history with Click so sometimes it is a bit hard to tell where the line for should be with regard to accepting something. Sounds @davidism is going to help you out.

kdeldycke:
Browsing i18n-related PR, just realized this one has some unit tests that I was suggesting implementing in https://github.com/pallets/click/pull/2902 (by @ic).

What I see here that is left to do is:
- rebase that PR on top of `main` branch
- drop the translation files as @v-- proposed (`the translations can easily be extracted into another package (or implemented differently in a later pull request)`), which will reduce the diff, and do not interfere with @davidism future plans to rely on Weblate
- maybe adopt @carmenbianca's proposal in #2706 to implement `install_translations` to decouple the translations, but keep the current `gettext` calls as fallback so nothing changes for previous users

@v-- are you interested in rebooting the work on this PR? I can take over and try to merge your code from this PR with the good ideas from https://github.com/pallets/click/issues/2706 .

v--:
@kdeldycke I have rebased my original pull request, but did not make any other changes. You can add/remove/change anything as you see fit.

One thing should be noted. The tests are failing, but the "fix" for this is not trivial. Unlike a year ago when I created the pull request, some messages such as "Show this message and exit." are now fixed and so dynamically changing the language is no longer possible without some refactoring. It's an unpleasant edge case where some strings are translated according to the new locale, but others are not.

ic:
@v-- Thank you for your work. Discovering your OP now, we had the exact same problems for the languages we want to support, and concur on your analysis and approach. In fact I implemented about the same code last year too! But backed off and focused on a narrower goal---first localise as many strings as possible, ideally all (#2902).

This was merged last week. I have checked on `main` for the string you take as example, and it is localised now:

```
src/click/decorators.py
548:    kwargs.setdefault("help", _("Show this message and exit."))
```

Any issue with this? (I for one forget to sync my fork with the original repo)

The PR may be missing recent new strings. It was decided to merge it for now, and likely roll forward as needed.

v--:
@ic I just rebased my changes again because of a merge conflict.

Anyhow, the problem I was describing concerns the _dynamic_ change of languages. See the failing unit test.

The gist is the following line from `src/click/decorators.py`:

```python
kwargs.setdefault("help", _("Show this message and exit."))
```

The default is set _after_ getting translated, but you can nevertheless change the language following that. Then that particular translation will stick to the original language, but most strings will use the new language.

There are simple ways to mitigate this (e.g. using a `default_factory` instead of `setdefault`), but since I am not a maintainer, I have no preference for whether to support dynamic translation at all.

ic:
Thanks @v-- , now very clear to me what the problem is. It seems the issues are bound to `src/click/decorators.py` (one way to deal with it could be to turn them into callables).

I understand this breaks tests _in the same runtime_, but I wonder whether it is a problem in practice. It could when the CLI is interactive and a user somehow changes the locale in the shell executing the CLI, but is it likely?

CLI are often (mostly?) one shot runs, where the execution environment does not change (much). So wondering whether there is a problem, even though the breaking test is technically correct.

PR #2880, opened 2025-04-08
Add format customization flow to `click.prompt` and `click.confirm` via `PromptBuilder` class

### LINKED ISSUE #2875: Method used to build prompts under the hood does not allow for full format customization
Currently method [`_build_prompt`](https://github.com/pallets/click/blob/main/src/click/termui.py#L60-L73) is implemented in very opinionated way regarding formatting how to display choices or default value. Considering how good of a job the rest of the features do in regards to potential extensions and customizations, I am surprised that the best way I found to customize this method is to overwrite it before use, for example as follows:

```python
import click

def my_own_build_prompts(*args, **kwargs):
    pass

click.termui._build_prompt = my_own_build_prompts
```

---
Even though it is "_technically_" possible to override, I'd argue the fact that in order to find it I had to dig deep into codebase and find the details of implementation feels "hacky" and "not intended" - and for sure error prone and not future-proof.

I am very happy to contribute and implement the solution, but I'd rather hear some opinions from the library owners beforehand whether this is something they'd find valuable.

AdamPiotrowski91:
This PR implements a base class which allows to customize prompt format when used directly via `click.prompt` or `click.confirm` methods.

It adds an optional parameter which can be an instance of a `PromptBuilder` class or one inheriting from it.

You can customize all of the below in any combination:

- how the prompt is built via overriding `build` method
- how the choices are shown via overriding `add_choices` method
- how the default is shown via overriding `add_default` method
- how to calculate if choices should be shown via overriding `should_add_choices` method
- how to calculate if default should be shown via overriding `should_add_default` method

---
fixes #2875

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,12 @@
 .. currentmodule:: click
 
+Version 8.2.1
+-------------
+
+Unreleased
+
+-   Create ``PrompBuilder`` class for customizing prompts display and add ``builder_cls`` parameter to ``prompt`` and ``confirm`` methods. :issue:`2875`
+
 Version 8.2.0
 -------------
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -41,6 +41,8 @@ Decorators
 Utilities
 ---------
 
+.. autoclass:: PromptBuilder
+
 .. autofunction:: echo
 
 .. autofunction:: echo_via_pager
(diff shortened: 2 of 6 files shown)


AdamPiotrowski91:
Tests pass, but I am not sure, based on the logs, why do the other jobs fail. Would appreciate some help in this regard.

PR #2890, opened 2025-05-02
Localisation of more strings

ic:
# Problem and proposal
Using Click in a multi-lingual package, we would like to localise more strings than currently possible.

In our work before this PR, we reach like:

```
> export LANG=ja_JP.UTF-8
> artefacts hello
Usage: artefacts hello [OPTIONS] プロジェクト名
Try 'artefacts hello --help' for help.

Error: Missing argument 'プロジェクト名'.
```

The package we develop yields the `artefacts` command, based on Click, works alright to translate our strings, as well as a range of strings in Click wrapped with `gettext`.

With this PR, we get:

```
> export LANG=ja_JP.UTF-8
> artefacts hello
使用法:  artefacts hello [OPTIONS] プロジェクト名
ヘルプについては 'artefacts hello --help' を試してください。

エラー：引数がありません 'プロジェクト名'.
```

Which looks like possibly full coverage of meaningful strings in Click.

Note this PR addresses two issues:
1. Several strings where not wrapped into `gettext`
2. Libraries like PyBabel (most users?) cannot work with f-strings, so the pathological approach to use the format method, and to deactivate the PyUpgrade U032 rule (converts to f-strings).

# Limitation, side-effects and discussion

It looks like part of this PR may be desired, as localisation is already in place---this mainly covers more of the strings (I'd say all the strings, but localisation does not always make sense).

However in the current approach I had to "deal" with f-strings and PyUpgrade, which may be undesired change for the project. On top of that it seems there is a bug in either Ruff or PyUpgrade in accepting ignore rules on multi-line commands. So a bunch of files get a file-global deactivation of the U032 rule, which means these files will not get checked for f-strings.

F-strings are assumedly desired, but given (1) [PyBabel does not and may not support them](https://github.com/python-babel/babel/issues/594), and (2) Click is library code, the project may (have to) accept the format method everywhere localisation is needed.

Recent discussions mention [f-strings may work in PyBabel from Python 3.12](https://github.com/python-babel/babel/issues/1001). We did not confirm it, as we want to support down to the oldest supported Python3. So the changed proposed here may be transient for a couple years (then PyUpgrade may be reactivated and let work).

# Related work

This PR only aims at localising more strings, so related to i18n issues.

On the way to this PR, we have considered a couple alternatives, notably trying to use the class API of Python's `gettext`. Some elements of discussion here may be useful to:

* https://github.com/pallets/click/pull/2864 by @v--
* https://github.com/pallets/click/issues/2706 by @carmenbianca

In fact, [Carmen's post](https://www.carmenbianca.eu/en/post/2024-04-14-how-to-set-up-python-internationalisation-with-click-poetry-forgejo-and-weblate/) helped solve an issue in using catalogues from different domains at runtime (thanks!).

# Checklist on [CONTRIBUTING](https://palletsprojects.com/contributing/pr)

- [ ] Add an entry in CHANGES.rst summarizing the change and linking to the issue.

At submission time, nothing checked here, as first would like to make sure this PR target is acceptable (likely not as-is). The tox-based checks all pass, though (i.e. running the `tox` command returns all green, except the skipped tests).


davidism:
Please resubmit this from your own account. Submitting from a repo that does not match the user causes GitHub to disable allowing maintainers to work on the PR branch.

PR #2893, opened 2025-05-10
drop end of life python versions

davidism:
Per our [version support policy](https://palletsprojects.com/versions):

> When a release is made, it supports all Python versions that are not within six months of end of life (EOL). See [Python's EOL calendar](https://devguide.python.org/versions/) for timing.

3.7 and 3.8 are already end of life. 3.9 is end of life in October.


mikenerone:
Umm, folks, clearly "not within six months of end of life" is intended to mean six months _after_ EOL. This latest release, 8.2.0, just dropped support for Python 3.9 five months _before_ its EOL.

PR #2902, opened 2025-05-12
Localisation of more strings

ic:
# Problem and proposal
Using Click in a multi-lingual package, we would like to localise more strings than currently possible.

In our work before this PR, we reach like:

```
> export LANG=ja_JP.UTF-8
> artefacts hello
Usage: artefacts hello [OPTIONS] プロジェクト名
Try 'artefacts hello --help' for help.

Error: Missing argument 'プロジェクト名'.
```

The package we develop yields the `artefacts` command, based on Click, works alright to translate our strings, as well as a range of strings in Click wrapped with `gettext`.

With this PR, we get:

```
> export LANG=ja_JP.UTF-8
> artefacts hello
使用法:  artefacts hello [OPTIONS] プロジェクト名
ヘルプについては 'artefacts hello --help' を試してください。

エラー：引数がありません 'プロジェクト名'.
```

Which looks like possibly full coverage of meaningful strings in Click.

Note this PR addresses two issues:
1. Several strings where not wrapped into `gettext`
2. Libraries like PyBabel (most users?) cannot work with f-strings, so the pathological approach to use the format method, and to deactivate the PyUpgrade U032 rule (converts to f-strings).

# Limitation, side-effects and discussion

It looks like part of this PR may be desired, as localisation is already in place---this mainly covers more of the strings (I'd say all the strings, but localisation does not always make sense).

However in the current approach I had to "deal" with f-strings and PyUpgrade, which may be undesired change for the project. On top of that it seems there is a bug in either Ruff or PyUpgrade in accepting ignore rules on multi-line commands. So a bunch of files get a file-global deactivation of the U032 rule, which means these files will not get checked for f-strings.

F-strings are assumedly desired, but given (1) [PyBabel does not and may not support them](https://github.com/python-babel/babel/issues/594), and (2) Click is library code, the project may (have to) accept the format method everywhere localisation is needed.

Recent discussions mention [f-strings may work in PyBabel from Python 3.12](https://github.com/python-babel/babel/issues/1001). We did not confirm it, as we want to support down to the oldest supported Python3. So the changed proposed here may be transient for a couple years (then PyUpgrade may be reactivated and let work).

# Related work

This PR only aims at localising more strings, so related to i18n issues.

On the way to this PR, we have considered a couple alternatives, notably trying to use the class API of Python's `gettext`. Some elements of discussion here may be useful to:

* https://github.com/pallets/click/pull/2864 by @v--
* https://github.com/pallets/click/issues/2706 by @carmenbianca

In fact, [Carmen's post](https://www.carmenbianca.eu/en/post/2024-04-14-how-to-set-up-python-internationalisation-with-click-poetry-forgejo-and-weblate/) helped solve an issue in using catalogues from different domains at runtime (thanks!).

# Checklist on [CONTRIBUTING](https://palletsprojects.com/contributing/pr)

- [ ] Add an entry in CHANGES.rst summarizing the change and linking to the issue.

At submission time, nothing checked here, as first would like to make sure this PR target is acceptable (likely not as-is). The tox-based checks all pass, though (i.e. running the `tox` command returns all green, except the skipped tests).

This PR supersedes #2890, because of [a problem with GitHub](https://github.com/pallets/click/pull/2890#issuecomment-2870531948).


Rowlando13:
I don't think we are willing to limit the use of f strings in the project at this time. Thoughts @davidism?

src/click/core.py
```diff
@@ -1,3 +1,11 @@
+# Ask Ruff to accept the format method on strings, and not let pyupgrade
+# always force f-strings. The latter are unfortunately not supported yet
+# by Babel, a localisation library.
+#
+# Note: Using `# noqa: UP032` on lines has not worked, so a file
```
davidism: None of these `noqa` marks are needed. Ruff is not trying to autoupgrade anything if I remove the comments, inline or file-level.

src/click/core.py
```diff
@@ -76,15 +84,15 @@ def _check_nested_chain(
         return
 
     if register:
-        message = (
-            f"It is not possible to add the group {cmd_name!r} to another"
-            f" group {base_command.name!r} that is in chain mode."
-        )
+        message = _(
```
davidism: We're only translating user-facing messages at this time, not developer-facing. Please remove all such translation markings.

davidism:
For user-facing messages, yes we can use the `_("").format()` pattern, f-strings can't be used for translation.

The `# noqa: UP036` mark does not seem to be needed, please remove this.

We're only translating user-facing messages at this time, not developer-facing. Please remove all such translation markings.

src/click/core.py
```diff
@@ -1114,9 +1124,9 @@ def format_help_text(self, ctx: Context, formatter: HelpFormatter) -> None:
 
         if self.deprecated:
             deprecated_message = (
-                f"(DEPRECATED: {self.deprecated})"
+                _("(DEPRECATED: {target})".format(target=self.deprecated))
                 if isinstance(self.deprecated, str)
-                else "(DEPRECATED)"
+                else _("(DEPRECATED)")
             )
             text = _("{text} {deprecated_message}").format(
```
davidism: This shouldn't have ever been marked for translation. It should be removed, since you've added the translation to the proper location above.

davidism:
Most of the changes here need to be rolled back, they are not part of the user-facing output.

ic:
Hey thanks for the review and feedback. A quick note to let you know I expect to improve the work here soon, most likely in November. Sorry for anyone watching and expecting this sooner.

src/click/_termui_impl.py
```diff
@@ -174,7 +174,13 @@ def format_eta(self) -> str:
             hours = t % 24
             t //= 24
             if t > 0:
-                return f"{t}d {hours:02}:{minutes:02}:{seconds:02}"
+                return "{d}{day_label} {h:02}:{m:02}:{s:02}".format(
+                    d=t,
+                    day_label=_("d"),
```
carmenbianca, line 178:
```suggestion
                return "{d}{day_label} {h:02}:{m:02}:{s:02}".format(
                    d=t,
                    # TRANSLATORS: The 'd' stands for 'day'.
                    day_label=_("d"),
```

src/click/types.py
```diff
@@ -853,7 +853,11 @@ def convert(
 
             return f
         except OSError as e:
-            self.fail(f"'{format_filename(value)}': {e.strerror}", param, ctx)
+            self.fail(
```
ic, line 891: Reformatting from `ruff 0.14.10`'s format command.

src/click/core.py
ic: Updated from former Ruff 0.8.1 to the latest 0.14.10, and as you reported, the UP032 do not trigger anymore.

All files cleaned from the comments, as well as the `noqa` marks.

src/click/core.py
ic: I have done a pass on all new translations and removed the ones that looked like dev-facing. How does it look now? I am not sure about a couple.

src/click/core.py
ic: Done!

AndreasBackx (approved):
This looks fine now. Though I'd love for someone else to give this a final eye because I've never really touched too much gettext stuff.

ic:
For the record and contribute to the review, we have been using this branch for a year in our CLI, without noticeable problem so far:

The CLI is for robotics engineering: `pip install artefacts-cli`

```
# export LANGUAGE=en_GB:en
# artefacts run-remote test_manipulation
Scanning source...
Job data compression: 5/5 Done
Job data uploading:   5/5 Done
Job meta uploading:   3/3 Done
Uploaded 42 MiB of compressed job data and 1 KiB of job metadata.
The new job will show up shortly at https://app.artefacts.com/artefacts/artefacts-toolkit-testsuite

# export LANGUAGE=ja_JP:ja
# artefacts run-remote test_manipulation
ソースをスキャン中...
ジョブデータの圧縮: 5/5 完了
ジョブデータのアップロード:   5/5 完了
ジョブメタデータのアップロード:   3/3 完了
圧縮されたジョブデータ 32 MiB 、およびジョブメタデータ 2 KiB のアップロードが完了しました。
新しいジョブは、まもなく https://app.artefacts.com/artefacts/artefacts-toolkit-testsuite に表示されます。
```

(if you try the package, the output will differ as working on it, but the translation is in use for a year already)

kdeldycke:
@ic how hard would it be to add a couple of unittests? At least to prove and lockdown that Click is properly translating strings? With unittests we can force future contributors to be aware of this support.

Other than that this PR looks good for inclusion into 8.4.0.

src/click/_winconsole.py
```diff
@@ -169,7 +170,7 @@ def _get_error_message(errno: int) -> str:
             return "ERROR_SUCCESS"
         elif errno == ERROR_NOT_ENOUGH_MEMORY:
             return "ERROR_NOT_ENOUGH_MEMORY"
-        return f"Windows error {errno}"
+        return _("Windows error: {error}").format(error=errno)
```
kdeldycke, line 173: I like how you unified the `Windows error:` message string here by adding a colon.

src/click/_termui_impl.py
kdeldycke: Are we using the `# TRANSLATORS:` standard anywhere else in the code? I don't think so, and my understanding is that we have to specify the `--add-comments=TRANSLATORS` option to have them included in the generated files right? Isn't there a better way to express the translated `d` here?

We can start using the `# TRANSLATORS:` convention but that would need a documentation update somewhere.

src/click/formatting.py
```diff
@@ -153,7 +153,7 @@ def write_usage(self, prog: str, args: str = "", prefix: str | None = None) -> N
             ``"Usage: "``.
         """
         if prefix is None:
-            prefix = f"{_('Usage:')} "
+            prefix = "{usage} ".format(usage=_("Usage:"))
```
kdeldycke, line 156: Maybe there is a subtlety I don't get but both `f"{_('Usage:')} "` and `"{usage} ".format(usage=_("Usage:"))` looks equivalent to me no?

kdeldycke:
Grepping through the code I just realized some default strings are not translated, like in:
- https://github.com/pallets/click/blob/831c8f0948af519e45b90801d7430ff25451f972/src/click/decorators.py#L399-L400
- https://github.com/pallets/click/blob/831c8f0948af519e45b90801d7430ff25451f972/src/click/core.py#L1139-L1143
- https://github.com/pallets/click/blob/831c8f0948af519e45b90801d7430ff25451f972/src/click/core.py#L2786-L2790
- https://github.com/pallets/click/blob/831c8f0948af519e45b90801d7430ff25451f972/src/click/core.py#L2600-L2607
- https://github.com/pallets/click/blob/831c8f0948af519e45b90801d7430ff25451f972/src/click/termui.py#L258
- https://github.com/pallets/click/blob/831c8f0948af519e45b90801d7430ff25451f972/src/click/termui.py#L266-L269

src/click/formatting.py
ic: This is about translation reuse (and not best retrospectively). The translated string has no trailing space. The format adds the space at this place in the code.

Not best, as I usually prefer to not include punctuation as much as possible, like the trailing colon. I did no check, but likely done to reduce the amount of change across the codebase.

ic:
@kdeldycke Thank you for the round of review.

1. Unit test, on it starting today. It looks like I can come up with some next week.
2. Translation strings like `Usage:` are often better be `usage` and let the code format and punctuate. Not done to minimise code change, but personal preference and suggestion, perhaps a later PR.
3. The default strings you list up should be included. Two questions:
  * The PR is one year old, and some may be new strings. If acceptable, perhaps working first toward merging this PR, then iterate? As writing tests by next week, the ones you found can be done at the same time.
  * An [earlier comment](https://github.com/pallets/click/pull/2902#discussion_r2296117964) by @davidism points out some strings are for developers only and should not be translated. It makes sense to me as dev is done in English anyway. It looks like only 2 of the ones you found enter this category. Best if we can confirm before I make us oscillating change/review.

src/click/formatting.py
kdeldycke: What I mean is in both `f"{_('Usage:')} "` and `"{usage} ".format(usage=_("Usage:"))`, the `_()` translation utility get the exact same `Usage:` string as parameter. I do not question the punctuation or extra space.

So effectively, these two lines does the exact same thing. So I was questioning why using a `.format()` method instead of the original `f`-style string. I think this change is the result of back and forth between you and @davidism when it was decided to only translate user-facing strings.

What I propose is simply reverting this change to not introduce more noise and make the merge easier.

kdeldycke:
> 1. Unit test, on it starting today. It looks like I can come up with some next week.

No worries, take your time! :) There is a couple of other PRs aligned for 8.4.0 that are going to take more time to land so we're good here! :)

> 2. Translation strings like `Usage:` are often better be `usage` and let the code format and punctuate. Not done to minimise code change, but personal preference and suggestion, perhaps a later PR.

Your attention to punctuation and separation of concern is the right one. See my comment on the review discussion at: https://github.com/pallets/click/pull/2902#discussion_r3181227847 . My proposal is to simply reduce the diff to make the merge to upstream even more easier.

> 3. The default strings you list up should be included. Two questions:
> 
> * The PR is one year old, and some may be new strings. If acceptable, perhaps working first toward merging this PR, then iterate? As writing tests by next week, the ones you found can be done at the same time.

It's better to keep this PR self-contained. Now that this PR has both your attention and mine, I will keep tracking it and helping you so it lands in the upcoming 8.4.0. If you need more time no worries. Just tell us with a quick comment here. If you find the merge too hard to do let me know. I can take over this PR to cleanup and prepare it for merging.

> * An [earlier comment](https://github.com/pallets/click/pull/2902#discussion_r2296117964) by @davidism points out some strings are for developers only and should not be translated. It makes sense to me as dev is done in English anyway. It looks like only 2 of the ones you found enter this category. Best if we can confirm before I make us oscillating change/review.

He is right that only user-facing strings should be translated. Re-reviewing the one I pointed out, yes, they should be translated as these are user-facing strings in the help screen: they're about options or parameters that are marked as deprecated.

src/click/formatting.py
ic: F-strings are not supported in Python for gettext/PyBabel.

I forgot the detail but it is something about low-level implementation of f-strings that prevents detecting them (something of the kind).

ic:
@kdeldycke Thank you for the detail. Proceeding accordingly.

A thought on unit testing: A couple generic tests on the `_` utility works, and wondering whether a CI workflow to check for missing translation would not be useful? Done a couple times in other projects, where `pybabel` can list missing translations, and help failing some CI jobs and add GH labels for missing translations, etc. Retrospectively such CI is more informative and useful to me, rather than unit tests. How does it sound?

davidism:
I don't really think this needs to be unit tested. Seems like a lot of work just to get this in. Maybe we can do that in another talk, but I'm not sure what those tests would reveal anyway.

src/click/formatting.py
kdeldycke: > F-strings are not supported in Python for gettext/PyBabel.
> 
> I forgot the detail but it is something about low-level implementation of f-strings that prevents detecting them (something of the kind).

OK so forget about my comment here and let's keep the change. This is just code style in the end, so we will document it later if needed. Keep the `.format()` style. Thanks for the feedback.

kdeldycke:
> I don't really think this needs to be unit tested. Seems like a lot of work just to get this in. Maybe we can do that in another talk, but I'm not sure what those tests would reveal anyway.

What I was thinking about is a kind of test just doing what @ic is demonstrating in its comment https://github.com/pallets/click/pull/2902#issuecomment-4348801118 . And have that test called `test_translation` or something generic like that, that is just a big and stupid string comparison of the output in Japanese. This would at least cover the high-level translation code path. Could be added to `test_basic.py`. Nobody cared about translation until @ic showed up. I can add that test later in a subsequent PR.

So @ic, no need to go to a complicated route like you proposed with `pybabel` deep introspection of the code. Just a high-level test as I describe is good enough, and better than nothing. Can you also squash your commits here?

ic:
Starting to proceed with the changes.

1. Squashing commits can be done on merge, or you have something specific in mind?
2. [Done] ~I do now the missing messages~, and
3. [Update] ~next on unit test if still wanted.~ Unit tests would amount to test unrelated `gettext` and `ngettext` functions from the `gettext` standard library. Can still do, but now thinking it is not useful to the test suite (it would be useful with custom use, but here plain use on purpose).
4. The CI approach is quite straight forward, with the advantage to be comprehensive. Something like (based on my working workflow):

```
pybabel extract src -o path/to/locales/base.pot # and other options
pybabel update  -i path/to/locales/base.pot -d path/to/locales
pybabel compile -d path/to/locales
msgattrib --untranslated path/to/locales/ja_JP/LC_MESSAGES/artefacts.po -o untranslated.po
if [[ -s untranslated.po ]]
then
  echo "Found  untranslated strings"
  echo "Please translate them and commit"
  gh pr edit ${{ github.event.pull_request.number }} --add-label "Empty translations"
  cat untranslated.po | grep msgid
  exit 1
else
  echo "All expected translations ready to go."
  gh pr edit ${{ github.event.pull_request.number }} --remove-label "Empty translations"
fi
```

Note: All tests pass according to CI here. They also pass on a local Linux machine. But on a Mac Intel, I have a single error with Python 3.14t, 3.13 stress, and tried also the random environment in Tox. It looks unrelated, but for reference:

```
tests/test_utils.py:438: in test_echo_via_pager
    assert pager == expected_pager, (
E   AssertionError: Unexpected pager output in test case 'Exception in generator function argument'
E   assert 'test' == ''
E
E     + test
```

src/click/_termui_impl.py
ic: I introduced that prefix, used only here.

One way to remove the comment and the "d" is to name use, say, `_("day_label")` and "translate" in the EN file as well as other supported languages.

ic: Reviewing the code, it seems it speaks for itself, so removed this exceptional comment.

kdeldycke: OK thanks for the update! :)

src/click/core.py
```diff
@@ -1136,14 +1136,13 @@ def get_short_help_str(self, limit: int = 45) -> str:
             text = ""
 
         if self.deprecated:
+            localised_deprectated = _("deprecated").upper()
```
kdeldycke, line 1139: Oh OK. I see how you decoupled the meaning (i.e. the `deprecated` word), from its style (the `upper()` call). That's neat.

src/click/termui.py
```diff
@@ -255,17 +255,17 @@ def confirm(
         text,
         prompt_suffix,
         show_default,
-        "y/n" if default is None else ("Y/n" if default else "y/N"),
+        _("y/n") if default is None else (_("Y/n") if default else _("y/N")),
```
kdeldycke: Maybe it's too much but why not split this into:
- `{_("y")/_("n")}`
- `{_("y").upper()/_("n")}`
- `{_("y")/_("n").upper()}`

The idea is to reuse the same `_("y")` and `_("n")` items you translated below? Or is this too much splitting?

ThiefMaster: Do people typically want Y/N in this type of CLI translated? E.g. do French speakaers really want O/N instead, or German speakers J/N? (I can tell you that as a German speakers I would never want this!)

kdeldycke:
> 4\. The CI approach is quite straight forward, with the advantage to be comprehensive. Something like (based on my working workflow):

Ok let's see that point in a future PR.

Thanks for your feedback on the tests, we will see that topic later in a future PR.

I just have a last comment on the `y/n` format, and once you answer that, I will squash that PR and merge it.

src/click/termui.py
kdeldycke: > Do people typically want Y/N in this type of CLI translated? E.g. do French speakaers really want O/N instead, or German speakers J/N? (I can tell you that as a German speakers I would never want this!)

I personally agree with you and find that awkward. But it is user facing, so in the name of applying a blanket translatable policy, let's allow this to be translated and changed. In the end this is about taste. So at least with this we allow CLI developers choose to apply the translation or not. The capability exist and is consistent, the rest is developers preferences. 🤷

ThiefMaster: But does a CLI developer get to choose? If they enabled translations in general, they'd also get this one. Which is also kind of a breaking change because it changes the behavior of the CLI. I would make it opt-in somehow.

ThiefMaster: I just tested in a docker container with debian and a german translation and how the `rm -i ...` prompt behaves: it accepts `j` and `ja` as yes, but `y` and `yes` also work. likewise with french where `o` and `oui` are acccepted, but also the english defaults.

So I strongly suggest we do the same here and at least always accept the standard values on top of additional ones coming from translations.

kdeldycke: > I just tested in a docker container with debian and a german translation and how the `rm -i ...` prompt behaves: it accepts `j` and `ja` as yes, but `y` and `yes` also work. likewise with french where `o` and `oui` are acccepted, but also the english defaults.
> 
> So I strongly suggest we do the same here and at least always accept the standard values on top of additional ones coming from translations.

Oh I see. Sorry I did not got your point in your first comment. Thanks for checking other CLIs and illustrating them. And so yes, I agree with you on that front and we should accept both the translated `_("y")` and `_("yes")` as well as the original `y` and `yes` as valid answer in the code below.

ic: Oh I got caught very lazy on this one. Thank you both for the feedback. I was actually wondering how people were workin in a different language... In non-alphabet languages, it seems to me developers (a tiny fraction of a population, right?) work with and expect y/n.

Interestingly asked Qwen (an LLM developed in China) how it is preferred there. It replies they expect y/n. The example it reports:

```
是否继续？(Y/n): 
确认安装？(y/N): 
删除文件？(Y/n): 
```

Same for Japanese. It goes as far as to list "best practices" like: "Never translate y/N or Y/n: Maintains cross-language consistency and avoids terminal width issues."

ic:
Experimented with translations on `y/n`, I concur with @ThiefMaster now. I did try to offer two options like `y/n` or custom, but there are really basic issues in general. For example "y" in Chinese is a single character (是/shi), but you need to type 2-3 keys before getting the character. Same in Japanese where はい/hai requires 3 keystrokes (and no one in Japan would associate "h" or "ha" to the idea of approval). Mismatch char/keys breaks the UX for some "popular" languages.

Another issue is default format. Language without notion of capitalisation do not match the `y/N` pattern. Many CLIs opt for the format `y/n [n]`, with the convention that square brakets mean optional entries, with default. Stashed an implementation here:

```python
# src/click/termui.py
# in the confirm function
def confirm(...):
    yes_mark = _("y")
    no_mark = _("n")
    def_mark = (
        "" if default is None else (f" ({yes_mark})" if default else f" ({no_mark})")
    )

    prompt = _build_prompt(
        text, prompt_suffix, show_default, f"{yes_mark}/{no_mark}{def_mark}"
    )

    while True:
        try:
            value = _readline_prompt(visible_prompt_func, prompt, err).lower().strip()
        except (KeyboardInterrupt, EOFError):
            raise Abort() from None
        if value in ("y", "yes", _("y"), _("yes")):   # <= here accepting y and translations
            rv = True
        elif value in ("n", "no", _("n"), _("no")):   # <= here accepting n and translations
            rv = False
...
```

This would lead to output like (chose round brackets for readability):

```
Prompt to stdin with no suffix [y/n (n)]

stdin へのプロンプト(サフィックスなし) [はい/いいえ (いいえ)]

标准输入提示(无后缀) [是/否 (否)]
```

As much as I'd like full localisation, this looks regular enough but not great. If Click users are essentially developers, y/n really looks the way to go. On option for the rare users who want customisation is best, though, perhaps out of scope of this PR (stashed the above, but need a mechanism to switch modes).

ic:
@kdeldycke Conclusion on the geeking out here: I propose to rollback the y/n commit (0b75b01ec37ea591d76a51d50604589c69163a43) for now, and freeze the PR scope.

The thinking:
* y/n translated is often not expected, or not working for non-alphabet languages (1 char is not 1 keystroke in many of them).
* Regular localisation would require really visible format change from `y/N` to `y/n [n]` to apply to most language.
* y/n can be future change, assuming Click users are essentially developers.

kdeldycke:
> I propose to rollback the y/n commit ([0b75b01](https://github.com/pallets/click/commit/0b75b01ec37ea591d76a51d50604589c69163a43)) for now, and freeze the PR scope.

Oh yes. That's full of edge-cases. Let's keep things simple for now. You can revert that commit, and I will merge after that. Thanks a lot for your deep tests!

ic:
> Oh yes. That's full of edge-cases. Let's keep things simple for now. You can revert that commit, and I will merge after that. Thanks a lot for your deep tests!

Rollback completed.

kdeldycke:
Thanks @ic for your patience and for your work! Merged upstream and will be part of Click 8.4.

Issue #2894, opened 2025-05-11
Click 8.2.0 ignores is_flag options that have a type


andy-maier:
Click 8.2.0 ignores the use of options defined with `is_flag` and `type=str`.

Of course, combining `is_flag` and `type=str` does not make sense, but it happened in one of our projects. Click before 8.2.0 ignored `type=str` and processed this as expected for `is_flag`.

Click 8.2.0 does not set a value for such an option.

I could trace this change to PR #2829 - If I install click 8.2.0 and manually undo the change in `core.py` that is done by that PR, it works as before.

To reproduce:

1. Have a file `main.py` with:

```
import click

@click.group(invoke_without_command=True)
@click.option('-x', '--transpose', type=str, is_flag=True,
              help='Transpose the output table.')
def cli(transpose):
    print(f"Debug: transpose={transpose}")
    return 0
```

2. Reproduce with click 8.2.0:
```
$ pip install click==8.2.0
$ python -c "import main,sys; sys.argv=['cli', '--transpose']; main.cli()"
Debug: transpose=None
```

3. Verify that click 8.2.0 with PR #2829 undone works:
```
# perform the undo in core.py (see below)
$ python -c "import main,sys; sys.argv=['cli', '--transpose']; main.cli()"
Debug: transpose=True
```

4. Verify that click 8.1.8 works:
```
$ pip install click==8.1.8
$ python -c "import main,sys; sys.argv=['cli', '--transpose']; main.cli()"
Debug: transpose=True
```

The "undo" of PR #2829 has the following changes in `core.py`, starting at line 2613:
```
        # Undo PR 2829: Added the following two lines
        if flag_value is None:
            flag_value = not self.default
        self.type: types.ParamType
        if is_flag and type is None:
            # Undo PR 2829: Removed the following two lines
            # if flag_value is None:
            #     flag_value = not self.default
            # Re-guess the type from the flag value instead of the
            # default.
            self.type = types.convert_type(None, flag_value)
```

Environment:

- Python version: 3.12.7 on macOS (it happens on all Python versions supported by click 8.2.0, but not on all OSs)
- Click version: 8.2.0

Based on our CI test run, the issue happens only on macOS and Windows, but not on Ubuntu: 
https://github.com/zhmcclient/zhmccli/actions/runs/14949544297. I did not investigate why it does not happen on Ubuntu.

The fix for projects using click is of course to remove `type` when `is_flag` is used.

[...]

PR #2912, opened 2025-05-13
Fixes flag default handling for boolean options

### LINKED ISSUE #2897: Boolean flags are always False
When upgrading from Click 8.1.8 to 8.2.0, I encountered a bug where boolean flags are not parsed correctly.

## Reproduction
The below program can reproduce the error 
```python
from sys import exit
import click

@click.command(help="help")
@click.option("-v", "--version", type=click.BOOL, default=False, is_flag=True)
@click.option("-h", "--help", type=click.BOOL, default=False, is_flag=True)
@click.option("-n", "--name", type=click.STRING, default="John Doe")
def main(help, version, name):
    print(f"{help=}, {version=}, {name=}")
    if help:
        print("Help")
        exit(0)
    if version:
        print("v1.0.0")
        exit(0)
    print(f"Hello, {name}")

if __name__ == "__main__":
    main()
```
- When running with click 8.1.8, all flags/options are parsed as expected
- On 8.2.0, boolean flags are not parsed and remain at their default value
  - e.g. `python main.py --help --name Robert` will print `help=False, version=False, name='Robert'`

Environment:
- OS: NixOS 24.11
- Python version: 3.12.8
- Click version: 8.2.0

I'm unsure if this is a regression in Click, or an issue on my end. Any help would be greatly appreciated. TIA!

john0isaac:
Ensures boolean flag options correctly determine their default values when not explicitly set.
Adds tests to validate proper behavior for various flag configurations, improving reliability.

Fixes #2897


src/click/core.py
```diff
@@ -2618,6 +2618,12 @@ def __init__(
             # default.
             self.type = types.convert_type(None, flag_value)
 
+        if is_flag and isinstance(self.type, types.BoolParamType):
+            # If the type is a boolean, we need to set the flag_value
+            # to True or False depending on the default.
+            if flag_value is None:
+                flag_value = not self.default
+
```
john0isaac, line 2626: this code was nested in https://github.com/pallets/click/pull/2829 causing the current error.

Click was executing it in case any of the previous conditions were not met but after nesting it, it was no longer reachable.

To not break the previous change I added `isinstance(self.type, types.BoolParamType)` so we preserve the introduced behaviour in `2829`

tests/test_defaults.py
```diff
@@ -84,3 +84,43 @@ def foo(name):
...
+    @click.option("--baz", is_flag=True, default=True, type=click.BOOL)
+    @click.option("--qux", is_flag=True, default=True, type=bool)
+    def cli(foo, bar, baz, qux):
+        assert isinstance(foo, bool)
+        assert isinstance(bar, bool)
+        assert isinstance(baz, bool)
+        assert isinstance(qux, bool)
+        click.echo(f"foo: {foo}, bar: {bar}, baz: {baz}, qux: {qux}")
+
+    result = runner.invoke(cli, [])
```
john0isaac, line 104: Not sure if this is the correct place for this test but I think it's required as this case was not covered introducing this breaking change.

src/click/core.py
john0isaac: If you do not use any type for click flags, click will automatically detect the flag and everything will work as it was before the breaking change working

john0isaac:
Although it's the same issue this won't fix https://github.com/pallets/click/issues/2894, the fix would be to revert back the changed lines completely but it will break the introduced behaviour in https://github.com/pallets/click/pull/2829

PR #2921, opened 2025-05-16
Fix documenting-arguments anchor in docs

meator:
Judging from other places in the documentation which link to this anchor, this anchor is misplaced:

https://click.palletsprojects.com/en/stable/parameters/:

image

https://click.palletsprojects.com/en/stable/arguments/:

image

I don't think that https://click.palletsprojects.com/en/stable/documentation/#command-short-help is relevant in both of these. https://click.palletsprojects.com/en/stable/documentation/#id1 seems to be a better fit.

### DIFF
--- a/docs/documentation.rst
+++ b/docs/documentation.rst
@@ -30,8 +30,6 @@ Simple example:
     invoke(hello, args=['--help'])
 
 
-.. _documenting-arguments:
-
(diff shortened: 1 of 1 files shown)


Rowlando13:
@meator Thanks for the PR!

PR #2929, opened 2025-05-19
Add a doc for supporting 8.1.x|8.2.0

sirosen:
This uses `get_metavar` as an example, demonstrates how to feature-detect based on
`kwargs` containing `"ctx"`, and shows simplified usage.
A note block indicates how this same strategy applies to other ``ParamType`` methods.

Section naming is setup to support:
- future documentation of other version transitions
- documentation of other transitional methods for 8.1 vs 8.2

---

There are lots of details here on which I have no particularly strong opinion (e.g., precise doc name or section names), and am happy to work through feedback.


davidism:
I converted this to Markdown since we now have myst-parser available. Then did some further editing. Added a link to our version policy, and a general description of feature detection.

PR #2930, opened 2025-05-19
Fix flag value set when is_flag=True and type is provided

### LINKED ISSUE #2894: Click 8.2.0 ignores is_flag options that have a type

### LINKED ISSUE #2897: Boolean flags are always False

kieranyyu:
When setting the value of a flag, check only if `is_flag` is `True` and `flag_value` is not provided when negating the default flag value.

#2829 introduced a bug by checking if the type of the flag is `None` before setting the value.
Since we still check if `is_flag` is set, objects without a `__bool__` method will not raise an error if `is_flag` is not set.

fixes #2894
fixes #2897

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,8 @@
 Version 8.2.1
 -------------
 
+-   Fix flag value handling for flag options with a provided type. :issue:`2894`
+    :issue:`2897` :pr:`2930`
 
 Version 8.2.0
 -------------
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2610,10 +2610,11 @@ class Option(Parameter):
             else:
                 self.default = False
 
+        if is_flag and flag_value is None:
+            flag_value = not self.default
+
         self.type: types.ParamType
         if is_flag and type is None:
-            if flag_value is None:
-                flag_value = not self.default
             # Re-guess the type from the flag value instead of the
             # default.
             self.type = types.convert_type(None, flag_value)
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -1008,6 +1008,31 @@ def test_type_from_flag_value():
 
 
 @pytest.mark.parametrize(
+    ("opts", "pass_flag", "expected"),
+    [
+        pytest.param({"type": bool}, False, "False"),
+        pytest.param({"type": bool}, True, "True"),
+        pytest.param({"type": bool, "default": True}, False, "True"),
+        pytest.param({"type": bool, "default": True}, True, "False"),
+        pytest.param({"type": click.BOOL}, False, "False"),
+        pytest.param({"type": click.BOOL}, True, "True"),
+        pytest.param({"type": click.BOOL, "default": True}, False, "True"),
+        pytest.param({"type": click.BOOL, "default": True}, True, "False"),
+        pytest.param({"type": str}, False, ""),
+        pytest.param({"type": str}, True, "True"),
+    ],
+)
+def test_flag_value_is_correctly_set(runner, opts, pass_flag, expected):
+    @click.command()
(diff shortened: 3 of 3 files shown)


docs/.DS_Store
dbaxa, line 1: Please remove this from your commit.

kieranyyu: Done

PR #2933, opened 2025-05-20
Flush `sys.stderr` when `CliRunner` finalizes

### LINKED ISSUE #2682: Revisiting stderr flush in CLIRunner
This bug is described in detail at https://github.com/pallets/click/issues/2647 and https://github.com/pallets/click/issues/2636.  I've read through those issues and am not convinced the current behavior is accurate or beneficial.

From: https://github.com/pallets/click/issues/2636#issuecomment-1826891936

> This isn't a bug. When using print in general, you have to flush to be sure all output is sent to stderr. That's one of the reasons Click provides echo instead.

From: https://github.com/pallets/click/issues/2647#issuecomment-1846440749

> Another way to put it is: you need to explain why your suggested behavior is correct. It's certainly different than current, but that doesn't mean it's more correct.

> What I know is that stdout is flushed, and stderr isn't. I don't know why that's the case, but I do know that this is the behavior I've encountered all over the place, and if I want stderr to show up at a specific time I need to flush manually. This is what click.echo does automatically. If what I know and click's test runner's current behavior is incorrect, then you need to show me sources that explain why. Otherwise, it would just mean we're hiding the actual behavior of the calls you're making.

I believe the current behavior of not flushing stderr is wrong because it results in semantics during testing that aren't present during a normal run of the command.  Take:

```python
import sys

import click


@click.command()
def cli():
    print("to stdout")
    print("to stderr", file=sys.stderr)


def test_cli():
    runner = CliRunner(mix_stderr=False)
    result = runner.invoke(cli, (), catch_exceptions=False)

    assert result.stdout.strip() == "to stdout"
    assert result.stderr.strip() == "to stderr"


if __name__ == "__main__":
    cli()
```

If you run this through a normal Python execution, you will __always__ get "to stdout" and "to stderr" on the console.  If there were times when "to stderr" were not printed to the console, it would be a bug in Python.  If I don't care when the output comes through, only that it shows up by the time the process has exited, then I don't need to take any extra steps.

However, if I run the test, its possible the second assert fails even though the first assert will always pass.  Additionally, I'd have to add a `sys.stderr.flush()` at the end of cli() but only to get the tests to pass.  It would never be needed when running the program in a normal Python process.  Furthermore, its basically impossible to add the flush only during testing because of where the flush() is needed in the current code.

From: https://github.com/pallets/click/issues/2647#issuecomment-1846430069

> You're welcome to use print to stderr, you just have to remember to flush in that case. Stdout is flushed, stderr isn't. That's true regardless of if you're using click.

> If we flush for you during tests, that's hiding the bug, not fixing it.

As I demonstrated in the code above, my program works just fine when ran normally.  There is no bug.  The tests only fail because invoke() is not matching the expected semantics of a normal execution of a click command in a Python process.

It would only be flushing at the end of an invoke, which matches what Python does at the end of program execution.  Therefore, I'm not sure I grok the resistance to adding this flush.  Its not a performance issue.  Its not a correctness issue.  Its not a backwards compatibility issue.  There are three issues now raised because of the unexpected behavior and countless other people have likely ran into it that didn't comment or raise a new issue.  So why not add it?  What's the downside?

sirosen:
Add a dedicated finalizer for CliRunner streams to collect up these values, which makes the scope of the `finally` block easy to read and reason about.

closes #2682

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@ Version 8.2.1
 -   Fix flag value handling for flag options with a provided type. :issue:`2894`
     :issue:`2897` :pr:`2930`
 -   Fix shell completion for nested groups. :issue:`2906` :pr:`2907`
+-   Flush ``sys.stderr`` at the end of ``CliRunner.invoke``. :issue:`2682`
 
 Version 8.2.0
 -------------
--- a/src/click/testing.py
+++ b/src/click/testing.py
@@ -510,6 +510,7 @@ class CliRunner:
                 exc_info = sys.exc_info()
             finally:
                 sys.stdout.flush()
+                sys.stderr.flush()
                 stdout = outstreams[0].getvalue()
                 stderr = outstreams[1].getvalue()
                 output = outstreams[2].getvalue()
--- a/tests/test_testing.py
+++ b/tests/test_testing.py
@@ -450,3 +450,24 @@ def test_isolation_stderr_errors():
         click.echo("\udce2", err=True, nl=False)
 
     assert err.getvalue() == b"\\udce2"
+
+
+def test_isolation_flushes_unflushed_stderr():
+    """An un-flushed write to stderr, as with `print(..., file=sys.stderr)`, will end up
+    flushed by the runner at end of invocation.
+    """
+    runner = CliRunner()
+
+    with runner.isolation() as (_, err, _):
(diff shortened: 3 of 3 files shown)


davidism:
After some review, I think it's better to leave it in the `finally` block for now, so I've updated the PR. Rebased to stable and added a change entry.

PR #2938, opened 2025-05-21
Fixed the description of `is_flag=False, flag_value=value`

BobKerns:
The text had "only if", but the example shows that the behavior "if only" the flag is given, is to return the `flag_value`.

This really confused me the first several times I read it, until I carefully studied the example.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -432,8 +432,8 @@ providing only the option's flag without a value will either show a
 prompt or use its `flag_value`.
 
 Setting `is_flag=False, flag_value=value` tells Click that the option
-can still be passed a value, but only if the flag is given the
-`flag_value`.
+can still be passed a value, but if only the flag is given, the
+value will be `flag_value`.
 
 ```{eval-rst}
 .. click:example::


davidism:
This says something significantly different now. Before I merge this, can you identify existing tests that demonstrate the behavior is what you've changed the docs to say? And perhaps add additional tests if they don't seem comprehensive? Click's various flag arguments combine in weird ways, so I want to be certain.

BobKerns:
I understand your caution; it does change the meaning of the text. It changes it to match the example and the observed behavior. Clearly one must change, so which is a reasonable question.

The example seems to be the correct and intended behavior; the text doesn't really make sense as a behavior.

The matching test in the suite is `test_option_with_optional_value`, which seems comprehensive enough. It covers supplying supplying no options, just the flag w/o a value, with a value, and those cases in combination with other flags. For example, it distinguishes between `--opt -a 42` and `--opt some-value -a 42`, where `-a` is an option, so if it appears immediately after, `--opt` was not given a value, but in the second case, `some-value` would be the value.

It looks to me like a simple mangled edit. The actual behavior is pretty clear and useful,

Rowlando13:
This correct. Thanks for the fix.

PR #2940, opened 2025-05-22
testing/CliRunner: Fix regression related to EOF introduced in 262bdf0

### LINKED ISSUE #2939: Click 8.2.1 CliRunner raises SystemExit(1) when using click.File's stdin support
In Click 8.2.1, when testing the behavior of a command-line which uses `click.File`'s stdin as an argument, `CliRunner.invoke()` always throws a `SystemExit(1)` with `Aborted !` on stderr.

Seems like a regression introduced by the change in #2934, as this `Aborted!` is generated when the main Click endpoint raises an `EofError`.

Reproducer :

```python
#!/usr/bin/env python3

import click


@click.command()
@click.argument("file_input", type=click.File("r"))
def cli(file_input):
    for line in file_input.readlines():
        print(line.rstrip())


def test_cli():
    result = CliRunner().invoke(cli, ["-"], input="test\n")
    assert result.exit_code == 0, str(result)
    assert result.output == "test\n"


if __name__ == "__main__":
    cli()
```

``` shell
nicolasd@pythagore:~/tmp/repro$ uv venv -p python3.13 venv
Using CPython 3.13.3 interpreter at: /usr/bin/python3
Creating virtual environment at: venv
Activate with: source venv/bin/activate
nicolasd@pythagore:~/tmp/repro$ . ./venv/bin/activate
(venv) nicolasd@pythagore:~/tmp/repro$ uv pip install pytest Click==8.2.1
Resolved 5 packages in 5ms
Installed 5 packages in 6ms
 + click==8.2.1
 + iniconfig==2.1.0
 + packaging==25.0
 + pluggy==1.6.0
 + pytest==8.3.5
(venv) nicolasd@pythagore:~/tmp/repro$ echo foo | python3 test_repro.py -
foo
(venv) nicolasd@pythagore:~/tmp/repro$ pytest test_repro.py 
===================================================================== test session starts =====================================================================
platform linux -- Python 3.13.3, pytest-8.3.5, pluggy-1.6.0
rootdir: /home/nicolasd/tmp/repro
collected 1 item                                                                                                                                              

test_repro.py F                                                                                                                                         [100%]

========================================================================== FAILURES ===========================================================================
__________________________________________________________________________ test_cli ___________________________________________________________________________

    def test_cli():
        result = CliRunner().invoke(cli, ["-"], input="test\n")
>       assert result.exit_code == 0, str(result)
E       AssertionError: <Result SystemExit(1)>
E       assert 1 == 0
E        +  where 1 = <Result SystemExit(1)>.exit_code

test_repro.py:16: AssertionError
=================================================================== short test summary info ===================================================================
FAILED test_repro.py::test_cli - AssertionError: <Result SystemExit(1)>
====================================================================== 1 failed in 0.09s ======================================================================
(venv) nicolasd@pythagore:~/tmp/repro$ uv pip install Click==8.2.0
Resolved 1 package in 3ms
Uninstalled 1 package in 1ms
Installed 1 package in 1ms
 - click==8.2.1
 + click==8.2.0
(venv) nicolasd@pythagore:~/tmp/repro$ pytest test_repro.py 
===================================================================== test session starts =====================================================================
platform linux -- Python 3.13.3, pytest-8.3.5, pluggy-1.6.0
rootdir: /home/nicolasd/tmp/repro
collected 1 item                                                                                                                                              

test_repro.py .                                                                                                                                         [100%]

====================================================================== 1 passed in 0.09s ======================================================================
```

Environment:

- Python version: 3.13
- Click version: 8.2.1

anlambert:
Commit 262bdf0 ensured to raise an `EOFError` exception on end of input to simulate tty behavior and avoid blocking prompt during tests when no more input is available.

However the introduced implementation has a side effect when testing a click command having a `File` type option or argument and when it is set to `stdin`: the command ends up with error due to the `Abort` exception being raised when the stdin `EOFError` exception is caught.

To prevent this undesirable side effect, prefer to raise the `EOFError` exceptions directly from the prompts functions inside the `CliRunner` class instead of doing it in the method overriding the iterator protcol for the` _NamedTextIOWrapper` class.

Restore previous implementation of a test broken by changes of 262bdf0.

Fixes #2939.


src/click/testing.py
```diff
@@ -348,7 +345,9 @@ def isolation(
         @_pause_echo(echo_input)  # type: ignore
         def visible_input(prompt: str | None = None) -> str:
             sys.stdout.write(prompt or "")
-            val = next(text_input).rstrip("\r\n")
+            if text_input_size and text_input.tell() == text_input_size:
+                raise EOFError()
+            val = text_input.readline().rstrip("\r\n")
```
kieranyyu, line 350: I would suggest simply moving the logic in `_NamedTextIOWrapper.__next__` to here:
```
try:
    val = next(text_input).rstrip("\r\n")
except StopIteration as e:
    raise EOFError() from e
```

This will do the same without having to check the input size. When I changed the relevant test, I thought that raising `Abort` at the end of stdin made sense since this is the behavior if the user uses Ctrl+C to end input but it seems that input can be ended cleanly with Ctrl+D.

anlambert: Simpler indeed, I have updated that PR accordingly.

kdeldycke:
Sorry @Rowlando13 for the late answer following your review request. But good call for merging this. I could not find a regression with the actual `main` branch .

CHANGES.rst
```diff
@@ -24,6 +24,7 @@ Unreleased
 -   Lazily import ``shutil``. :pr:`3023`
 -   Properly forward exception information to resources registered with
     ``click.core.Context.with_resource()``. :issue:`2447` :pr:`3058`
+-   Fix regression related to EOF handling in CliRunner. :issue:`2939`:pr:`2940`
```
webknjaz, line 27: Lacking whitespace in between two refs ^

webknjaz: #3078

PR #2944, opened 2025-05-25
Fix _pipepager()/_tempfilepage() to work with multi-call binaries

### LINKED ISSUE #2943: _pipepager() fails with mutli-call binaries
Programs such as BusyBox, Toybox and Coreutils (also gzib, bzip etc) in multi-call mode derive their identity from the symlink. Resolving the symlink causes them to misbehave.

Environment:

- Python version: python3.12
- Click version: 8.2.1

ghost:
Programs such as BusyBox, Toybox and Coreutils (also gzib, bzip etc) in multi-call mode derive their identity from the symlink. Resolving the symlink causes them to misbehave.

fixes #2943

### DIFF
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -429,8 +429,11 @@ def _pipepager(
     cmd_filepath = shutil.which(cmd)
     if not cmd_filepath:
         return False
-    # Resolves symlinks and produces a normalized absolute path string.
-    cmd_path = Path(cmd_filepath).resolve()
+
+    # Produces a normalized absolute path string.
+    # multi-call binaries such as busybox derive their identity from the symlink
+    # less -> busybox. resolve() causes them to misbehave. (eg. less becomes busybox)
+    cmd_path = Path(cmd_filepath).absolute()
     cmd_name = cmd_path.name
 
     import subprocess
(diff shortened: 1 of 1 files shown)


stefreak:
Makes sense, this looks good to me 👍

Rowlando13:
This looks like this will work but it's not obvious that calling `.resolve()` will break things for BusyBox, etc. Please add a comment and a test.

stefreak:
@Rowlando13 How can we ensure the presence of Busybox for the test?

Rowlando13:
The issue says that it is from resolving symlinks and that multiple utilities exhibit the behavior. I would imagine one of those utilities would installed on   A GitHub runner, but I don't know that they are. I was going to leave it up to the person who submitted the issue.

ghost:
Would be simulating a multicall-binary in shell or as a python script good enough? That is what I planned for the test. Multicall-binaries read argv[0] and we can do that in shell or python. 

But of course running the whole test-suite against busybox would prevent any kind busybox based bugs in the future. I‘ll look into that too.

Rowlando13:
I think in Python would be best. I would guess it is hard to guarantee everything that a shell gives you from inside pytest since it does a lot of funny things, but I don't know for sure. @davidism might know. I am pretty sure we don't want to have busybox as part of our test suite, since it would make things slower and I don't think any of the core contributors have experience with it.

davidism:
It's fine without a test. Please make the comment very specific about why we use `absolute` and not `resolve`, so that we don't change it back later.

ghost:
> It's fine without a test. Please make the comment very specific about why we use `absolute` and not `resolve`, so that we don't change it back later.

I updated this PR, but I got curious on how difficult it is to add busybox tests. It seems quite straight forward so I added this proposal: #3101

davidism:
Comment looks good. I'm not going to add the busybox tests, but it's a good reference, thanks for looking into it.

PR #2946, opened 2025-05-25
Expect alternate errors in test_file_surrogates()

### LINKED ISSUE #2945: test_file_surrogates() covers not all file-system types
Not all filesystems react with FileNotFoundError to surrogates. Inside podman the filesystem with `OSError: [Errno 84] Illegal byte sequence: '\udcff'`.

On zfs:

```shell
» python3
Python 3.12.10 (main, Apr  9 2025, 08:28:36) [GCC 14.2.0] on linux
>>> from pathlib import Path
>>> a = Path("\udcff")
>>> a.open("r")
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.12/pathlib.py", line 1013, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '\udcff'
>>>
```

In podman overlayfs + zfs:

```shell
~/ephemeral/click $ python
Python 3.12.10 (main, May 21 2025, 16:23:36) [GCC 14.2.0] on linux
>>> from pathlib import Path
>>> a = Path("\udcff")
>>> a.open("r")
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.12/pathlib.py", line 1013, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [Errno 84] Illegal byte sequence: '\udcff'
````

Environment:

- Python version: 3.12
- Click version: 8.2.1

ghost:
Not all file systems trigger a FileNotFoundError when encountering surrogates. Inside Podman the file system triggers an `OSError: [Errno 84] Illegal byte sequence: '\udcff'`.

fixes #2945

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -226,7 +226,12 @@ def test_path_surrogates(tmp_path, monkeypatch):
 def test_file_surrogates(type, tmp_path):
     path = tmp_path / "\udcff"
 
-    with pytest.raises(click.BadParameter, match="�': No such file or directory"):
+    # - common case: �': No such file or directory
+    # - special case: Illegal byte sequence
+    # The spacial case is seen with rootless Podman. The root cause is most
+    # likely that the path is handled by a user-space program (FUSE).
+    match = r"(�': No such file or directory|Illegal byte sequence)"
+    with pytest.raises(click.BadParameter, match=match):
         type.convert(path, None, None)
 
 


rhizoome:
> Thanks for the PR! Can use please add a comment that specifies the FileNotFoundError as the common case and the OsError as a special case for podman.

I tried a few filesystems that supposedly should not handle surrogates, but all of them work. Even overlayfs (which is used in podman), so the only explanation seems to be, that in rootless podman it passes through a user-space program (FUSE). Let me know if that comment is now too detailed.

Rowlando13:
Looks good.

PR #2947, opened 2025-05-26
Convert docs from reST to MyST

### LINKED ISSUE #2886: Move current docs to myst
Move any sections of the docs to myst using rst-myst command line tool.

denuoweb:
This PR switches the documentation from reST to MyST.


davidism:
Thanks for putting in time to work on this. We need to clean it up so this PR focuses only on the syntax change and nothing else. That will make it easier to review.

- [ ] First two commits need to be combined, `git blame` is more likely to treat those as renames, or to detect changes across files, if they're in the same commit.
- [ ] Don't change any writing, same reason as above.
- [ ] Don't use autodoc2, I still prefer the results from autodoc.
- [ ] Don't use linkify extension, all links should be using syntax rather than detection.
- [ ] Don't bring in new extensions such as sphinx-design.
- [ ] Bring back the two extensions that were removed.

PR #2949, opened 2025-05-26
Move Why doc from reST to MyST

denuoweb:
This PR switches the Why documentation from reST to MyST.

Assists with https://github.com/pallets/click/issues/2886


Rowlando13:
These PRs should target stable. @denuoweb. I made the change for you.

PR #2950, opened 2025-05-27
Alter readthedocs to prep for building to build html, htmlzip, and pdf docs. 

Rowlando13:


davidism:
Please revert this. I deliberately do not do this.

Issue #2952, opened 2025-05-27
Flag value is always true when setting any envvar


alfechner:
Click 8.2.1 includes fixes related to boolean flags.

We’re encountering a specific issue described in [this comment](https://github.com/pallets/click/issues/2897#issuecomment-2873272069). The comment appears in the thread for a different but related issue ([#2897](https://github.com/pallets/click/issues/2897)), focused on a separate aspect of boolean flag handling. Maybe the issue was overlooked because the discussion centered on the main issue.

alfechner:
I'm using the `envvar` option and I set the `envvar` to `False`, however it's always `true` for me.

```python
# hi.py
import click

@click.command()
@click.option("--dry-run/--no-dry-run", default=False, envvar="DRY_RUN")
def main(dry_run: bool):
    print(type(dry_run))
    print(f"Dry run: {dry_run}")

if __name__ == "__main__":
    main()
```

```python
export DRY_RUN=false

python hi.py
```
Output
```python
<class 'bool'>
Dry run: True
```

davidism:
@kieranyyu you've recently become the flag value expert, any chance you could take a look at this one? No worries if not!

It looks like `flag_value` is `True` and that's triggering incorrect behavior in `resolve_envvar_value` during `handle_parse_result`. Not sure what it was before the recent change, or what it's "supposed" to be.

kdeldycke:
I can confirm the issue as a regression introduced by Click 8.2.1.

I developed an [extensive suite of unittests in Click Extra](https://github.com/kdeldycke/click-extra/blob/main/tests/test_envvar.py) to cover edge cases of envvar handling in Click. Essentially [around case-sensitivity and boolean values](https://github.com/kdeldycke/click-extra/blob/9b909adb2cb3b9a1f3c0c8ca2cf54d3d51bf96d0/tests/test_envvar.py#L106-L182).

I can try to port them for Click 8.2.2.

kdeldycke:
I just realized there a lot of uncovered edge-cases with flag options, so I'm attempting to fix all of these in https://github.com/pallets/click/pull/2956

alfechner:
@kieranyyu thanks! I don't know how "far" you want to go but how about considering to drop support for this kind of options: `--dry-run/--no-dry-run`? I really have a hard time to see the value. As I said before, the negative case will never be set IMHO.

davidism:
We're not dropping support for positive/negative flags. There are many use cases including:

* Defaults can be dynamic.
[...]

PR #2956, opened 2025-05-28
Fix reconciliation of envvar with `default`, `flag_value` and `type` parameters for flag options

kdeldycke:
While digging into internals of flag option management in the context of fixing #2952, I stumbled upon lots of edge-cases.

This is a WIP to identify and fix all these issues, essentially because the absence of value is mixed-up with the value of absence.

This PR:
- Fix reconciliation of `default`, `flag_value` and `type` parameters for flag options.
- Fix parsing of flag's environment variables.
- Tests all combinations of these parameters.

Checklist:

Closes:
- #2952

Fix edge-cases introduced in:
- #2788
- #2746

Refs:
- #2930
- #2788
- #2746


davidism:
We should also thoroughly document all the interactions once we settle on something.

kdeldycke:
> > I stumbled upon lots of edge-cases.
> 
> An extreme understatement! 😆 If we can once and for all consolidate how these arguments all affect each other, that would be amazing.

Ahah! 😄 If I can only increase the coverage and fix a couple of issues, I'll be happy. Refactoring will be easier after some incremental improvements.

This PR is still a WIP. I'll work on it progressively in the next few days and try to push it as far as I can without breaking everything! 😅

src/click/types.py
```diff
@@ -663,21 +664,38 @@ class BoolParamType(ParamType):
...
+        if isinstance(value, bool):
+            return value
 
-        if norm in {"0", "false", "f", "no", "n", "off"}:
-            return False
+        # import pdb
+
+        # pdb.set_trace()
+
+        bool_states: dict[str, bool] = RawConfigParser.BOOLEAN_STATES | {
```
kieranyyu: It seems that unpacking `RawConfigParser.BOOLEAN_STATES` with `**` or using `t.cast()` could fix the typing issue

kieranyyu (commented):
Test cases look very comprehensive!

src/click/types.py
kdeldycke: Ah yes, that's good alternatives. I'm not opinionated about which is the preferred way here.

kdeldycke:
> Test cases look very comprehensive!

But I'm not finished yet! Still have some edge-cases to tests! 😅

kdeldycke:
> @kdeldycke This is excellent! I struggled with just reconciling a few edgecases when I was fleshing out the docs.

I'll probably add a couple of lines in the docs once I've ironed out this PR. At least to explain the rules by which to expect the different arguments to behave.

kdeldycke:
This PR is ready to be reviewed and merged: all tests are passing! :)

src/click/core.py
```diff
@@ -2554,7 +2554,6 @@ def __init__(
         if help:
             help = inspect.cleandoc(help)
 
-        default_is_missing = "default" not in attrs
```
davidism, line 2557: Thank goodness. If we can actually get rid of this line, it removes one of the barriers I kept running into when consolidating signatures and improving typing.

kdeldycke: Ahah happy to help! 🤗

Rowlando13:
@davidism did you review all of this or just leave a comment? If so I am good with merging it. I assume this would go in Click 8.3.0?

Rowlando13:
@kieranyyu would you mind reviewing this pr?

src/click/core.py
```diff
@@ -2588,47 +2587,51 @@ def __init__(
...
 
-        if is_flag and default_is_missing and not self.required:
-            if multiple:
-                self.default = ()
-            else:
-                self.default = False
+        if is_flag:
+            # Set missing default for flags if not explicitly required or prompted.
+            if self.default is None and not self.required and not self.prompt:
+                self.default = () if multiple else False
```
davidism: I try to avoid ternary expressions, since coverage can't currently measure both branches. I know we don't test with coverage by default, but it is available. Especially in situations like this, knowing what branches were tested seems relevant.

src/click/core.py
```diff
@@ -2588,47 +2587,51 @@ def __init__(
...
+            # Not a flag, but when used as a flag it shows a prompt.
             elif self._flag_needs_value:
-                # Not a flag, but when used as a flag it shows a prompt.
                 is_flag = False
-            else:
-                # Implicitly a flag because flag options were given.
-                is_flag = bool(self.secondary_opts)
+            # Implicitly a flag because flag options were given.
+            elif self.secondary_opts:
+                is_flag = True
```
davidism, line 2598: What's the reason for the change here? It seems like the comments were moved, and `is_flag` can remain `None`, but that doesn't actually seem to matter?

src/click/core.py
```diff
@@ -2588,47 +2587,51 @@ def __init__(
...
         self.default: t.Any | t.Callable[[], t.Any]
 
-        if is_flag and default_is_missing and not self.required:
-            if multiple:
-                self.default = ()
-            else:
-                self.default = False
+        if is_flag:
+            # Set missing default for flags if not explicitly required or prompted.
+            if self.default is None and not self.required and not self.prompt:
```
davidism, line 2608: I just want to be absolutely sure that `self.default is None` results in the same behavior as the previous `default_is_missing` kwargs check. I'm assuming some of the tests deal with this, but it's hard to pick out which ones are relevant from the huge list.

src/click/types.py
```diff
@@ -6,6 +6,7 @@
 import stat
 import sys
 import typing as t
+from configparser import RawConfigParser
```
davidism: Don't pull in `configparser` for this, it implies a level of support/parity that I don't think is intended.

src/click/types.py
```diff
@@ -661,23 +662,41 @@ def _clamp(self, bound: float, dir: t.Literal[1, -1], open: bool) -> float:
...
     name = "boolean"
 
+    bool_states: dict[str, bool] = dict(**RawConfigParser.BOOLEAN_STATES) | {
+        "t": True,
+        "f": False,
+        "y": True,
+        "n": False,
+        # Absence of value is considered False.
+        "": False,
+        # XXX Possible states to add in the future:
```
davidism: Let's not suggest other values, because that's an endless bikeshed. Having a dict here that users can override with their own values is good enough. Although I don't think we should document that as a public API (yet).

src/click/core.py
kdeldycke: Removing that condition will not let secondary flags to be defined with a minimal set of parameters, as they'll be considered as non-boolean types, making these two test fails:
```pytb
tests/test_info_dict.py:187: in <module>
    click.Option(["--cache/--no-cache", "-c/-u"]),
src/click/core.py:2657: in __init__
    raise TypeError("Secondary flag is not valid for non-boolean flag.")
E   TypeError: Secondary flag is not valid for non-boolean flag.
```
```pytb
tests/test_options.py:1267: in <module>
    pytest.param(Option(["-a/-A"]), True, id="secondary option [implicit flag]"),
src/click/core.py:2657: in __init__
    raise TypeError("Secondary flag is not valid for non-boolean flag.")
E   TypeError: Secondary flag is not valid for non-boolean flag.
```

Moving the comment is to align it with other comments layout in the conditions above for readability.

src/click/types.py
kdeldycke: OK let's remove that dependency then.

src/click/types.py
kdeldycke: Agreed!

kdeldycke: Fixed in https://github.com/pallets/click/pull/2956/commits/2f794921bdb8e65b095e6b000a7955822be38a64

src/click/types.py
kdeldycke: Fixed in https://github.com/pallets/click/pull/2956/commits/2f794921bdb8e65b095e6b000a7955822be38a64

src/click/core.py
kdeldycke: Fixed in https://github.com/pallets/click/pull/2956/commits/94347130897a4c9807af08f0aeb1986152f916e0

src/click/core.py
kdeldycke: I understand, the matrix of cases is a bit hard to read.

But I preferred to flat it out instead of resorting to a set of `@pytest.mark.parametrize` variations, to keep the `test_flag_value_and_default()` test as straightforward as possible. There's a compromise here: either we keep the list of variation complex, or the body of the test complex. I choose to keep the body of the test simple, as making it complex tends to mirror the very same logic we are trying to test from the core implementation.

Anyway, the situation is thouroughly tested. See how the basic parameter `{"type": bool}` (i.e. the legacy `default_is_missing` case), is then complemented with all the other possible variations:
- `{"type": bool, "default": True}`
- `{"type": bool, "default": False}`
- `{"type": bool, "default": None}`

kdeldycke:
Given how @davidism is uncomfortable with the removal of `default_is_missing`, maybe we can target the `8.3` release instead of the `8.2.2` one (also suggested by @Rowlando13)?

Rowlando13:
@davidism Could you please weigh in? I think 8.3.0 makes sense.

potiuk:
> @davidism Could you please weigh in? I think 8.3.0 makes sense.

This one fixes #2952  - which is very annoying for CI  - and reason why we have < 8.2.0 for click in Airflow. So it would be worth to fix it in 8.2.2 and pretty quickly - also beacaue people think this is a FEATURE not a BUG - someone already created a PR documenting behaviour of ENVVAR with flags/bools thinking that it is how it is supposed to be https://github.com/pallets/click/pull/2988

potiuk:
BTW. We tested this one in Airflow and it looks good and indeed solves the #2952 issue. I can confirm that.

davidism:
Oh sorry, I missed the latest messages here. It looks like this doesn't change expected behavior, only fixes issues, so this should be fine for 8.2.2.

Rowlando13:
@kdeldycke Great work! The community feedback from this already looks very good. You are good to merge into stable. You should have perms, if not then let me know and I will merge!

kdeldycke:
> BTW. We tested this one in Airflow and it looks good and indeed solves the #2952 issue. I can confirm that.

Thanks @potiuk for testing this PR! Airflow being popular means my PR seems sound enough to be merged upstream! :)

> @kdeldycke Great work! The community feedback from this already looks very good. You are good to merge into stable. You should have perms, if not then let me know and I will merge!

Ok I'll try to merge it myself. Should be a good test for the process. Thanks for the approval! 👍

kdeldycke:
Hmmm @Rowlando13 ... What's your preferred way here? A proper merge commit or a rebase? On `stable` or `main`? (feel free to do it if you have the answer so we don't waste time with back and forth).

davidism:
Since it's a bug fix it goes in stable. Since you don't have any work commits and no merge conflicts (you already squashed/rebased), you can do a regular merge commit.

davidism:
Tagged you in a few other issues that might be affected by this.

PR #2960, opened 2025-06-02
Move Utils and Unicode-support docs from reST to MyST

denuoweb:
This PR switches the Utils and Unicode-support documentation from reST to MyST.


denuoweb:
Added "click." after {func} and before the function names for proper calling without the click module header.

davidism:
In general, prefer adding a `currentmodule` directive rather than adding `click.` or `click.sub.` in front of references. Otherwise links look verbose or redundant. It's possible to use this directive in MyST, although it's a little more annoying to write. Alternatively, if the names are unambiguous, you can add `.` in front of them and Sphinx will find the reference automatically. If you've been doing this in pages that have already been merged, please go back and correct that.

denuoweb:
@davidism I added back the currentmodule in myst form and removed the extra click. additions. This was the first instance of it so no need to modify the previously merged files. @Rowlando13 These two docs should now be ready for you to review.

docs/unicode-support.md
```diff
@@ -0,0 +1,100 @@
+```{currentmodule} click
```
Rowlando13, line 3: Move `# Unicode Support` to top of file.

docs/utils.md
```diff
@@ -0,0 +1,393 @@
+```{currentmodule} click
+```
+
+# Utilities
```
Rowlando13: Move # Utilities to top of file.

docs/utils.md
```diff
@@ -0,0 +1,393 @@
...
+directly over it:
+
+```python
+with click.progressbar(length=total_size,
+                       label='Unzipping archive') as bar:
+    for archive in zip_file:
+        archive.extract()
+        bar.update(archive.size)
+```
+
```
Rowlando13: Use regular inline links where they are used rather than this.

denuoweb: > Use regular inline links where they are used rather than this.

This was a result of the rst2myst tool. I modified them to be like this:
```console
[Link](https://link.com)
```

denuoweb:
@Rowlando13 I moved the headers to the top and also made the two links inline.

PR #2972, opened 2025-06-07
Don't include envvar in error hint when envvar not configured

### LINKED ISSUE #2971: Error hint shows env var when one doesn't exist
#2696 added the environment variable to the error hint when `show_envvar` is set. However, it does not check to see if an `envvar` exists.

click-test.py
```python
import click

@click.command()
@click.option("--foo", envvar="FOO", show_envvar=True, type=click.types.BOOL)
@click.option("--bar", show_envvar=True, type=click.types.INT)
def main(foo: bool, bar: int) -> None:
    print(f"{foo=} {bar=}")

if __name__ == '__main__':
    main()
```

Examples of running this file where each argument is invalid. And one example of the help output.
```
$ python ./click-test.py --bar aaa
Usage: click-test.py [OPTIONS]
Try 'click-test.py --help' for help.

Error: Invalid value for '--bar' (env var: 'None'): 'aaa' is not a valid integer.
$ python ./click-test.py --foo aaa
Usage: click-test.py [OPTIONS]
Try 'click-test.py --help' for help.

Error: Invalid value for '--foo' (env var: 'FOO'): 'aaa' is not a valid boolean.
$ python ./click-test.py --help
Usage: click-test.py [OPTIONS]

Options:
  --foo BOOLEAN  [env var: FOO]
  --bar INTEGER
```

I would expect the error hint for `--bar` to act like the help output by not displaying any information about an environment variable.

Environment:

- Python version: 3.10.12
- Click version: 8.2.1

plannigan:
Add a check for if `envvar` is `None` before appending to the error hint text.

fixes #2971

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,12 @@
 .. currentmodule:: click
 
+Version 8.2.0
+-------------
+
+Unreleased
+
+- Don't include envvar in error hint when not configured. :issue:`2971`
+
 Version 8.2.1
 -------------
 
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2676,7 +2676,7 @@ class Option(Parameter):
 
     def get_error_hint(self, ctx: Context) -> str:
         result = super().get_error_hint(ctx)
-        if self.show_envvar:
+        if self.show_envvar and self.envvar is not None:
             result += f" (env var: '{self.envvar}')"
         return result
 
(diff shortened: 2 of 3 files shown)


paveldikov:
I am interested in this PR, and noticed it's still unreviewed. Is there anything that can be done to unblock this?

Rowlando13:
@kdeldycke Would you mind reviewing this?

kdeldycke:
Hmmm. Envvar would need a bit of refactoring right and left. I can propose something for a bigger minor/major release later. In the meantime, this PR is not wrong, solve a edge-case and adds a unit test delimiting the current behavior. So I'm good to merge it to the 8.2.2 release.

kdeldycke:
I just fixed the merge conflict. And all tests are passing. This PR is ready to be merged into stable.

PR #2980, opened 2025-06-17
Documentation reST to MyST

Veebaa:
Converted documentation.rst to documentation.md

Part of issue #2886


docs/documentation.md
```diff
@@ -1,17 +1,18 @@
-Help Pages
-===================
+# Help Pages
 
+```{eval-rst}
```
Rowlando13: This can be done in myst directly.

Veebaa: From what i could find .. currentmodule:: is a Sphinx directive that only exists in reStructuredText, eval-rst is the most straight forward to use for this.

Otherwise, would you prefer :
1. autodoc-style directives that explicitly import modules e.g. ```{autofunction} click.echo.
2.  Or fully qualify the reference, e.g. :func:click.echo

Rowlando13: \`\`\`{currentmodule} click
\`\`\`

Rowlando13: It is hard to find so I listed it above.

Veebaa: All done :+1:

PR #2988, opened 2025-06-27
docs: clarify boolean flag behavior with envvar and add example

### LINKED ISSUE #2952: Flag value is always true when setting any envvar

### LINKED ISSUE #2967: Document use cases for --option/--no-option pattern

KarinaNi:
This PR improves the documentation for boolean flags when used with environment variables. It adds a new subsection under the "Boolean" options section to explain how `envvar=` behaves with `--flag/--no-flag` patterns.

It includes a runnable example showing how to ensure proper parsing of values like `"false"` and `"0"` by using `type=click.BOOL` and passing the default as a string.

fixes #2967 
fixes #2952

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -311,24 +311,7 @@ If you want to define an alias for the second option only, then you will need to
 
     invoke(info, args=['--help'])
 ```
-### Boolean Flags and Environment Variables
 
-You can use 'envvar=...' with boolean flags like '--debug/--no-debug', but it has some caveats. By default, Click treats  **any set environment variable as "flag present"**, even if the value is `"false"` or `"0"`.
-
-```{eval-rst}
-.. click:example::
-
-    @click.command()
-    @click.option('--dry-run/--no-dry-run', default='False', envvar='DRY_RUN', type=click.BOOL)
-    def cli(dry_run):
-        click.echo(f"Dry run: {dry_run}")
-
-.. click:run::
-
-    invoke(cli, env={"DRY_RUN": "false"})
-    invoke(cli, env={"DRY_RUN": "true"})
-
-```
 ## Flag Value
 
 To have an flag pass a value to the underlying function set `flag_value`. This automatically sets `is_flag=True`. To set a default flag, set `default=True`. Setting flag values can be used to create patterns like this:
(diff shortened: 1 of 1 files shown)


potiuk:
I think it's not what @davidism wants to do - as far as I understand, the behaviour of click for 8.2.0 and 8.2.1 is really a bug and we should bring back the old behaviour (and we hit it as well in Airflow) - we are about to provide a PR fixing it

Rowlando13:
Please read my comments. This is not what you asked to do.

KarinaNi:
I apologize for the inconvenience.  I was wondering if the current changes about explaining the use cases of double flag (positive/negative) is more close to what you were looking for. Any advice/suggestion is appreciated. Let me know if I am doing anything wrong.

docs/options.md
```diff
@@ -334,6 +334,33 @@ To have an flag pass a value to the underlying function set `flag_value`. This a
     invoke(info, args=['--lower'])
     invoke(info)
 ```
+### Positive / Negative Flags
```
SurenNihalani, line 337: Would calling them boolean flags be sufficient? I'm not aware of click's nomenclature so grep around to figure out which term is more commonly used in the repo

docs/options.md
```diff
@@ -334,6 +334,33 @@ To have an flag pass a value to the underlying function set `flag_value`. This a
...
+
+Click supports paired boolean flags like `--debug/--no-debug` or `--dry-run/--no-dry-run`, which let users explicitly enable or disable a feature. These are especially useful when clarity or override behavior matters.
```
SurenNihalani, line 339: Use only single example. Second example doesn't seem to add value

docs/options.md
SurenNihalani: This code is powering: https://click.palletsprojects.com/en/stable/options/

We'll want to move this under the boolean flag on 280.

docs/options.md
```diff
@@ -334,6 +334,33 @@ To have an flag pass a value to the underlying function set `flag_value`. This a
...
+
+- **When defaults change dynamically**
+  If your program determines defaults at runtime, users may want to override them with either form of the flag.
```
SurenNihalani, line 342: Maybe show an example when you can change the default dynamically. Like, default=(os.environ["FORCE_VALUE"] or sys.isatty())

Rowlando13:
Yes I was looking for docs addition for why the --flag/--no-flag pattern exists.

kdeldycke:
I guess this PR can be revisited once #2956 get merged, as I added to the latter a detailed explanation on how the values are parsed and normalized in this case: https://github.com/pallets/click/pull/2956/files#diff-2e046795c81fb2bcc2667705f88c6421fa988047f147ddb2333ec8da5f638b7b

The mechanism is quite dense, with the two concepts or flag activation/deactivation and option's value. So maybe there is a way to make things more approachable through this PR once  #2956 get merged.

Issue #2911, opened 2025-05-13
`Choice` prefers `Enum.name` even for `StrEnum` where it should use the value


jwodder:
Consider the following code:

```python
from enum import Enum
import click


class Color(str, Enum):
    RED = "red"
    GREEN = "green"
    BLUE = "blue"

    def __str__(self) -> str:
        return self.value


@click.command()
@click.option("-C", "--color", type=click.Choice(list(Color)), show_default=True)
def main(color: Color) -> None:
    print(repr(color))


if __name__ == "__main__":
    main()
```

With Click 8.1.8, running this script with `-C red` works as expected, but with Click 8.2.0, it instead fails with:

```
Usage: enum-choice01a.py [OPTIONS]
Try 'enum-choice01a.py --help' for help.

Error: Invalid value for '-C' / '--color': 'red' is not one of 'RED', 'GREEN', 'BLUE'.
```

This was observed with Python 3.13.3 on macOS 14.7.5.

qdegraaf:
Could this be the culprit: https://github.com/pallets/click/pull/2796/files#diff-dcb534e6a7591b92836537d4655ddbd2f18e3b293c3420144c30a9ca08f65c4eR298 ? 

And if so, is this behaviour intentional? If not, I'd be happy to open a PR with an additional test for `Enum`'s that are also `str`'s and tweak the line to return the string value in that additional case. I am new to the repo though so could be missing context.

qdegraaf:
I've replicated the example as a test and see now that it is by design. `case_sensitive: bool = True` by default. @jwodder setting `case_sensitive` to `False` in `click.Choice` makes your example above work for me. 

I guess the follow-up question is whether this default is desirable. I could see quite a few
```python
class SomeStrEnum(StrEnum)
    FOO = "foo"
    BAR = "bar"

class SomeOtherStrEnum(str, Enum)
    BLA = "bla"
    ETC = "etc"
```
that are currently being used in `click.Choice` that all implicitly used the `str` values of the `Enum` values. 

We experienced a similar, but not exactly related issue for `StrEnum` default values in the `typer.Option` wrapper around `click.Choice` where something like:
```python
@dataclass
class CLIOptions
    foo_preference: Annotated[
        SomeStrEnum,
        typer.Option( "-f", "--foo-preference"),
    ] = SomeStrEnum.FOO
```
[...]

PR #3004, opened 2025-07-20
Fix rendering of Enum choices' default

kdeldycke:
This PR fix (and test) the rendering of a variety of `Enum`-based choices. Specifically their default value in the help/usgae screen.

Additionally, it adds a couple of unit tests to lock-in and document the current behavior of choices.

Refs:
 - #2911
- #2928


kdeldycke:
I think this fix is safe for inclusion into the upcoming 8.2.2 release.

kdeldycke:

PR #3011, opened 2025-07-22
Improve discoverability of Context.max_content_width

learn-more:
It took me a while to find this option (because I was not familiar with the `Context` class at all,
so I figured this would be an improvement in the documentation.


Rowlando13:
Docs fixes go on stable.

PR #3013, opened 2025-07-24
Fix completions for quoted/escaped parameters in Fish

### LINKED ISSUE #2995: Incorrect handling of quoted & escaped parameters for Fish completions
Completions for parameters in quotes, partially quoted or with escaped characters (e.g. whitespace) don't work properly for Fish shell.

I believe the issue is that for fish completion, the incomplete paramater is passed through the `COMP_CWORD` environment variable and it's expected that it will be contained in `COMP_WORDS` as the last parameter. However, before doing this match `COMP_WORDS` is parsed, which quotes and unescapes the parameters, while `COMP_CWORD` is not causing the match to fail. See code [here](https://github.com/pallets/click/blob/91ac02718af3a0b88dbe282501c06642da0785fe/src/click/shell_completion.py#L392).

## How to replicate

Consider the following command

```python
@click.command()
@click.argument('bar', type=click.Choice(('hello world',)))
def main(bar: str):
    pass
```

Trying to complete `foo hello` in Fish will correctly return completion for `foo hello\ world`.

The following incomplete commands will not return any completions
- `foo hello\ `
- `foo "hello`
- `foo "hello"`

Environment:

- Python version: 3.13.5
- Click version: 8.2.1

HansLehnert:
Fix completions for Fish shell when the value being completed has spaces
or is enclosed in quotes

Fixes #2995

### DIFF
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -405,6 +405,8 @@ class FishComplete(ShellComplete):
     def get_completion_args(self) -> tuple[list[str], str]:
         cwords = split_arg_string(os.environ["COMP_WORDS"])
         incomplete = os.environ["COMP_CWORD"]
+        if incomplete:
+            incomplete = split_arg_string(incomplete)[0]
         args = cwords[1:]
 
         # Fish stores the partial word in both COMP_WORDS and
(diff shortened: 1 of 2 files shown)


Rowlando13:
Can you please rebase so that it only contains your commit. I changed the branch to stable.

Rowlando13:
@AndreasBackx Would you mind reviewing this?

HansLehnert:
I rebased the changes

AndreasBackx:
Looks sane, I'm not super familiar with Fish itself and how it handles the `COMP_CWORD`. I'm more familiar with bash/zsh but this doesn't _seem_ wrong and it does fix the issue. Thanks @HansLehnert!

Rowlando13:
@HansLehnert could you please make a separate PR for changes. I did not see there wasn't one.

AndreasBackx:
I'll take care of the CHANGES mention, don't worry about it as I merged this without checking. @HansLehnert

Issue #2092, opened 2021-10-10
click.confirm does not cooperate with readline


alex-ball:
This follows on from #665.

Consider this code again, using `click.confirm` instead of `click.prompt`:

```python
import click, readline
click.confirm("Input")
```
- Type anything then backspace
    - **Actual behaviour:** the entire line is erased
    - **Expected behaviour:** only the entered character is erased
- Type return only
    - **Actual behaviour:** no newline is printed to the terminal (a following `click.echo()` or `print()` will appear to be ignored)
    - **Expected behaviour:** a newline is printed to the terminal

The solution from PR #1836 seems to work in the case of `click.confirm` as well:

```python
    while True:
        try:
            # Write the prompt separately so that we get nice
            # coloring through colorama on Windows
            echo(prompt.rstrip(" "), nl=False, err=err)
            value = visible_prompt_func(" ").lower().strip()
```

Environment:

- Python version: 3.8.10
- Click version: 8.0.1

PR #3021, opened 2025-07-28
Fix rendering when `prompt_suffix` is empty

### LINKED ISSUE #3019: click.prompt and click.confirm argument prompt_suffix no longer works when suffix is empty
After #1836 and #2093 were merged the behavior of the `prompt_suffix` argument in the `click.prompt` and `click.confirm` functions changed.

# Before these changes in click version 7.1.2

Prior to the #1836 and #2093 PRs when an empty `prompt_suffix` was passed, no suffix was applied to the prompt.

Here's how `prompt` used to behave

```python
>>> import click
>>> click.prompt("test", prompt_suffix="")
testfoo
'foo'
>>> click.prompt("test", prompt_suffix=" ")
test foo
'foo'
```

Notice that when a `prompt_suffix of `""` is passed, there is no suffix after the prompt text of `test`. This is illustrated by how the text typed by the user (in this case `foo`) displays immediately after the prompt without a space character.

# Current behavior in click 8.0.0 and newer

```python
>>> import click
>>> click.prompt("test", prompt_suffix="")
test foo
'foo'
>>> click.prompt("test", prompt_suffix=" ")
test foo
'foo'
```

Notice now that if you tell `prompt` to use no suffix, it still applies a space character suffix.

This is because the fixes in #1836 and #2093 presumes that the last character of the prompt is a space character. It then [`rstrip`s the space character and adds the space character back in](https://github.com/pallets/click/pull/1836/files#diff-1f88635ce1d1f86bb7f9de29ef053bf8fc628a13530d1d2c6a69d0ad637e32cdR130-R133). This doesn't work if the last character of the prompt is not a space character though.

This matters because it makes it impossible to do prompts like the following example

# Example

```python
import click
click.prompt("What IP address would you like? : 192.168.1.", prompt_suffix="")
```

This would have the user enter the last octet of an IP and make it clear that they can only set something within the displayed subnet. The problem is that the resulting display looks like this

```
What IP address would you like? : 192.168.1. 123
```

# Note

The fixes in these two PRs can't just be reverted because the `readline` behavior that inspired them still exists.

Environment:

- Python version: 3.13
- Click version: 8.0.3

gene1wood:
This changes the implementation of the fixes in #1836 and #2093 to address the problem that they created when `click.prompt` or `click.confirm` has a `prompt_suffix` value which is empty.

This still fixes the problems reported in #665 and #2092 but does so without modifying the prompt and adding a trailing space.

Previously, the problem was solved by trimming all trailing space characters from the prompt and then adding a single trailing space to every prompt. This was only visible if the last character in the prompt wasn't a space character. In those cases a prompt like "This is my prompt" would render as "This is my prompt ". Additionally, in the previous solution, if the `err=True` argument was passed, that final trailing space character that was added would be sent to STDOUT while the rest of the prompt would go to STDERR. This also wasn't very apparent since it was a space character.

This different implementation uses the last character of the prompt to work around #665 and #2092 instead of adding a space character. This fixes most of the potentially confusing behavior. It does retain one confusing behavior which is described above, where if you use `err=True` the last character of the prompt is sent to STDOUT instead of STDERR. Previously, this last character sent to STDOUT would always be a space character. Now if the last character of the prompt is not a space character, that final character will go to STDOUT.

For example if you call

```python
click.prompt("bar", prompt_suffix="", err=True)
```

The `ba` will go to STDERR and the `r` will go to STDOUT. Previously `bar` would go to STDERR and an added ` ` would go to STDOUT.

This odd behavior, which is already present but just applies to a space character, is worth accepting so that the `prompt_suffix` correctly renders the suffix as it did in version 7.1.2 and prior.

This also adds unit tests for cases where `prompt_suffix` is empty.

This fixes #3019


kdeldycke:
I agree with you that having the last character of the prompt being sent to `<stdout>` instead of `<stderr>` is confusing.

But your PR has the merit of:
- making the behavior more deterministic
- not hiding the edge-case visually by resorting to spaces
- adding unit-tests on cases that were not covered before

For me this is a case of "worse is better", in which we gradually tends to the ideal behavior. The ideal behavior being a refactor of the `prompt()` / `confirm()` methods.

I am in favor of merging it in `stable` for a Click 8.3.1 release.

But before that, could you:
- add a `.. versionchanged:: 8.3.1` item in the docstring of each function to explain the change in behavior,
- add a changelog entry to `CHANGES.rst`,
- and change the branch target from `main` to `stable` please?

gene1wood:
@kdeldycke Sure thing

I've 
* Re-added the note to `CHANGES.rst`
* Changed the PR to be based off of [`stable`](https://github.com/pallets/click/tree/stable)
* Updated docstrings for `confirm` and `prompt`
  * Note : I tried to find a sweet spot between too verbose and too succinct with the docstring additions. If you'd like to see more or less detail let me know and I can tweak them.

src/click/termui.py
```diff
@@ -119,6 +119,13 @@ def prompt(
                          show_choices is true and text is "Group by" then the
                          prompt will be "Group by (day, week): ".
 
+    .. versionchanged:: 8.3.1
```
davidism, line 122: Shorten this to "a space is no longer appended to the prompt".

gene1wood: Sure thing. Done.

kdeldycke (approved):

PR #3023, opened 2025-07-31
Properly lazily import shutil

ofek:
I agree with the intention to lazily import but it wasn't actually happening. Locally I get the following on Python 3.12:

```
❯ python -m timeit -n 1 -r 1 "import shutil"
1 loop, best of 1: 8.09 msec per loop
```

### DIFF
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -17,7 +17,6 @@ import typing as t
 from gettext import gettext as _
 from io import StringIO
 from pathlib import Path
-from shutil import which
 from types import TracebackType
 
 from ._compat import _default_text_stdout
@@ -235,8 +234,6 @@ class ProgressBar(t.Generic[V]):
         ).rstrip()
 
     def render_progress(self) -> None:
-        import shutil
-
         if self.hidden:
             return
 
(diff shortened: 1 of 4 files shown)


ofek:
Okay, that's everything!

davidism:
Can you target this at stable instead?

davidism:
There's a test that checks what modules are imported, you can probably remove shutil from that list as well.

ofek:
I changed to stable and updated that test! If it's fine with you, I'm interested in the coming weeks/months making sure that everything (within reason) is lazily imported.

davidism:
Definitely, it's a bit annoying to code, but I think the speed is worth it for the CLI. We want to distinguish between what's going to be imported no matter what (general Python startup / used everywhere anyway) versus what only comes up in certain use cases. That's what that import test is for, so we know if something gets imported unintentionally, and can review imports from time to time.

Rowlando13:
@ofek Can you rebase onto current stable. There are my commits that merge stable into master on the branch. Same thing happened to me and I am not sure how exactly.

ofek:
Done!

Rowlando13:
@ofek Could you please add a change entry for this on another PR. I did not see there was not one.

AndreasBackx:
I'll take care of the CHANGES mention, don't worry about it.

Issue #2483, opened 2023-04-05
Missing auto-generated environment variables in help screen & case-sensitivity


kdeldycke:
Here is a simple CLI that mixes auto-generated and multiple environment variables:
```python
from click import command, echo, option, pass_context

@command(context_settings={"auto_envvar_prefix": "yo", "show_default": True})
@option("--flag/--no-flag", envvar=["FlAg", "sUper"], show_envvar=True)
@pass_context
def my_cli(ctx, flag):
    echo(f"Flag value: {flag}")

if __name__ == "__main__":
    my_cli()
```

This produces the following help screen:
```shell-session
$ python ./multiple_envvars.py --help
Usage: multiple_envvars.py [OPTIONS]

Options:
  --flag / --no-flag  [env var: FlAg, sUper; default: no-flag]
  --help              Show this message and exit.
```

Now let's try a couple of different envvars:
```shell-session
$ FlAg=1 python ./multiple_envvars.py
Flag value: True

$ FlAg=0 python ./multiple_envvars.py
Flag value: False
```

```shell-session
$ sUper=1 python ./multiple_envvars.py
Flag value: True

$ sUper=0 python ./multiple_envvars.py
Flag value: False
```

```shell-session
$ yo_FLAG=1 python ./multiple_envvars.py
Flag value: False

$ YO_FLAG=1 python ./multiple_envvars.py
Flag value: True

$ YO_FLAG=0 python ./multiple_envvars.py
Flag value: False
```

The surprise here is the discovery process of the auto-generated `YO_FLAG`. I expected it to respect the case-sensitivity of others. but everything auto-generated is uppercased.

So I wonder if we should enforce upper-casing all env vars, whatever their source.

And additionally, shouldn't we add the auto-generated `YO_FLAG` to the help screen to improve discoverability?

For reference, I ran these test on macOS / zsh, and with the latest `click 8.3.1`.

kdeldycke:
Things are a little bit messier than I initially expected.

For instance, if I register explicitly the `yo_FLAG` as an environment variable on the option, it is properly recognised with its own particular mixed-case. Which highlight the difference of treatment between registered envvar and auto-generated envvar.

I currently mapped all these cases and mismatches on a test case here: https://github.com/kdeldycke/click-extra/blob/48f980b0d6fb1e7ddb914fe03f34d3460c8ef0ca/click_extra/tests/test_parameters.py#L62-L98

[...]

Issue #2012, opened 2021-07-16
`flag_value` is passed through `ParamType.convert`


jgadling:
In upgrading from click 7.1.2 to 8.0.1, our click-based CLI started failing with type errors. We have click options whose default values are classes, and as of 8.0.1 they're being converted to strings.

This script is able to reproduce the behavior change:

```python
import click

class Class1:
    pass

class Class2:
    pass

@click.command()
@click.option("--cls1", "config_cls", flag_value=Class1, default=True)
@click.option("--cls2", "config_cls", flag_value=Class2)
def test(config_cls):
    print(config_cls)
    print(type(config_cls))

if __name__ == "__main__":
    test()

```

Behavior with click 7.1.2:
```shell-session
# pip freeze | grep click
click==7.1.2
# python test.py
<class '__main__.Class1'>
<class 'type'>
```

Behavior with 8.0.1:
```shell-session
# pip freeze | grep click
click==8.0.1
# python test.py
<class '__main__.Class1'>
<class 'str'>
```

The type of `config_cls` should be the class we specified, not 'str'.

Environment:

- Python version: 3.9
- Click version: 8.0.1

MHannila:
This seems to have already appeared in the `8.0.0` version.  
Ran into this myself using enum member as the `flag_value` and managed to get an integer to convert to string as well so I did some testing (with `8.0.1`).

A simple workaround solution for most cases seems to be just defining the `type`, but there are also issues if you have multiple flags for the same argument that are of different types. Then even defining the types doesn't help. That hopefully isn't a problem for many though, because why would you even do that.
```python
from enum import Enum

import click
import pkg_resources


class EnumClass(Enum):
    FOO = 1
    BAR = 2


@click.command()
@click.option("--enum-no-type", flag_value=EnumClass.FOO, default=True)
@click.option("--enum-with-type", flag_value=EnumClass.FOO, type=EnumClass, default=True)
@click.option("--integer", flag_value=1, default=True)
@click.option("--integer-as-int", 'multitype_borked', flag_value=1, default=True)
@click.option("--integer-as-str", 'multitype_borked', flag_value='1')
@click.option("--integer-as-int-with-type", 'multitype_borked2', flag_value=1, type=int, default=True)
@click.option("--integer-as-str-with-type", 'multitype_borked2', flag_value='1', type=str)
def test(**kwargs):
    for name, value in kwargs.items():
        print(name, value, type(value))


if __name__ == '__main__':
[...]

PR #3030, opened 2025-08-06
Differentiate absence of value and value of absence for `default` and `flag_value` arguments

### LINKED ISSUE #1992: Binary flags vs `default=None`: semantics and issue with prompt
The current behavior is:
- if `default` is not provided, the option default is set to `False`
- if `default` is set to `None`, default is set to `None`.

I think this may be accidental behavior but it's not senseless: settings `default=None` you can force the prompt to require an explicit value to be provided (it'll be `[y/n]` rather than `[y/N]`). 

Nonetheless, I noticed that if you provide `default=None` to a flag not having a secondary option, the prompt message is broken: it won't include "[y/n]" at the end.

_Originally posted by @janluke in https://github.com/pallets/click/issues/1965#issuecomment-873600435_

### LINKED ISSUE #2164: Support default for arguments with multiple values
I'd like to be able to specify a default for arguments with multiple values.

A currently failing test case:
```
import pytest
import click

@pytest.fixture(scope="function")
def runner(request):
    return CliRunner()

def test_multivalue_args_with_default(runner):
    @click.command()
    @click.argument('files', nargs=-1, default=['arg1'])
    def cmd(files):
        for filename in files:
            click.echo(filename)

    result = runner.invoke(cmd, ['--'])
    assert result.output == 'arg1\n'
```

Fails with:

```
TypeError: 'default' is not supported for nargs=-1.
```

### LINKED ISSUE #2514: make `flag_value=None` actually set flag value to `None`
Right now, click uses `None` as a sentinel for the `flag_value` keyword in the `@click.option` decorator, which triggers the default behavior where click tries to figure out the flag value by itself (either `True` or `False`). This has the unfortunate effect that you cannot tell click to use `None` as a flag value.

My command function has a keyword with the type `bool | None`, and I want my CLI to offer three flags, where one of them sets it to `True`, a second sets it to `False`, and a third sets it to `None` (which is also the default). Right now, there does not seem to be a straightforward way to do this in click.

Supporting `None` as a value can be implemented trivially by using some other sentinel that is guaranteed to be unique and unused by outsiders, which is a fairly common pattern in libraries where `None` is accepted as a possible/meaningful value for some keyword rather than to mean "this keyword was not specified". Example:

```python
_unspecified = object()

# The `@click.option` decorator definition
def option(..., flag_value=_unspecified, ...):
    # ....
    if flag_value is _unspecified:
        # figure out flag value
    else:
        # accept flag value as-is (including if it's None)
```

Even better, the sentinel can be made a public feature by giving it a more meaningful name like this:

```python
AUTO = object()

# The `@click.option` decorator definition
def option(..., flag_value=AUTO, ...):
    # ....
    if flag_value is AUTO:
        # figure out flag value
    else:
        # accept flag value as-is (including if it's None)
```

Then, users of the library can explicitly request an automatic value for the flag like this:

```python
import click

@click.command
@click.option('-x', flag_value=click.AUTO)
def cmd(x: bool) -> None:
    ...
```

Note that I know I can make a non-flag option that accepts an argument and converts it to one of `True`, `False` or `None`, but the three flags approach is far more preferred and more appropriate for this use case. (And I suppose many other people's use cases as well.)

Also, while there might be some workarounds for this, like using any other arbitrary three constants as flag values, but having to do so instead of simply being able to supply `None` as a flag value is kind of awkward (especially when you take type annotations into account), so I think that in the grand scheme of things, it would probably be much better if click supported something elementary like this, rather than forcing the user to come up with awkward workarounds.

In any case, thank you in advance.

### LINKED ISSUE #2610: The `flag_value` shadows `default` in `click.option()`
There are two problems:

1. The `flag_value` shadows `default` in `click.option()`
2. The default type is omitted, the option converts to `str` anyway.

About 2., The click docs says in the option section: "If no type is provided, the type of the default value is used."
But it is not True. I have to use type=frozenset to make the option's value not be "frozenset()", which breaks my function iterating it.

And the 1. is the big deal.
I provide my option code below and a [minimal reproducer script.](https://github.com/mojopaa/menv/blob/main/notes/reproducer.py)
I'm making a venv for Mojo language, and looking at new venv code on [cpython.](https://github.com/python/cpython/blob/main/Lib/venv/__init__.py)

If click can not fix this, it can not do things like Brett Cannon does in venv cli design. That's not good.
I want `scm_ignore_files` to be `frozenset()` if the flag is provided, else it is default to `frozenset(["git"])`.

Why is default not useful and the flag_value is always used?

```
@click.option(
    "--without-scm-ignore-files",
    "scm_ignore_files",
    is_flag=True,
    type=frozenset,
    flag_value=frozenset(),
    default=frozenset(["git"]),
    help="Skips adding SCM ignore files to the environment "
    "directory (Git is supported by default).",
)
```

Thanks for watching!

Environment:

- Python version: Python 3.10.12
- Click version: 8.1.7

### LINKED ISSUE #3024: Click 8.2.2 breaks auto-reloader-active on flask --debug
On upgrading click to version 8.2.2, `flask --debug run` no longer automatically enables the reloader.  If I downgrade click to 8.2.1, it works again.

```
$ uv pip install flask
Using Python 3.13.2 environment at: venv
Resolved 7 packages in 2ms
Installed 7 packages in 3ms
 + blinker==1.9.0
 + click==8.2.2
 + flask==3.1.1
 + itsdangerous==2.2.0
 + jinja2==3.1.6
 + markupsafe==3.0.2
 + werkzeug==3.1.3

$ cat app.py 
from flask import Flask

app = Flask(__name__)

if __name__ == '__main__':
    app.run()

$ flask run --debug
 * Debug mode: on
WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.
 * Running on http://127.0.0.1:5000
Press CTRL+C to quit
^C

$ uv pip install click==8.2.1
Using Python 3.13.2 environment at: venv
Resolved 1 package in 1ms
Uninstalled 1 package in 0.65ms
Installed 1 package in 1ms
 - click==8.2.2
 + click==8.2.1

$ flask run --debug
 * Debug mode: on
WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.
 * Running on http://127.0.0.1:5000
Press CTRL+C to quit
 * Restarting with stat
 * Debugger is active!
 * Debugger PIN: 575-957-529
```

[From my trace, it looks like the issue affects the debugger as well, but I haven't been using or testing that.]

This appears to be connected to pallets/click#2952, but I'm not sure if the issue is now in click or in how flask is using click.

The issue occurs because [the --reload/--no-reload option](https://github.com/pallets/flask/blob/85c5d93cbd049c4bd0679c36fd1ddcae8c37b642/src/flask/cli.py#L898) is now defaulting to `False` when it is not specified on the command line, meaning the `if reload is None:` condition is not triggered to activate the reloader based on the debug flag.

Environment:

- Python version: 3.13.3
- Flask version: 3.1.1

kdeldycke:
Tl;Dr: this PR makes user-provided values (like `None`) first-class citizens.

### Description

- Introduce the `UNSET` sentinel value to track when `default` and `flag_value` have been explicitly set by the user on `Parameter`, `Option` and `Argument`
- Allow `None` or any other Python value to be used as `default` and `flag_value` 
- Fix dozens of flag-adjacent edge-cases
- Adds a new class to create self-documenting sentinel objects
- Adds comments and docstrings for internal methods (this should help write better higher-level documentation and tutorial in Sphinx)
- Rename some obscure internal variables to make code more approachable to external contributors
- Adds 200+ unittests
- Streamline code
- Fix typos
- Old-school hand-made code, LLMs only used as a glorified auto-complete

### Context

The pandora box was open in Click 8.2.0 with an attempt to always returns the `flag_value` if an option gets activated by its environment variable:
- https://github.com/pallets/click/issues/2746
- https://github.com/pallets/click/pull/2788

Another patch in Click 8.2.0 tried to fix how `flag_value` gets aligned to its `default`:
- https://github.com/pallets/click/pull/2829

The effect of these patches were reported after 8.2.0 release, and uncovered how tangled the flag logic is regarding its `default`, `flag_value`, `type` and `envvar` parameters:
- https://github.com/pallets/click/discussions/2863
- https://github.com/pallets/click/issues/2894
- https://github.com/pallets/click/issues/2897
- https://github.com/pallets/click/issues/2952

A first attempt to fix these issues was made in Click 8.2.1:
- https://github.com/pallets/click/pull/2930

As I was [aware of some of these edge-cases in 2023](https://github.com/pallets/click/issues/2483), I tackled this challenge and made a big PR to reconcile everything:
- https://github.com/pallets/click/pull/2956

This was released as Click 8.2.2. But it was not enough and even more nastier edge-cases were spotted (see the *Progress* section below). 8.2.2 was finally yanked to prevent widespread frustration.

The situation cannot be fixed with trivial changes as we are confronted with internal design limits. This PR is trying to fix all this mess.

### Progress

Currently playing Whac-A-Mole, but ultimate goal is to fix the following issues:

- [x] https://github.com/pallets/click/issues/1992

  **Status**: ✅ 100% fixed

  **Proof**: 
  - https://github.com/pallets/click/blob/14e6ceeb571e644440f96eca9350d490e6cadfd0/tests/test_termui.py#L492-L554

- [x] https://github.com/pallets/click/issues/2012

  **Status**: ✅ 100% fixed

  **Proof**: 
  - https://github.com/pallets/click/blob/da074218d9ef2cb1ee814cf70f383f005747b703/tests/test_options.py#L1749-L1817
  - https://github.com/pallets/click/blob/f9d006e7909108ae4049158950ee39cfa1348876/tests/test_options.py#L1850-L1864
  - https://github.com/pallets/click/blob/f9d006e7909108ae4049158950ee39cfa1348876/tests/test_options.py#L1727-L1744

- [x] https://github.com/pallets/click/issues/2164

  **Status**: ✅ 100% fixed

  **Proof**:
  - https://github.com/pallets/click/blob/3bd9e22cb7290b1e592ee1c1ed6050eea2438861/tests/test_arguments.py#L388-L392

- [x] https://github.com/pallets/click/issues/2514

  **Status**: ✅ 100% fixed

  **Proof**:
  - https://github.com/pallets/click/blob/14e6ceeb571e644440f96eca9350d490e6cadfd0/tests/test_options.py#L1278-L1279

- [x] https://github.com/pallets/click/issues/2610

  **Status**: ✅ 100% fixed

  **Proof**:
  - https://github.com/pallets/click/blob/14e6ceeb571e644440f96eca9350d490e6cadfd0/tests/test_options.py#L1697-L1718
  - https://github.com/pallets/click/issues/3036
  
- [x] https://github.com/pallets/click/issues/3024

  **Status**: ✅ 100% fixed

  **Proof**:
    - https://github.com/pallets/click/blob/14e6ceeb571e644440f96eca9350d490e6cadfd0/tests/test_options.py#L1254-L1255
    - https://github.com/pallets/click/blob/14e6ceeb571e644440f96eca9350d490e6cadfd0/tests/test_options.py#L1656-L1694
    - https://github.com/pallets/click/blob/9fcd409f8e9e233e26fab08fdd57c23e9273ef95/tests/test_options.py#L1697-L1722
    - https://github.com/pallets/click/blob/373429e75223b548e11f96fee487e33a623a151f/tests/test_basic.py#L279-L297
    - https://github.com/pallets/click/blob/373429e75223b548e11f96fee487e33a623a151f/docs/options.md?plain=1#L317-L329
    - https://github.com/pallets/click/blob/b6db38e974a58159b1e960994db6a6f3cc286793/tests/test_options.py#L1725-L1746


Rowlando13:
From just a quick look at this, it is pretty cool!

kdeldycke:
> From just a quick look at this, it is pretty cool!

I solved 99% of issues and all tests are passing. I am now trying to hunt for even more obscure edge-cases and adding more tests.

So I'm going to polish it in the next couple of days before requesting a proper review. In the meantime, I'm continue hacking on it in the open.

davidism:
Great work on tracking all this down, been following all week.

kdeldycke:
> Great work on tracking all this down, been following all week.

I was indeed working the full week on that! 😮‍💨

src/click/utils.py
```diff
@@ -29,6 +30,27 @@
...
+        `PEP 661 - Sentinel Values <https://peps.python.org/pep-0661/>`_.
+    """
+
+    UNSET = object()
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}.{self.name}"
+
+
+UNSET = Sentinel.UNSET
```
edgarrmondragon: This should be falsy, right?

kdeldycke: This is just an alias to not have to write `Sentinel.UNSET` everywhere.

edgarrmondragon: That doesn't address my question, though. `Sentinel.UNSET` should be falsy then:

```python
>>> from click.utils import UNSET
>>> value = UNSET
>>> if not value:
...     print('No value provided!')
... else:
...     print(value)
...
Sentinel.UNSET
```

kdeldycke: Ok, I get it. Just added that behavior to the sentinel: https://github.com/pallets/click/pull/3030/commits/6eff766f077856563244d696c1c49ff3f4cbdd65

edgarrmondragon: Thanks!

src/click/utils.py
```diff
@@ -43,6 +43,9 @@ class Sentinel(enum.Enum):
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}.{self.name}"
 
+    def __bool__(self):
+        return False
```
Rowlando13: I am not sure that we want this. Feels like we want to be explicit whether something is a sentinel all the way down. It does end up with some boiler plate, but the click code is very long lived and I can imagine lots of spots where we don't foresee wanting to know which of the 3 positions we have, user set, default, or unset, and later we want it.

Rowlando13: @edgarrmondragon @kdeldycke thoughts?

src/click/core.py
```diff
@@ -2186,7 +2155,12 @@ def to_info_dict(self) -> dict[str, t.Any]:
             "required": self.required,
             "nargs": self.nargs,
             "multiple": self.multiple,
-            "default": self.default,
+            # XXX Normalize the default to None if UNSET. This is acceptable for
+            # now as to_info_dict() has been designed to be used for documentation
+            # purposes, and UNSET is not a user-facing value.
+            # TODO: Make UNSET a first class citizen in Click 9.0, remove this
```
davidism: I don't think we should make `UNSET` public. It should remain an implementation detail to make the internal tracking work, but users should never be passing or receiving it.

src/click/utils.py
davidism: I agree. The sentinel seems purely intended for internal tracking detection, I don't see a situation where we shouldn't be concerned with whether the value is the sentinel. I also don't see a situation where the sentinel should be passed or received by the user.

edgarrmondragon: I also agree that the sentinel should not concern the user, but I mainly worry about existing code that relies on `None` being the default value for `flag_value`.

Some code may be checking the value explicitly against `None`, and other code may be checking if it's falsy.

For example, this is very similar to something we have in production:

```python
# click_3030.py
import click
from dataclasses import dataclass
from typing import Literal


@dataclass
class Config:
    format: Literal["json", "console"] = "console"


def _get_output_format(ctx, param, value):
    if value is None:
        return Config().format
    return value


@click.command()
@click.option("--json", "output_format", flag_value="json", callback=_get_output_format)
@click.option("--console", "output_format", flag_value="console", callback=_get_output_format)
def main(output_format):
    click.secho(f"Output format: {output_format} ({type(output_format)})", fg="green")


if __name__ == "__main__":
    main()
```

It works as expected in 8.2.1:

```console
uvx --with 'click==8.2.1' python click_3030.py
Output format: console (<class 'str'>)
```

but breaks here

```console
uvx --with 'git+https://github.com/pallets/click@refs/pull/3030/head' python click_3030.py
Output format: None (<class 'NoneType'>)
```

This is fixed if I pass `default=None` explicitly, but I think requiring that from the developer is a breaking change.

Rowlando13: At this point, there is going to be some breakage. Every other attempt to fix some edge case, uncovers a new one. I leave it up to @kdeldycke to figure out breakage / fix things trade off. If he wants the behavior, of the sentinel to be falsely, then it's fine. I only opposed it, because it seemed like it was suggested to make it more convenient for users, which I think that we have agreed are not getting access to it.

kdeldycke:
OK I have finished all my changes, all tests are passing, so I'm requesting a proper review.

I will keep the commits as-is during the review to not break the precise references to code changes.

Once everyone is ready for the merge, I'll squash everything and do the rebase dance on top of `stable`/`main`.

src/click/core.py
kdeldycke: > I don't think we should make `UNSET` public. It should remain an implementation detail to make the internal tracking work, but users should never be passing or receiving it.

Good to me. I removed that mention and hide `UNSET` to the user everywhere I can. See: https://github.com/pallets/click/blob/64c84dd3841f5976155c049d42424e9fb68d4b48/src/click/core.py#L2147-L2152

src/click/utils.py
kdeldycke: > I am not sure that we want this. Feels like we want to be explicit whether something is a sentinel all the way down. It does end up with some boiler plate, but the click code is very long lived and I can imagine lots of spots where we don't foresee wanting to know which of the 3 positions we have, user set, default, or unset, and later we want it.

@Rowlando13 Yes, that was my initial intention: to make things extremely explicit. This helped me identified lots of edge-cases over the past few days.

But now that I think I've finally fixed everything, I'm OK to make `UNSET` falsy. It is syntactic sugar that could make the code easier to comprehend for a first-time user. If we find a regression or another edge-case, I'll pile it on the other unit tests and try to fix it.

kdeldycke: > I agree. The sentinel seems purely intended for internal tracking detection, I don't see a situation where we shouldn't be concerned with whether the value is the sentinel. I also don't see a situation where the sentinel should be passed or received by the user.

@davidism Sorry but I am tired and cannot parse these sentences! 😅 So are you for or against making `UNSET` falsy? 😄

kdeldycke: > ```python
> @click.option("--json", "output_format", flag_value="json", callback=_get_output_format)
> @click.option("--console", "output_format", flag_value="console", callback=_get_output_format)
> ```

@edgarrmondragon Hmmm. Let me play a bit with your example. Maybe there's something here...

davidism: I'm saying that we should:

* Never expose the sentinel to the user. They should not know it exists, or be able to interact with it.
* Always explicitly check for the sentinel, rather than relying on it being falsey. Similarly, I prefer explicitly checking `is None` rather than relying on none being falsey.

Rowlando13:
@kdeldycke We usually leave the commits as is so people can follow your train of thought. If you were going clean up the commits, so they are easier to follow that is also okay. I will merge stable into main prior to you switching branches to make the changeover easy.

CHANGES.rst
```diff
@@ -5,19 +5,29 @@ Version 8.2.x
 
 Unreleased
```
Rowlando13: - rework flag value passing by disentangling use of default. Formerly defaults for ``flag_value`` were set by passing ``default=True``, now set by passing ``default='desired_value'``. **This change breaks all default settings for flag value type flags.**  In all other places in code, ``default`` takes on the default value. Fixing this inconsistency allows the fixing of the linked issues and supports a more consistent api.  :issue:`1992`   :issue:`2012` :issue:`2514`  :issue:`2610`  :issue:`3024`   :pr:`3030`

src/click/utils.py
```diff
@@ -29,6 +30,46 @@
 R = t.TypeVar("R")
```
Rowlando13: Move to _utils.py. Remove ```__bool__```.

tests/test_arguments.py
```diff
@@ -251,13 +245,37 @@ def cmd(arg):
     assert "Missing argument 'ARG'." in result.output
```
Rowlando13, line 247: Add comment for what this is testing so you can just look at the comment without having to read whole test.

src/click/utils.py
kdeldycke: > > ```python
> > @click.option("--json", "output_format", flag_value="json", callback=_get_output_format)
> > @click.option("--console", "output_format", flag_value="console", callback=_get_output_format)
> > ```
> 
> @edgarrmondragon Hmmm. Let me play a bit with your example. Maybe there's something here...

The use of callback was not covered by unittests. So I added one at: https://github.com/pallets/click/blob/763eb5977b9a4c59d5f3975a6fa193a720acf870/tests/test_options.py#L1365-L1396

I think it is OK to let the `UNSET` sentinel reach the `callback`.  Which let the CLI developer decide how to handle the different cases of `None`, `UNSET`, or any other value. Callbacks are used for a lot of funky use-cases, as an escape-hatch from Click default pipeline. So we can keep it under the responsibility of the CLI developer.

src/click/core.py
```diff
@@ -2352,6 +2352,9 @@ def process_value(self, ctx: Context, value: t.Any) -> t.Any:
             raise MissingParameter(ctx=ctx, param=self)
 
         if self.callback is not None:
+            # The callback is receiving the default value as-is, which let the
+            # developer decide how to handle the different cases of None, UNSET, or any
+            #  other value.
```
Rowlando13: Make sense. So this exposes to UNSET in the callback? If so, it looks like we will have to expose UNSET as public. Is this [this callback](https://click.palletsprojects.com/en/stable/advanced/#callbacks-for-validationl). @kdeldycke @davidism

src/click/utils.py
kdeldycke: > * Never expose the sentinel to the user. They should not know it exists, or be able to interact with it.
> * Always explicitly check for the sentinel, rather than relying on it being falsey. Similarly, I prefer explicitly checking `is None` rather than relying on none being falsey.

OK, I backtracked on making `UNSET` falsy and removed that code: https://github.com/pallets/click/pull/3030/commits/f4b5b5219759fe74826d9df48bd16449d17bc466

src/click/core.py
kdeldycke: > Make sense. So this exposes to UNSET in the callback? If so, it looks like we will have to expose UNSET as public. Is this [this callback](https://click.palletsprojects.com/en/stable/advanced/#callbacks-for-validationl). @kdeldycke @davidism

That's another choice to make. It depends on how we want to hide `UNSET`. Should we hide it to **CLI users**? Absolutely.

Should we hide it to **CLI developers**? I don't think so. Developers are responsible, so if they want to do funky stuff with callbacks out of the default Click pipeline, let them do it.

CHANGES.rst
kdeldycke: Done: https://github.com/pallets/click/pull/3030/commits/7b809cfe57b1189df79a9a3faf30de6e0f4e86aa

src/click/utils.py
kdeldycke: `__bool__` removed in: https://github.com/pallets/click/pull/3030/commits/f4b5b5219759fe74826d9df48bd16449d17bc466

`UNSET` moved to `_utils` in: https://github.com/pallets/click/pull/3030/commits/d8c2d433d66c0cf434384128c4b38fca03e23718

src/click/core.py
kdeldycke: Just documented the `UNSET` on callback in: https://github.com/pallets/click/pull/3030/commits/e6870a68090926a0c49b79243fef9acf6550809e

tests/test_arguments.py
kdeldycke: Done in https://github.com/pallets/click/pull/3030/commits/ae20270b8764e67fccdf6716e2b852febab9d05c

kdeldycke:
> @kdeldycke We usually leave the commits as is so people can follow your train of thought. If you were going clean up the commits, so they are easier to follow that is also okay. I will merge stable into main prior to you switching branches to make the changeover easy.

As I was hinting in my previous comment, this work is more playing Whac-A-Mole than pure green-field engineering! 🙃 So my workflow is more like hack, hack, hack, fix, document, hack, unittest, remove stuff, shit, backtrack, hack, cleanup. 😅

I'll try to rework the history and reduce it from the 43 commits monster (and growing) it is right now. But I fear I'll just get frustrated fast as everything is influencing everything (i.e. I doubt I can find a couple of intermediary states in which tests are all passing). 😖 So I'll squash it and call it a day. 😐

Rowlando13:
@kdeldycke sounds good.

Rowlando13:
@davidism Do you think you will have time to review part or all of this in the next few weeks? If not, @AndreasBackx would you mind reviewing?

docs/options.md
```diff
@@ -314,15 +314,15 @@ If you want to define an alias for the second option only, then you will need to
...
+To have an flag pass a value to the underlying function set `flag_value`. This automatically sets `is_flag=True`. To influence the value-less behavior, force its value with `default='upper'`. Setting flag values can be used to create patterns like this:
 
 ```{eval-rst}
 .. click:example::
 
     import sys
 
     @click.command()
-    @click.option('--upper', 'transformation', flag_value='upper', default=True)
+    @click.option('--upper', 'transformation', flag_value='upper', default='upper')
```
davidism: Is there no way to continue supporting `default=True` here? It seems like we could say `if flag_value is not None and default is True: default = flag_value`.

davidism:
Working through reviewing this now.

src/click/core.py
```diff
@@ -1024,7 +1025,7 @@ def get_help_option_names(self, ctx: Context) -> list[str]:
     def get_help_option(self, ctx: Context) -> Option | None:
         """Returns the help option object.
 
-        Skipped if :attr:`add_help_option` is ``False``.
+        Skipped if :attr:`add_help_option` is `False`.
```
davidism: ``code`` double backticks are the correct syntax for code in rST, which the docstrings must still be written in. I think I set the default role so that `code` single backticks also render as code. In general, leave cleaning that up for another time, unless you're rewriting the sentence it's in as well. Having it peppered throughout the PR makes it a bit harder to pay attention to the important changes.

(This is just a comment, you don't have to change it back unless you want to.)

src/click/core.py
```diff
@@ -2015,14 +2016,16 @@ class Parameter:
     :param is_eager: eager values are processed before non eager ones.  This
                      should not be set for arguments or it will inverse the
                      order of processing.
-    :param envvar: a string or list of strings that are environment variables
-                   that should be checked.
+    :param envvar: environment variable(s) that are used to provide a default value for
```
davidism, line 2019: Whenever I rewrite `param` blocks, I always re-indent to four spaces on subsequent lines, it's a lot easier to format. See the next param, for example.

src/click/core.py
```diff
@@ -2175,6 +2144,13 @@ def to_info_dict(self) -> dict[str, t.Any]:
         Use :meth:`click.Context.to_info_dict` to traverse the entire
         CLI structure.
 
+        .. versionchanged:: 8.3.0
+            Returns `None` for the :attr:`default` if it is :attr:`UNSET`.
```
davidism: Don't mention `UNSET` here. "Default is shown as `None` if it was not set." should be fine for the change entry.

Or perhaps we could add a `has_default` key so documentation using this method can determine whether the `None` is because it's actually `None` or not set.

src/click/core.py
```diff
@@ -2175,6 +2144,13 @@ def to_info_dict(self) -> dict[str, t.Any]:
...
+
+            We explicitly hide the :attr:`UNSET` value to the user, as we choose to
```
davidism: I don't think we need this level of explanation in the change entry. Either suggestion above is probably enough.

src/click/core.py
```diff
@@ -2268,29 +2244,51 @@ def add_to_parser(self, parser: _OptionParser, ctx: Context) -> None:
...
-            value = self.value_from_envvar(ctx)
-            source = ParameterSource.ENVIRONMENT
+        If the parser did not produce a value from user input, the value is either
+        sourced from the environment variable, the default map, or the parameter's
+        default value. In that order of precedence.
 
-        if value is None:
-            value = ctx.lookup_default(self.name)  # type: ignore
-            source = ParameterSource.DEFAULT_MAP
+        If no value is found, the value is returned as :attr:`UNSET`.
```
davidism: I know this is a somewhat internal method, but we probably shouldn't mention `UNSET` here. "If no value is found, an internal sentinel value is returned." perhaps?

src/click/core.py
```diff
@@ -2268,29 +2244,51 @@ def add_to_parser(self, parser: _OptionParser, ctx: Context) -> None:
...
+        """
+        # If the value is set, it means the parser produced a value from user input.
+        value = opts.get(self.name, UNSET)  # type: ignore
+        source = (
+            ParameterSource.COMMANDLINE
+            if value is not UNSET
+            else ParameterSource.DEFAULT
```
davidism, line 2272: The comment doesn't explain this check. Should it say something like "Source is command line if parser produced a value, otherwise it's unset and uses the default default value."

src/click/termui.py
```diff
@@ -220,6 +220,8 @@ def confirm(
     .. versionadded:: 4.0
         Added the ``err`` parameter.
     """
+    assert default in (True, False, None), "default must be True, False, or None"
```
davidism: This is already handled by the type annotation.

src/click/core.py
```diff
@@ -2346,6 +2354,22 @@ def value_is_missing(self, value: t.Any) -> bool:
...
+
+        1. Type cast the value using :meth:`type_cast_value`.
+        2. Check if the value is missing (see: :meth:`value_is_missing`), and raise
+           :exc:`MissingParameter` if it is required.
+        3. If a :attr:`callback` is set, call it to have the value replaced buy the
+           result of the callback.
+
+        .. versionchanged:: 8.3
+
+            The :attr:`callback` is now receiving the :attr:`UNSET` sentinel if the
```
davidism: In general, change entries need to have less explanation. Usage information belongs in the main docstring. In the case of anything that mentions `UNSET`, perhaps it should be confined to an "Advanced Method Customization" page, instead of in the API docs.

src/click/core.py
```diff
@@ -2357,71 +2381,108 @@ def process_value(self, ctx: Context, value: t.Any) -> t.Any:
...
+                    echo(style(message, fg="red"), err=True)
+
             try:
                 value = self.process_value(ctx, value)
             except Exception:
                 if not ctx.resilient_parsing:
                     raise
 
-                value = None
+                # Ignore values that we have a hard time processing.
```
davidism: What does "hard time" mean here?

src/click/core.py
```diff
@@ -2127,40 +2130,6 @@ def __init__(
                     f" type {self.type!r}, but it was {nargs}."
                 )
 
-            # Skip no default or callable default.
```
davidism, line 2130: I'm having trouble finding where this logic moved to. Is it checked at processing time rather than creation time now? Or was it unneeded and removed?

src/click/core.py
```diff
@@ -2357,71 +2381,108 @@ def process_value(self, ctx: Context, value: t.Any) -> t.Any:
...
+        .. caution::
+
+            The raw value extracted from the environment is not normalized and is
+            returned as-is. Any normalization or reconciliation is performed later by
+            the :class:`Parameter`'s :attr:`type`.
+        """
+        if not self.envvar:
+            return None
+
+        envvars = [self.envvar] if isinstance(self.envvar, str) else self.envvar
```
davidism: I know the previous code looks a little verbose, but it avoids constructing a list in the much, much more common case of a single key. Avoiding extra work as much as reasonably possible is particularly important in Click, where lowering startup and processing time affects the user experience.

I can't say for certain whether this affects performance meaningfully, but easier to save these sorts of refactors for dedicated PRs.

src/click/core.py
```diff
@@ -2357,71 +2381,108 @@ def process_value(self, ctx: Context, value: t.Any) -> t.Any:
...
+    def value_from_envvar(self, ctx: Context) -> str | cabc.Sequence[str] | None:
+        """Process the raw environment variable string for this parameter.
 
-        if rv is not None and self.nargs != 1:
-            rv = self.type.split_envvar_value(rv)
+        Returns the string as-is or splits it into a sequence of strings if the
+        parameter is expecting multiple values (i.e. its :attr:`nargs` property is set
+        to a value other than ``1``).
+        """
+        raw_value = self.resolve_envvar_value(ctx)
```
davidism: Don't refactor variable names unless you're actually rewriting the code, otherwise save it for a dedicated PR. It makes it harder to evaluate what code is actually changing and needs review.

`rv` typically stands for `return_value`, not `raw_value`. I agree it's not a clear variable, and I do try to clean it up when I'm rewriting code. But I'd go for `env_value`, or maybe just `value`.

src/click/core.py
```diff
@@ -2357,71 +2381,108 @@ def process_value(self, ctx: Context, value: t.Any) -> t.Any:
...
                 if not ctx.resilient_parsing:
                     raise
 
-                value = None
+                # Ignore values that we have a hard time processing.
+                value = UNSET
 
-        if self.expose_value:
-            ctx.params[self.name] = value  # type: ignore
+        if self.expose_value and self.name not in ctx.params:
```
davidism: Why skip if `name` is already in `params`?

src/click/core.py
```diff
@@ -2967,105 +3030,129 @@ def prompt_for_value(self, ctx: Context) -> t.Any:
...
-            option's type should happen later.
+        """:class:`Option` resolves its environment variable the same way as
+        :func:`Parameter.resolve_envvar_value`, but it also supports
+        :attr:`Context.auto_envvar_prefix`. If we could not find an environment from
+        the :attr:`envvar` property, we fallback on :attr:`Context.auto_envvar_prefix`
+        to build dynamiccaly the environment variable name using the
+        :python:`{ctx.auto_envvar_prefix}_{self.name.upper()}` template.
         """
-        rv = super().resolve_envvar_value(ctx)
+        raw_value = super().resolve_envvar_value(ctx)
```
davidism: Again, don't refactor variable names unless the line of code is also changed.

src/click/core.py
```diff
@@ -2941,12 +2996,18 @@ def prompt_for_value(self, ctx: Context) -> t.Any:
...
-        # If this is a prompt for a flag we need to handle this
-        # differently.
+        # A boolean flag can use a simplified [y/n] confirmation prompt.
         if self.is_bool_flag:
+            # If we have no boolean default, we force the user to explicitly provide
+            # one.
+            if default in (UNSET, None):
+                default = None
+            else:
+                default = bool(default)
```
davidism, line 3080: When would this be the case? They shouldn't be entering a non-bool default for a bool flag.

tests/test_options.py
```diff
@@ -170,38 +172,120 @@ def cli(message):
...
+        ([], True, 1, None, ()),
+        ([], True, 1, UNSET, ()),
+        # Number of values are kept as-is in the default.
+        ([], True, 1, [1], (1,)),
+        ([], True, 1, [1, 2], (1, 2)),
+        ([], True, 1, [1, 2, 3], (1, 2, 3)),
+        ([], True, 1, [1.1, 2.2], (1.1, 2.2)),
+        ([], True, 1, ["1", "2"], ("1", "2")),
+        ([], True, 1, [None, None], (None, None)),
+        # XXX Why items of the defaults are converted to strings and not left as
```
davidism: Is this something you figured out? Should we open another issue?

tests/test_options.py
```diff
@@ -1455,31 +1688,252 @@ def cli(one, two):
         runner.invoke(cli, [])
 
 
+def test_custom_type_click_class_flag_value(runner):
+    """A reproduction of
```
davidism: Describe what the test is doing in the docstring, the link should be extra context only.

tests/test_termui.py
```diff
@@ -485,3 +487,196 @@ def cmd(arg1):
...
+    ("opt_decls", "opt_params", "args", "prompt", "input", "expected"),
+    # Boolean flag prompt cases.
+    [("--flag/--no-flag", *case_params) for case_params in BOOLEAN_FLAG_PROMPT_CASES]
+    # Non-boolean flag prompt cases.
+    + [("--flag", *case_params) for case_params in FLAG_VALUE_PROMPT_CASES],
+)
+def test_flag_value_prompt(
+    runner, opt_decls, opt_params, args, prompt, input, expected
+):
+    """Covers concerns raised in issue https://github.com/pallets/click/issues/1992."""
```
davidism: Describe what the test does in the docstring, the link should be extra context only.

tests/test_options.py
```diff
@@ -1100,188 +1232,293 @@ def test_type_from_flag_value():
...
+        (UNSET, [], "A real UNSET"),
+        ("random", [], "random"),
+        (None, ["--js"], "js"),
+        (None, ["--xml"], "xml"),
+    ],
+)
+def test_flag_value_callback(runner, default, args, expected):
+    """
+    Reproduction of the issue reported in
+    https://github.com/pallets/click/pull/3030#discussion_r2271571819
```
davidism, line 1498: Describe what the test does in the docstring, the link should only be extra context.

tests/test_options.py
```diff
@@ -1455,31 +1688,252 @@ def cli(one, two):
...
+class Class2:
+    pass
+
+
 @pytest.mark.parametrize(
-    ("multiple", "nargs", "default", "expected_message"),
+    ("opt_params", "args", "expected"),
     [
+        # Reproduction of cases from
+        # https://github.com/pallets/click/issues/3024#issuecomment-3146480714
```
davidism: Describe what's being tested, not only a link.

tests/test_options.py
```diff
@@ -1455,31 +1688,252 @@ def cli(one, two):
...
+            [],
+            EngineType.OSS,
+        ),
+        (
+            {"type": EngineType, "is_flag": True, "default": EngineType.OSS},
+            ["--pro"],
+            EngineType.OSS,
+        ),
+        # Reproduction of cases from
+        # https://github.com/pallets/click/issues/2012#issuecomment-892437060
```
davidism: Describe what's being tested, not only a link.

tests/test_options.py
```diff
@@ -1455,31 +1688,252 @@ def cli(one, two):
...
+        assert re.match(expected, result.output)
+    else:
+        assert result.output == repr(expected)
+
+
+@pytest.mark.parametrize(
+    ("opt1_params", "opt2_params", "args", "expected"),
+    [
+        ###
+        ### Reproduces https://github.com/pallets/click/issues/3024#issuecomment-3146508536
```
davidism: Describe what's being tested, not only a link.

tests/test_options.py
```diff
@@ -1455,31 +1688,252 @@ def cli(one, two):
...
+            EngineType.OSS,
+        ),
+        (
+            {"flag_value": EngineType.OSS, "type": UNPROCESSED},
+            {"flag_value": EngineType.PRO, "type": UNPROCESSED},
+            ["--opt2"],
+            EngineType.PRO,
+        ),
+        ###
+        ### Reproduces https://github.com/pallets/click/issues/2012#issue-946471049
```
davidism: Describe what's being tested, not only a link.

tests/test_options.py
```diff
@@ -1455,31 +1688,252 @@ def cli(one, two):
...
+            re.compile(r"<test_options.Class1 object at 0x[0-9A-Fa-f]+>"),
+        ),
+        (
+            {"flag_value": Class1, "type": UNPROCESSED, "default": Class2},
+            {"flag_value": Class2, "type": UNPROCESSED},
+            [],
+            re.compile(r"<test_options.Class2 object at 0x[0-9A-Fa-f]+>"),
+        ),
+        ###
+        ### Reproduces https://github.com/pallets/click/issues/2012#issuecomment-892437060
```
davidism: Describe what's being tested, not only a link.

tests/test_options.py
```diff
@@ -1455,31 +1688,252 @@ def cli(one, two):
...
+
+    result = runner.invoke(cli, args)
+    if isinstance(expected, re.Pattern):
+        assert re.match(expected, result.output)
+    else:
+        assert result.output == repr(expected)
+
+
+def test_custom_type_frozenset_flag_value(runner):
+    """A reproduction of https://github.com/pallets/click/issues/2610"""
```
davidism: Describe what's being tested, not only a link.

tests/test_basic.py
```diff
@@ -261,6 +276,27 @@ def cli(flag):
     assert result.output == expect
 
 
+@pytest.mark.parametrize(
+    ("default", "args", "expect"),
+    (
+        # See: https://github.com/pallets/click/issues/3024#issuecomment-3146199461
```
davidism: Describe what's being tested, not only a link.

davidism:
I was trying to figure out exactly what you did to address #2012, and I think you just added tests for what's already happening, didn't change anything? Can you confirm how you addressed it.

davidism:
To be clear, despite the large number of review comments, the overall change and tests look good.

src/click/core.py
kdeldycke: These changes are based on ancient knowledge I acquired maybe 15 years ago. In my recollection simple ticks were used in rST for literals, and double-ticks for code. I just made this update out of habits. 😅

I just reverted these changes to reduce the noise and stick with double-ticks: https://github.com/pallets/click/pull/3030/commits/8645e70266479a4c3abd9b821eca608971a8c051

src/click/core.py
kdeldycke: Fixed in: https://github.com/pallets/click/pull/3030/commits/c442c05c315e61b38932e6d25a4e12dfc533882e

src/click/core.py
kdeldycke: Addressed in https://github.com/pallets/click/pull/3030/commits/634549fc180a4c7655eafedce5b84a9b140b0a35

src/click/core.py
kdeldycke: Ok, I reverted to the original structure of the code in https://github.com/pallets/click/pull/3030/commits/44c2bea54ef1f4d153138be2db9537a7f442d461

src/click/core.py
kdeldycke: OK, I moved the detailed explanation to a comment to keep the change entry concise: https://github.com/pallets/click/pull/3030/commits/cf0e5242b3adfbfbeaff64b24a398b57d15ae94b

src/click/core.py
kdeldycke: Done in: https://github.com/pallets/click/pull/3030/commits/60feda3c9240dcb7f362874bb6ae709eea151983

src/click/core.py
kdeldycke: Ah yes. That's what I tried to imply by the comment above. But you are right, let's make this more explicit. Fixed in https://github.com/pallets/click/pull/3030/commits/b39590fd8560e3ee6bec411ddd575761bdb1a113

src/click/core.py
kdeldycke: Addressed in https://github.com/pallets/click/pull/3030/commits/f561f1880e6bbc8f2d199d50f08dfa2002090596

src/click/core.py
kdeldycke: I was just trying to comment on the presence of the `except Exception` statement here, but could not infer @mitsuhiko intention from his https://github.com/pallets/click/commit/4fa40cf5051f249829600f1b731822729cd50e56 commit 11 years ago! 😅

kdeldycke: Do you want me to remove it?

src/click/core.py
kdeldycke: Removing that `and self.name not in ctx.params` condition here, we make all the tests with dual options fail:
```
FAILED tests/test_basic.py::test_flag_value_dual_options[True-args2-True] - assert 'None' == "'True'"
FAILED tests/test_basic.py::test_flag_value_dual_options[upper-args3-upper] - assert 'None' == "'upper'"
FAILED tests/test_options.py::test_flag_value_dual_callback[None-args0-A real None] - assert "'A real UNSET'" == "'A real None'"
FAILED tests/test_options.py::test_flag_value_dual_callback[random-args2-random] - assert "'A real UNSET'" == "'random'"
FAILED tests/test_options.py::test_custom_type_flag_value_dual_options[opt1_params3-opt2_params3-args3-True] - assert 'None' == "'True'"
FAILED tests/test_options.py::test_custom_type_flag_value_dual_options[opt1_params6-opt2_params6-args6-True] - assert 'None' == "'True'"
FAILED tests/test_options.py::test_custom_type_flag_value_dual_options[opt1_params8-opt2_params8-args8-True] - AssertionError: assert 'None' == 'True'
FAILED tests/test_options.py::test_custom_type_flag_value_dual_options[opt1_params11-opt2_params11-args11-'<test_options.Class1 object at 0x[0-9A-Fa-f]+>'] - assert None
FAILED tests/test_options.py::test_custom_type_flag_value_dual_options[opt1_params12-opt2_params12-args12-'<test_options.Class2 object at 0x[0-9A-Fa-f]+>'] - assert None
FAILED tests/test_options.py::test_custom_type_flag_value_dual_options[opt1_params13-opt2_params13-args13-<test_options.Class1 object at 0x[0-9A-Fa-f]+>] - AssertionError: assert None
FAILED tests/test_options.py::test_custom_type_flag_value_dual_options[opt1_params14-opt2_params14-args14-<test_options.Class2 object at 0x[0-9A-Fa-f]+>] - AssertionError: assert None
FAILED tests/test_options.py::test_custom_type_flag_value_dual_options[opt1_params15-opt2_params15-args15-1] - AssertionError: assert 'None' == '1'
FAILED tests/test_options.py::test_custom_type_flag_value_dual_options[opt1_params16-opt2_params16-args16-1] - AssertionError: assert 'None' == '1'
```

That's because this condition is preventing parameters competing for the same variable name to override each other.

I updated the explanation in https://github.com/pallets/click/pull/3030/commits/2d1ceffe936a38ff2bfe8102f3a2188ba72e1e6f

src/click/core.py
kdeldycke: I finally reworked the documentation of that method in https://github.com/pallets/click/pull/3030/commits/2d1ceffe936a38ff2bfe8102f3a2188ba72e1e6f

src/click/termui.py
kdeldycke: Ah yes, good catch. Just a left-over of my work. Removed in https://github.com/pallets/click/pull/3030/commits/11bfb2c9e246b91ea35ed75e23c9aad80240a3d7

tests/test_basic.py
kdeldycke: Done in: https://github.com/pallets/click/pull/3030/commits/c498ced0e7b595880b70a81767ae86902714ee1c

tests/test_options.py
kdeldycke: Done in: https://github.com/pallets/click/pull/3030/commits/c498ced0e7b595880b70a81767ae86902714ee1c

tests/test_options.py
kdeldycke: Done in: https://github.com/pallets/click/pull/3030/commits/c498ced0e7b595880b70a81767ae86902714ee1c

src/click/core.py
```diff
@@ -2337,7 +2337,15 @@ def convert(value: t.Any) -> t.Any:  # tuple[t.Any, ...]
         return convert(value)
 
     def value_is_missing(self, value: t.Any) -> bool:
-        if value is None:
+        """
```
davidism: Start docstrings on the same line as the quote, not on the next line.

davidism: Any methods that were not previously documented did not show up in the API docs, so were not considered public. Add `:meta private:` at the bottom of each of these method's docstrings, until we determine what we want to make public.

src/click/core.py
```diff
@@ -2346,6 +2354,19 @@ def value_is_missing(self, value: t.Any) -> bool:
...
+
+        1. Type cast the value using :meth:`type_cast_value`.
+        2. Check if the value is missing (see: :meth:`value_is_missing`), and raise
+           :exc:`MissingParameter` if it is required.
+        3. If a :attr:`callback` is set, call it to have the value replaced by the
+           result of the callback. The callback is receiving the value as-is, which let
+           the developer decide how to handle the different cases.
+
+        .. versionchanged:: 8.3
+            The :attr:`callback` gets an internal sentinel if the parameter was not set
```
davidism: This doesn't seem correct? User code should not be interacting with `UNSET`, and command/param callbacks are user code.

tests/test_options.py
kdeldycke: Fixed in https://github.com/pallets/click/pull/3030/commits/354788eaba85d2ef405b300186a3752c2bc4dbaa

tests/test_options.py
kdeldycke: Fixed in https://github.com/pallets/click/pull/3030/commits/354788eaba85d2ef405b300186a3752c2bc4dbaa

tests/test_options.py
kdeldycke: Done in: https://github.com/pallets/click/pull/3030/commits/7b84990ead4a663e4b57d5cdb83eca1e508c3d78

tests/test_options.py
kdeldycke: Done in: https://github.com/pallets/click/pull/3030/commits/7b84990ead4a663e4b57d5cdb83eca1e508c3d78

tests/test_options.py
kdeldycke: Done in: https://github.com/pallets/click/pull/3030/commits/7b84990ead4a663e4b57d5cdb83eca1e508c3d78

tests/test_options.py
kdeldycke: Done: https://github.com/pallets/click/pull/3030/commits/472bf383bcf403487b939242ddcd0bbb54257bb3

src/click/core.py
kdeldycke: Done in: https://github.com/pallets/click/pull/3030/commits/e9e50023f87474939d99596a90b9c8d4335c03c5

tests/test_termui.py
kdeldycke: Dene: https://github.com/pallets/click/pull/3030/commits/9f7d99100aa8b1b81d8f011de679557e19e0c3d7

tests/test_options.py
davidism: There's a few other places with `# XXX` notes, are these things that still need to be addressed?

src/click/core.py
kdeldycke: Done in: https://github.com/pallets/click/pull/3030/commits/0acca45214a66eefe50716ecb0bcfcefa8fcc6eb

kdeldycke: Done in: https://github.com/pallets/click/pull/3030/commits/0acca45214a66eefe50716ecb0bcfcefa8fcc6eb

Rowlando13:
@kdeldycke Once you address all of @davidism comments. I will merge!

src/click/core.py
kdeldycke: Re-adding these checks, the following tests fails:
```
FAILED tests/test_arguments.py::test_nargs_star - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_tup - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_tup_composite[opts0] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_tup_composite[opts1] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_tup_composite[opts2] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_tup_composite[opts3] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_envvar[2--None] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_envvar[2-a-Takes 2 values but 1 was given.] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_envvar[2-a b-expect2] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_envvar[2-a b c-Takes 2 values but 3 were given.] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_envvar[-1-a b c-expect4] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_envvar[-1--expect5] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_envvar_only_if_values_empty - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_empty_nargs - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_eat_options - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_star_ordering - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_nargs_specified_plus_star_ordering - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[Yo-Error: Invalid value for '[A]...': Value must be an iterable.] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[-Error: Invalid value for '[A]...': Value must be an iterable.] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[True-Error: Invalid value for '[A]...': Value must be an iterable.] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[False-Error: Invalid value for '[A]...': Value must be an iterable.] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[12-Error: Invalid value for '[A]...': Value must be an iterable.] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[7.9-Error: Invalid value for '[A]...': Value must be an iterable.] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default6-Error: Invalid value for '[A]...': Takes 2 values but 0 were given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[Sentinel.UNSET-Error: Missing argument 'A...'.] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default8-Error: Invalid value for '[A]...': Takes 2 values but 0 were given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default9-Error: Invalid value for '[A]...': Takes 2 values but 1 was given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default10-Error: Invalid value for '[A]...': Takes 2 values but 3 were given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default11-Error: Invalid value for '[A]...': Takes 2 values but 0 were given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default12-Error: Invalid value for '[A]...': Takes 2 values but 1 was given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default13-Error: Invalid value for '[A]...': Takes 2 values but 3 were given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default14-Error: Invalid value for '[A]...': Takes 2 values but 0 were given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default15-Error: Invalid value for '[A]...': Takes 2 values but 1 was given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default16-Error: Invalid value for '[A]...': Takes 2 values but 3 were given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default17-Error: Invalid value for '[A]...': Takes 2 values but 0 were given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default18-Error: Invalid value for '[A]...': Takes 2 values but 1 was given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default19-Error: Invalid value for '[A]...': Takes 2 values but 3 were given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default20-Error: Invalid value for '[A]...': Takes 2 values but 0 were given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default21-Error: Invalid value for '[A]...': Takes 2 values but 1 was given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_arguments.py::test_bad_defaults_for_nargs[default22-Error: Invalid value for '[A]...': Takes 2 values but 3 were given.] - ValueError: 'default' length must match nargs=2.
FAILED tests/test_chain.py::test_group_arg_behavior - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_commands.py::test_other_command_invoke_with_defaults - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_commands.py::test_unprocessed_options - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_defaults.py::test_multiple_flag_default - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_invalid_nargs - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_nargs_tup_composite_mult - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_multiple_required - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_good_defaults_for_multiple[args7-True-1-Sentinel.UNSET-expected7] - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_good_defaults_for_multiple[args24-True-2-Sentinel.UNSET-expected24] - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_bad_defaults_for_multiple[True-1-Yo-Error: Invalid value for '-a': Value must be an iterable.] - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_bad_defaults_for_multiple[True-1--Error: Invalid value for '-a': Value must be an iterable.] - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_bad_defaults_for_multiple[True-1-True-Error: Invalid value for '-a': Value must be an iterable.] - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_bad_defaults_for_multiple[True-1-False-Error: Invalid value for '-a': Value must be an iterable.] - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_bad_defaults_for_multiple[True-1-12-Error: Invalid value for '-a': Value must be an iterable.] - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_bad_defaults_for_multiple[True-1-7.9-Error: Invalid value for '-a': Value must be an iterable.] - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_bad_defaults_for_multiple[False-2-42-Error: Invalid value for '-a': Value must be an iterable.] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_bad_defaults_for_multiple[True-2-default7-Error: Invalid value for '-a': Value must be an iterable.] - ValueError: 'default' must be a list of lists when 'multiple' is true and 'nargs' != 1.
FAILED tests/test_options.py::test_bad_defaults_for_multiple[True-2-default8-Error: Invalid value for '-a': Value must be an iterable.] - ValueError: 'default' must be a list of lists when 'multiple' is true and 'nargs' != 1.
FAILED tests/test_options.py::test_bad_defaults_for_multiple[True-2-default9-Error: Invalid value for '-a': Value must be an iterable.] - ValueError: 'default' must be a list of lists when 'multiple' is true and 'nargs' != 1.
FAILED tests/test_options.py::test_multiple_envvar - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_nargs_envvar - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_do_not_show_default_empty_multiple - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_option_with_optional_value[None-expect0] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args1-expect1] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args2-expect2] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args3-expect3] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args4-expect4] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args5-expect5] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args6-expect6] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args7-expect7] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args8-expect8] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args9-expect9] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args10-expect10] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args11-expect11] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args12-expect12] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_option_with_optional_value[args13-expect13] - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_options.py::test_multiple_option_with_optional_value - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_options.py::test_invalid_flag_combinations[kwargs0-'count' is not valid with 'multiple'.] - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_shell_completion.py::test_flag_option_with_nargs_option - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_shell_completion.py::test_option_multiple - ValueError: 'default' must be a list when 'multiple' is true.
FAILED tests/test_shell_completion.py::test_option_nargs - ValueError: 'default' must be a list when 'nargs' != 1.
FAILED tests/test_shell_completion.py::test_argument_nargs - ValueError: 'default' must be a list when 'nargs' != 1.
```

That logic at instantiation is trying to replicate a bunch of implied constraints that is already embedded and disseminated into the processing pipeline. Aligning this logic to the pipeline is a lot of work and consist in emulating the pipeline itself. I don't think this instantiation checks are of any value compared to the maintenance cost. Especially since these statements are under a `__debug__` conditional and all these cases are now covered by unittests.

docs/options.md
kdeldycke: There's probably a way to do it by sparkling the code with conditionals here and there.

But I fear it will make the documentation even harder to read. I'll need to add a bunch of these kind of `.. warning` admonitions everywhere flag value are mentioned:

> [WARNING]
> To support legacy usage of the option API, an option specifying a `flag_value` is not allowed to have its `default` explicitly sets to `True`. Because in that case, `default=True` will be replaced by `default=flag_value`.

kdeldycke: Ok, I managed to keep the legacy behavior in https://github.com/pallets/click/pull/3030/commits/06847da9126a99c88bf347adf213d33bfe4e4af2

tests/test_options.py
kdeldycke: > There's a few other places with `# XXX` notes, are these things that still need to be addressed?

Partly addressed in https://github.com/pallets/click/pull/3030/commits/06847da9126a99c88bf347adf213d33bfe4e4af2#diff-e9b43a3662176468b449eee04c777e29d67212bfc0dc0ca42a03b32d970b0dea

kdeldycke: > Is this something you figured out? Should we open another issue?

That was a mystery I solved later. Just updated the documentation and tests in: https://github.com/pallets/click/pull/3030/commits/bb2a1d952cd79151dc8ce859ac65230ff95278d3

src/click/core.py
kdeldycke: > When would this be the case? They shouldn't be entering a non-bool default for a bool flag.

Nothing prevent you to declare an option that is auto-detected as a boolean flag, is allow to prompt but still declare a non-boolean default:

```python
@option("--flag/--no-flag", "prompt": True, "default": "foo")

```

This is the case covered by the `tests/test_termui.py::test_flag_value_prompt[--flag/--no-flag-opt_params11-args11-[Y/n]--]` case.

I.e. the one where:

- `opt_params = {"prompt": True, "default": False}`
- `args = []`
- `prompt = "[Y/n]"`
- `input = ""`
- `expected = True`

With this casting, we expect the prompt to show as `Flag [Y/n]:`, and not entering a value for input, makes the option takes `True` as the default is set to a truthy value (`foo`). This is required to align the prompt behavior (defaulting to `[Y/n]`, i.e. `True`) with the value of the default.

See the set of cases described by:
```python
    # Random string default is treated as a truthy value, so it prompts with [Y/n].
    ({"prompt": True, "default": "foo"}, [], "[Y/n]", "", True),
    ({"prompt": True, "default": "foo"}, [], "[Y/n]", "y", True),
    ({"prompt": True, "default": "foo"}, [], "[Y/n]", "n", False),
```

Without this condition, we get this error:
```shell-session
❯ uv run pytest             
(...)
________ test_flag_value_prompt[--flag/--no-flag-opt_params11-args11-[Y/n]--True] ________

runner = <click.testing.CliRunner object at 0x10500ead0>, opt_decls = '--flag/--no-flag', opt_params = {'default': 'foo', 'prompt': True}, args = [], prompt = '[Y/n]', input = '', expected = True

    @pytest.mark.parametrize(
        ("opt_decls", "opt_params", "args", "prompt", "input", "expected"),
        # Boolean flag prompt cases.
        [("--flag/--no-flag", *case_params) for case_params in BOOLEAN_FLAG_PROMPT_CASES]
        # Non-boolean flag prompt cases.
        + [("--flag", *case_params) for case_params in FLAG_VALUE_PROMPT_CASES],
    )
    def test_flag_value_prompt(
        runner, opt_decls, opt_params, args, prompt, input, expected
    ):
        """Check how flag value are prompted and handled by all combinations of
        ``prompt``, ``default``, and ``flag_value`` parameters.
    
        Covers concerns raised in issue https://github.com/pallets/click/issues/1992.
        """
    
        @click.command()
        @click.option(opt_decls, **opt_params)
        def cli(flag):
            click.echo(repr(flag))
    
        invoke_options = {"standalone_mode": False}
        if input is not None:
            assert isinstance(input, str)
            invoke_options["input"] = f"{input}\n"
    
        result = runner.invoke(cli, args, **invoke_options)
    
        if expected in (MissingParameter, BadParameter):
            assert isinstance(result.exception, expected)
            assert not result.output
            assert result.exit_code == 1
    
        else:
            expected_output = ""
            if prompt is not None:
                # Build the expected prompt.
                assert isinstance(prompt, str)
                expected_prompt = f"Flag {prompt}: " if prompt else "Flag: "
    
                # Add the user input to the expected output.
                assert isinstance(input, str)
                expected_output += f"{expected_prompt}{input}\n"
    
                if expected is INVALID:
                    expected_output += "Error: invalid input\n"
    
                # The prompt is expected to be repeated.
                if expected in (REPEAT, INVALID):
                    expected_output += expected_prompt
    
            if expected not in (REPEAT, INVALID):
                expected_output += f"{expected!r}\n"
    
>           assert result.output == expected_output
E           AssertionError: assert 'Flag [Y/n]: \n' == 'Flag [Y/n]: \nTrue\n'
E             
E               Flag [Y/n]: 
E             - True

tests/test_termui.py:701: AssertionError
===== short test summary info =====
FAILED tests/test_termui.py::test_flag_value_prompt[--flag/--no-flag-opt_params11-args11-[Y/n]--True] - AssertionError: assert 'Flag [Y/n]: \n' == 'Flag [Y/n]: \nTrue\n'
```

I added an explanation in: https://github.com/pallets/click/pull/3030/commits/96e4c7599025eaefa58b7e804805e366f511a219

src/click/core.py
kdeldycke: I made that change explicitely as discussed here: https://github.com/pallets/click/pull/3030#discussion_r2280279005

But I can hide `UNSET` to the callback.

kdeldycke: Just hide the UNSET from the callback in https://github.com/pallets/click/pull/3030/commits/029bbeda76aa824149f110408d56de7b402a87ed

kdeldycke:
> I was trying to figure out exactly what you did to address #2012, and I think you just added tests for what's already happening, didn't change anything? Can you confirm how you addressed it.

I initially covered all the cases discussed over there to highlight your comment that, indeed, due to your changes in the pipeline, these types were fed as string type. In which case using `click.UNPROCESSED` was the solution. So for me #2012 was solved, as in: "*this is now expected in the new Click 8.x pipeline, now use `UNPROCESSED`*". And I just confirmed/locked down the new behavior by unit tests.

But since we re-establish the legacy case of options with `flag_value` and `default=True` in https://github.com/pallets/click/commit/06847da9126a99c88bf347adf213d33bfe4e4af2, we reverted back to solving the initial complain of #2012 as it was in Click 7.x.

kdeldycke:
I just addressed all the comments. I'm ready for a final review.

Rowlando13:
Looks like there is a docs build issue. I am going to merge then open it in another pr. @davidism I looked through all the comments and they were addressed. If you think they were not, I would suggest opening an issue and we can fix it there. The chain has gotten so long that it is hard to follow. I might have missed one.

Rowlando13:
@kdeldycke Awesome work!

kdeldycke:
Arf no don't merge yet! I wanted to squash some commits first! 😂 Well, whatever. I don't mind. `¯\_(ツ)_/¯`

davidism:
Sorry, wasn't ready to merge this yet either. I still want to understand what's going on with `flag_value="value", default=True`. Why is this considered legacy? There's a lot of strong language in the docs and code now implying that it's barely supported, but as far as I can tell is completely fine. `default=True` says "use this option's flag value as the default value". What's the issue with this?

I'm also not following the explanation https://github.com/pallets/click/pull/3030#issuecomment-3217081685 of `UNPROCESSED` and how it relates to this.

davidism:
@Rowlando13 don't merge huge work commit chains like this in the future. It makes it _way_ harder to use git blame in the future to figure out what the actual intended change was, as now we have to track the change through all the work and review as well. Either squash manually into a few commits, or use the Squash Merge feature in GitHub to create one commit.

Rowlando13:
@davidism My bad. Will do in the future. I thought we wanted to preserve the process.

Rowlando13:
@davidism shall I revert the merge?

davidism:
No it's fine, not a big deal, just something to keep in mind.

kdeldycke:
> Sorry, wasn't ready to merge this yet either. I still want to understand what's going on with `flag_value="value", default=True`. Why is this considered legacy? There's a lot of strong language in the docs and code now implying that it's barely supported, but as far as I can tell is completely fine. `default=True` says "use this option's flag value as the default value". What's the issue with this?

Ah yes, sorry for the "strong langage"! 😅 I'm too deep into Click internals to have good judgment about what makes a visible breaking change from a mere behind the scene arrangement. So thanks for the feedbacks you made over the past few weeks (like the `:meta private:` and so on).

I just made a new PR to not declare `default=True` as a legacy case: https://github.com/pallets/click/pull/3049

kdeldycke:
> @Rowlando13 don't merge huge work commit chains like this in the future. It makes it _way_ harder to use git blame in the future to figure out what the actual intended change was, as now we have to track the change through all the work and review as well. Either squash manually into a few commits, or use the Squash Merge feature in GitHub to create one commit.

I guess @davidism and I had a mutual implied understanding that PRs are OK to be messy while we work things up. Once we all agree about the quality, there's always time before merging to clean-up.

For me being messy is part of the process, especially on big changes like this. I am not afraid of doing stupid stuff, making mistakes and backtracking on code. And doing all of that in the open. Writing code that way allow me to gather feedback early. And course-correct if I am not going in the right direction.

Anyway, that is also part of learning to work together. No big deal indeed! I'll make sure to address everything so we can release 8.3.0 together and move on at a more relaxed pace! :)

mcepl:
Not only I completely agree with @davidism , but I would go much further. Sit down and read http://who-t.blogspot.com/2009/12/on-commit-messages.html ... it is not about letting your dirty laundry hanging around (and nobody is interested in that), it is about creating good foundation for yourself when you study the change a year later.

kdeldycke:
> I'm also not following the explanation [#3030 (comment)](https://github.com/pallets/click/pull/3030#issuecomment-3217081685) of `UNPROCESSED` and how it relates to this.

Hmm. Re-reading #2012 I think I overlooked a detail. Let's re-open it while I investigate.

PR #3049, opened 2025-08-25
Tone-down the case of `default=True` for flags in documentation

kdeldycke:
This PR reduce the strong langage around the behavior of `default=True` aligning to `flag_value` for flag options. There is not need to declare that case legacy as it has been Click's behavior for a long time.

This address issues uncovered by @davidism in https://github.com/pallets/click/pull/3030#issuecomment-3218244596 .

### Context

This is a follow-up on #3030, which has been merged a little bit early and needs some polishing before Click 8.3.0 release.

Amends: 06847da9126a99c88bf347adf213d33bfe4e4af2

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,14 +5,16 @@ Version 8.3.x
 
 Unreleased
 
--   Rework relationship between ``flag_value`` and ``default``: the value given to
-    ``default`` is now left untouched, and keep the value it receive. So
-    ``default=<desired_value>`` is respected and ``<desired_value>`` is passed on as-is
-    to the CLI function. With the exception of flag options, where setting
-    ``default=True`` maintain the legacy behavior of defaulting to the ``flag_value``.
-    This allow ``default`` to be of any type, including ``bool`` or ``None``, fixing
-    inconsistencies reported in: :issue:`1992` :issue:`2012` :issue:`2514`
-    :issue:`2610` :issue:`3024` :pr:`3030`
+-   **Improved flag option handling**: Reworked the relationship between ``flag_value``
+    and ``default`` parameters for better consistency:
+
+    * The ``default`` parameter value is now preserved as-is and passed directly
+      to CLI functions (no more unexpected transformations)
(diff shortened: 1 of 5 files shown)


kdeldycke:
@Rowlando13 As you are deep into the documentation side of Click, I'll be happy to have your feedback about the general wording here! :)

kdeldycke:
@Rowlando13 This is ready to be merged upstream! :)

davidism:
Good to merge, but I did mean to ask, is there a legitimate problem I'm not seeing here regarding typing/behavior? Or is it only a matter of being explicit with the code?

kdeldycke:
> Good to merge, but I did mean to ask, is there a legitimate problem I'm not seeing here regarding typing/behavior? Or is it only a matter of being explicit with the code?

Nothing about typing in this PR as most of the internals are already typed to `t.Any`.

Wait. Are you referring to typing as in *text editing* and not *Python's type system*? 😂

In which case I'm just requesting a quick review as I am not a native English speaker, so there's a non-zero chance my prose can end up a bit convoluted. 😬

davidism:
I just meant that in the original comments/discussion, you seemed to be saying there was an issue with the default. Is this something we should change in the future, or just a matter of consistency in `type_value` having the same type instead of `True`?

kdeldycke:
> I just meant that in the original comments/discussion, you seemed to be saying there was an issue with the default. Is this something we should change in the future, or just a matter of consistency in `type_value` having the same type instead of `True`?

You mean `flag_value` instead of `type_value ` right?

In which case I was just warning that a flag cannot have its default value set to `True`, and be active by default (i.e. `default=True`) while at the same time having a `flag_value` set. Because it is that `flag_value` that will be used for default. That is more a matter of semantics as pointing to by @janluke in https://github.com/pallets/click/issues/1992.

Maybe we will refine the concepts in the future with a better separation of concerns but that's really far away. We have more important things to do. So no big deal.

kdeldycke:

PR #3051, opened 2025-08-27
WIP: Rough out the idea of reference tables for kwargs and situations

### LINKED ISSUE #2835: Document existing parser fully
The existing Click API is very large and a lot of edge cases are undocumented. Additionally there are lots of combinations between parameters and options. 

I am not sure what is the best way to document all the combinations. There are enough combinations that it is hard to remember them all. Maybe a table with with tests that correlate to the table entry?

Rowlando13:


kdeldycke:
I really like the exhaustivity of the table. Because we can rely on it to track inconsistencies and as a truth table for tests. I think I will reuse that pattern to document the behavior of boolean flags in https://github.com/pallets/click/pull/3239

Rowlando13:
@kdeldycke Thanks. After doing this, I realized there are just too many arguments to resolve in a good way, because there are just so many edge cases. Not sure exactly what do about it yet, but plan to detail it out in an issue.

Rowlando13:
To resolve this will need to keep options and arguments public API but have a private API that breaks arguments and options down into more parts. To solve the many silent failures the current setup causes, will need to have a --check setting that can be run during tests to warn people.

kdeldycke:
@Rowlando13 if this is becoming too overwhelming, starts by looking at the `@parametrize` cases of the unittests: they're the ground truth. And maybe a better way of understanding the general rules and the edge-cases.

Issue #3028, opened 2025-08-05
`click.Choice` prompt with `StrEnum` displays enum values but requires enum names


lodo1995:
## How to replicate:

main.py:
```python
import click
from enum import StrEnum

class MyOpt(StrEnum):
    NAME_1 = "value-1"
    NAME_2 = "value-2"
    NAME_3 = "value-3"

@click.command()
@click.option("--my-opt", type=click.Choice(MyOpt), prompt = True)
def main(my_opt):
    pass

if __name__ == "__main__":
    main()

```
executed with:
```
> uv init click-strenum-bug
> cd click-strenum-bug
> uv add click
> uv run main.py
My opt (value-1, value-2, value-3): value-1
Error: 'value-1' is not one of 'NAME_1', 'NAME_2', 'NAME_3'.
```
The prompt suggests the `StrEnum` values, but the validation requires the `StrEnum` names.

## Expected output

Either of the following would be acceptable.

`StrEnum` values used consistently:
```
> uv run main.py
My opt (value-1, value-2, value-3): asdf
Error: 'asdf' is not one of 'value-1', 'value-2', 'value-3'.
```
`StrEnum` names used consistently:
```
> uv run main.py
My opt (NAME_1, NAME_2, NAME_3): asdf
Error: 'asdf' is not one of 'NAME_1', 'NAME_2', 'NAME_3'.
```

## Environment

- Python version: 3.13
- Click version: 8.2.1

Rowlando13:
This is fixed in #3004 and released in 8.2.2, but we had to yank it because of #3024. Look for this to be available and fixed in 8.2.3.

PR #3053, opened 2025-08-30
Add more tests on `Enum` rendering

kdeldycke:
This is just some more unit-tests to demonstrate that Click use `Enum`s names instead of their values.

These tests particularly highlight `Enum`'s whose keys mismatch their values. And add checks on the error messages proposing the accepted values for an option when a random one is provided by the user.

Refs:
- https://github.com/pallets/click/issues/2911
- https://github.com/pallets/click/pull/3004
- https://github.com/pallets/click/issues/3028


kdeldycke:

PR #3058, opened 2025-09-11
Forward exception information to resources registered in a context

### LINKED ISSUE #2447: Exception information isn't passed through to context.with_resource managed resources
When writing a simple tool with a bunch of commands updating a database, I was using a context manager in each command that would begin a transaction, and then commit it on success, or rollback on error.

Initially, I thought I'd be able to avoid some boilerplate on every subcommand by adding the session as a managed resource via ctx.with_resource(context_manager).  However, while this works in the success case, it will also commit the transaction if an unexpected exception occurs, since the exception information doesn't seem to be being passed through to the context manager.

Eg. given the below context manager:

    class TestContextManager:
        def __enter__(self):
            print("ENTER")

        def __exit__(self, exc_type, exc_val, traceback):
            print("EXIT", exc_type, exc_val, traceback)

Using it in a regular context manager via:

    with TestContextManager():
        raise Exception()

will correctly print the exc_type, exc_val and traceback parameters when an exception occurs.  But using it via:

    @click.group()
    @click.pass_context
    def cli(ctx):
        ctx.obj = ctx.with_resource(TestContextManager())

These will always be `None`, and so any context manager that needs to act differently on the occurrence of an exception vs success will always treat this as success.  (And likewise, try/except blocks will never trigger in function-style ones using the `@contextmanager` decorator.)

Would it be possible for the exception information to be passed through to managed resources so that context managers that treat success/failure cases differently operate correctly?

Environment:

- Python version: 3.10.9
- Click version: 8.1.3

Jazzinghen:
Updated `Context.close` and `Context.__exit__` to allow forwarding exceptions to the `__exit__` methods of the resources registered using `Context.with_resource`.

Added an extra function to handle both `Context.close` and `Context.__exit__`, called `Context._close_with_exception_info` that accepts all the parameters that `__exit__` and moved the `Content.close` logic there.

Added relevant tests.

- fixes #2447

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,8 @@ Unreleased
 -   Show correct auto complete value for nargs option in combination with flag option :issue:`2813`
 -   Fix handling of quoted and escaped parameters in Fish autocompletion. :issue:`2995` :pr:`3013`
 -   Lazily import ``shutil``. :pr:`3023`
+-   Properly forward exception information to resources registered with
+    ``click.core.Context.with_resource()``. :issue:`2447` :pr:`3058`
 
 Version 8.2.2
 -------------
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -483,11 +483,14 @@ class Context:
         exc_type: type[BaseException] | None,
         exc_value: BaseException | None,
         tb: TracebackType | None,
-    ) -> None:
+    ) -> bool | None:
         self._depth -= 1
+        exit_result: bool | None = None
         if self._depth == 0:
-            self.close()
+            exit_result = self._close_with_exception_info(exc_type, exc_value, tb)
         pop_context()
+
+        return exit_result
 
     @contextmanager
     def scope(self, cleanup: bool = True) -> cabc.Iterator[Context]:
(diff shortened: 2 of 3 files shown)


Jazzinghen:
I am not entirely sure what test is failing.

davidism:
Failing test seems unrelated. Change looks good.

I'm not entirely certain whether this makes more sense as a bug fix (exit wasn't capturing exception) vs feature (exit now captures exception). It's technically a change in behavior, but the old behavior was incomplete. I _think_ it's fine to target it at stable, just wanted to point out it's borderline.

Jazzinghen:
> Failing test seems unrelated. Change looks good.
> 
> I'm not entirely certain whether this makes more sense as a bug fix (exit wasn't capturing exception) vs feature (exit now captures exception). It's technically a change in behavior, but the old behavior was incomplete. I _think_ it's fine to target it at stable, just wanted to point out it's borderline.

Ah, yes, this is something I was thinking about as well. The issue was marked as bug, but it's partly a bug, party a "new feature"?
I targeted `stable` only because the referenced issue was marked as Bug.

Rowlando13:
@Jazzinghen I moved it to main so I can merge it and have it go out with 8.3.0.  There were a lot of changes made to main that are not on stable. Can you check that changes still look right.

Jazzinghen:
> @Jazzinghen I moved it to main so I can merge it and have it go out with 8.3.0.  There were a lot of changes made to main that are not on stable. Can you check that changes still look right. 

I'll rebase it and check it in an hour!

Jazzinghen:
Ok, I rebased the branch against `main` and tested it

Rowlando13:
Also just caught it. Please add to changes.rst.

Jazzinghen:
> Also just caught it. Please add to changes.rst.

Done!
Is the style correct?

Rowlando13:
Yes!

PR #3063, opened 2025-09-15
Move Option-decorators doc from reST to MyST

denuoweb:
This PR switches the Option-decorators documentation from reST to MyST .


docs/option-decorators.md
```diff
@@ -30,25 +32,24 @@ useful for password input:
 .. click:run::
 
     invoke(encode, input=['secret', 'secret'])
+```
 
 Because this combination of parameters is quite common, this can also be
 replaced with the :func:`password_option` decorator:
```
Rowlando13: This did not get converted over to myst.

denuoweb: fixed

PR #3079, opened 2025-09-22
Fix default handling to defer `UNSET` normalization

### LINKED ISSUE #3071: `default=True` on feature flags is order-sensitive
The following code:

```python
import click


@click.command()
@click.option("--red", "color", flag_value="red")
@click.option("--green", "color", flag_value="green", default=True)
@click.option("--blue", "color", flag_value="blue")
def main(color: str) -> None:
    print(repr(color))


if __name__ == "__main__":
    main()
```

when run with no arguments using click 8.2.1 outputs `'green'`, which is what I expect.  If click 8.3.0 is used instead, then it outputs `None` — but if the `--green` option is moved above the `--red` option in the script, click 8.3.0 will output `'green'`.

I believe that the script should output `'green'` regardless of the declaration order of the options.

Environment:

- Python version: 3.13.7
- Click version: 8.2.1 and 8.3.0

sirosen:
When options are being parsed, multiple options may evaluate to have defaults of `UNSET`, and some may evaluate to have explicitly set defaults of `None`. The normalization of `UNSET -> None` is now deferred until all options have been parsed, so that it is possible to accurately evaluate and save the first non-UNSET default value.

A new test exercises this functionality by pointing two flags at the same parameter declaration, and showing that their behavior is not order sensitive when only one of the flags has a default.

resolves #3071

---

The PR template says to add a note in `CHANGES.rst` for the change, but there's no header for the next release, so I wasn't sure where to put a change note. If a maintainer gives me a pointer in terms of what to do, I'm happy to amend with a change note. :slightly_smiling_face:

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1226,6 +1226,12 @@ class Command:
         for param in iter_params_for_processing(param_order, self.get_params(ctx)):
             _, args = param.handle_parse_result(ctx, opts, args)
 
+        # after handling *all* parameters, do a pass over the params to rewrite
+        # `UNSET`
+        for name, value in ctx.params.items():
+            if value is UNSET:
+                ctx.params[name] = None
+
         if args and not ctx.allow_extra_args and not ctx.resilient_parsing:
             ctx.fail(
                 ngettext(
@@ -2538,17 +2544,17 @@ class Parameter:
             # We skip adding the value if it was previously set by another parameter
             # targeting the same variable name. This prevents parameters competing for
             # the same name to override each other.
-            and self.name not in ctx.params
+            and (self.name not in ctx.params or ctx.params[self.name] is UNSET)
         ):
             # Click is logically enforcing that the name is None if the parameter is
             # not to be exposed. We still assert it here to please the type checker.
             assert self.name is not None, (
                 f"{self!r} parameter's name should not be None when exposing value."
             )
-            # Normalize UNSET values to None, as we're about to pass them to the
-            # command function and move them to the pure-Python realm of user-written
-            # code.
-            ctx.params[self.name] = value if value is not UNSET else None
+            # Do not normalize UNSET values to None at this point
+            # It needs to be done, but doing so too early impacts multiple parameters
+            # targeting the same variable name
+            ctx.params[self.name] = value
 
         return value, args
 
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1226,8 +1226,15 @@ class Command:
         for param in iter_params_for_processing(param_order, self.get_params(ctx)):
             _, args = param.handle_parse_result(ctx, opts, args)
 
-        # after handling *all* parameters, do a pass over the params to rewrite
-        # `UNSET`
+        # We now have all parameters' values into `ctx.params`, but the data may contain
+        # the `UNSET` sentinel.
+        # Convert `UNSET` to `None` to ensure that the user doesn't see `UNSET`.
+        #
+        # Waiting until after the initial parse to convert allows us to treat `UNSET`
+        # more like a missing value when multiple params use the same name.
+        # Refs:
+        # https://github.com/pallets/click/issues/3071
+        # https://github.com/pallets/click/pull/3079
         for name, value in ctx.params.items():
             if value is UNSET:
                 ctx.params[name] = None
@@ -2551,9 +2558,6 @@ class Parameter:
             assert self.name is not None, (
                 f"{self!r} parameter's name should not be None when exposing value."
             )
-            # Do not normalize UNSET values to None at this point
-            # It needs to be done, but doing so too early impacts multiple parameters
-            # targeting the same variable name
             ctx.params[self.name] = value
 
         return value, args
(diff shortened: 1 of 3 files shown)


src/click/core.py
```diff
@@ -2528,17 +2534,17 @@ def handle_parse_result(
...
                 f"{self!r} parameter's name should not be None when exposing value."
             )
-            # Normalize UNSET values to None, as we're about to pass them to the
-            # command function and move them to the pure-Python realm of user-written
-            # code.
-            ctx.params[self.name] = value if value is not UNSET else None
+            # Do not normalize UNSET values to None at this point
+            # It needs to be done, but doing so too early impacts multiple parameters
+            # targeting the same variable name
+            ctx.params[self.name] = value
```
sirosen, line 2561: Re-checking my changes, I'm not sure if I should keep this comment. Very open to a nudge about rewriting or even just removing it.

kdeldycke:
Your unit tests clearly demonstrate the issue and your changes don't trigger any other tests covering the `default` behavior. So for me this is enough to prove that your PR is really fixing the issue.

I'm pretty sure there is a more elegant way to solve it but for inclusion in a 8.3.1 hotfix release, this is good to me. We'll have plenty of time to refactor the code later. At this stage, I consider Click is still in the accumulation phase of external use-cases. So more unit tests the better! :)

sirosen:
> I'm pretty sure there is a more elegant way to solve it

I agree, but the way I see it that's a change for 8.4 or 9.0 .

There are too many different places which are independently responsible for choosing between `UNSET` and `None`. I think the right solve is to separate the data which is stored (where you might have `UNSET`) from the data which is _viewed_.

For example, changes like this:
```python
class Option(Parameter):
    _default: t.Any | UNSET

    @property
    def default(self):
        if self._default is UNSET:
            return None
        return self._default
```

But I don't know how that looks across *all* usage sites.

Rowlando13:
@sirosen I added the new section for you. In the future, you can just add a new section. If it is not there, it just means the maintainer who released forgot to add it after releasing.

src/click/core.py
kdeldycke: You are safe to remove the whole comment, as the original code before the introduction of the `UNSET` sentinel did not need any explanation: https://github.com/pallets/click/commit/b64ea07128a6368b5f6f93035c75d5693c7ba572#diff-11ba83cac151f7b24a1ed7c31a2a522d24d190cfa43199aa478d1e9cd2e6c610L2423-R2485

The `ctx.params[self.name] = value` statement feels natural as we have the `# Add parameter's value to the context` comment at the start of the `if` statement.

src/click/core.py
```diff
@@ -1216,6 +1216,12 @@ def parse_args(self, ctx: Context, args: list[str]) -> list[str]:
         for param in iter_params_for_processing(param_order, self.get_params(ctx)):
             _, args = param.handle_parse_result(ctx, opts, args)
 
+        # after handling *all* parameters, do a pass over the params to rewrite
+        # `UNSET`
```
kdeldycke: Feel free to be extremely explicit here in your comment, something like:

```
At this point we have completed the handling of all parameters. There is no risk of competing parameters sharing the same name to override each other. So we can proceed with explicitly hiding the :attr:`UNSET` value to the user. Refs:
https://github.com/pallets/click/issues/3071
https://github.com/pallets/click/pull/3079
```

kdeldycke:
> There are too many different places which are independently responsible for choosing between `UNSET` and `None`. I think the right solve is to separate the data which is stored (where you might have `UNSET`) from the data which is _viewed_.

100% this. I could not find an way to do this cleanly in #3030 so I had to draw the line. Addressing this is indeed a refactor.

In the mean time, your fix is in line with the policy we adopted:
- to hide `UNSET` in `to_info_dict()` method in the original PR:
  - https://github.com/pallets/click/pull/3030/files#diff-11ba83cac151f7b24a1ed7c31a2a522d24d190cfa43199aa478d1e9cd2e6c610L2677-R2820
  - https://github.com/pallets/click/pull/3030#discussion_r2270002483
- to hide `UNSET` in the `callback` in the original PR:
  - https://github.com/pallets/click/commit/029bbeda76aa824149f110408d56de7b402a87ed
  - https://github.com/pallets/click/pull/3030#discussion_r2291553604

So once you tweaked the code comments in the review I left you, this PR is to me good to be merged in a 8.3.1 hotfix release.

src/click/core.py
sirosen: I didn't apply an exact copy of this text, but same basic meaning and level of detail.
If you want to tweak it further, I'm happy to take any suggestion, but since the message was the same, I put it in as a co-authored commit. :smile:

sirosen:
I rebased and added the missing changelog note. I think this is ready!

src/click/core.py
kdeldycke: Ahah yeah, no worries. It's good as it is! :) 👍 

I just realized that rewording the comment was just a way for me to make sure that I understood your changes and did not left any edge-case! 😅

kdeldycke:
Good call @Rowlando13 on merging this! :)

PR #3090, opened 2025-09-25
Ensure that `type_cast_value` never gets `UNSET`

### LINKED ISSUE #3069: `Sentinel` is passed to `type_cast_value`
Starting in version 8.3.0, a `Sentinel` object is passed in to the `type_cast_value` function of the `Parameter` class and its derivatives if no value is supplied.

Reproduction:

```python
# cli_bug.py
import click


class CSVOption(click.Option):
    def type_cast_value(self, ctx, value):
        if value:
            return value.split(",")
        return value


@click.command(name="run")
@click.option("--csv", cls=CSVOption)
def main(csv):
    click.echo(f"{csv!r}")

if __name__ == "__main__":
    main()
```

```shell
$ uv run --with 'click==8.3.0' cli_bug.py

Installed 1 package in 1ms
  File "/home/coder/workspace/scratch/click-8_3_0-bug/cli_bug.py", line 17, in <module>
    main()
  File "/home/coder/.cache/uv/archive-v0/9Q5N3REh-YZUOJbO3iHRv/lib/python3.10/site-packages/click/core.py", line 1462, in __call__
  File "/home/coder/.cache/uv/archive-v0/9Q5N3REh-YZUOJbO3iHRv/lib/python3.10/site-packages/click/core.py", line 1382, in main
  File "/home/coder/.cache/uv/archive-v0/9Q5N3REh-YZUOJbO3iHRv/lib/python3.10/site-packages/click/core.py", line 1206, in make_context
    self.parse_args(ctx, args)
  File "/home/coder/.cache/uv/archive-v0/9Q5N3REh-YZUOJbO3iHRv/lib/python3.10/site-packages/click/core.py", line 1217, in parse_args
    _, args = param.handle_parse_result(ctx, opts, args)
  File "/home/coder/.cache/uv/archive-v0/9Q5N3REh-YZUOJbO3iHRv/lib/python3.10/site-packages/click/core.py", line 2516, in handle_parse_result
  File "/home/coder/.cache/uv/archive-v0/9Q5N3REh-YZUOJbO3iHRv/lib/python3.10/site-packages/click/core.py", line 2401, in process_value
    value = self.type_cast_value(ctx, value)
  File "/home/coder/workspace/scratch/click-8_3_0-bug/cli_bug.py", line 7, in type_cast_value
    return value.split(",")
AttributeError: 'Sentinel' object has no attribute 'split'
```

```shell
uv run --with 'click<8.3.0' cli_bug.py
None
```

Based on comments in #3065 and #3068 it would make sense to pass through `None` here given that this function is public. At a minimum, a changelog denoting the new behavior would be nice.

Environment:

- Python version: 3.10
- Click version: 8.3.0

sirosen:
In order to ensure that user-defined `type_cast_value` hooks are never
called with a value of `UNSET`, it is necessary to intercept `UNSET`
values and, in the particular cases in which `UNSET` needs special
handling, preempt any call to `type_cast_value()`.

`process_value()` is updated to check before calling `type_cast_value`
and to skip the call entirely if an `UNSET` value is detected. Passing
a value of `None` would not have the same effect for `click.Option`,
which previously had logic which is specific to the `value is UNSET`
case, and which cannot be reliably detected if a `None` value is
passed in.

This means that `type_cast_value()` is no longer guaranteed to be called
during argument processing, but when it is called, it's guaranteed not to
see an `UNSET` value. New tests explore this only from the perspective of
checking if `UNSET` is ever passed, not in terms of pinning down when/if
the hook gets called.

resolves #3069

-----

I think this change is worth considering as a fix for #3069, but I'm not
certain that it's good because it seems to trade one kind of incompatibility,
whether or not `UNSET` gets passed, for a different one: whether or not
`type_cast_value()` gets called at all.

If `type_cast_value()` is supposed to never see an `UNSET` value, and
`process_value` is "holding an `UNSET` value, then you might think that all
we need is

```python
type_cast_value(..., value if value is not UNSET else None)
```

but `Option.type_cast_value` throws a wrench in that, since it needs to
see the `UNSET` sentinel to properly determine its behavior.

I chose to take the same approach for handling vis-a-vis `multiple` and
`nargs=-1` in `Parameter`, on the grounds that it seemed "better aligned" with
what has to happen in `Option`.


src/click/core.py
```diff
@@ -2338,7 +2338,7 @@ def type_cast_value(self, ctx: Context, value: t.Any) -> t.Any:
         """Convert and validate a value against the parameter's
         :attr:`type`, :attr:`multiple`, and :attr:`nargs`.
         """
-        if value in (None, UNSET):
+        if value is None:
```
kdeldycke, line 2341: I like this as it returns `type_cast_value` to its original code before my #3030 PR.

kdeldycke (approved):
OK I understand the trade off and I am ok with them given that this PR:
- pragmaticcaly solve #3069
- adds dozen of new unit tests over uncovered edge-cases

It is not worse than the current state, as:
- in-place `multiple` checks and normalization for tuples already happens a lot (i.e. `self.multiple or self.nargs == -1`)
- other in-place checks on `UNSET` have been added in upcoming 8.3.1 for compatibility reasons

Overall it is a net gain as it will keep the community happy and provide grounds for upcoming refactors in 9.x.

kdeldycke:
Oh and also: I checked this branch against some private code and I could not find any issues.

kdeldycke:
Looking in my uncommitted tentative code of refactoring some part of the option parsing, I also identified the `type_cast_value` of `Option` to be unnecessary and misplaced. So yes, I am even more convinced that this is the direction to go.

kdeldycke:
This PR is ready to be merged upstream.

dotlambda:
It would be nice if you released version 8.3.1 with this fix.

PR #3091, opened 2025-09-26
Move License and Index doc pages from reST to MyST

denuoweb:
This PR switches the License and Index documentation pages from reST to MyST .


Rowlando13:
Remove the index file. That will have to be converted last. I could not get a markdown index file to work with rst files.

PR #3092, opened 2025-09-26
Convert api.rst to api.md

Veebaa:
Converted api.rst to api.md

#2886

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,217 +1,353 @@
-API
-===
+# API
 
-.. module:: click
+```{currentmodule} click
+```
 
 This part of the documentation lists the full API reference of all public
 classes and functions.
 
-.. contents::
-   :depth: 1
-   :local:
+```{contents}
+:depth: 1
+:local: true
+```
 
-Decorators
-----------
+## Decorators
 
+```{eval-rst}
 .. autofunction:: command
+```
 
+```{eval-rst}
 .. autofunction:: group
+```
 
+```{eval-rst}
 .. autofunction:: argument
+```
 
+```{eval-rst}
 .. autofunction:: option
+```
 
+```{eval-rst}
 .. autofunction:: password_option
+```
 
+```{eval-rst}
 .. autofunction:: confirmation_option
+```
 
+```{eval-rst}
 .. autofunction:: version_option
+```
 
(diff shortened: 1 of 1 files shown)


Rowlando13:
Please try to convert the rst directives to myst. Not sure if those directives are supported, but I think they are.

denuoweb:
> Please try to convert the rst directives to myst. Not sure if those directives are supported, but I think they are.

https://github.com/pallets/click/blob/27aaed3fe5bcd6adedd6e91de234914af9859cf1/docs/conf.py#L16

The .. autoclass:: and .. autofunction:: are autodoc directives from the sphinx.ext.autodoc extension which relies on Docutils’ reST parser, so to use these auto directives in MyST the current sphinx.ext.autodoc extension would need to be switched to [sphinx-autodoc2](https://sphinx-autodoc2.readthedocs.io/en/latest/autodoc_diff.html) for full MyST docstring compatibility. What Veebaa has might be the best it can be according to [6.2.sphinx.ext.autodoc](https://myst-parser.readthedocs.io/en/latest/syntax/code_and_apis.html#sphinx-ext-autodoc)

@Veebaa there is a minor issue on line 315 switch `/shell-completion` to {doc}`/shell-completion` .

docs/api.rst
```diff
@@ -1,217 +1,353 @@
...
 .. autoclass:: OptionParser
    :members:
 
+```
 
-Shell Completion
-----------------
+## Shell Completion
 
-See :doc:`/shell-completion` for information about enabling and
```
denuoweb, line 192: Switch to MyST {doc}

Rowlando13:
In that case, I will review it when I can get in front of my computer. We may migrate to the extension you suggested, but I would like to explore that once all the files are converted to markdown.

denuoweb:
> In that case, I will review it when I can get in front of my computer. We may migrate to the extension you suggested, but I would like to explore that once all the files are converted to markdown.

@Rowlando13 Switching the extension could necessitate additional updates to the existing converted files so you could accept this PR now and I will review all converted .md myst files for conversion compliance when you start looking to make the extension update. 

@Veebaa You may need to skip over converting changes.rst.

PR #3101, opened 2025-09-30
add a test based on busybox to github workflows

ghost:
add a test based on busybox to github workflows

In #2944 there has been a discussion whether or not to test the PR. Currently it is proposed to just to add comments so the fix doesn't get reverted by accident. But I got curious and wanted to see how difficult it is to add a busybox based test. Alpinelinux uses busybox by default and there is a convenient github-action to test on alpine.

Of course the maintainers are free to decide, I only want to mention, that if this test becomes a burden, they can always just remove it and we are back on status quo.

NOTE: This will fail till #2944 is merged.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -30,6 +30,18 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
       - run: uv run --locked tox run -e ${{ matrix.tox || format('py{0}', matrix.python) }}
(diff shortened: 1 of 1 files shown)


davidism:
Thanks, but I don't want to add this extra complexity. There's a lot of shell/terminal related stuff that's tricky to really test, so I'm happy with comments in those parts for now. It's a good reference if we want to add it later.

PR #3102, opened 2025-10-02
Don't test using a file in docs/

m-u-xyz:
The docs/ directory may not be present when building, and click itself may or may not be in a src/ subdirectory.
Thus, test with a file in tests/.


Rowlando13:
@smurfix How did you find this issue? When would you be building that you have tests but not docs?

m-u-xyz:
Building my `asyncclick` fork as a Debian package. It assembles the package in a build directory and runs the tests from there.

Rowlando13:
Thanks for the fix and good to know!

PR #3104, opened 2025-10-07
Fixes #3084 — Restore optional value behavior for options with flag_value

### LINKED ISSUE #3084: Bug: Optional value not optional anymore
### Outline

As stated in [official documentation](https://click.palletsprojects.com/en/stable/options/#optional-value): 
> Setting `is_flag=False, flag_value=value` tells Click that the option can still be passed a value, but if only the flag is given, the value will be `flag_value`.

Since 8.3.0 this doesn't work anymore, instead `SystemExit` with "Error: Option '--xyz' requires an argument." is thrown.

### Replication of bug

Example taken from documentation:
```python
@click.command()
@click.option("--name", is_flag=False, flag_value="Flag", default="Default")
def hello(name):
    click.echo(f"Hello, {name}!")
```

```shell
$ hello --name
Error: Option '--name' requires an argument.
```

### Expected behavior

The code example above should work as
```shell
$ hello --name
Hello, Flag!
```

It's not even working in the generated documentation anymore: while [documentation up to Sep, 10th](https://web.archive.org/web/20250910222023/https://click.palletsprojects.com/en/stable/options/#optional-value) shows the expected and correct behavior, [current documentation](https://click.palletsprojects.com/en/stable/options/#optional-value) demonstrates the bug.

- Python version: 3.11
- Click version: 8.3.0

trippinganymess:
This PR:
- Corrects Option.__init__ logic to ensure options using `flag_value` and `is_flag=False` work as documented, accepting just the flag and using the `flag_value` as the option's value.
- Adds comprehensive tests in `tests/test_optional_value_bug.py` that fail before and pass after this fix, demonstrating the bug and its resolution.
- Updates code documentation with a new .. versionchanged:: note, and adds a changelog entry summarizing the change.

This restores the intended and documented API behavior, aligning Click’s implementation with user expectations and preventing regressions in downstream projects


tests/test_optional_value_bug.py
davidism, line 1: Don't create a completely new file to test one bug fix. All that's needed is probably one new parametrization of an existing test.

Also, the docstrings, comments, and verbosity of the tests is not really helping at all.

src/click/core.py
```diff
@@ -2625,6 +2625,13 @@ class Option(Parameter):
     :param hidden: hide this option from help outputs.
     :param attrs: Other command arguments described in :class:`Parameter`.
 
+    .. versionchanged:: 8.3.0
```
davidism: This version is wrong.

src/click/core.py
```diff
@@ -2625,6 +2625,13 @@ class Option(Parameter):
...
+        Fixed a regression where options with ``flag_value`` and ``is_flag=False``
```
davidism: This change comment is way too verbose. "Restore optional value behavior when `flag_value` is set."

CHANGES.rst
```diff
@@ -6,6 +6,7 @@ Version 8.3.x
 Unreleased
 
 -   Don't discard pager arguments by correctly using subprocess.Popen. :issue:`3039` :pr:`3055`
+-   Fix: Restore documented behavior for options with flag_value and is_flag=False (#3084).
```
davidism: Follow the style in the rest of the changelog. Don't add a tag prefix, use the correct reference, add this to a new section rather than an already released version.

Also, the PR title, commit message, docstring, and this line are all different wordings of the same fix. Be consistent.

test.sh
davidism, line 1: Remove this file.

davidism (changes requested):

davidism:
I appreciate that you submitted a fix for this. However, your use of AI really detracted from the quality of the PR. Please avoid using AI, or at least carefully review the results and compare them to the project style before submitting.

The initial PR description is also a wordy mess that AI tends to produce.

"fixes #..." should be in the PR description, not a tag in the title.

You haven't configured git correctly, so your commits aren't linked to your GitHub account.

The commit message, PR title, changelog entry, and docstring changelog entry are all worded differently.

trippinganymess:
I will try to be more careful next time, Thank you so much for the review

davidism:
This needs to be rebased onto stable instead of main as well.

trippinganymess:
Hi, I think I have made all the relevant changes, I am sorry for being negligence before. It was my first contribution to open source I went on chasing for a quick win. please lemme know if there is anything still wrong I will try to make it right.

Thank you!!

kdeldycke:
I propose to close this PR has it has been superseded by https://github.com/pallets/click/pull/3152 , which has been merged upstream.

kdeldycke:
Just doubled-check the content of that PR against #3152, so yes, it is safe to close.

PR #3106, opened 2025-10-12
Move Extending-click page from reST to MyST

denuoweb:
This PR switches the Extending Click documentation page from reST to MyST .


docs/extending-click.md
```diff
@@ -116,10 +110,11 @@ prefix for a command. If there was a command called ``push``, it would accept
             # always return the full command name
             _, cmd, args = super().resolve_command(ctx, args)
             return cmd.name, cmd, args
+```
 
 It can be used like this:
 
-.. click:example::
+```python
```
denuoweb, line 117: Switched to python because click:example:: code is broken with the above "It can be used like this:" text.

PR #3122, opened 2025-10-22
Document help page exit codes

### LINKED ISSUE #2909: Exit code changed for command showing help message when arguments are not provided
Defining a simple entry-point group:

```
from __future__ import annotations

import click

from .sys_info import run as sys_info


@click.group()
def run() -> None:
    """Main package entry-point."""  # noqa: D401


run.add_command(sys_info)
```

And executing the corresponding command without an arguments shows the help message, e.g.

```
$ my_click_cmd
Usage: my_click_cmd [OPTIONS] COMMAND [ARGS]...

  Main package entry-point.

Options:

Commands:
  sys-info  Run sys_info() command.
```

But now returns exit code 2 instead of 0 as in `click<8.2.0`. `$ my_click_cmd --help`, providing explicitly the `--help` option returns the same message with exit code 0.

Is this an intended backward incompatible change (undocumented in the changelog), or an introduced bug?

Rowlando13:


kdeldycke:
LGTM! Let's merge!

And I guess with that explanation in the doc, you can close the related issue #2909.

PR #3129, opened 2025-10-27
Fix test_edit to work with BSD sed

### LINKED ISSUE #3128: `test_edit` fails with BSD sed
I'm running 8.3.0 tests with Python 3.14.0 and I see this failure:

```
=================================== FAILURES ===================================
__________________________________ test_edit ___________________________________
tests/test_termui.py:397: in test_edit
    assert reopened_file.read() == "aTest\nbTest"
E   AssertionError: assert 'aTest\nbTest\n' == 'aTest\nbTest'
E
E       aTest
E     - bTest
E     + bTest
E     ?      +
=========================== short test summary info ============================
FAILED tests/test_termui.py::test_edit - AssertionError: assert 'aTest\nbTest\n' == 'aTest\nbTest'

    aTest
  - bTest
  + bTest
  ?      +
============ 1 failed, 1284 passed, 21 skipped, 1 xfailed in 5.26s =============
```

To reproduce the issue I just downloaded the 8.3.0 sdist from PyPI, unpacked it and then started `tox -e py314`.

Please note that on this platform the `sed` is not GNU sed, but BSD sed.  When I force to use GNU sed in the test, then all tests pass.

Environment:

- Python version: 3.14.0
- Click version: 8.3.0
- OS: OpenIndiana

mtelka:
Fixes #3128.


Rowlando13:
Please add a comment.

mtelka:
> Please add a comment.

@Rowlando13 could you please elaborate what kind of comment you mean?  Thank you.

Rowlando13:
Explain in short why the test was altered. Specifically note the os and reason. If you don't we may unintentionally break the test again or add a similar breaking test.

mtelka:
> Explain in short why the test was altered. Specifically note the os and reason. If you don't we may unintentionally break the test again or add a similar breaking test.

Here in the PR, or in the sources?

Rowlando13:
If by sources you mean in tests/test_termui.py, then yes.

mtelka:
> If by sources you mean in tests/test_termui.py, then yes.

Okay, will do.

In the meantime, POSIX says this about [sed](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/sed.html):

```
In default operation, sed cyclically shall append a line of input, less its 
terminating <newline> character, into the pattern space. Reading from input
shall be skipped if a <newline> was in the pattern space prior to a D command 
ending the previous cycle. The sed utility shall then apply in sequence all 
commands whose addresses select that pattern space, until a command starts the
next cycle or quits. If no commands explicitly started a new cycle, then at the
end of the script the pattern space shall be copied to standard output (except
when -n is specified) and the pattern space shall be deleted. Whenever the
pattern space is written to standard output or a named file, sed shall
immediately follow it with a <newline>.
```
So it looks like GNU sed does not follow this accurately.

mtelka:
> Please add a comment.

Done. @Rowlando13 please let me know if you'd prefer different wording there.  Thank you.

PR #3137, opened 2025-11-11
Provide altered context to callbacks to hide `UNSET` values as `None`

### LINKED ISSUE #3136: Sentinel available in context during option callbacks
Sentinel is an implementation detail that is not intentionally exposed. However, it is in the context params during a callback: sentinel_in_callback.py

```
import click

def _validate(ctx, param, value):

    you_sentinel = ctx.params.get("you")
    you_too_sentinel = ctx.params.get("you_too")
    print(f"you: {you_sentinel}")
    print(f"you_too: {you_too_sentinel}")
    print(f"param: {param}")
    print(f"value: {value}")
    return value

@click.group()
def basic():
    pass

@basic.command()
@click.argument('you', required=False)
@click.option('--you-too', )
@click.option('--yall', callback=_validate)
@click.pass_context
def hello(ctx, you, you_too, yall):
    click.echo('Hello World!')

if __name__ == '__main__':
    basic()
    # python -m sentinel_in_callback hello
``` 

``` 
> python -m sentinel_in_callback hello
you: Sentinel.UNSET
you_too: Sentinel.UNSET
param: <Option yall>
value: None
Hello World!
```

kdeldycke:
PR to fix:
- https://github.com/pallets/click/issues/3136
- https://github.com/pallets/flask/issues/5836

This highlight a side-effect introduced in:
- https://github.com/pallets/click/pull/3079
- https://github.com/pallets/click/issues/3071

CC @dotlambda

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@ Unreleased
     the ``Context.invoke()`` method. :issue:`3066` :issue:`3065` :pr:`3068`
 -   Fix conversion of ``Sentinel.UNSET`` happening too early, which caused incorrect
     behavior for multiple parameters using the same name. :issue:`3071` :pr:`3079`
+-   Hide ``Sentinel.UNSET`` values as ``None`` when looking up for other parameters
+    through the context inside parameter callbacks. :issue:`3136` :pr:`3137`
 -   Fix rendering when ``prompt`` and ``confirm`` parameter ``prompt_suffix`` is
     empty. :issue:`3019` :pr:`3021`
 -   When ``Sentinel.UNSET`` is found during parsing, it will skip calls to
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2440,7 +2440,37 @@ class Parameter:
             # to None.
             if value is UNSET:
                 value = None
-            value = self.callback(ctx, self, value)
+
+            # Search for parameters with UNSET values in the context.
+            unset_keys = {k: None for k, v in ctx.params.items() if v is UNSET}
+            # No UNSET values, call the callback as usual.
+            if not unset_keys:
+                value = self.callback(ctx, self, value)
+
+            # Legacy case: provide a temporarily manipulated context to the callback
+            # to hide UNSET values as None.
+            #
+            # Refs:
+            # https://github.com/pallets/click/issues/3136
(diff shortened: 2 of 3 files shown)


kdeldycke:
Ok I found a solution to reconcile the 2 issues. It's a bit dirty, but it works without side-effects.

mgorny:
Thanks! I can confirm that this fixes the issues I've seen in `black`.

Rowlando13:
The tests look good.

Rowlando13:
@davidism Can you review this?

Rowlando13:
@kdeldycke Also add change entry.

davidism:
A ridiculous fix 😅, but I'm fine with it.

Maybe in the next release we can figure out some internal API separate from the external one, rather than this bit of magic.

davidism (approved):

kdeldycke:
> A ridiculous fix 😅, but I'm fine with it.

Oh yes. A bit ashamed by that. That's why I tried to be very explicit in delimiting that fix. Anyone reaching that implementation will understand that this part needs some refactor.

> Maybe in the next release we can figure out some internal API separate from the external one, rather than this bit of magic.

Oh yes. You initiated that direction by helping me identify private methods in the previous months, with `:meta private:` tags in docstrings. Maybe we can go further and start prefixing functions, methods and properties with underscore `_`, as a hint for stuff you are not supposed to mess with. Let's discuss that once things settle down.

kdeldycke:
> @kdeldycke Also add change entry.

Ah yes, good call. Done.

kdeldycke:
Also, just for reference, this PR pass all my tricky hacks in Click Extra. 😇

kdeldycke:
> > Maybe in the next release we can figure out some internal API separate from the external one, rather than this bit of magic.
> 
> Oh yes. You initiated that direction by helping me identify private methods in the previous months, with `:meta private:` tags in docstrings. Maybe we can go further and start prefixing functions, methods and properties with underscore `_`, as a hint for stuff you are not supposed to mess with. Let's discuss that once things settle down.

Also, the advantage of all that `UNSET` saga is that we can now easily find where in the code base of Click there is expectation of value normalization. Just grep on `UNSET`! 😅🫣

Issue #3110, opened 2025-10-16
CliRunner.invoke raises a "ValueError: I/O operation on closed file"


getzze:
Hi,
I am a maintainer of [subliminal](https://github.com/Diaoul/subliminal) that uses `click` for its CLI. Some tests for the CLI are failing since the upgrade to click 8.3.0.
I tracked the failures to a regression in click 8.3.0 due to #2991. Reverting this commit makes my tests pass with click 8.3.0.

The failing pytest tests are of the following form:

```
def test_cli_download(tmp_path: os.PathLike[str]) -> None:
    runner = CliRunner()

    result = runner.invoke(subliminal_cli, ['download', '-l', 'en', '-p', 'podnapisi', 'video.mkv'])

    # Curiously this test is not failing, subtitle is even downloaded
    assert result.exit_code == 0

    # The following two asserts are failing because `result.output` show a long error message.
    assert result.output.startswith('Collecting videos')
    assert result.output.endswith('Downloaded 1 subtitle\n')

```

When I only run this test function with `pytest`, all the assertions pass, but when I run the whole file that has plenty of test functions with the same structure, all the functions fail at the `result.output` assert (although not consistently...)

The error message in `result.output` is like this:
 
```
--- Logging error ---
Traceback (most recent call last):
  File "/usr/lib/python3.13/logging/__init__.py", line 1154, in emit
    stream.write(msg + self.terminator)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
ValueError: I/O operation on closed file.
[...]
  File "src/subliminal/score.py", line 157, in compute_score
    logger.debug('Found matches %r', matches)
Message: 'Found matches %r'
Arguments: ({'country', 'episode', 'season', 'series', 'year'},)
```

It seems the logging files are closed after one `CliRunner.invoke` call and its giving rise to failures in the other tests.

edgarrmondragon:
I'm seeing similar output in some projects. I think that ironically #2991 introduced a different race condition, because the stream is closed by the time `logging` tries to write a message to it.

cc @neutrinoceros

neutrinoceros:
Thanks for the ping. I'm not an expert here, so please take everything I say with a grain of salt, but I do notice there was a test that needed updating for click's own CI to survive my patch. Could you try using the `CliRunner.isolation` context manager in your tests, see if it resolves the new race ?

getzze:
[...]

PR #3139, opened 2025-11-12
`NamedTextIOWrapper`: stop closing buffer

pjonsson:
The NamedTextIOWrapper is closing the buffers
that are owned by the StreamMixer, so
implement a close() method that prevents
it from doing that.

Refs #824
Refs #2993

Previous PR for the same issue: #2991

Just removing the `__del__` on the `StreamMixer`
makes the test in #2993 fail for me when using
the free-threaded Python 3.14 in uv. The test
starts passing again when adding the no-op
`close()` on `_NamedTextIOWrapper`.


tests/test_testing_logging.py
```diff
@@ -0,0 +1,25 @@
...
+import click
+from click.testing import CliRunner
+
+
+# Minimized version of:
+# https://github.com/pallets/click/issues/824#issuecomment-3027263262
+#
+# Test needs to be run as "pytest --log-cli-level 30 tests/test_testing_logging.py"
+# to test the intended functionality.
+def test_runner_logger():
```
pjonsson: I need help with how to get this test file to also be run with `pytest --log-cli-level 30 tests/test_testing_logging.py` automatically.

Edit: my thought was to try to prevent future regressions by adding this test and running it with the right `--log-cli-level` parameter to pytest, but just having the test in the test suite is also an improvement over the current state since it makes it easier to run the test by hand.

kdeldycke:
FYI, these changes does not fix the issues uncovered from https://github.com/pallets/click/pull/3151.

pjonsson:
Someone commented about their success with this in my fork: https://github.com/pjonsson/click/commit/0813e7b329e0c67c08c7841a61675d56ef405590#commitcomment-172085876

kdeldycke:
> FYI, these changes does not fix the issues uncovered from #3151.

So it turns out #3151 has nothing to do with this issue. In the end the race condition highlighted here is a problem with `CliRunner`'s I/O stream lifecycle management. Specifically, who owns the streams, and who is responsible for closing them.

The root cause is dual ownership: both `StreamMixer` and `_NamedTextIOWrapper` (via `TextIOWrapper.__del__`) try to close the same buffer. So @pjonsson's fix is the right one.

So this PR:
- Closes #824
- Reverts #2991
- Reproduce #2993
- Closes #3110
- Supersed #3140

I have stress tests that can show the issue. I can add a commit on top of this PR to cover this fix.

kdeldycke:
So I pushed a comprehensive regular and stress tests on top of that PR at: https://github.com/pallets/click/pull/3139/changes/20f8c1a1a976bb83350c45d61f9d6735e263c29a

kdeldycke:
@pjonsson feel free to review my test suite to confirm we cover your case.

kdeldycke:
Here is the timeline of the issue:

```text
  #654 (2016, stdout proxy)  ─────────────────────────────────────┐
  #824 (2017, original)  ─────────────────────────────────────────┤ Historical
  #951 (2018, WIP PR, closed) ────────────────────────────────────┘ context

  Click 8.2 introduces StreamMixer + _NamedTextIOWrapper
            │
            ▼
  #2993 ── race in StreamMixer finalization (multi-threaded)
    │      │
    │      ▼
    └─► #2991 (merged) ── adds __del__ to order close
                │
                ▼
          #3110 ── __del__ causes regression with logging
          #824 (reopened) ── same symptom, different trigger
                │
                ├──► #3139 (open) ── correct fix: no-op close()
                ├──► #3140 (open) ── interim fix: guard checks
                └──► pjonsson/click@0813e7b ── confirms #3139 works
```

Tl;Dr: issue was [lying around since 2016](https://github.com/pallets/click/issues/654), I was [bitten by it in 2025](https://github.com/pallets/click/issues/824#issuecomment-3217917881), re-ignited the interest by [introducing `StreamMixer` and `_NamedTextIOWrapper` in 8.2.0](https://github.com/pallets/click/pull/2523), and then it took 3 attempts to fix it before @pjonsson addressed the root cause here.

pjonsson:
> @pjonsson feel free to review my test suite to confirm we cover your case.

The public "reproducer" I have is https://github.com/opendatacube/datacube-explorer/, but we're using a branch with the fix since a couple of weeks ago so it definitely works for that problem.

Thank you for making all the tests for this PR!

kdeldycke:

PR #3141, opened 2025-11-13
Fix Ruff UP038 warning

pjonsson:
Running "ruff format" gives
the warning:

warning: The following rules have been removed and ignoring them has no effect:
    - UP038

so remove the ignore.


davidism:
The lower bound isn't needed, instead only remove the ignore.

PR #3158, opened 2025-11-24
Add upgrade guide section.

Rowlando13:


Rowlando13:
@kdeldycke Let me know your thoughts.

kdeldycke:
Yeah the structure is good to me. Let's see if that will get adoption for the next releases.

PR #3162, opened 2025-11-29
Fix the indentation size in code snippets

gr1seo:
Use the same indentation size (4 spaces) for python code snippets, and remove the misplaced newline.

### DIFF
--- a/docs/commands-and-groups.rst
+++ b/docs/commands-and-groups.rst
@@ -255,24 +255,24 @@ the prefix is ``WEB``, then the variable is ``WEB_RUN_SERVER_HOST``.
 
 .. click:example::
 
-   @click.group()
-   @click.option('--debug/--no-debug')
-   def cli(debug):
-       click.echo(f"Debug mode is {'on' if debug else 'off'}")
+    @click.group()
+    @click.option('--debug/--no-debug')
+    def cli(debug):
+        click.echo(f"Debug mode is {'on' if debug else 'off'}")
 
-   @cli.command()
-   @click.option('--username')
-   def greet(username):
-       click.echo(f"Hello {username}!")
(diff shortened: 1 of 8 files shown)


Rowlando13:
How did you discover this? Did you use some tool? Seems like adjusting spaces in random files.

gr1seo:
I found missing indentation in the Utilities section under [Finding Application Folders](https://click.palletsprojects.com/en/stable/utils/#finding-application-folders). After fixing it, I rechecked the docs using these regexes:

`^\.\. click\:(?:example|run)\:\:\n{1,2}^(?: {1,3}| {5,7}| {9,11}| {13,15})[\w@]`
`^```python\n{1,2}^ {1,}[\w@]` (added after an additional recheck following your comment, second commit 😅)
`^(?:```python|\:caption\:.*\.py)\n{1,2}^(?: {1,3}| {5,7}| {9,11}| {13,15})[\w@]`
`^ *(?:class|def|with|for|while|if|elif|else).*?\:\n{1,2}(?: {0,3}| {5,7}| {9,11}| {13,15})[\w@]`

I also reviewed the files with visible spaces enabled. The fix was done manually, without any tools, just the IDE’s find functionality.

Rowlando13:
@gr1seo Thanks for find these. Unfortunately I am not going to accept them at this time. Restructured text is a bit finicky with stuff and can silently fail with formatting, so it would require me to go through and manually view every section of the docs that correlates to the change. As it is everything works. I am planning on implementing a formatter in the future to handle these issues. I will leave this PR is draft for reference.

kdeldycke:
This is likely to be fixed in the future, as I know @Rowlando13 is looking for stable solution to auto-format the docs.

PR #3166, opened 2025-12-09
Document show_envvar in Help Pages

RyanSquared:
I had to go to the API documentation to find out how to enable this feature. I would've also liked to add the relevant information to the "Option" page, but without seeing anything for `show_default` on the Option page as well, I decided it may be out of scope for this PR.

Credit to @meejah for checking the documentation first, and noticing its absence.

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -145,6 +145,22 @@ For single option boolean flags, the default remains hidden if the default value
    invoke(dots, args=['--help'])
 ```
 
+## Showing Environment Variables
+
+To control the appearance of environment variables pass `show_envvar`.
+
+```{eval-rst}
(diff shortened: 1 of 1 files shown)


meejah (commented):
FWIW, I approve of this addition to the docs

Rowlando13:
Looks good!

RishitaGupta-0:
Thanks for the explanation!

I faced the same confusion while trying to run one of the Click examples.
I ran the example directly using `python colors.py --help`, and I didn’t see any help output, which was a bit confusing at first.

After reading this discussion, it makes more sense now.
A small clarification like this is really helpful for beginners who are new to Click and CLI tools.

PR #3180, opened 2026-01-01
Feature: Add usage string to info_dict

### LINKED ISSUE #2992: Add output of get_usage() to get_info_dict()
## Context 

I am generating some of my user-facing documentation using `get_info_dict()`, and was looking for a way to automate the usage as well without having to hardcode it in the docstrings. 

## Potential solution

Using the calls of `get_info_dict()` to add a `usage` key with the value being the output of `get_usage()`.

Other relevant issues: #2762 

## Workaround

For now I'm using `get_info_dict()`, then looping through the context to get groups and commands, and using `get_usage()` on each, then adding it to that dictionary. The code looks something like this (with the for loop matching the structure of my CLI):

```python
    info = ctx.to_info_dict()
    for cmd_n1 in ctx.command.list_commands(ctx):
        cmd2 = ctx.command.get_command(ctx, cmd_n1)
        sub_ctx = ctx._make_sub_context(cmd2)
        usage = cmd2.get_usage(sub_ctx).replace("Usage: ", "")
        info["command"]["commands"][cmd_n1]["usage"] = usage
        if isinstance(cmd2, click.Group):
            # Groups
            for cmd_n2 in cmd2.list_commands(ctx):
                cmd3 = cmd2.get_command(ctx, cmd_n2)
                sub_ctx3 = sub_ctx._make_sub_context(cmd3)
                usage = cmd3.get_usage(sub_ctx3).replace("Usage: ", "")
                info["command"]["commands"][cmd_n1]["commands"][cmd_n2]["usage"] = usage

```

AmoghDeshmukh:
The Group class's to_info_dict method calls this parent method, so it will automatically get the usage string. It also recursively calls to_info_dict on its sub-commands, so the change will propagate through the entire command tree.
 Added new line -  "usage": self.get_usage(ctx)
Closes #2992

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -985,6 +985,7 @@ class Command:
             "short_help": self.short_help,
             "hidden": self.hidden,
             "deprecated": self.deprecated,
+            "usage": self.get_usage(ctx),
         }
 
     def __repr__(self) -> str:


Rowlando13:
Thank you for your efforts, but the tests don't pass and we are not accepting any features until we can clean things up with 9.0.

PR #3186, opened 2026-01-15
fix: `_termui_impl.open_url()` — 'start' on Windows is a cmd built-in, not an executable

### LINKED ISSUE #3164: click.launch() does not work for (non local) urls on Windows.
Environment:

- Python version: 3,13
- Click version: 8.3.1
- OS: Windows (11)

click.launch() (precisely: click._termui_impl.open_url()) does not work for (non local) urls on Windows.
e.g. click.launch("http://www.python.org") does not work.

The patch:

```diff
--- click/_termui_impl.py.org	2025-12-03 16:55:34.202910500 +0100
+++ click/_termui_impl.py	2025-12-03 21:44:31.351834200 +0100
@@ -707,7 +707,7 @@
             args.append("")
             args.append(url)
         try:
-            return subprocess.call(args)
+            return subprocess.call(args, shell=True)
         except OSError:
             # Command not found
             return 127
```

karpierz:
fix: _termui_impl.open_url() — 'start' on Windows is a cmd built-in command, not an executable

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,9 @@ Unreleased
     non-shadowed help option names, so ``Try '... -h'`` no longer points to a
     subcommand option that shadows ``-h``. All surviving names are shown
     (``-h/--help``). :issue:`2790` :pr:`3208`
+-   Use :func:`os.startfile` on Windows to open URLs in :func:`open_url`,
+    replacing the ``start`` built-in which cannot be invoked without
+    ``shell=True``. :issue:`3164` :pr:`3186`
 
 Version 8.3.3
 -------------
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -720,17 +720,16 @@ def open_url(url: str, wait: bool = False, locate: bool = False) -> int:
         if locate:
             url = _unquote_file(url)
             args = ["explorer", f"/select,{url}"]
+            try:
+                return subprocess.call(args)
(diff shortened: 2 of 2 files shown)


kdeldycke:
`shell=True` has been eliminated entirely in Click code base with #3245. So there is no reason to re-introduce it. Can you give use more details about the reason we need this?

kdeldycke:
OK got it. Click code is broken on Windows because `start` is a `cmd.exe` built-in, not a standalone executable. So without a `shell=True` the `subprocess.call` will fail. So that's a legitimate bug report.

But there is a better fix. Instead of re-introducing `shell=True`, we can use [`os.startfile()`](https://docs.python.org/3/library/os.html#os.startfile).

kdeldycke:
I just re-implemented the fix. This PR is ready to be merged in the next devlopment cycle of Click.

Rowlando13:
I will fix changelog then it's ready to merge.

kdeldycke:
> I will fix changelog then it's ready to merge.

I just fixed the merge conflicts and rebased on top of `main`. This PR is ready.

davidism:
~Check that this still works on cygwin. I know we don't officially support it, but I think some of this code being changed was there in part due to someone wanting to use it from cygwin or msys2 maybe?~

Never mind, missed the literal next line of the diff.

davidism (approved):

PR #3187, opened 2026-01-16
Fix BytesWarning in Path.convert when comparing dash values

### LINKED ISSUE #2877: [Python forward compat] click.types.Path.covert: BytesWarning: Comparison between bytes and string
So I've hit this through pip-tools' `pip-compile` executable some time ago but forgot to report immediately.

I tend to run Python with strictness like `python -I -s -E -bb -Werror -m piptools compile --some --args`. And that's how I've found this warning. I imagine, it's easy to address by changing the check here: https://github.com/pallets/click/blob/c9f7d9d/src/click/types.py#L929C64-L929C75. Perhaps, by converting the data type.

Here's a minimal repro. In needs a modern enough Python for `-bb` IIRC.
```console
$ python -Im pip install Click
$ python -bb -Werror
Python 3.12.9 (main, Feb 22 2025, 21:46:24) [GCC 14.2.1 20241221] on linux
>>> from click.types import Path
>>> Path(allow_dash=True).convert('', '', '')
  File "<stdin>", line 1, in <module>
  File "/path/to/venv/lib/python3.12/site-packages/click/types.py", line 865, in convert
    is_dash = self.file_okay and self.allow_dash and rv in (b"-", "-")
                                                     ^^^^^^^^^^^^^^^^^
BytesWarning: Comparison between bytes and string
```

Environment:

- Python version: 3.12.9
- Click version: 8.1.8

VedantMadane:
Fixes a BytesWarning that occurs when running Python with the -bb flag and using Path(allow_dash=True).convert(). The original code compared rv in (b'-', '-') which mixes bytes and string comparisons. This causes BytesWarning when Python is run with -bb. The fix checks the type of the value first and compares with the appropriate type. Closes #2877

### DIFF
--- a/src/click/types.py
+++ b/src/click/types.py
@@ -973,12 +973,10 @@ class Path(ParamType):
     ) -> str | bytes | os.PathLike[str]:
         rv = value
 
-        # Check for dash without mixing bytes and str comparisons to avoid
-        # BytesWarning when running Python with -bb flag.
-        if isinstance(rv, bytes):
-            is_dash = self.file_okay and self.allow_dash and rv == b"-"
-        else:
-            is_dash = self.file_okay and self.allow_dash and rv == "-"
+        # Only compare with str "-" since value type is str | os.PathLike[str].
+        # The original code used `rv in (b"-", "-")` which caused BytesWarning
+        # when running Python with -bb flag (issue #2877).
+        is_dash = self.file_okay and self.allow_dash and rv == "-"
 
         if not is_dash:
             if self.resolve_path:
--- a/src/click/types.py
+++ b/src/click/types.py
@@ -972,14 +972,11 @@ class Path(ParamType):
         ctx: Context | None,
     ) -> str | bytes | os.PathLike[str]:
         rv = value
-        # Check for dash without mixing bytes and string comparison to avoid BytesWarning.
-        # When Python is run with -bb flag, comparing rv in (b"-", "-") would raise
-        # BytesWarning. Check type first to compare with the appropriate literal.
-        is_dash = (
-            self.file_okay
-            and self.allow_dash
-            and (rv == b"-" if isinstance(rv, bytes) else rv == "-")
-        )
+        # Use simple string comparison instead of `rv in (b"-", "-")` to avoid
+        # BytesWarning when Python runs with -bb flag (which treats bytes/str
+        # comparison as an error). The input type is str | os.PathLike[str],
+        # so bytes are not expected here.
+        is_dash = self.file_okay and self.allow_dash and rv == "-"
 
         if not is_dash:
             if self.resolve_path:
--- a/src/click/types.py
+++ b/src/click/types.py
@@ -973,7 +973,12 @@ class Path(ParamType):
     ) -> str | bytes | os.PathLike[str]:
         rv = value
 
-        is_dash = self.file_okay and self.allow_dash and rv in (b"-", "-")
+        # Check for dash without mixing bytes and str comparisons to avoid
+        # BytesWarning when running Python with -bb flag.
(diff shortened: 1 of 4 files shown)


src/click/types.py
```diff
@@ -973,7 +973,12 @@ def convert(
     ) -> str | bytes | os.PathLike[str]:
         rv = value
```
webknjaz, line 974: Typing says this is `str`. It's probably never `bytes`. The type checker would've caught it. So it's probably fine to drop the byte comparison. Though, there may be a case for explicitly checking and erroring our on unexpected input.

webknjaz (commented):
Additionally, such changes typically require tests. Also, it's probably a good idea to add `-bb` to the Python invocation that spawns `pytest`.

src/click/types.py
```diff
@@ -973,12 +973,10 @@ def convert(
...
 
-        # Check for dash without mixing bytes and str comparisons to avoid
-        # BytesWarning when running Python with -bb flag.
-        if isinstance(rv, bytes):
-            is_dash = self.file_okay and self.allow_dash and rv == b"-"
-        else:
-            is_dash = self.file_okay and self.allow_dash and rv == "-"
+        # Only compare with str "-" since value type is str | os.PathLike[str].
+        # The original code used `rv in (b"-", "-")` which caused BytesWarning
```
davidism: This comment isn't relevant to the code anymore.

tests/test_types.py
```diff
@@ -255,3 +255,28 @@ def test_choice_get_invalid_choice_message():
     choice = click.Choice(["a", "b", "c"])
     message = choice.get_invalid_choice_message("d", ctx=None)
     assert message == "'d' is not one of 'a', 'b', 'c'."
+
+
+def test_path_allow_dash_no_bytes_warning():
```
davidism, line 254: This test is no longer relevant, it's testing something that will never happen.

VedantMadane:
Removed the irrelevant comment and test. Thanks for the review!

tests/test_types.py
webknjaz: Isn't that method public? It might make sense to add a runtime check and error out. Or is that not worth it?

webknjaz:
> Removed the irrelevant comment and test.

The PR comment wasn't irrelevant, though. I'm pretty sure David was talking about just the code comment. Removing such comments makes PR context difficult to follow.

---

@davidism so what do you think about erroring out on byte warnings in CI? In scope or not?

VedantMadane:
### Changes Made:

**Simplified the fix:**
- The `Path.convert()` method signature is `value: str | os.PathLike[str]` which does NOT include `bytes`
- My `isinstance` check and bytes handling was unnecessary defensive programming for a case that shouldn't happen
- **Reverted to simple:** `rv == "-"` without the isinstance check

**Updated the test:**
- Removed the `b'-'` (bytes dash) test case as it tests invalid input per the type signature
- Kept the string dash test and regular path test
- Both verify no BytesWarning with `-bb` flag

**Kept the CI configuration:**
- `pyproject.toml` still runs tests with `python -bb` flag  
- `PYTHONWARNINGS=error::BytesWarning`
- This catches any future BytesWarning issues

All tests pass with `python -bb` flag.

ThiefMaster:
Do not use an LLM to reply to maintainer feedback. Thanks.

VedantMadane:
@ThiefMaster what is your opinion, erroring out on byte warnings in CI was out of scope or not?

davidism:
It's pretty clear you're using LLM tools instead of actually understanding the issue and code and making the appropriate changes. The change here should have been to one line of code and some clean up of comments, not something that required LLMs. This is a waste of energy and our time.

Issue #2899, opened 2025-05-11
running tests in parallel with `tox p` fails


davidism:
Running tests in parallel with `tox p` fails, meaning I have to run all local test envs serially right now. It seems that the detected terminal environment is different in parallel mode, causing the pager tests to fail.

```
FAILED tests/test_utils.py::test_echo_via_pager[test0-cat] - AssertionError: Unexpected pager output in test case 'Plain string argument'
```

davidism:
It worked until the changes to the pager, and only fail on the pager test. I made sure everything ran correctly with `tox p` in all projects, it makes pre-checking a PR way faster.

Rowlando13:
I agree. What did you do? I was just planning on looking some tests

Rowlando13:
There are multiple instances of this: 

```python
def test_file_args(runner):
    @click.command()
    @click.argument("input", type=click.File("rb"))
    @click.argument("output", type=click.File("wb"))
    def inout(input, output):
        while True:
            chunk = input.read(1024)
            if not chunk:
                break
            output.write(chunk)

    with runner.isolated_filesystem():
        result = runner.invoke(inout, ["-", "hello.txt"], input="Hey!")
        assert result.output == ""
        assert result.exit_code == 0
        with open("hello.txt", "rb") as f:
            assert f.read() == b"Hey!"

        result = runner.invoke(inout, ["hello.txt", "-"])
        assert result.output == "Hey!"
        assert result.exit_code == 0
```

Seems like the `isolated_filesystem()` would not be thread safe.

davidism:
Seems to work now.

kdeldycke:
> Running tests in parallel with `tox p` fails, meaning I have to run all local test envs serially right now. It seems that the detected terminal environment is different in parallel mode, causing the pager tests to fail.
> 
> ```
> FAILED tests/test_utils.py::test_echo_via_pager[test0-cat] - AssertionError: Unexpected pager output in test case 'Plain string argument'
> ```

This part has been already fixed by https://github.com/pallets/click/pull/3238 which is available in Click 8.3.3.

kdeldycke:
> Seems like the `isolated_filesystem()` would not be thread safe.

It is not. So I move this concern to a dedicated issue at: https://github.com/pallets/click/issues/3501

That way this current issue can be marked as closed and addressed in Click 8.4.2 once https://github.com/pallets/click/pull/3499 is merged, as it will make the tests properly running in parallels, as the original issue suggested.

PR #3190, opened 2026-01-21
remove subprocess.Popen, replace with fake_pager

itsmiketorres:

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -196,3 +196,84 @@ def test_prompts():
 Prompts will be emulated so they write the input data to
 the output stream as well. If hidden input is expected then this
 does not happen.
+
+## Running Click's Test Suite
+
+This section covers running Click's own test suite, which is useful for contributors.
+
+### Parallel Test Execution with Tox
+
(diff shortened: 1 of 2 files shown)


itsmiketorres:
#2899 Please take a look at this branch and lmk if this fix is acceptable. Not seeing hyperfast execution with tox -p auto (approx 10 processes on my laptop takes about 70-80s), seems a bit faster with setting processes to 4 or 6. All tests are passing though with this change (both initial and subsequent runs as previously only the latter were passing). 

FWIW Opus 4.5 located the issue, I was just steering the ship. Please feel free to take over from here and make fixes.

davidism:
Please do not use LLMs to generate PRs. This is a huge mess of code (and a bunch of docs that don't belong) that is impossible to review.

itsmiketorres:
FWIW I thought this was opened as a Draft

seems like the issue is with       

```
force_subprocess_stdout = patch.object(
            subprocess,
            "Popen",
            partial(subprocess.Popen, stdout=f),
        )
 ```

Hope this helps someone

PR #3191, opened 2026-01-22
Clarify option decorator argument wording

cnaples79:
## Summary
- clarify that option decorators accept one or more option names and can optionally specify the function argument name
- explain the common inference pattern for dashed option names mapping to underscored parameter names
- rewrite the basic example description to state the default optional behavior and type

## Rationale
- improves docs readability per issue #3175 by fixing confusing grammar and clarifying defaults

## Changes
- update docs/options.md with clearer language around argument inference and default option behavior

Tests
- not run (docs-only change)

Fixes #3175

Issue #3175, opened 2025-12-25
Improve documentation


Rowlando13:
Pick a section of the docs and improve it. Look for
- Typos
- inconsistencies - somewhere else in the docs it says something different 

Or improve 
- coverage of click's behaviors
- clarity of language 
- clarity of examples - each example should clearly and simply show the behavior explained 

Post on this issue which section you are taking

developer-rpai:
@Rowlando13 I've found few duplicates and problems with clarification of arguments, in General Reference section. Can you please assign me this issue, as I would like to cover more sections once initial iterations are approved

Rowlando13:
The general reference section is 20 plus files. Lets start with one. Which one or would you like to work on?

AmoghDeshmukh:
Hello, I want to contribute on this issue. Should I proceed with it?

developer-rpai:
> The general reference section is 20 plus files. Lets start with one. Which one or would you like to work on?

I found a typo and a logical inconsistency in documentation pages for `src/click/core.py` and `src/click/decorators.py` modules . Specifically in options and arguments CLI parameters.
Additionally, can you please clear this confusion?

1. Arguments class states that arguments are required by default when no default value is provided (and nargs > 0) and arguments become optional when a default value is set. But the note on required arguments (in `docs/arguments.rst`) states the following:
"It is possible to make an argument required by setting ``required=True``.  It is not recommended since we think command line tools should gracefully degrade into becoming no ops. "
- this implies that arguments are not required by default.

Mariakourbi:
Hi! I’ve read through the discussion above and I’d like to help work on this issue as my first contribution.
If no one is currently working on it, could you please let me know if I can take it and how you’d suggest I start?
Thank you!

cnaples79:
Working on docs/options.md to clarify option decorator argument wording (inference, defaults) for this issue.

brihall:
I know someone already submitted a PR for `options.md`, but there were some unaddressed grammar and readability issues that I thought deserved attention.

DevThilag:
Hi! I would like to improve the Quickstart section in the documentatation.

adenbad:
I would like to be added to this issue

Alita41:
I'd like to improve clarity and correct minor grammatical issues in the Parameters and Parameter Names sections of the documentation.

pdonizete:
Hi! I’d like to work on this issue. I’ll take **docs/prompts.md** (one file) to improve clarity and examples.

[...]

PR #3192, opened 2026-01-22
revise 'Options' documentation

brihall:
#3175 

If #3191 is approved I'd be happy to go back and handle any merge conflicts caused by this PR's changes.

### Clarity improvements
- The "Option Decorator" section refers to "positional arguments" of the `click.option()` decorator and to the "function argument" of the decorated callback function. Using "positional arguments" as shorthand for "positional arguments passed to the `option()` decorator" introduces ambiguity, because the decorated function can also accept positional arguments (and in fact does in every provided example). Using "function" as shorthand for "decorated function" is also ambiguous. References to the positional arguments for `click.option()` are changed to "parameter declarations" (for consistency with the API docs), and "function argument" is replaced with the more specific "callback argument".

- When describing the steps to infer the callback argument name from the parameter declarations, the following is misleading (possibly outdated?):

  > 1. If a positional argument name does not have a prefix, it is chosen.

  The actual criteria is that the argument is a valid identifier, i.e. `str.isidentifier()` returns `True` (see `Option._parse_decls()`). A string that is "prefixed" with `_` would therefore be chosen. The step is updated to reflect this and links to Python's "Lexical analysis" documentation. 

- Several sections use "pass in" back-to-back when referencing `option()` arguments and command line usage. This makes it unclear when the command line usage is being referenced. When referring to `option()`, instead use "declare" for positional arguments or "set" for keyword arguments.

### Consistency improvements
- All instances of "underlying function" are replaced with "decorated function".

- References to "dash(es)" are replaced with `-` or `--`. This establishes consistency with the "Other Prefix Characters" section and allows for several sentences to be reworded for concision.

### Miscellaneous
Minor grammar and concision edits throughout.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -24,7 +24,7 @@ Useful and often used kwargs are:
 
 ## Option Decorator
 
-The {func}`option()` decorator accepts parameter declarations as positional arguments. Parameter declarations can consist of any number of option names and, optionally, a single identifier matching an argument name in the decorated callback function.
+The {func}`option()` decorator is usually passed two positional arguments: the option name and the decorated function argument name.
 
 ```{eval-rst}
 .. click:example::
@@ -40,7 +40,7 @@ The {func}`option()` decorator accepts parameter declarations as positional argu
     invoke(echo, args=['--help'])
 ```
 
-If an identifier for the callback argument is not declared, then Click will try to infer it. A simple way to name an option is by prepending `--` to the callback argument name and converting underscores to `-`; this ensures that Click can correctly infer the callback argument name. Consequently, {func}`option()` only requires declarations for the option name(s).
+However, if the decorated function argument name is not passed in, then Click will try to infer it. A simple way to name the option so that Click will infer it correctly is by taking the function argument, adding two dashes to the front and converting underscores to dashes.
 
 ```{eval-rst}
 .. click:example::
@@ -55,13 +55,11 @@ If an identifier for the callback argument is not declared, then Click will try 
     invoke(echo, args=['--string-to-echo', 'Hi!'])
 ```
 
-More formally, Click will try to infer the callback argument name as follows:
+More formally, Click will try to infer the decorated function argument name as follows:
 
-1. If a positional argument is a valid [Python identifier], it is chosen.
-2. If multiple positional arguments are prefixed with `--`, the one that was declared first is chosen.
+1. If a positional argument is a valid [Python identifier](https://docs.python.org/3/reference/lexical_analysis.html#identifiers) (and thus does not have dashes), it is chosen.
+2. If multiple positional arguments are prefixed with `--`, the first one declared is chosen.
 3. Otherwise, the first positional argument prefixed with `-` is chosen.
-
-[Python identifier]: https://docs.python.org/3/reference/lexical_analysis.html#identifiers
 
 To get the argument name, the chosen positional argument is converted to lower case, a leading `-` or `--` is removed if found, and any remaining `-` characters are replaced with `_`.
 
@@ -178,7 +176,7 @@ And on the command line:
     invoke(putitem, args=['--item', 'peter', '1338'])
 ```
 
-By using a tuple literal as `type`, `nargs` gets automatically set to the
+By using a tuple literal as the `type`, `nargs` gets automatically set to the
 length of the tuple and the {class}`click.Tuple` type is automatically
 used. The above example is thus equivalent to this:
 
@@ -456,7 +454,7 @@ every colon (`:`), and for Windows, splitting on every semicolon (`;`).
 
 ## Other Prefix Characters
 
-Click can deal with prefix characters besides `-` for options, including `/` and `+`. Note that alternative prefix characters are generally used very sparingly if at all within POSIX.
+Click can deal with prefix characters besides `-` for options, including `/` and `+`, as well as others. Note that alternative prefix characters are generally used very sparingly if at all within POSIX.
 
 ```{eval-rst}
 .. click:example::


Rowlando13:
I will have to take a closer look at this. At first glance you made some good grammatical fixes, but also made some changes that obscure what is meant. Sometimes it's best not to use the most technical terms available because people may not know what you mean.

brihall:
I appreciate the feedback, and after re-reading I definitely agree that some of the changes need more work (particularly the Option Decorator section). I have some ideas for improvements that might be a nicer middle ground, but I'll wait for your closer review before making any changes.

docs/options.md
```diff
@@ -24,7 +24,7 @@ Useful and often used kwargs are:
 
 ## Option Decorator
 
-Click expects you to pass at least two positional arguments to the option decorator. They are option name and function argument name.
+The {func}`option()` decorator accepts parameter declarations as positional arguments. Parameter declarations can consist of any number of option names and, optionally, a single identifier matching an argument name in the decorated callback function.
```
Rowlando13, line 28: The {func}`option()` decorator is usually passed two positional arguments, the option name and the decorated function name.

docs/options.md
```diff
@@ -40,7 +40,7 @@ Click expects you to pass at least two positional arguments to the option decora
     invoke(echo, args=['--help'])
 ```
 
-However, if you don't pass in the function argument name, then Click will try to infer it. A simple way to name your option is by taking the function argument, adding two dashes to the front and converting underscores to dashes. In this case, Click will infer the function argument name correctly so you can add only the option name.
+If an identifier for the callback argument is not declared, then Click will try to infer it. A simple way to name an option is by prepending `--` to the callback argument name and converting underscores to `-`; this ensures that Click can correctly infer the callback argument name. Consequently, {func}`option()` only requires declarations for the option name(s).
```
Rowlando13, line 44: However, if the decorated function name is not passed in, then Click will try to infer it. A simple way to name the option so that Click will infer it correctly is by taking the function argument, adding two dashes to the front and converting underscores to dashes

docs/options.md
```diff
@@ -24,7 +24,7 @@ Useful and often used kwargs are:
```
Rowlando13, line 24: I generally avoid using callback unless I have to since a lot of python developers are not software engineers and so probably won't know what it means.

docs/options.md
```diff
@@ -55,21 +55,21 @@ However, if you don't pass in the function argument name, then Click will try to
     invoke(echo, args=['--string-to-echo', 'Hi!'])
 ```
 
-More formally, Click will try to infer the function argument name by:
+More formally, Click will try to infer the callback argument name as follows:
```
Rowlando13, line 59: More formally, Click will try to infer the decorated function name as follows:

Rowlando13:
@brihall I will got through some of the PR. Will hopefully get to the rest sometime soon. Don't be put off by the lots of changes, I like what I see so far!

docs/options.md
brihall: Agreed that "decorated function" is better than "callback function"

docs/options.md
brihall: Yes, this is much more concise than the initial change. I think it's also important to refer to an "argument name" rather than just "decorated function name". To me, the "decorated function name" in the example is `echo`, not `string_to_echo`.

```suggestion
The {func}`option()` decorator is usually passed two positional arguments: the option name and the argument name from the decorated function.
```

docs/options.md
brihall: With the proposed change to line 27, is simply "argument name" clear enough? "Decorated function argument name" is more specific but also a bit of a mouthful.

```suggestion
However, if the argument name is not passed in, then Click will try to infer it from the option name. A simple way to name the option so that Click will infer it correctly is by taking the function argument, adding two dashes to the front and converting underscores to dashes.
```

docs/options.md
brihall:
```suggestion
More formally, Click will try to infer the argument name as follows:
```

brihall:
@Rowlando13 Not put off at all. I really appreciate you taking the time to give feedback, and the changes were definitely necessary. I applied some small changes in the unreviewed sections so that they're more aligned with your first round of suggestions.

docs/options.md
Rowlando13: The {func}option() decorator is usually passed two positional arguments, the option name and the decorated function argument name.

docs/options.md
Rowlando13: However, if the decorated function argument name is not passed in, then Click will try to infer it. A simple way to name the option so that Click will infer it correctly is by taking the function argument, adding two dashes to the front and converting underscores to dashes

Rowlando13: I prefer to be specific in this case since people jump in and out of the docs.

docs/options.md
Rowlando13: More formally, Click will try to infer the decorated function argument name as follows:

docs/options.md
```diff
@@ -55,21 +55,23 @@ However, if you don't pass in the function argument name, then Click will try to
...
 
-1. If a positional argument name does not have a prefix, it is chosen.
-2. If a positional argument name starts with with two dashes, the first one given is chosen.
-3. The first positional argument prefixed with one dash is chosen otherwise.
+1. If a positional argument is a valid [Python identifier], it is chosen.
```
Rowlando13: If a positional argument is a valid [Python identifier] (and thus does not have dashes), it is chosen.

docs/options.md
```diff
@@ -55,21 +55,23 @@ However, if you don't pass in the function argument name, then Click will try to
...
+2. If multiple positional arguments are prefixed with `--`, the one that was declared first is chosen.
```
Rowlando13: If multiple positional arguments are prefixed with `--`, the first one declared is chosen.

docs/options.md
```diff
@@ -55,21 +55,23 @@ However, if you don't pass in the function argument name, then Click will try to
...
+3. Otherwise, the first positional argument prefixed with `-` is chosen.
 
-The chosen positional argument is converted to lower case, up to two dashes are removed from the beginning, and other dashes are converted to underscores to get the function argument name.
+[Python identifier]: https://docs.python.org/3/reference/lexical_analysis.html#identifiers
```
Rowlando13: Inline this. Python identifier is not likely to be reused.

docs/options.md
```diff
@@ -177,7 +178,7 @@ And on the command line:
     invoke(putitem, args=['--item', 'peter', '1338'])
 ```
 
-By using a tuple literal as type, `nargs` gets automatically set to the
+By using a tuple literal as `type`, `nargs` gets automatically set to the
```
Rowlando13: By using a tuple literal as the `type`, `nargs` gets automatically set to the

docs/options.md
```diff
@@ -321,7 +322,7 @@ If you want to define an alias for the second option only, then you will need to
 
 ## Flag Value
 
-To have an flag pass a value to the underlying function set `flag_value`. This automatically sets `is_flag=True`. To mark the flag as default, set `default=True`. Setting flag values can be used to create patterns like this:
+To have a flag pass a value to the decorated function set `flag_value`. This automatically sets `is_flag=True`. To mark the flag as default, set `default=True`. Setting flag values can be used to create patterns like this:
```
Rowlando13, line 323: Stopped reviewing here.

docs/options.md
```diff
@@ -457,8 +456,7 @@ every colon (`:`), and for Windows, splitting on every semicolon (`;`).
 
 ## Other Prefix Characters
 
-Click can deal with prefix characters besides `-` for options. Click can use
-`/`, `+` as well as others. Note that alternative prefix characters are generally used very sparingly if at all within POSIX.
+Click can deal with prefix characters besides `-` for options, including `/` and `+`. Note that alternative prefix characters are generally used very sparingly if at all within POSIX.
```
Rowlando13: Click can deal with prefix characters besides `-` for options, including `/` and `+`, as well as others.

Rowlando13: This is an intentional effort to document / and +, and that there are other but not name them. As it stands many other prefix characters are supported, but I would like to keep it an implementation detail.

Rowlando13:
@brihall Review complete.

brihall:
@Rowlando13 Thanks! Pushed your suggestions

kdeldycke:
@Rowlando13 is that one ready to be merged into `stable` for 8.3.2?

Rowlando13:
@kdeldycke It is. I just have not had a chance to confirm all the requested changes were made so I can click merge.

Rowlando13:
@brihall I finally got a chance to do a final read through and hit merge. I really appreciate the effort. This section is way better on account of you help.

PR #3199, opened 2026-01-31
fix: lookup_default returns None instead of UNSET sentinel

### LINKED ISSUE #3145: `lookup_default` returns `Sentinel.UNSET` instead of `None`

# Reproduce

```python
#!/usr/bin/env -S uv run
# /// script
# ///
"""
Run with uv installed:
"""
import importlib.metadata
import sys
import click

        if call:
        else:

        if default is not None:
            return default
        if group:
            return group.get(name)
        return default

def describe(value):
    cls = value.__class__
    return {
        "repr": repr(value),
    }

def main():
    ctx.default_map = default_map

    info = describe(val)
    assert val is None, (
    )

if __name__ == "__main__":
    try:
        main()
    except AssertionError as e:
        print("AssertionError:", e)
        sys.exit(1)

```

# Expected

Environment:

- Python version: 3.14
- Click version: 8.30, 8.3.1

atulk-code:
Context.lookup_default() is a public API method that users can override or call directly. In Click 8.3.0, commit 1c20dc6 deferred UNSET normalization which caused lookup_default to return the internal UNSET sentinel instead of None when a parameter was not in the default_map.

This broke the public API contract - users expect None, not an internal sentinel object.

This fix:
- Adds internal _lookup_default() method that returns UNSET for internal use
- Modifies public lookup_default() to normalize UNSET to None
- Updates internal callers to use _lookup_default()
- Adds regression tests to verify the fix

Fixes #3145

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -702,8 +702,28 @@ class Context:
         :param call: If the default is a callable, call it. Disable to
             return the callable instead.
 
+        .. versionchanged:: 8.3.2
+            Returns ``None`` instead of internal sentinel when no default
+            is found in the default map.
+
         .. versionchanged:: 8.0
             Added the ``call`` parameter.
+        """
+        value = self._lookup_default(name, call)
+        # Normalize UNSET to None for public API compatibility.
+        # Internal code uses _lookup_default directly to distinguish
+        # between "no value" (UNSET) and "explicitly set to None".
+        if value is UNSET:
+            return None
+        return value
+
+    def _lookup_default(self, name: str, call: bool = True) -> t.Any:
+        """Internal method to get default from default_map.
+
+        Returns UNSET sentinel if no default is found. Use lookup_default()
+        for public API which normalizes UNSET to None.
+
+        :meta private:
         """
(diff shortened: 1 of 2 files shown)


kdeldycke:
This looks a bit AI-generated. Don't know what pallets policy is on that front. In the mean time @atulk-code, can you confirm or deny it?

As for the solution, it is not bad. But the problem is adding a private `_lookup_default()` method, then referencing it in the code defeat the purpose of having a public `lookup_default()` in Click API. What I mean is that we expect Click itself to invoke the public `lookup_default()` everywhere in its code, so the user has the opportunity to alter its behavior. This is exactly what the user was complaining about in #3145.

You should add a unittest that is reproducing #3145 exactly, with its custom `CustomClickContext. lookup_default` method to understand the limitations of your solution.

davidism:
This is like the 6th person to AI generate this PR. I'm not interested in that.

kdeldycke:
I'm back from other rabbit holes so I'll try to get back to work on Click. I guess I can fix that for good based on my feedback.

PR #3208, opened 2026-02-14
fix help hint showing shadowed option name

### LINKED ISSUE #2790: Nested command option shadowing -h prints wrong suggestion
Click prints an incorrect suggestion when using a combination of nested command and required argument. This happens when the nested command shadows the `-h` option shortcut of the main command which Click apparently tolerates.

## Replication

See https://github.com/tumidi/click-subcommand-help-message.

```
$ click-test-cli foo
Usage: click-test-cli foo [OPTIONS] REQUIRED_ARG
Try 'click-test-cli foo -h' for help.

Error: Missing argument 'REQUIRED_ARG'.
```

But, actually running the suggested `click-test-cli foo -h` prints

```
$ click-test-cli foo -h
```

Interestingly, Click correctly detects the shadowing of the option when printing the foo command help and prints a consistent help page.

```
$ click-test-cli foo --help
Usage: click-test-cli foo [OPTIONS] REQUIRED_ARG

Options:
  -h, --host TEXT
```

### Python code

```py
@click.group(context_settings={"help_option_names": ["-h", "--help"]})
def cli():
    pass


@click.command()
@click.argument("required_arg")
@click.option("--host", "-h", "host", default="localhost")
def foo(required_arg, host):
    click.echo(f"Foo command required_arg={required_arg} host={host}")


cli.add_command(foo)
```

## Possible solutions

Fix the generated message, so it prints `Try 'click-test-cli foo --help' for help.` in this case.

Environment:

- Python version: 3.12.6
- Click version: 8.2.0.dev0 d73083e

Mr-Neutr0n:
Fixes #2790.

When a subcommand defines an option that shadows one of the configured help option names (e.g. `-h`), the error hint was still suggesting the shadowed name:

```
Try 'cli foo -h' for help.
```

But `-h` is now the subcommand's `--host` option, not help. Running `cli foo -h` gives `Error: Option '-h' requires an argument.` instead of showing help.

The fix uses `get_help_option_names()` which already exists and filters out names claimed by other parameters. So now the hint picks a working alternative like `--help`:

```
Try 'cli foo --help' for help.
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,35 @@
 .. currentmodule:: click
 
+Version 8.4.0
+-------------
+
+Unreleased
+
+-   :class:`ParamType` typing improvements. :pr:`3371`
+
+    -   :class:`ParamType` is now a generic abstract base class,
+        parameterized by its converted value type.
+    -   :meth:`~ParamType.convert` return types are narrowed on all
+        concrete types (``str`` for :class:`STRING`, ``int`` for
+        :class:`INT`, etc.).
+    -   :meth:`~ParamType.to_info_dict` returns specific
+        :class:`~typing.TypedDict` subclasses instead of
+        ``dict[str, Any]``.
+    -   :class:`CompositeParamType` and the number-range base are now
+        generic with abstract methods.
+-   Split string values from ``default_map`` for parameters with ``nargs > 1``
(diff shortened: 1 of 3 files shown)


kdeldycke:
I just reworked that PR to be more resilient against various cases of default option names overriding. Also added unittests to illustrate all these edge-cases.

PR #3210, opened 2026-02-17
Add glossary of terms to documentation

### LINKED ISSUE #3077: Add glossary of terms to click reference docs
Add a section to the general reference docs for glossary of terms. 
Should contain ambiguous terms that defines how we use them in our docs. It allows us to be internally consistent in our usage 
- Option 
- flag 
- command line application 
- command line utility 
- terminal

by22Jy:
## Summary
Adds a glossary section to Click's documentation defining key terms used throughout the docs.

## Changes
- Added `docs/glossary.rst` with definitions for 10 key terms
- Updated `docs/index.rst` to include the glossary in General Reference section

## Terms Defined
- **Argument**: Positional parameter passed to a command
- **Callback**: Function invoked when a parameter is processed
- **Command**: Function decorated with `@click.command()` or `@click.group()`
- **Command Line Interface (CLI)**: Text-based interface for program interaction
- **Context**: Object carrying state between commands
- **Flag**: Boolean option without a value
- **Group**: Command containing subcommands
- **Multi-command**: Command grouping multiple subcommands
- **Option**: Named parameter with dash prefix
- **Parameter**: General term for options and arguments

Closes #3077

## Checklist
- [x] Documentation builds correctly
- [x] Follows existing documentation style
- [x] Terms are clear and consistent with Click's usage

### DIFF
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -17,7 +17,7 @@ This glossary defines key terms used throughout Click's documentation.
        See :doc:`/parameters` for more information.
 
    Command
-       A function decorated with ``@click.command()`` or ``@click.group()`` 
+       A function decorated with ``@click.command()`` or ``@click.group()``
        that can be invoked from the command line. Commands can accept options
        and arguments.
 
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -21,9 +21,19 @@ This glossary defines key terms used throughout Click's documentation.
        that can be invoked from the command line. Commands can accept options
        and arguments.
 
+   Command Line Application
+       A Python program built with Click that provides a command line interface.
+       Command line applications use commands, options, and arguments to interact
+       with users through the terminal.
+
    Command Line Interface (CLI)
        A text-based interface for interacting with a program through commands,
        options, and arguments. Click helps you build CLIs in Python.
+
+   Command Line Utility
+       A command line application designed to perform a specific task or set of
+       tasks. Command line utilities are often used for system administration,
+       file manipulation, or development workflows.
 
    Context
        An object that carries state between commands in a group hierarchy.
@@ -51,3 +61,8 @@ This glossary defines key terms used throughout Click's documentation.
        The general term for both options and arguments that can be passed
        to a command. Parameters are defined using decorators like
        ``@click.option()`` or ``@click.argument()``.
+
+   Terminal
+       A text-based interface for interacting with the operating system.
+       Also known as a command prompt, console, or shell. Command line
+       interfaces are used within a terminal to run commands and applications.
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -1,0 +1,53 @@
+Glossary
+========
+
+This glossary defines key terms used throughout Click's documentation.
+
+.. glossary::
+   :sorted:
+
+   Argument
+       A positional parameter passed to a command without a name prefix.
+       Arguments are identified by their position in the command line.
+       For example, in ``mycli command file.txt``, ``file.txt`` is an argument.
+
+   Callback
+       A function that is invoked when a parameter is processed. Callbacks can be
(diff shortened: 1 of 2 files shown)


nrschroeder:
This looks good ... the glossary seems like it will make the docs a lot clearer.

One thing I noticed while reading the issue is that it also mentioned terms like “terminal”, “command line application”, and “command line utility”. Since those show up in different places in the docs as well, it might be helpful to include them here too for completeness.

by22Jy:
Hi @nrschroeder, thank you for the great suggestion! I've added the three additional terms you mentioned:

- **Terminal**: A text-based interface for interacting with the operating system
- **Command Line Application**: A Python program built with Click that provides a CLI
- **Command Line Utility**: A command line application designed to perform specific tasks

These terms will definitely help make the documentation more comprehensive. Let me know if you'd like any adjustments to the definitions!

docs/glossary.rst
```diff
@@ -0,0 +1,68 @@
...
+       A boolean option that doesn't take a value (e.g., ``--verbose`` or ``-v``).
+       Flags are either present (True) or absent (False).
+
+   Group
+       A command that contains other commands (subcommands). Created with
+       ``@click.group()``. Groups allow organizing commands hierarchically.
+
+   Multi-command
+       A command that groups multiple subcommands together. Groups and
+       multi-commands are similar concepts in Click.
```
AndreasBackx, line 53:
```suggestion

   Multi-command
       Alias for Group, used to have a slightly different implementation.
       Planned for removal in 9.0. 
```

docs/glossary.rst
```diff
@@ -0,0 +1,68 @@
...
+       tasks. Command line utilities are often used for system administration,
+       file manipulation, or development workflows.
+
+   Context
+       An object that carries state between commands in a group hierarchy.
+       The context is created for each command invocation and can be accessed
+       via the ``@click.pass_context`` decorator.
+
+   Flag
+       A boolean option that doesn't take a value (e.g., ``--verbose`` or ``-v``).
```
AndreasBackx, line 44: I'm not too familiar with how glossaries work in Sphinx. But will "option" here link to the glossary for that word down below automatically or does that need to be done explicitly here and elsewhere in the code? I hope it's the latter.

docs/glossary.rst
```diff
@@ -0,0 +1,68 @@
...
+       and arguments.
+
+   Command Line Application
+       A Python program built with Click that provides a command line interface.
+       Command line applications use commands, options, and arguments to interact
+       with users through the terminal.
+
+   Command Line Interface (CLI)
+       A text-based interface for interacting with a program through commands,
+       options, and arguments. Click helps you build CLIs in Python.
```
AndreasBackx, line 31: Correct me if I'm wrong here. But I feel like "CLI" is common nomenclature for the application, regardless of what the true acronym stands for.

```suggestion
   Command-line Application
       Application meant to be interacted with using a command-line interface.
       Often referred to as a CLI as well, even though it's not the true acronym. 

   Command-line Interface (CLI)
       A text-based interface for interacting with a program through commands,
       options, and arguments. While CLI is its acronym, a CLI often refers to
       an application written for the command-line interface.
```

docs/glossary.rst
```diff
@@ -0,0 +1,68 @@
...
+
+   Command Line Utility
+       A command line application designed to perform a specific task or set of
+       tasks. Command line utilities are often used for system administration,
+       file manipulation, or development workflows.
+
```
AndreasBackx, line 37:
```suggestion
```

Personally I'd remove this to avoid confusion.

AndreasBackx (changes requested):
Would love to primarily get some input from others here. I think this is great to have however, thank you!

davidism:
Unfortunately gen AI, which we don't accept

PR #3211, opened 2026-02-17
`FuncParamType` should use `ValueError` message in `self.fail()`

### LINKED ISSUE #3105: FuncParamType should use ValueError for self.fail(message)
FuncParamType currently just calls self.fail(value,...) if it self.func(value) raise a ValueError.  This is missing the point of the self.fail which is to give better feedback.  
Using typer so something like 
```python

@dataclass
class MyClass:
    myvalue: str
     @staticmethod
      def from_str(cls, input:Any) -> 'MyClass':
              if input != 'works':
                  raise ValueError(f'input was  {input=}, should be works')
              return cls(input)
@app.command('test', help='get snapshot')
async def test(
    mydata: Annotated[
        MyClass, typer.Argument(parser=MyClass.from_str)
    ],
) -> None:
```

should give str(error) not str(input)

fix would be 
```python
class FuncParamType(ParamType):
    def __init__(self, func: t.Callable[[t.Any], t.Any]) -> None:
        self.name: str = func.__name__
        self.func = func

    def to_info_dict(self) -> dict[str, t.Any]:
        info_dict = super().to_info_dict()
        info_dict["func"] = self.func
        return info_dict

    def convert(
        self, value: t.Any, param: Parameter | None, ctx: Context | None
    ) -> t.Any:
        try:
            return self.func(value)
        except ValueError as e:
            self.fail(str(e), param, ctx)
```
happy to submit a PR if this makes sense.  Rather not have to create a customclass when `FuncParamType` is so close.  Not sure what else `FuncParamType` there there for, as there isn't really any documentation of why it exists, but typer uses it.

StevOti:
Fixes #3105

When FuncParamType.convert() caught a ValueError, it discarded the exception message and only passed the input value to self.fail(). This meant users lost context about why the conversion failed.

This fix captures and uses the ValueError message, providing better error feedback to users. Falls back to the input value only if the message is empty (backward compatibility).

Includes regression test ensuring the message is surfaced in the BadParameter exception.


AndreasBackx (approved):

AndreasBackx:
@kdeldycke It seems like you've wanted to include this in stable instead of main so I updated the CHANGES and added 8.3.4 but won't merge it yet as I'm not up to date with plans re 8.3.4.

Rowlando13:
The current plan is to release 8.4.0. No more 8.3.x.

kdeldycke:
> @kdeldycke It seems like you've wanted to include this in stable instead of main so I updated the CHANGES and added 8.3.4 but won't merge it yet as I'm not up to date with plans re 8.3.4.

I manually changed the target of all PRs from `main` to `stable` after the discussion in https://github.com/pallets/click/discussions/3337 . The idea was to evaluate them for inclusion for the 8.3.3 release, which targetted `stable`. That's how BTW I ended up realizing your old `typing/parameter` and `typing/paramtype` branches were doable for inclusion with some effort to clean them up (see: https://github.com/pallets/click/discussions/3329).

Then @Rowlando13 switch the merging activity from `stable` to `main` a couple of days ago to quickstart the 8.4.0 release cycle. At least that's my interpretation: https://github.com/pallets/click/pull/3371#issuecomment-4349966413 .

So this PR should go back to target `main` if we want to merge it to 8.4.0.

Rowlando13:
@kdeldycke @AndreasBackx I don't think we have a particular process for deciding the next release. I have just been eyeballing the PRs that @kdeldycke has done or reviewed and said what I think the next release should be. With this one I was planning to do a 8.3.x release, but some of the work definitely required a feature release and main and stable were just synchronized, so just easier to do it all on main, so they can all go in.

tests/test_types.py
```diff
@@ -51,6 +51,18 @@ def test_range_fail(type, value, expect):
     assert expect in exc_info.value.message
 
 
+def test_func_param_type_uses_value_error_message():
```
kdeldycke: @StevOti can you `@parametrize` the test you added here to test the empty message fallback?

StevOti: I've updated the test to use @pytest.mark.parametrize, covering both the non-empty message case and the empty message fallback.

kdeldycke:
Thanks @StevOti for the fix. I'll merge that upstream.

PR #3212, opened 2026-02-17
Fix lookup_default return None

### LINKED ISSUE #3145: `lookup_default` returns `Sentinel.UNSET` instead of `None`

StevOti:
Fixes #3145

### Problem
In Click >= 8.3, `Context.lookup_default()` returns the internal
`Sentinel.UNSET` object when a key is missing. Historically (<= 8.2.1),
the public API returned `None`.

### Reproduction
- main: `ctx.lookup_default("missing_key") -> Sentinel.UNSET`
- Click 8.2.1: `ctx.lookup_default("missing_key") -> None`

### Root Cause
Regression introduced by 1c20dc6 (“Fix default handling to defer UNSET normalization ”)
Let the internal sentinel leak through the public API.

### Solution
- Add internal `_lookup_default()` that returns `UNSET` for internal use
- Normalize public `lookup_default()` to return `None`
- Update internal call sites to use `_lookup_default()`
- Add regression tests covering missing keys, callable defaults, and internal behavior

### Safety
- Internal sentinel semantics preserved
- Regression tests added
- Full test suite run locally (see note below)

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -685,6 +685,27 @@ class Context:
             self.obj = rv = object_type()
         return rv
 
+    def _lookup_default(self, name: str, call: bool = True) -> t.Any | object:
+        """Internal method for looking up defaults. Returns Sentinel.UNSET
+        if not found, for use by internal library code that needs to
+        distinguish between "not set" and "set to None".
+
+        :param name: Name of the parameter.
+        :param call: If the default is a callable, call it. Disable to
+            return the callable instead.
+
+        .. versionadded:: 8.3
+        """
+        if self.default_map is not None:
+            value = self.default_map.get(name, UNSET)
+
+            if call and callable(value):
+                return value()
+
+            return value
+
+        return UNSET
+
     @t.overload
     def lookup_default(
         self, name: str, call: t.Literal[True] = True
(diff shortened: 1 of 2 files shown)


StevOti:
I verified the regression and historical behaviour:

- main: lookup_default("missing_key") -> Sentinel.UNSET
- Click 8.2.1: lookup_default("missing_key") -> None

Full test suite on Windows/Python 3.14.2: 1 failure in tests/test_imports.py::test_light_imports
because ctypes._layout is imported (not in ALLOWED_IMPORTS). This seems unrelated
to the lookup_default change; happy to rerun in a different env if preferred.

PR #3217, opened 2026-02-19
Fix typo: principle to principal in documentation

realMelTuc:
Fixes a typo where "principle" was incorrectly used instead of "principal" in the parameters documentation.

### DIFF
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -5,7 +5,7 @@
 ```{currentmodule} click
 ```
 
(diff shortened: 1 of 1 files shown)


ThiefMaster:
Doc fixes should go into the `stable` branch; please rebase (`git rebase --onto stable main`) and change the target branch of your PR (edit button next to the title).

davidism:
The real fix here would be to remove that word altogether, it implies there's non-principal parameter types as well, which there aren't. Go ahead and make a new PR against stable that does that instead.

PR #3224, opened 2026-02-20
Hide `Sentinel.UNSET` values as `None` in `lookup_default()`

### LINKED ISSUE #3145: `lookup_default` returns `Sentinel.UNSET` instead of `None`

kdeldycke:
This PR address issue #3145, in which the `UNSET` sentinel is leaking into `lookup_default()` when users have overriding it.

This changes prevent `lookup_default()` to return `UNSET` as-is, and masquerade it as `None`. To account for this change in `lookup_default()` behavior, I had to update some internal checks around defaults.

Additionally, to prevent any side-effects, I added a couple of tests to fill the gaps in coverage of `lookup_default()`, especially around treatment of absence of value and value of absence.

> [!IMPORTANT]
> I did use LLMs (Claude Opus 4.6) in the process of creating this PR, but as a co-worker to criticize my changes and modifications, not as a vibe-coding agent. This PR was carefully written, edited and reviewed by me.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Unreleased
 
 -   Fix handling of ``flag_value`` when ``is_flag=False`` to allow such options to be
     used without an explicit value. :issue:`3084`
+-   Hide ``Sentinel.UNSET`` values as ``None`` when using ``lookup_default()``.
+    :issue:`3136` :pr:`3199` :pr:`3202` :pr:`3209` :pr:`3212` :pr:`3224`
 
 Version 8.3.1
 --------------
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -706,14 +706,14 @@ class Context:
             Added the ``call`` parameter.
         """
         if self.default_map is not None:
-            value = self.default_map.get(name, UNSET)
+            value = self.default_map.get(name)
 
             if call and callable(value):
                 return value()
 
             return value
 
-        return UNSET
+        return None
 
     def fail(self, message: str) -> t.NoReturn:
         """Aborts the execution of the program with a specific error
@@ -2278,9 +2278,12 @@ class Parameter:
         .. versionchanged:: 8.0
             Added the ``call`` parameter.
         """
-        value = ctx.lookup_default(self.name, call=False)  # type: ignore
+        name = self.name
+        value = ctx.lookup_default(name, call=False) if name is not None else None
 
-        if value is UNSET:
+        if value is None and not (
+            ctx.default_map is not None and name is not None and name in ctx.default_map
+        ):
             value = self.default
 
         if call and callable(value):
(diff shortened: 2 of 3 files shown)


kdeldycke:
This PR replace AI-slop from: #3199, #3202, #3209, #3212 and #3215.

kdeldycke:
Also note that the unnatural ``None`` dance around the results of ``lookup_default`` were made to please ``mypy``.

davidism:
This is fine for the bug fix, but you're right that this is getting messy. We need to figure out a long term solution.

kdeldycke:
@Rowlando13, should we merge? to `main` or `stable`?

Rowlando13:
Main. @kdeldycke Getting this fix is good, but it would be better if we try to make larger fixes to simplify the issues Click has accumulated.

Rowlando13:
9.0 should be next release.

kdeldycke:
Looks like @davidism is in favor on Discord of getting 8.3.2 out first, which could include this fix, which as limited blast radius. 😁

kdeldycke:
FYI, just tested this fix against my own Click Extra codebase and could not find any regression: https://github.com/kdeldycke/click-extra/commit/4bbd3875e291f5035f43c5d160126b90a8031c63

Rowlando13:
Sounds good.

sankarcn:
If the context setting is defined like below:
CONTEXT_SETTINGS = dict(
    default_map={'runserver': {'port': UNSET}}
)
And we set the option for the command like below
@click.option('--port', default=8000)

The default for port is picked as None instead of 8000

davidism:
`UNSET` is not a public API, so I'm not clear why you'd use that. Don't add the key to the default map if you don't want to change the default.

kdeldycke:

kdeldycke:
> `UNSET` is not a public API, so I'm not clear why you'd use that. Don't add the key to the default map if you don't want to change the default.

This is a reasonable assumption as in dict, you can materialize the absence of value by the absence of the key. But even if @sankarcn is a bot, it can still point to a gap in the tests around `default_map`. I will explore that in https://github.com/pallets/click/pull/3240

sankarcn:
@kdeldycke I am a human. Sorry, I could not reply to you and considered your reply as a guidance and I did not have anything to add. This was one of the edge cases, I came across when testing the PR, for a different purpose.

kdeldycke:
> @kdeldycke I am a human. Sorry, I could not reply to you and considered your reply as a guidance and I did not have anything to add. This was one of the edge cases, I came across when testing the PR, for a different purpose.

Ahah sorry, paranoia is high these days among open-source maintainers. Maybe you can tell us more about this use case. If you stumble upon it, there is a high chance others will. So I'd like to make a case so we can reduce the blast-radius of `UNSET`. Let's discuss this at: https://github.com/pallets/click/pull/3240

PR #3225, opened 2026-02-21
Fix callable ``flag_value`` being instantiated when used as a default

### LINKED ISSUE #3121: Using a class as a `flag_value` causes the default to be instantiated in click 8.3.0
Consider the following code:

```python
import click

class Foo:
    pass

class Bar:
    pass

@click.command()
@click.option("--foo", "ty", flag_value=Foo, type=click.UNPROCESSED, default=True)
@click.option("--bar", "ty", flag_value=Bar, type=click.UNPROCESSED)
def main(ty):
    click.echo(repr(ty))

if __name__ == "__main__":
    main()
```

When run with no arguments using click 8.3.0, this prints `<__main__.Foo object at 0x10bed92b0>`, indicating that `Foo` has been instantiated, which is not what I want.  When using an older version of click, the script outputs `<class '__main__.Foo'>`, which is what I want.

If the script is passed an explicit `--foo` or `--bar` option, a class is output, indicating that the problem lies with the default handling.

Changing the `default=True` to `default=Foo` does not help.

kdeldycke:
This fix issue #3121, where a `flag_value` set to a class produce an instance of that class in the particular case that it is explicitly set as the default.

The fix consist in lazily resolving the default value at the time of use, with a branch dedicated to callables.

I extended the unittests to `default_map` to demonstrate the current state of callable presence in them.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ Unreleased
     logging interaction, multi-threaded safety, and sequential invocation
     isolation. Add high-iteration stress tests behind a ``stress`` marker
     with a dedicated CI job. :pr:`3139`
+-   Fix callable ``flag_value`` being instantiated when used as a default via
+    ``default=True``. :issue:`3121` :pr:`3201` :pr:`3213` :pr:`3225`
 
 Version 8.3.1
 --------------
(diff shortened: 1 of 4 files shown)


kdeldycke:
This PR replace AI-slop from: #3201 and #3213

Rowlando13:
@kdeldycke Can you please fix merge conflicts.

kdeldycke:
@Rowlando13 rebased and ready to merge! :)

kdeldycke:
Had to rebase it once again. Done.

kdeldycke:

PR #3227, opened 2026-02-21
Enhance command pipelines documentation

DevThilag:
Updated documentation for command pipelines and example usage.


davidism:
This doesn't appear to be an improvement, especially the change to the group example.

PR #3233, opened 2026-02-27
merge main into stable to synchronize commits.

Rowlando13:


davidism:
This doesn't seem right. The only time main merges into stable is when a new feature release is created. Stable (fix branch) should only merge up to main (feature branch).

Rowlando13:
I was aware of that. The commits don't have any file changes. Is that an issue?

davidism:
Huh, how did that happen? Is stable somehow already synced with main? I guess as long as the 8.3.2 work is still distinct, everything's fine. Just weird.

Issue #2847, opened 2025-01-23
Long option completion with = broken


brandon-avantus:
When using Click shell completion in bash and zsh, long option completion is not working properly when = is used to separate the option from the value (e.g., `command --long-option=value` vs. `command --long-option value`). In zsh, the option being completed is replaced with the incomplete value or matched value, essentially "gobbling" up the option. In bash, the option isn't completed at all. I did not test fish.

## Steps to reproduce the issue

1. Create the following script and place it in the bin directory of an activated virtual environment with *click* installed:

```python
# gobble

#!/usr/bin/env python
import click


@click.command
@click.option("--color", type=click.Choice(["auto", "always", "never"]), default=None,
              help="Control colors in output.")
@click.option("--name",
              help="Name of something.")
def main(color: str | None, name: str | None) -> None:
    """Test long options."""
    click.echo(f"{color=}, {name=}")


if __name__ == "__main__":
    main()
```

2. Open a zsh shell and enable script completion: `eval "$(_GOBBLE_COMPLETE=zsh_source gobble)"`

3. Attempt to tab complete the `--color` option using both styles

| before tab completion | after tab completion | offered completions |
|--------------------------|----------------------------------------|------------------------------|
| gobble --color= | gobble | |
| gobble --color=a | gobble a | |
| gobble --color=al | gobble always | |
| gobble --color  | gobble --color | auto always never |
| gobble --color a | gobble --color a | auto always |
| gobble --color al | gobble --color always | |

4. Open a bash shell and enable script completion: `eval "$(_GOBBLE_COMPLETE=bash_source gobble)"`

5. Attempt to tab complete the `--color` option using both styles

| before tab completion | after tab completion | offered completions |
|--------------------------|----------------------------------------|------------------------------|
| gobble --color= | gobble --color= | auto always never|
| gobble --color=a | gobble --color=a | |
| gobble --color=al | gobble --color=al | |
| gobble --color  | gobble --color | auto always never |
| gobble --color a | gobble --color a | auto always |
| gobble --color al | gobble --color always | |

## Expected behavior

| before tab completion | after tab completion | offered completions | alternate completions |
|--------------------------|----------------------------------------|------------------------------|-------------------------------|
| gobble --color= | gobble --color= | auto always never| --color=auto --color=always --color=never |
| gobble --color=a | gobble --color=a | auto always | --color=auto --color=always |
| gobble --color=al | gobble --color=always | | |
| gobble --color  | gobble --color | auto always never | |
[...]

PR #3237, opened 2026-03-01
Fix shell completion for `--option=value` format

Simon99:
## Summary

Fixes #2847

When using `--option=value` style completion in bash/zsh, the shell passes `--option=partial` as the incomplete value. After resolving completions, the code was returning plain values (e.g., `auto`) without the option prefix. This caused the shell to replace `--option=` with just `auto`, resulting in `command auto` instead of `command --option=auto`.

## The Problem

```bash
# User types:
command --color=<TAB>

# Before fix: Shell receives 'auto', replaces '--color=' with 'auto'
# Result: 'command auto' ❌

# After fix: Shell receives '--color=auto'  
# Result: 'command --color=auto' ✅
```

## The Fix

In `ShellComplete.complete()`, detect when the incomplete value is in `--option=...` format and prepend the `--option=` prefix to each completion value.

## Testing

- Added test cases for bash and zsh completion with `--option=` format
- All 56 existing tests pass
- Verified manually that `--option value` (space) format still works correctly

### DIFF
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -296,7 +296,25 @@ class ShellComplete:
         completion.
         """
         args, incomplete = self.get_completion_args()
+
+        # Fix #2847: When completing ``--option=value`` style, the shell
+        # passes ``--option=partial`` as the incomplete value. After
+        # resolving completions, we need to prepend ``--option=`` to each
+        # completion so the shell replaces the whole token correctly.
+        option_prefix = ""
+        if "=" in incomplete:
+            name, _, _ = incomplete.partition("=")
+            if name.startswith("-"):
+                option_prefix = name + "="
+
         completions = self.get_completions(args, incomplete)
+
+        if option_prefix:
+            for item in completions:
+                # Only add prefix to values, not to option names
+                if not str(item.value).startswith("-"):
+                    item.value = option_prefix + str(item.value)
+
         out = [self.format_completion(item) for item in completions]
         return "\n".join(out)
 
--- a/tests/test_shell_completion.py
+++ b/tests/test_shell_completion.py
@@ -559,3 +559,48 @@ def test_files_closed(runner) -> None:
             assert not current_warnings, "There should be no warnings to start"
             _get_completions(cli, args=[], incomplete="")
             assert not current_warnings, "There should be no warnings after either"
+
+
+@pytest.mark.parametrize(
+    ("shell", "env", "expect"),
+    [
+        # Test --option= format (issue #2847)
+        # When user types --color=<TAB>, completion should include --color= prefix
+        (
+            "bash",
+            {"COMP_WORDS": "cli --color=", "COMP_CWORD": "1"},
+            "plain,--color=red\nplain,--color=green\nplain,--color=blue\n",
+        ),
+        (
+            "bash",
+            {"COMP_WORDS": "cli --color=r", "COMP_CWORD": "1"},
+            "plain,--color=red\n",
+        ),
+        # Space format should still work without prefix
+        (
+            "bash",
+            {"COMP_WORDS": "cli --color ", "COMP_CWORD": "2"},
+            "plain,red\nplain,green\nplain,blue\n",
+        ),
+        # Zsh tests
+        (
+            "zsh",
+            {"COMP_WORDS": "cli --color=", "COMP_CWORD": "1"},
+            "plain\n--color=red\n_\nplain\n--color=green\n_\nplain\n--color=blue\n_\n",
+        ),
+    ],
+)
+@pytest.mark.usefixtures("_patch_for_completion")
+def test_option_equals_completion(runner, shell, env, expect):
+    """Test that --option=value style completion works correctly.
+
+    Fixes https://github.com/pallets/click/issues/2847
+    """
+    @click.command("cli")
+    @click.option("--color", type=click.Choice(["red", "green", "blue"]))
+    def cli(color):
+        pass
+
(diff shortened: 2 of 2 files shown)


kdeldycke:
I like the extensive `test_option_equals_completion` test cases which prove the fix is working. That's good enough for me to be merged upstream and target 8.3.2.

tests/test_shell_completion.py
```diff
@@ -559,3 +559,49 @@ def cli(ctx, config_file):
             assert not current_warnings, "There should be no warnings to start"
             _get_completions(cli, args=[], incomplete="")
             assert not current_warnings, "There should be no warnings after either"
+
+
+@pytest.mark.parametrize(
+    ("shell", "env", "expect"),
+    [
+        # Test --option= format (issue #2847)
```
davidism, line 567: Don't mention the issue number, all information should be in the comment if it's needed at all.

tests/test_shell_completion.py
```diff
@@ -559,3 +559,49 @@ def cli(ctx, config_file):
...
+            "bash",
+            {"COMP_WORDS": "cli --color=", "COMP_CWORD": "1"},
+            "plain,--color=red\nplain,--color=green\nplain,--color=blue\n",
+        ),
+        (
+            "bash",
+            {"COMP_WORDS": "cli --color=r", "COMP_CWORD": "1"},
+            "plain,--color=red\n",
+        ),
+        # Space format should still work without prefix
```
davidism, line 579: This test is redundant. It's the default behavior already tested by other cases.

tests/test_shell_completion.py
```diff
@@ -559,3 +559,49 @@ def cli(ctx, config_file):
...
+        (
+            "bash",
+            {"COMP_WORDS": "cli --color ", "COMP_CWORD": "2"},
+            "plain,red\nplain,green\nplain,blue\n",
+        ),
+        # Zsh tests
```
davidism, line 585: The fix doesn't differ based on shell type so this test is unnecessary.

tests/test_shell_completion.py
```diff
@@ -559,3 +559,49 @@ def cli(ctx, config_file):
...
+            {"COMP_WORDS": "cli --color=", "COMP_CWORD": "1"},
+            "plain\n--color=red\n_\nplain\n--color=green\n_\nplain\n--color=blue\n_\n",
+        ),
+    ],
+)
+@pytest.mark.usefixtures("_patch_for_completion")
+def test_option_equals_completion(runner, shell, env, expect):
+    """Test that --option=value style completion works correctly.
+
+    Fixes https://github.com/pallets/click/issues/2847
```
davidism, line 597: Don't include issues all information should be in the comment if needed at all.

src/click/shell_completion.py
```diff
@@ -296,7 +296,25 @@ def complete(self) -> str:
         completion.
         """
         args, incomplete = self.get_completion_args()
+
+        # Fix #2847: When completing ``--option=value`` style, the shell
```
davidism, line 300: Don't include the issue all information should be in the comment. 

This comment is extremely verbose for what it's actually saying don't use AI to write comments

src/click/shell_completion.py
```diff
@@ -296,7 +296,25 @@ def complete(self) -> str:
...
+        # passes ``--option=partial`` as the incomplete value. After
+        # resolving completions, we need to prepend ``--option=`` to each
+        # completion so the shell replaces the whole token correctly.
+        option_prefix = ""
+        if "=" in incomplete:
+            name, _, _ = incomplete.partition("=")
+            if name.startswith("-"):
```
davidism, line 307: Why is this check here?

src/click/shell_completion.py
```diff
@@ -296,7 +296,25 @@ def complete(self) -> str:
...
+                option_prefix = name + "="
+
         completions = self.get_completions(args, incomplete)
+
+        if option_prefix:
+            for item in completions:
+                # Only add prefix to values, not to option names
+                if not str(item.value).startswith("-"):
```
davidism, line 315: Why does the value need to be converted to a string here?

src/click/shell_completion.py
```diff
@@ -296,7 +296,25 @@ def complete(self) -> str:
...
+
+        # Fix #2847: When completing ``--option=value`` style, the shell
+        # passes ``--option=partial`` as the incomplete value. After
+        # resolving completions, we need to prepend ``--option=`` to each
+        # completion so the shell replaces the whole token correctly.
+        option_prefix = ""
+        if "=" in incomplete:
+            name, _, _ = incomplete.partition("=")
+            if name.startswith("-"):
+                option_prefix = name + "="
```
davidism, line 308: Use F strings not concatenation.

src/click/shell_completion.py
```diff
@@ -296,7 +296,25 @@ def complete(self) -> str:
...
+
         completions = self.get_completions(args, incomplete)
```
davidism, line 310: How is this working? Shouldn't we need to add the modified option token to the arguments list?

davidism (changes requested):

davidism:
@kdeldycke this looks like AI junk to me. We shouldn't merge this.

kdeldycke:
> @kdeldycke this looks like AI junk to me. We shouldn't merge this.

Oh yes you are right. I did not evaluate the fix, I was just looking at the tests and compared them to the problem described in https://github.com/pallets/click/issues/2847. Which looked good enough to me. Should have been more explicit in my previous comment.

PR #3151, opened 2025-11-19
Randomized and parallel tests

kdeldycke:
This PR helps detection of leaks and flaky tests by running tests in randomized order and parallel.

Results:

- Multiple cores are properly detected and tests are run in parallel:

  ```
  created: 4/4 workers
  4 workers [1336 items]
  ```

- Fix `test_echo_via_pager` test with was leaky and polluting others.

PR #3238, opened 2026-03-02
Fix pager test leaks

kdeldycke:
This PR fix an issue with the `test_echo_via_pager` test, which is polluting other tests and is leaky.

This issue was uncovered while playing with parallels and random tests in https://github.com/pallets/click/pull/3151#issuecomment-3976773942 .

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -9,8 +9,6 @@ from decimal import Decimal
 from fractions import Fraction
 from functools import partial
 from io import StringIO
-from pathlib import Path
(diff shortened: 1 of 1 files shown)


davidism:
Great, this will fix the issue I pointed out at some point where `tox parallel` stopped working for Click.

kdeldycke:

kdeldycke:
For the record, just added a missing changelog entry in https://github.com/pallets/click/commit/1339fd3323357119a9c7a6326c788f80295954ce

PR #3251, opened 2026-03-06
Fix pager buffering delay with generator input

### LINKED ISSUE #3242: Flush stdin after writing to a pager
I would love to see text immediately in this toy example:
```python
import click
import os
import time

def _generator():
    for i in range(10000):
        time.sleep(0.01)
        yield f"{i+1:07}\n"

@click.command
def main():
    os.environ.setdefault("LESS", "-XFR")
    click.echo_via_pager(_generator())

main()
```
As it is currently, nothing will show until 512 lines have been `yield`ed.
I suspect a flush after this line would do the trick:
https://github.com/pallets/click/blob/cdab890e57a30a9f437b88ce9652f7bfce980c1f/src/click/_termui_impl.py#L468

Related: #2542 ☹️

G1DO:
When using `echo_via_pager()` with a generator, content doesn't appear in the pager until ~8KB of data accumulates in the pipe buffer. This means you can yield hundreds of lines before anything shows up on screen.

The fix is to pass `bufsize=1` to `subprocess.Popen` in `_pipepager()`, which enables line buffering in text mode. Each line gets sent to the pager as soon as it's written instead of sitting in the buffer.

Closes #3242

### DIFF
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -458,6 +458,7 @@ def _pipepager(
         env=env,
         errors="replace",
         text=True,
+        bufsize=1,
     )
     assert c.stdin is not None
     try:


davidism:
Appears to have been AI generated, then edited to hide that. All that just to generate a single line PR.

PR #3255, opened 2026-03-08
This PR updates docs to clarify Click’s flag special-case: for is_flag=True options with flag_value, default=True means “activated by default” and therefore defaults to flag_value. For negative boolean flags, prefer the --foo/--no-foo pattern.

shivamtiwari3:


kdeldycke:
Low-effort PR, but just for the sake of it, something similar has already been proposed at: https://github.com/pallets/click/pull/3239/changes#diff-2e046795c81fb2bcc2667705f88c6421fa988047f147ddb2333ec8da5f638b7bR386-R390

PR #3256, opened 2026-03-10
Show custom error message in prompt with `hide_input=True`

### LINKED ISSUE #2809: @click.Option with 'hide_input' = True (used for password) and custom 'type' parameter don't show custom error message.
I use @click.Option with 'hide_input' = True and custom 'type' parameter for password prompt with raise click.BadParameter (it inheriths class UsageError). And if the password failed the code don't show custom error message, but standard message "Error: The value you entered was invalid"

Standard message comes from termui.py file: ~172 line - 
```
        except UsageError as e:
            if hide_input:
                echo(_("Error: The value you entered was invalid."), err=err)
            else:
                echo(_("Error: {e.message}").format(e=e), err=err)  # noqa: B306
            continue
```
Because it checks hide_input parameter and if it True (and it is really True as i hide_input because of the password) it give the error "Error: The value you entered was invalid". When hide_input is False - it gives me my custom message.

If i use click.ClickException exception class it give me my error, but stops the promt, so i have to start from the beginning.
If i make hide_input = False - the password is visible on the screen and i recieve my custom message. But it is wrong for password entry.

Here is the example how to replicate my error. 
Run the code and enter "1234".
The message i expect "Error: Your password is week. Use 'Minimum eight characters, at least one letter and one number"

```
import click
import re

class PassRequirement(click.ParamType):
    """Click check class for password"""

    name = "Password requirements"

        pw_req = "Minimum eight characters, at least one letter and one number"
        pw_strength = r"^(?=.*[A-Za-z])(?=.*\d)[A-Za-z\d]{8,}$"

        if re.match(pw_strength, value):
            return value

        self.fail(f"Your password is week. Use {pw_req!r}", param, ctx)

    def fail(self, message="", param=None, ctx=None):
        raise click.BadParameter(message, ctx, param)
        # raise click.ClickException(message) - # works well, shows my message, but stops the promt, so i have to start the command from the beginning.

PASS_REQUIREMENT = PassRequirement()

@click.command()
@click.option("-password", help="Password", prompt=True, confirmation_prompt=True, hide_input=True,
              type=PASS_REQUIREMENT)
def pw(password):
    print(f'PASSWORD: {password}')


if __name__ == '__main__':
    pw()

```

Environment:
- Python version: 3.12.7
- Click version: 8.1.7

doiken:
fixes #2809

`prompt()` with `hide_input=True` always showed a generic error message, even for custom types. 
                                         
This change masks the input value using `repr(value)` matching
(consistent with built-in types' `{value!r}` pattern) and shows
custom messages that don't contain the input value as-is.

### DIFF
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -203,7 +203,20 @@ def prompt(
             result = value_proc(value)
         except UsageError as e:
             if hide_input:
-                echo(_("Error: The value you entered was invalid."), err=err)
+                repr_val = repr(value)
+                if repr_val in e.message:
+                    # Built-in type pattern: mask the repr'd value.
+                    msg = e.message.replace(repr_val, "'***'")
+                elif value in e.message:
+                    # Raw value found: could be a coincidental or
+                    # unquoted match. Ambiguous, use generic.
+                    msg = _("The value you entered was invalid.")
+                else:
+                    # Value not found: show as-is, assuming custom
+                    # types with hide_input=True avoid leaking input.
+                    msg = e.message
+
+                echo(_("Error: {msg}").format(msg=msg), err=err)
             else:
                 echo(_("Error: {e.message}").format(e=e), err=err)
             continue
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -4,6 +4,7 @@ import collections.abc as cabc
 import inspect
 import io
 import itertools
+import re
 import sys
 import typing as t
 from contextlib import AbstractContextManager
@@ -51,6 +52,27 @@ _ansi_colors = {
     "bright_white": 97,
 }
 _ansi_reset_all = "\033[0m"
+
+
+_HIDDEN_INPUT_MASK = "'***'"
+
+
+def _mask_hidden_input(message: str, value: str) -> str:
+    """Replace occurrences of ``value`` in ``message`` with a fixed mask.
+
+    Both ``repr(value)`` (the form built-in :class:`ParamType` errors use
+    via ``{value!r}``) and the raw value are masked. The raw-value pass
+    uses word-boundary lookarounds so a substring like ``"1"`` does not
+    match inside ``"10"``, and ``"ent"`` does not match inside
+    ``"Authentication"``. The empty string is skipped to avoid matching
+    at every boundary.
+    """
+    message = message.replace(repr(value), _HIDDEN_INPUT_MASK)
+    if value:
+        message = re.sub(
+            rf"(?<!\w){re.escape(value)}(?!\w)", _HIDDEN_INPUT_MASK, message
+        )
+    return message
 
 
 def hidden_prompt_func(prompt: str) -> str:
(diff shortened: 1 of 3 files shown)


src/click/termui.py
```diff
@@ -203,7 +203,20 @@ def prompt_func(text: str) -> str:
...
             if hide_input:
-                echo(_("Error: The value you entered was invalid."), err=err)
+                repr_val = repr(value)
+                if repr_val in e.message:
+                    # Built-in type pattern: mask the repr'd value.
+                    msg = e.message.replace(repr_val, "'***'")
+                elif value in e.message:
+                    # Raw value found: could be a coincidental or
+                    # unquoted match. Ambiguous, use generic.
+                    msg = _("The value you entered was invalid.")
```
AndreasBackx: I understand why this is done. Though should we not do the same as the repr value branch?

Let's say there's an error message saying "PASSWORD is too short". I'd rather have a bit of a weirder message like "*** is too short" than "The value you entered was invalid.".

@kdeldycke @Rowlando13 what do you peeps think? I think otherwise this looks great.

AndreasBackx (approved):

src/click/termui.py
kdeldycke: I am more concerned about the fragility of the `repr_val in e.message` matching. What happens if your are matching on substrings that have nothing to do with your value?

kdeldycke:
So exactly what I was thinking @AndreasBackx : this fix is way too naive. See the number of failing edge-cases I was able to add.

So now: either I try to fix it locally on top of that PR, or I can implement a better long-term solution. Maybe by modifying how `UsageError` manage the raw value. But I have not studied the consequences and impact yet.

AndreasBackx:
I've got two thoughts as well. We could instead of a ValueError do a custom error like you say, but require the message to be a literal meaning you cannot interpolate the input value in it.

Or... we wrap the input value in a wrapper that has its own str and repr implementation that would obfuscate the real value. The real value would be some subfield perhaps. Though this would be a breaking change and I don't think it's very pretty.

kdeldycke:
> I've got two thoughts as well. We could instead of a ValueError do a custom error like you say, but require the message to be a literal meaning you cannot interpolate the input value in it.
> 
> Or... we wrap the input value in a wrapper that has its own str and repr implementation that would obfuscate the real value. The real value would be some subfield perhaps. Though this would be a breaking change and I don't think it's very pretty.

I think the subfield looks like the ultimate, realiable solution. But it would introduce too much changes in my opinion. And I am not sure if it will look clean or not. I propose to defer that exploration in another PR.

In the mean time, I implemented a solution based on regexp so the hidden heuristic is more robust and cover all cases. You can see the edge-cases I handle with this regexp in the associated unittests. Just pushed that on top of the PR.

So this PR is ready to be reviewed and merged unless someone has feedbacks.

doiken:
Thanks so much for taking the time to push the regex-based solution and the extra test coverage!

AndreasBackx (approved):

PR #3165, opened 2025-12-07
Support `show_default` string in prompts

MirrorDNA-Reflection-Protocol:
## Summary

Fixes #2836

When `show_default` is set to a string on an `Option` with `prompt=True`, the string was only used in the help text but not in the actual prompt. Now the custom string is also displayed in the prompt, matching the behavior of help text.

## Before

```python
@click.option('--name', default='default', show_default='show_default', prompt=True)
```

Help: `--name TEXT  [default: (show_default)]` ✓
Prompt: `Name [default]:` ✗ (shows actual default, not custom string)

## After

Help: `--name TEXT  [default: (show_default)]` ✓  
Prompt: `Name [(show_default)]:` ✓ (now shows custom string)

## Changes

1. **`_build_prompt()`**: Updated to accept `str | bool` for `show_default`. When it's a string, display it in parentheses (matching the help text format).

2. **`prompt()`**: Updated signature to accept `str | bool` for `show_default` parameter, and updated docstring.

3. **`Option.prompt_for_value()`**: Now passes `show_default` to `prompt()` regardless of type (previously only passed it when it was a bool).

## Testing

Manually tested with the reproduction case from the issue. The prompt now correctly shows the custom `show_default` string.

PR #2837, opened 2025-01-06
When prompting, use `show_default` as the default if its a string

thejcannon:
This PR just makes sure the `default` handed to `.prompt` is the `show_default` if its a string (to match the help).

fixes #2836

- [ ] Add tests that demonstrate the correct behavior of the change. Tests
  should fail without the change.
- [ ] Add or update relevant docs, in the docs folder and in code.
- [ ] Add an entry in CHANGES.rst summarizing the change and linking to the issue.
- [ ] Add `.. versionchanged::` entries in any relevant code docs.

PR #3262, opened 2026-03-13
Fix `show_default` string not displayed in prompts

### LINKED ISSUE #2836: `show_default` as a string isn't used in prompts as the default shown

Help: 👍 
```
Usage: foo.py [OPTIONS]

Options:
```

Prompt: 👎 
```
Name [default]:
```

```
Name [show_default]:
```

Environment:

- Python version: 3.11.6
- Click version: 8.1.6

oaksprout:
## Summary

When `show_default` is set to a custom string on an option with `prompt=True`, the help text correctly displays the custom string (e.g. `[default: (show_default)]`), but the interactive prompt displays the actual default value instead.

This PR fixes the inconsistency so the prompt also displays the custom `show_default` string:

```python
@click.option("--name", default="default", show_default="custom label", prompt=True)
```

**Before:** `Name [default]: `
**After:** `Name [custom label]: `

### Changes

- Updated `_build_prompt` in `termui.py` to accept `show_default` as `bool | str` and display the string when provided
- Updated `prompt()` signature to accept `show_default: bool | str`
- Updated `Option.prompt_func` in `core.py` to pass string `show_default` values through to `prompt()`
- Added regression test

Fixes #2836

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -3145,10 +3145,10 @@ class Option(Parameter):
                 default = bool(default)
             return confirm(self.prompt, default)
 
-        # If show_default is set to True/False, provide this to `prompt` as well. For
-        # non-bool values of `show_default`, we use `prompt`'s default behavior
+        # Pass show_default to prompt. When it's a string, the prompt will
+        # display that string instead of the actual default value.
         prompt_kwargs: t.Any = {}
-        if isinstance(self.show_default, bool):
+        if self.show_default is not None:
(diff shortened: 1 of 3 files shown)


kdeldycke:
This PR has been closed as a copy of both https://github.com/pallets/click/pull/3165, https://github.com/pallets/click/pull/2837 and https://github.com/pallets/click/pull/3280

kdeldycke:
A PR is ready to be merged at: https://github.com/pallets/click/pull/3165

PR #3271, opened 2026-03-19
Fix usage string for invoke_without_command groups to mark subcommand as optional

### LINKED ISSUE #3059: With invoke_without_command set commands are not marked as optional
Having

```python
def main():
    if ctx.invoked_subcommand is not None:
        return
    # do something by default
```

I expect `--help` to output:

```
Usage: app.py [OPTIONS] [COMMAND] [ARGS]...
```

But it outputs:

```
Usage: app.py [OPTIONS] COMMAND [ARGS]...
```

Environment:

- Python version: 3.11.8
- Click version: 8.2.0

siddmoparti:
Fixes #3059

When a group is created with `invoke_without_command=True`, the help usage string currently shows:

    COMMAND [ARGS]...

which incorrectly implies that a subcommand is required. However, in this configuration, the group can be invoked without a subcommand.

This change updates the default `subcommand_metavar` so that the usage string correctly reflects optional behavior:

    [COMMAND] [ARGS]...

The change is limited to non-chain groups with `invoke_without_command=True`, and does not affect existing behavior for other group configurations.

A regression test has been added to verify that the usage output includes `[COMMAND] [ARGS]...` when `--help` is invoked.

Let me know if any adjustments are needed.

### DIFF
--- a/repro.py
+++ b/repro.py
@@ -1,0 +1,14 @@
+import click
+
+@click.group(invoke_without_command=True)
+@click.pass_context
+def main(ctx):
+    if ctx.invoked_subcommand is not None:
+        return
+
+@main.command()
+def sub():
+    pass
+
+if __name__ == "__main__":
+    main()
--- a/repro.py
+++ b/repro.py
@@ -1,14 +1,17 @@
 import click
 
+
 @click.group(invoke_without_command=True)
 @click.pass_context
 def main(ctx):
     if ctx.invoked_subcommand is not None:
         return
 
+
 @main.command()
 def sub():
     pass
+
 
 if __name__ == "__main__":
     main()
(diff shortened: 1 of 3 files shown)


themavik (commented):
Reviewed the changes — the approach looks correct and follows the existing patterns.

kdeldycke:
@Rowlando13 , you removed the author original comment at https://github.com/pallets/click/issues/3059#issuecomment-4086646890 . Does this mean this PR is AI slop? CC @davidism

Rowlando13:
@kdeldycke Not sure. @siddmoparti Did you do this using AI or are you just new?

kdeldycke:
> @kdeldycke Not sure. @siddmoparti Did you do this using AI or are you just new?

No feedback for a couple of month, so I propose to close that PR as invalid. I will produce a new one for https://github.com/pallets/click/issues/3059

PR #3280, opened 2026-03-22
Fix show_default string not shown in prompt (#2836)

### LINKED ISSUE #2836: `show_default` as a string isn't used in prompts as the default shown

r266-tech:
## Description

Fixes #2836

When `show_default` is a string and `prompt=True`, the custom string was ignored and the actual default value was shown in the prompt instead.

**Before:**
```
$ python cli.py
Name [default]:
```

**After:**
```
$ python cli.py
Name [custom]:
```

## Changes

- `src/click/core.py`: Pass `show_default` to `prompt()` regardless of its type (bool or str), not just for bool values
- `src/click/termui.py`: Updated `_build_prompt()` and `prompt()` signature to accept `str` for `show_default`. When it is a string, it is displayed in the prompt bracket instead of the formatted default value
- `tests/test_options.py`: Added test `test_show_default_string_in_prompt`

All existing tests pass (1320 passed, 22 skipped).

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -3145,10 +3145,9 @@ class Option(Parameter):
                 default = bool(default)
             return confirm(self.prompt, default)
 
-        # If show_default is set to True/False, provide this to `prompt` as well. For
-        # non-bool values of `show_default`, we use `prompt`'s default behavior
+        # Pass show_default to prompt regardless of its type (bool or str).
         prompt_kwargs: t.Any = {}
-        if isinstance(self.show_default, bool):
+        if self.show_default is not None:
             prompt_kwargs["show_default"] = self.show_default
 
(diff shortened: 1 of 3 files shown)


kdeldycke:
This is a copy of https://github.com/pallets/click/pull/3165, https://github.com/pallets/click/pull/2837 and https://github.com/pallets/click/pull/3262

kdeldycke:
A PR is ready to be merged at: https://github.com/pallets/click/pull/3165

PR #3295, opened 2026-03-30
fix: show string show_default value in prompt, not just in help text

### LINKED ISSUE #2836: `show_default` as a string isn't used in prompts as the default shown

juliosuas:
## Summary

Fixes #2836.

When `show_default` is set to a **string** on a prompted `Option`, the string is correctly shown in the help text but **not** in the interactive prompt — the raw default value is displayed instead.

## Root Cause

In `Option.prompt_for_value()`, the `show_default` value is only forwarded to `termui.prompt()` when it is a `bool`:

```python
# Before
prompt_kwargs: t.Any = {}
if isinstance(self.show_default, bool):
    prompt_kwargs["show_default"] = self.show_default
```

When `show_default` is a string, nothing is passed to `prompt()`, which falls back to showing the real `default` value.

## Fix

When `show_default` is a `str`, pass it as the `default` argument (which controls the label shown in brackets) and set `show_default=True`:

```python
if isinstance(self.show_default, str):
    prompt_kwargs["show_default"] = True
    prompt_default = self.show_default   # display the custom string, not the real default
elif isinstance(self.show_default, bool):
    prompt_kwargs["show_default"] = self.show_default
```

## Before / After

```python
@click.command()
@click.option('--name', default='real_default', show_default='my_label', prompt=True)
def cmd(name): ...
```

| | Prompt |
|---|---|
| **Before** | `Name [real_default]:` |
| **After** | `Name [my_label]:` |

The help text already showed `[default: (my_label)]` correctly — this brings the prompt into parity.

## Testing

Added a manual test that verifies the custom string appears in the prompt output.

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -3145,17 +3145,25 @@ class Option(Parameter):
                 default = bool(default)
             return confirm(self.prompt, default)
 
-        # If show_default is set to True/False, provide this to `prompt` as well. For
-        # non-bool values of `show_default`, we use `prompt`'s default behavior
+        # If show_default is set to True/False, provide this to `prompt` as well.
+        # If show_default is a string, display that string as the default label
+        # in the prompt (e.g. "Name [show_default]: ") instead of the raw default
+        # value. This matches the behaviour already implemented for the help text.
         prompt_kwargs: t.Any = {}
-        if isinstance(self.show_default, bool):
+        prompt_default = None if default is UNSET else default
+
+        if isinstance(self.show_default, str):
+            # Use the custom string as the displayed default in the prompt.
+            prompt_kwargs["show_default"] = True
+            prompt_default = self.show_default
+        elif isinstance(self.show_default, bool):
             prompt_kwargs["show_default"] = self.show_default
 
         return prompt(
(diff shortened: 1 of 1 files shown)


juliosuas:
Wanted to add some context in case it's helpful for review:

The issue is subtle — Python 3.14 added `_check_help()` inside `add_argument()` which accesses `action.help` during parser construction. Previously this only happened at `parse_args(['--help'])` time. The PR's fix specifically targets the `prompt_for_value()` path, which is separate from the help-text formatting already handled by the `help` property setter.

I also confirmed the fix works by writing a quick test: `parser.parse_args([])` no longer triggers the getter, but calling `--help` still invokes it and returns the formatted string. Let me know if you'd like me to add that as a formal test case.

kdeldycke:
This is already covered by https://github.com/pallets/click/pull/3165

PR #3299, opened 2026-04-01
Fix speculative empty string check

### LINKED ISSUE #3298: semver.Version as default causing an error
Any instance of `semver.Version` set as `default` in an option will lead to an error when Click tries to show it in the help text.
The problem seems to be that `Version` allows comparison to strings, and Click, when building the help text, tries to check the default value against an empty string `""`, which cannot be parsed by `semver.Version`:

```
  File ".venv/lib/python3.12/site-packages/click/core.py", line 3091, in get_help_extra
    elif default_value == "":
         ^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/semver/version.py", line 52, in wrapper
    return operator(self, other)
           ^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/semver/version.py", line 467, in __eq__
    return self.compare(other) == 0
           ^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/semver/version.py", line 394, in compare
    other = cls.parse(other)
            ^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/semver/version.py", line 644, in parse
    raise ValueError(f"{version} is not valid SemVer string")
ValueError:  is not valid SemVer string
```

This can be replicated with the following code by trying to print the help text:

<details>

```python
import click
from semver import Version


class SemverType(click.ParamType):
    name = "semver"

    def convert(self, value: str | Version, param: click.Parameter | None, ctx: click.Context | None) -> Version:
        if isinstance(value, Version):
            return value

        try:
            return Version.parse(value)
        except ValueError as e:
            raise self.fail(str(e), param, ctx) from e


@click.command()
@click.option(
    "--version",
    type=SemverType(),
    default=Version(1, 0, 0),
    show_default=True,
)
def cli(version: Version | None) -> None:
    print(version)


if __name__ == "__main__":
    cli()
```

</details>

By quickly inspecting the code, I'd say this could be fixed by checking whether `default_value` is actually a string before comparing it to `""`. 

Environment:

- Python version: 3.12
- Click version: 8.3.1
- semver version: 3.0.4

kdeldycke:
Fix #3298 by checking against a speculative empty string.

That not unreasonable given the hard-coded nature of the empty string.

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -3110,7 +3110,7 @@ class Option(Parameter):
                 )[1]
             elif self.is_bool_flag and not self.secondary_opts and not default_value:
                 default_string = ""
-            elif default_value == "":
+            elif isinstance(default_value, str) and default_value == "":
                 default_string = '""'
             else:
                 default_string = str(default_value)
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -1259,12 +1259,36 @@ def test_show_default_string(runner):
     assert "[default: (unlimited)]" in message
 
 
-def test_show_default_with_empty_string(runner):
-    """When show_default is True and default is set to an empty string."""
-    opt = click.Option(["--limit"], default="", show_default=True)
+class _StrictEq:
+    """Object whose ``__eq__`` raises on string comparison (like semver.Version)."""
+
+    def __eq__(self, other):
+        if isinstance(other, str):
+            raise ValueError("cannot compare to string")
+        return NotImplemented
+
+    def __str__(self):
+        return "strict"
+
+
+@pytest.mark.parametrize(
+    ("default", "expected"),
+    [
+        ("", '[default: ""]'),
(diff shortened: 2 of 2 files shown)


kdeldycke:
I could not find any regression regarding default values so this PR is ready to be merged.

Rowlando13:
Can you add a changelog entry.

kdeldycke:
> Can you add a changelog entry.

Ah you're right. Just did that in https://github.com/pallets/click/commit/1339fd3323357119a9c7a6326c788f80295954ce .

I am not touching `stable` and always use a PR so you can review it, but for such an obvious fix I took the liberty to push it directly to `stable`. Also used the opportunity to add a new `8.3.3` section and lint some syntax.

PR #3314, opened 2026-04-05
AI junk

kaiisfree:
## Summary

- When completing `--option=value` style input, the shell passes `--option=partial` as the incomplete value. `_resolve_incomplete` correctly splits this and resolves completions for the value part, but `complete()` returns only the bare values (e.g. `auto`). The shell then replaces the whole `--option=auto` token with just `auto`.
- Detect the `--option=` prefix in `complete()` and prepend it to each completion value before formatting, so the shell receives `--color=auto` instead of `auto`.

Fixes #2847. A previous attempt at this fix was made in #3237, which was closed. This is a fresh approach that addresses the review feedback from that PR.

## Test plan

- [x] Added parametrized test for `--option=` and `--option=partial` completion in bash
- [x] All 54 shell completion tests pass
- [x] Full test suite passes (1387 passed, 21 skipped, 1 xfailed)
- [x] Pre-commit hooks pass (ruff, ruff format, etc.)

### DIFF
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -296,7 +296,25 @@ class ShellComplete:
         completion.
         """
         args, incomplete = self.get_completion_args()
+
+        # When the incomplete value is ``--option=partial``, the option
+        # prefix needs to be prepended to each completion value so the
+        # shell replaces the whole token correctly.
+        option_prefix = ""
+
+        if "=" in incomplete:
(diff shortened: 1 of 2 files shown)


ThiefMaster:
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

Please do not dump such PRs onto other people's project unless the project explicitly welcomes AI-generated PRs (**Pallets doesn't**).

PR #3245, opened 2026-03-05
Document and fix command sanitizing with `shlex.split`

kdeldycke:
Also removes last use of `shell=True` use for command invokation for defense-in-depth.

### Context

I was trying to eliminate the last usage of `shell=True` in `subprocess.Popen`, then had to resort to using `shlex.split` in `edit_files`. This led me into the rabbit hole of `shlex.split` usage in `_termui_impl.py` file. Which I use as an opportunity to document past choices from older issues and PRs.

### Tests

I also use that opportunity to collect test cases discussed in older related issues and PRs to illustrate and cover how we are expecting command path to be unquoted and interpreted on POSIX systems and Windows.

### Related

- #1026
- #1477
- #2775
- https://github.com/pallets/click/commit/43a7d70f1e2a4a929af99b848612e0fd478e6e50

PR #3330, opened 2026-04-10
Path normalization documentation clean-up + `shlex.split` extensive tests

kdeldycke:
This is a follow up on #3245, which was merged to `stable` a bit too early before some documentation issues has been caught.

So this PR:
- Move all references to issues and PR, as well as developer-centric details, out of their published docstrings and to Python code comments
- Add some more edge-cases to illustrate the behavior of path normalization
- Add explicit test cases to demonstrate and verify the behavior of the underlying `shlex.split` function without `posix=False`

### DIFF
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -367,21 +367,7 @@ class ProgressBar(t.Generic[V]):
 
 
 def pager(generator: cabc.Iterable[str], color: bool | None = None) -> None:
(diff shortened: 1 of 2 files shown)


kdeldycke:
As you can see in this PR: all cases passes on the 3 platforms, including Windows.

Rowlando13 (approved):

kdeldycke:
Thanks @Rowlando13 for the merge ! :)

Issue #2779, opened 2024-09-24
Wrong error message when wrong multicharacter short option is passed


ocervell:
Sometime it is usefull to have not only single char short option like `-d` but something like `-dbg`.

They are already working as-is (even though not documented), but when passing a wrong option such as `-dbgwrong`, the error displays: `No such option: -d` instead of `No such option: -dbg`.

Rowlando13:
@davidism I have only seen single character short options. Do we intentionally support multi character short options?

ocervell:
Hi @davidism, 
any updates here ? It seems we do accept multi-char short options, so we should have the proper error message when they are not found.

davidism:
This is the correct error message. Short options can be stacked, `-dbg` is equivalent to `-d -b -g`, so the first unregistered option raises an error.

Rowlando13:
Just reopening for docs addition.

Rowlando13:
Add docs section in options section with:
- Title: combining short options
- the short option combining behavior explanation and example 
- note about multi character short options not being supported

LouisLau-art:
I opened a PR adding the requested docs section about stacking short options and clarifying that multi-character short options aren’t supported: https://github.com/pallets/click/pull/3200

Rowlando13:
@LouisLau-art Please read the activity before opening a PR. #3179 was already submitted.

LouisLau-art:
Sorry about the duplicate PR noise — I missed that #3179 was already submitted. I’ve closed #3200 to avoid duplication. Thanks for the heads-up.

kdeldycke:
So, to reframe the discussion: the root cause is a documentation issue, there is no approval by the maintainers to change the parser.

This will be closed by #3179, or we can even re-open #3200 by @LouisLau-art if the author of #3179 (@aadelb) is unresponsive to feedbacks.

kdeldycke:
If this issue is about documentation, I still think the origin lies in a surprising error message. So in the mean time, I experimented with Click UI and tried to make the error message in this case verbosier. It is integrated with [Click Extra](https://github.com/kdeldycke/click-extra) >= 7.12.0 so you can use Click Extra if you want to get this feature. See the code at: https://github.com/kdeldycke/click-extra/commit/344a90787247dc95e42fa9bff364662ab4f83807

Rowlando13:
[...]

PR #3335, opened 2026-04-12
fix: correct error message for unknown options that start with a multi-char short option

DoriAlagem:
## Summary

Fixes #2779.

When a multi-character short option like `-dbg` is registered, passing an unrecognised option that merely **starts** with it (e.g. `-dbgwrong`) caused a misleading error:

```
Error: No such option: -d
```

The user-visible string reported only the first character (`-d`) because `_process_opts` fell back to `_match_short_opt`, which iterates the argument character-by-character and raises `NoSuchOption` on the very first character that isn't a known short option.

The expected message is:

```
Error: No such option: -dbgwrong
```

## Root cause

In `_process_opts`, after a long-option match fails for a single-dash argument, the code dispatches to `_match_short_opt` for character-by-character parsing. When the argument actually started with a known long option (single-dash multi-char, e.g. `-dbg`), this dispatch produces a confusing error naming only the first character.

## Fix

Before calling `_match_short_opt`, check whether `norm_long_opt` starts with any key in `self._long_opt`. If it does, skip the character loop and re-raise the original `NoSuchOption` (which already contains the full unrecognised token).

A regression test is included in `tests/test_options.py`.

## Test plan

- [x] New test `test_multichar_short_option_error_message` verifies the full option name appears in the error output
- [x] Existing test suite for the option parser still passes

### DIFF
--- a/src/click/parser.py
+++ b/src/click/parser.py
@@ -490,6 +490,15 @@ class _OptionParser:
             # short option code and will instead raise the no option
             # error.
             if arg[:2] not in self._opt_prefixes:
+                # If the arg starts with a known long option (e.g. "-dbg")
+                # followed by extra characters, don't fall back to per-character
+                # short option parsing.  That would produce a misleading
+                # "No such option: -d" instead of "No such option: -dbgX".
+                if any(norm_long_opt.startswith(k) for k in self._long_opt):
+                    if not self.ignore_unknown_options:
+                        raise
+                    state.largs.append(arg)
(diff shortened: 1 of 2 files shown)


DORI2001:
Closing due to failing CI — tests fail across all Python versions (3.10–3.14, Mac, Windows, PyPy). The fix logic needs more investigation before this can be in a mergeable state. Please reopen if this is still relevant.

PR #3338, opened 2026-04-14
Use 'import typing as t' instead of 'import typing' (#3164)

TanishqGupta1205:


ThiefMaster:
You did not actually make these changes yourself, but rather used some LLM or broken tool, did you?

- There is some absolute nonsense, and your "fix" commit made some of it worse... `import typing as t as t as t` and `import typing as t as t_extensions as te` for example (if you wrote it yourself, please explain what made you write this)
- Your commit message references a completely unrelated issue

PR #3339, opened 2026-04-15
Report full arg when single-dash token isn't a short flag

MukundaKatta:
## Summary

When a user defines a multi-character single-dash option (e.g. `-dbg`) and mistypes it as `-dbgwrong`, click currently reports:

```
Error: No such option: -d
```

because long-option matching fails, the code falls back to `_match_short_opt`, which greedy-splits the argument character by character and stops at the first unknown single char.

This PR makes the fallback only kick in when the first character after the prefix is actually registered as a short option. Otherwise the original `NoSuchOption` is re-raised with the full argument, which also lets the existing close-match machinery suggest the right option:

```
Error: No such option: -dbgwrong Did you mean -dbg?
```

## What's unchanged

Short-option combining still works exactly the same:

- `-abc` where `-a`, `-b`, `-c` are all registered flags → combined, all three enabled
- `-abZ` where `-a`, `-b` registered but `-Z` isn't → still reports `No such option: -Z` (first char IS registered, so the split path runs, and the specific bad char gets named)

## Tests

Added two regression tests in `tests/test_options.py`:

1. `test_unknown_multichar_short_option_reports_full_arg` — guards the #2779 scenario
2. `test_short_option_combining_still_works` — guards that `-abc`-style combining isn't regressed

Full suite passes: `1386 passed, 22 skipped, 30000 deselected, 1 xfailed`.

## Changelog

Added an entry under an `Unreleased` section. Happy to move it under whichever future release number the maintainers prefer.

Fixes #2779

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,15 @@
 .. currentmodule:: click
 
+Unreleased
+----------
+
+-   Report the full argument (and close-match suggestion) when a
+    single-dash token looks like a typo of a multi-character option
+    (``-dbgwrong`` for ``-dbg``) instead of greedily splitting it into
+    single-char shorts. Short-option combining (``-abc``) is unaffected.
+    :issue:`2779`
+
+
 Version 8.3.2
 --------------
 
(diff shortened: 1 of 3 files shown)


kdeldycke:
AI slop

PR #3340, opened 2026-04-15
Fix #2779: Wrong error message when wrong multicharacter short opti...

mattiagaggi:
## Summary
Fix the error message when an invalid multi-character single-dash option is passed. Previously, `-dbgwrong` would report `No such option: -d` because the parser tried to match each character individually as a short option. Now it correctly reports `No such option: -dbgwrong`.

## Changes
- In `_match_short_opt` in `parser.py`, when a character fails to match a short option, check if any registered single-dash long option (e.g. `-dbg`) is a prefix of the original argument. If so, raise `NoSuchOption` with the full argument string instead of just the single unmatched character.
- Added test verifying that `-dbg` (correct usage) still works and that `-dbgwrong` produces an error message referencing the full argument.

## Testing
- `pytest tests/test_options.py::test_multicharacter_short_option_wrong_option_error` passes
- Existing option-related tests remain green

Closes #2779

### DIFF
--- a/src/click/parser.py
+++ b/src/click/parser.py
@@ -401,6 +401,18 @@ class _OptionParser:
                 if self.ignore_unknown_options:
                     unknown_options.append(ch)
                     continue
+
+                # If a registered single-dash long option (like -dbg)
+                # is a prefix of the original argument, report the
+                # error using the full argument instead of just the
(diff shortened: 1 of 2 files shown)


kdeldycke:
AI slop

PR #3345, opened 2026-04-16
Fix error message for unrecognised multi-char short options

Ai-chan-0411:
## Problem

When using multi-character short options like `-dbg` (which Click supports but doesn't document), passing an unrecognised argument such as `-dbgwrong` produces a misleading error:

```
No such option: -d
```

instead of the expected:

```
No such option: -dbgwrong
Did you mean -dbg?
```

This happens because multi-character short options are stored internally as long options. When `_match_long_opt` fails, the parser falls back to `_match_short_opt`, which iterates character-by-character and reports only the first character (`-d`) in the error.

Reported in #2779.

## Fix

In `_process_opts`, catch the `NoSuchOption` from `_match_short_opt`. If the error refers to the very first character (meaning no valid single-char options were consumed at all), re-raise the original long-option error instead — it contains the full argument name and the "Did you mean …?" suggestion.

## Test

Added `test_multichar_short_option_wrong_value_error` which registers `-dbg` as a flag, invokes the CLI with `-dbgwrong`, and asserts that the error names `-dbgwrong` and suggests `-dbg`.

All existing tests pass.

### DIFF
--- a/src/click/parser.py
+++ b/src/click/parser.py
@@ -482,7 +482,7 @@ class _OptionParser:
         # like "-foo" to be matched as long options.
         try:
             self._match_long_opt(norm_long_opt, explicit_value, state)
-        except NoSuchOption:
+        except NoSuchOption as long_opt_exc:
             # At this point the long option matching failed, and we need
             # to try with short options.  However there is a special rule
             # which says, that if we have a two character options prefix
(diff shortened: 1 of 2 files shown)


kdeldycke:
AI slop

Ai-chan-0411:
Thank you for the feedback. I understand the concern and appreciate your maintainance of click. I will ensure future contributions are more carefully tailored to the project's standards and context.

PR #3348, opened 2026-04-17
fix: reject NaN values in FloatRange

### LINKED ISSUE #3347: FloatRange accepts NaN, bypassing range validation
## Bug

`click.FloatRange` accepts `NaN` as a valid value, bypassing range validation entirely. In contrast, `inf` and `-inf` are correctly rejected.

## Reproduction

```python
import click

@click.command()
@click.option("--temperature", type=click.FloatRange(-50.0, 150.0))
def cmd(temperature):
    click.echo(f"Processing: {temperature}C")

runner = CliRunner()

# NaN passes validation (bug)
result = runner.invoke(cmd, ["--temperature", "nan"])
print(result.exit_code)  # 0 — should be 2
print(result.output)     # "Processing: nanC"

# inf is correctly rejected
result = runner.invoke(cmd, ["--temperature", "inf"])
print(result.exit_code)  # 2 — correct
```

## Root cause

`_NumberRangeBase.convert` checks `rv < min` and `rv > max`. Since `NaN < 0.0` and `NaN > 1.0` both evaluate to `False`, the check `lt_min or gt_max` is `False`, so NaN passes through.

This also bypasses `clamp=True` — the NaN value is returned unclamped.

## Impact

Any CLI using `FloatRange` for input validation (temperature ranges, ratios, sensor data thresholds) will accept NaN and pass it to downstream code, where it silently corrupts calculations. The whole point of `FloatRange` is to reject invalid values, and NaN is not a valid number for any range.

## Environment

click 8.3.2, Python 3.12

PewterZz:
Fixes #3347

`FloatRange` currently accepts NaN because NaN comparisons (`NaN < min`, `NaN > max`) always return `False`, causing the range check to pass. This bypasses validation entirely, including `clamp=True` mode.

The fix adds a `math.isnan` check in `FloatRange.convert` before returning the value. NaN is not a valid number for any range, so it should always be rejected.

**Change summary:**
- Override `convert` in `FloatRange` to check for NaN after calling `super().convert()`
- Fail with "nan is not a valid number" if detected
- 6 new test cases covering bounded, unbounded, clamped, and open-bound FloatRange configurations

All 1391 tests pass.

### DIFF
--- a/src/click/types.py
+++ b/src/click/types.py
@@ -648,6 +648,22 @@ class FloatRange(_NumberRangeBase, FloatParamType):
         if (min_open or max_open) and clamp:
             raise TypeError("Clamping is not supported for open bounds.")
 
+    def convert(
+        self, value: t.Any, param: Parameter | None, ctx: Context | None
+    ) -> t.Any:
+        import math
+
+        rv = super().convert(value, param, ctx)
+
+        if math.isnan(rv):
+            self.fail(
+                _("{value} is not a valid number.").format(value=rv),
+                param,
+                ctx,
+            )
+
+        return rv
+
     def _clamp(self, bound: float, dir: t.Literal[1, -1], open: bool) -> float:
         if not open:
             return bound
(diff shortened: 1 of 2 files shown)


davidism:
AI slop. I don't see why it matters if we accept all float values, and if certain values that are known not to compare don't compare.

PR #3364, opened 2026-04-21
Split string values from `default_map` for multi-value parameters

### LINKED ISSUE #2745: Modifying `default_map` in callback behaves differently between click 7.1.2 and 8.0.0
Hi,

I'm trying to upgrade an application from click 7 to click 8 but I'm running into issues.
I was able to reduce the issue to a minimal reproducible example.

The following program behaves differently in 7.1.2 and 8.0.0

```
import click

def load_settings(ctx, param, value):
    # breakpoint()
    if (value is not None):
        ctx.default_map['general_foo'] = value + 'foo'
        ctx.default_map['general_bar'] = value + 'bar'

def sanitize_foo(ctx, param, val):
    click.echo('sanitizing foo')
    return val

def sanitize_bar(ctx, param, val):
    click.echo('sanitizing bar')
    return val

def print_help(ctx, param, value):
    if value:
        ctx.exit()

@click.command()
@click.option('--general-foo', '--foo', default=None, nargs=2, callback=sanitize_foo)
@click.option('--general-bar', '--bar', default=None, callback=sanitize_bar)
@click.option('--settings', callback=load_settings, default=None, is_eager=True)
@click.pass_context
def hello(ctx, settings, general_bar, general_foo):
    click.echo(general_foo)
    click.echo(general_bar)

@click.group()
@click.option('--settings', hidden=True)
@click.pass_context
def cli(ctx, settings):
    ctx.default_map['hello'] = { 'settings': settings }

cli.add_command(hello)

if __name__ == '__main__':
    cli(default_map={})
```

In click 7.1.2:
```
(.venv) dezsiszabi@SZABI-PC:~/projects/click-test$ python3 main.py hello --settings hello
sanitizing foo
sanitizing bar
('h', 'e', 'l', 'l', 'o', 'f', 'o', 'o')
hellobar
```

In click 8.0.0 (and later):
```
(.venv) dezsiszabi@SZABI-PC:~/projects/click-test$ python3 main.py hello --settings hello
Usage: main.py hello [OPTIONS]
Try 'main.py hello --help' for help.

Error: Invalid value for '--general-foo' / '--foo': Value must be an iterable.
```

I assume this difference was introduced intentionally, but I'm having a hard time adapting my code and even after reading the changelog multiple times I'm not sure which change in 8.0.0 is causing it exactly.

Note: the application is basically trying to do this https://jwodder.github.io/kbits/posts/click-config/
I assume the original developer of the application used this guide as well, but unfortunately this is now breaking with 8.0.0 :)

Any help is appreciated.

Environment:

- Python version: 3.12.3
- Click version: 8.0.0 to latest

kdeldycke:
The issue reported in #2745 is still present on `stable` branch, in which data from `default_map` is naively processed when `nargs > 1` by `type_cast_value` and unpack each characters from a string.

The fix consist in evaluating these data as if they are environment variables, where we are in a strict string-domain.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,26 @@
 .. currentmodule:: click
 
+Version 8.4.0
+-------------
+
+Unreleased
+
+-   :class:`ParamType` typing improvements. :pr:`3371`
+
+    -   :class:`ParamType` is now a generic abstract base class,
+        parameterized by its converted value type.
+    -   :meth:`~ParamType.convert` return types are narrowed on all
+        concrete types (``str`` for :class:`STRING`, ``int`` for
+        :class:`INT`, etc.).
+    -   :meth:`~ParamType.to_info_dict` returns specific
+        :class:`~typing.TypedDict` subclasses instead of
+        ``dict[str, Any]``.
+    -   :class:`CompositeParamType` and the number-range base are now
+        generic with abstract methods.
+-   Split string values from ``default_map`` for parameters with ``nargs > 1``
+    or :class:`Tuple` type, matching environment variable behavior.
+    :issue:`2745` :pr:`3364`
+
 Version 8.3.3
 -------------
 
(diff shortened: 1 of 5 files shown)


Rowlando13:
Approved! Will merge once 8.3.3 goes out.

kdeldycke:
> Approved! Will merge once 8.3.3 goes out.

Take your time, there is no rush! I am aligning all the obvious bug fixes to purge the backlog for the next dev cycle.

In the mean time, just added a "Multi-value parameters" section to the docs as you suggested in https://github.com/pallets/click/issues/2745#issuecomment-4292508532

kdeldycke:
Thanks for the merge!

PR #3371, opened 2026-04-22
`ParamType` and other typing improvements

kdeldycke:
PR to track and study the unmerged changes from @AndreasBackx that were left in the stale `typing/paramtype`. See the discussion at https://github.com/pallets/click/discussions/3329 for context.

I was able to salvage and rebase on top of `stable` the majority of the initial changes like:
- making `ParamType` a generic ABC and introducing `ParamTypeInfoDict`
- adding `ParamTypeInfoDict`
- updating `FuncParamType` to a generic
- narrowing of `convert()` return types
- `CompositeParamType` generic with abstract `arity`

Stuff that were made obsolete from PRs merged upstream since the last commit in 2024:
- `BoolParamType`: rewritten in #2956
- `Choice` normalization: already part of #2796
- `File` docstring: already cleaned up in #2586

Other stuff that I skipped as too strict:
- `_compat.py` `t.Any` to `t.AnyStr`
- `File` generic with `t.AnyStr`
- `_is_file_like` with `t.AnyStr`

I also extracted from the original `typing/paramtype` the refactor of `convert_type` which is not strictly related to typing improvements and live in its own PR for later evaluation at: https://github.com/pallets/click/pull/3372

Related to https://github.com/pallets/click/pull/2805

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2149,7 +2149,7 @@ class Parameter:
     def __init__(
         self,
         param_decls: cabc.Sequence[str] | None = None,
-        type: types.ParamType | t.Any | None = None,
+        type: types.ParamType[t.Any] | t.Any | None = None,
         required: bool = False,
         # XXX The default historically embed two concepts:
         # - the declaration of a Parameter object carrying the default (handy to
@@ -2181,7 +2181,7 @@ class Parameter:
         self.name, self.opts, self.secondary_opts = self._parse_decls(
             param_decls or (), expose_value
         )
-        self.type: types.ParamType = types.convert_type(type, default)
+        self.type: types.ParamType[t.Any] = types.convert_type(type, default)
 
         # Default nargs to what the type tells us if we have that
         # information available.
@@ -2648,7 +2648,7 @@ class Parameter:
         """Return a list of completions for the incomplete value. If a
         ``shell_complete`` function was given during init, it is used.
         Otherwise, the :attr:`type`
-        :meth:`~click.types.ParamType.shell_complete` function is used.
+        :meth:`~click.types.ParamType[t.Any].shell_complete` function is used.
 
         :param ctx: Invocation context for this command.
         :param incomplete: Value being completed. May be empty.
@@ -2749,7 +2749,7 @@ class Option(Parameter):
         multiple: bool = False,
         count: bool = False,
         allow_from_autoenv: bool = True,
-        type: types.ParamType | t.Any | None = None,
+        type: types.ParamType[t.Any] | t.Any | None = None,
         help: str | None = None,
         hidden: bool = False,
         show_choices: bool = True,
@@ -2825,7 +2825,7 @@ class Option(Parameter):
             if type is None:
                 # A flag without a flag_value is a boolean flag.
                 if flag_value is UNSET:
-                    self.type: types.ParamType = types.BoolParamType()
+                    self.type: types.ParamType[t.Any] = types.BoolParamType()
                 # If the flag value is a boolean, use BoolParamType.
                 elif isinstance(flag_value, bool):
                     self.type = types.BoolParamType()
--- a/src/click/types.py
+++ b/src/click/types.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+import abc
 import collections.abc as cabc
 import enum
 import os
 import stat
 import sys
 import typing as t
+import uuid
 from datetime import datetime
 from gettext import gettext as _
 from gettext import ngettext
@@ -27,7 +29,12 @@ if t.TYPE_CHECKING:
 ParamTypeValue = t.TypeVar("ParamTypeValue")
 
 
-class ParamType:
+class ParamTypeInfoDict(t.TypedDict):
+    param_type: str
+    name: str
+
+
+class ParamType(t.Generic[ParamTypeValue], abc.ABC):
     """Represents the type of a parameter. Validates and converts values
     from the command line or Python into the correct type.
 
@@ -59,7 +66,7 @@ class ParamType:
     #: Windows).
     envvar_list_splitter: t.ClassVar[str | None] = None
 
-    def to_info_dict(self) -> dict[str, t.Any]:
+    def to_info_dict(self) -> ParamTypeInfoDict:
         """Gather information that could be useful for a tool generating
         user-facing documentation.
 
@@ -85,9 +92,10 @@ class ParamType:
         value: t.Any,
         param: Parameter | None = None,
         ctx: Context | None = None,
-    ) -> t.Any:
+    ) -> ParamTypeValue | None:
         if value is not None:
             return self.convert(value, param, ctx)
+        return None
 
     def get_metavar(self, param: Parameter, ctx: Context) -> str | None:
         """Returns the metavar default for this param if it provides one."""
@@ -101,7 +109,7 @@ class ParamType:
 
     def convert(
         self, value: t.Any, param: Parameter | None, ctx: Context | None
-    ) -> t.Any:
+    ) -> ParamTypeValue:
         """Convert the value to the correct type. This is not called if
         the value is ``None`` (the missing value).
 
@@ -121,7 +129,7 @@ class ParamType:
         :param ctx: The current context that arrived at this value. May
             be ``None``.
         """
-        return value
+        return value  # type: ignore[no-any-return]
 
     def split_envvar_value(self, rv: str) -> cabc.Sequence[str]:
         """Given a value from an environment variable this splits it up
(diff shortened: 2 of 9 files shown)


src/click/core.py
```diff
@@ -2149,7 +2149,7 @@ class Parameter:
     def __init__(
         self,
         param_decls: cabc.Sequence[str] | None = None,
-        type: types.ParamType | t.Any | None = None,
```
AndreasBackx, line 2152: Next logical step: Parameter[T]? Of course not for this PR though.

src/click/types.py
```diff
@@ -121,7 +129,7 @@ def convert(
         :param ctx: The current context that arrived at this value. May
             be ``None``.
         """
-        return value
+        return value  # type: ignore[no-any-return]
```
AndreasBackx: I think this function should become an abc method, no? Because the point is that you override this, right?

src/click/types.py
```diff
@@ -160,23 +168,25 @@ def shell_complete(
...
 
-class CompositeParamType(ParamType):
+class CompositeParamType(ParamType[ParamTypeValue]):
     is_composite = True
 
     @property
-    def arity(self) -> int:  # type: ignore
-        raise NotImplementedError()
+    @abc.abstractmethod
+    def arity(self) -> int: ...  # type: ignore[override]
```
AndreasBackx, line 178: Do you know why this type ignore is here in the first place? I wonder if we can get rid of it? It might be there because in the before, the type was incorrect because we never returned int, but now it's an abstract property so we might be able to remove the ignore?

src/click/types.py
```diff
@@ -160,23 +168,25 @@ def shell_complete(
...
+
+
+class FuncParamTypeInfoDict(ParamTypeInfoDict):
+    func: t.Callable[[t.Any], t.Any]
 
 
-class FuncParamType(ParamType):
+class FuncParamType(ParamType[t.Any]):
```
AndreasBackx: I guess it's out of scope for this PR to make thie genericf in its entirety?

src/click/types.py
```diff
@@ -502,29 +528,28 @@ def __init__(
...
 
         rv = super().convert(value, param, ctx)
         lt_min: bool = self.min is not None and (
             operator.le if self.min_open else operator.lt
-        )(rv, self.min)
+        )(rv, self.min)  # type: ignore[arg-type]
         gt_max: bool = self.max is not None and (
             operator.ge if self.max_open else operator.gt
-        )(rv, self.max)
+        )(rv, self.max)  # type: ignore[arg-type]
```
AndreasBackx: We can likely get rid of the type ignore this way:

```suggestion
		min = self.min
		lt_min: bool = min is not None and (
            operator.le if self.min_open else operator.lt
        )(rv, min)
        max = self.max
        gt_max: bool = max is not None and (
            operator.ge if self.max_open else operator.gt
        )(rv, max)
```

Because technically there can be a race condition between checking and using self.min/max.

src/click/types.py
```diff
@@ -1057,7 +1096,11 @@ def shell_complete(
         return [CompletionItem(incomplete, type=type)]
 
 
-class Tuple(CompositeParamType):
+class TupleInfoDict(ParamTypeInfoDict):
+    types: cabc.Sequence[ParamTypeInfoDict]
+
+
+class Tuple(CompositeParamType[tuple[t.Any, ...]]):
```
AndreasBackx, line 1107: As above, I guess out of scope for now?

Though I wonder whether type checkers would see this as all values in the tuple being the same type or not because of the `...`. We should perhaps add a test somewhere for this?

src/click/types.py
```diff
@@ -1071,25 +1114,26 @@ class Tuple(CompositeParamType):
...
 
     @property
-    def name(self) -> str:  # type: ignore
+    def name(self) -> str:  # type: ignore[override]
         return f"<{' '.join(ty.name for ty in self.types)}>"
 
     @property
-    def arity(self) -> int:  # type: ignore
+    def arity(self) -> int:  # type: ignore[override]
         return len(self.types)
```
AndreasBackx, line 1136: Curious to see the typing issue here as well. What does the full typing error say?

AndreasBackx (approved):
We should probably update the example in the docs as well: https://click.palletsprojects.com/en/stable/parameter-types/. Maybe also inline the version updates in the docs.

Approving as this is already an improvement and I don't see any regressions from popping up, besides perhaps the tuple type checking?

src/click/types.py
kdeldycke: Your suggestion does not completely eliminates the `ignore` as it is also silencing `rv: ParamTypeValue`. We can still adopt your code change to narrow it down.

So I reused it and extended it to the `_clamp` block.

As for the race condition, it looks like `NumberRange` instances are not mutable so it looks fine I guess.

kdeldycke: Done in https://github.com/pallets/click/pull/3371/changes/6e981edefaeafa6c16b478dc131fba4d936024f1 .

src/click/types.py
kdeldycke: Making this an abstract method raises `TypeError` in same cases when `convert` is not overrided.

See:
- https://github.com/pallets/click/blob/8bd8b4a074c55c03b6eb5666edc44a9c43df38a2/tests/test_formatting.py#L250-L252
- https://github.com/pallets/click/blob/8bd8b4a074c55c03b6eb5666edc44a9c43df38a2/tests/test_info_dict.py#L272-L273
- https://github.com/pallets/click/blob/8bd8b4a074c55c03b6eb5666edc44a9c43df38a2/docs/shell-completion.md?plain=1#L123-L130

If we making `convert` abstract it breaks these 3 patterns.

kdeldycke: Addressed in: https://github.com/pallets/click/pull/3371/changes/3e2d92c590c34c92e116acdda71e7678ca50895e

src/click/types.py
kdeldycke: If we remove the ignore statements, we get:
```pytb
src/click/types.py:1131: error: Cannot override writeable attribute with read-only property  [override]
        def name(self) -> str:
        ^~~~~~~~~~~~~~~~~~~~~~
src/click/types.py:1135: error: Cannot override writeable attribute with read-only property  [override]
        def arity(self) -> int:
        ^~~~~~~~~~~~~~~~~~~~~~~
```

src/click/core.py
kdeldycke: > Next logical step: Parameter[T]? Of course not for this PR though.

I will let you implement it in a subsequent PR then! 😝 On a more serious note your instinct seems right. Adding `Parameter[T]` look like annotating ~200/400 lines of code. Looks non trivial even for a 8.4 minor release. I want to increase the chance of having this PR merged for 8.4 so we can get rid of your global `typing/paramtype` branch in that repo.

src/click/types.py
kdeldycke: Contrary to `Parameter[T]` this seems small enough to be doable. Let's try...

kdeldycke: Done in https://github.com/pallets/click/pull/3371/changes/78a4efcdce1067566fa58771afbaeb18e2c1cf0f

src/click/types.py
kdeldycke: I propose to defer that to another attempt. Look a bit too intrusive to implement, like `Parameter[T]`.

src/click/types.py
kdeldycke: Removing the ignore we get:
```pytb
src/click/types.py:178: error: Cannot override writeable attribute with read-only property  [override]
        def arity(self) -> int: ...
```

So look like `ParamType.arity` is writeable but the property here is read-only.

kdeldycke:
> We should probably update the example in the docs as well: https://click.palletsprojects.com/en/stable/parameter-types/. Maybe also inline the version updates in the docs.

Just updated the docs at: https://github.com/pallets/click/pull/3371/changes/ded5b692472629154ab5a2f63bd7cdd4deea3146

kdeldycke:
Thanks @AndreasBackx for reviewing your own code and the approval! 😎 Unless you find some quick win to add, I propose to consider this PR ready for final review/merge.

Again, the main objective here is to clear the backlog of PRs and close global branches like `typing/paramtype`.

kdeldycke:
Any reason @Rowlando13 that you merged into `main` and not `stable`? Anyway, thanks for the merge and removal of `typing/paramtype`!

kdeldycke:
> Any reason @Rowlando13 that you merged into `main` and not `stable`?

Just answered my question based on yesterday's activity: seems like we target `main` because we stated an `8.x` release cycle (`8.4.0` in this case), and it seems the tradition is to target `stable` for `8.3.x`/`8.4.x` kind of releases.

Rowlando13:
Yep

Issue #3360, opened 2026-04-21
Empty output from `HelpFormatter.write_usage` for a program without arguments


Melebius:
If no `args` are passed to `HelpFormatter.write_usage`, the Usage line is not printed.

### Reproduction

```python
import click

f = click.HelpFormatter()
f.write_usage("program")
print(f.getvalue())
```

### Expected output
```
Usage: program
```

### Actual output
(an empty line)

### Environment

- Python version: 3.12.3
- Click version: 8.3.1

### Additional information

We use `click` for a CLI program with interactive mode. We are about to use `click.HelpFormatter` to print help for internal commands, some of which do not accept any arguments (`exit`, for example).

kdeldycke:
Look somewhat related to an old edge-case about `write_usage` that was closed by the original author at: https://github.com/pallets/click/issues/1550 . Worth investigating I guess...

kdeldycke:
A fix is being worked on at: https://github.com/pallets/click/pull/3434

kdeldycke:
Lol. We had 13 AI-generated PRs for this issue: #3433, #3402, #3400, #3378, #3367, #3413, #3409, #3386, #3365, #3374, #3401, #3410, #3375.

PR #3375, opened 2026-04-23
fix: handle empty args in HelpFormatter.write_usage

achbj:
## Summary

Fixes #3360 - Empty output from `HelpFormatter.write_usage` for a program without arguments.

When `HelpFormatter.write_usage()` is called with only a program name and no arguments, the output was empty (just a newline) instead of showing the expected usage line like `"Usage: program "`.

### Changes

- Modified `write_usage()` in `src/click/formatting.py` to handle the case when `args` is empty
- Added a condition to check if `args` is truthy before attempting to wrap the arguments text
- When `args` is empty, just writes the usage prefix directly
- Added three regression tests in `tests/test_formatting.py` to verify the fix

### Root Cause

The issue was that `wrap_text("", ...)` returns an empty string, so nothing was written when `args` was empty. The code was trying to wrap an empty string and getting nothing as a result.

### Test Plan

- [x] Run `pytest tests/test_formatting.py -v` - all tests pass
- [x] Run `pytest tests/` - all 1387 tests pass
- [x] Manual verification with reproduction script from issue

```python
import click

f = click.HelpFormatter()
f.write_usage("program")
print(f.getvalue())
# Expected: "Usage: program \n"
# Before fix: "\n"
# After fix: "Usage: program \n"
```

### DIFF
--- a/src/click/formatting.py
+++ b/src/click/formatting.py
@@ -158,27 +158,31 @@ class HelpFormatter:
         usage_prefix = f"{prefix:>{self.current_indent}}{prog} "
         text_width = self.width - self.current_indent
 
-        if text_width >= (term_len(usage_prefix) + 20):
-            # The arguments will fit to the right of the prefix.
-            indent = " " * term_len(usage_prefix)
-            self.write(
-                wrap_text(
-                    args,
-                    text_width,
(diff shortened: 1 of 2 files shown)


kdeldycke:
AI slop

PR #3396, opened 2026-05-01
Fix sentinel typing and its uses in parser

AndreasBackx:
I was working on improving typing by making `Parameter` generic. Though while I was working on that, I noticed that the way we type the sentinel values leads to them being perceived as `Any` as they're not properly typed. Or so I understand. It's past 2AM here so I might be entirely misunderstanding as I'm not entirely familiar with sentinel best practices.

This messed with typing in `Parameter` as where I replaced `Any` with `T_UNSET` and then a generic type, it basically meant `GenericType | Any` thus `Any`.

So in the first commit here, I think I corrected the issue which then lead to surfacing some typing issues in the parser:

```
➜ uv run mypy
src/click/parser.py:85: error: Unsupported operand types for < ("int" and "Literal[Sentinel.UNSET]")  [operator]
            elif nargs > 1:
                 ^
src/click/parser.py:85: note: Left operand is of type "int | Literal[Sentinel.UNSET]"
src/click/parser.py:86: error: Argument 1 to "_fetch" has incompatible type "deque[str]"; expected "deque[str | Literal[Sentinel.UNSET]]"  [arg-type]
                x = [_fetch(args) for _ in range(nargs)]
                            ^~~~
src/click/parser.py:86: error: Argument 1 to "range" has incompatible type "int | Literal[Sentinel.UNSET]"; expected "SupportsIndex"  [arg-type]
                x = [_fetch(args) for _ in range(nargs)]
                                                 ^~~~~
src/click/parser.py:94: error: Unsupported operand types for > ("int" and "Literal[Sentinel.UNSET]")  [operator]
            elif nargs < 0:
                 ^
src/click/parser.py:94: note: Left operand is of type "int | Literal[Sentinel.UNSET]"
src/click/parser.py:108: error: Argument 1 to "tuple" has incompatible type "list[str | tuple[str | Literal[Sentinel.UNSET], ...] | Literal[Sentinel.UNSET]]"; expected
"Iterable[Sequence[str | None] | None]"  [arg-type]
        return tuple(rv), list(args)
                     ^~
src/click/parser.py:194: error: Generator has incompatible item type "int"; expected "bool"  [misc]
                holes = sum(1 for x in value if x is UNSET)
                            ^
src/click/parser.py:194: error: Non-overlapping identity check (left operand type: "str | None", right operand type: "Literal[Sentinel.UNSET]")  [comparison-overlap]
                holes = sum(1 for x in value if x is UNSET)
                                                ^~~~~~~~~~
src/click/parser.py:382: error: Incompatible types in assignment (expression has type "Literal[Sentinel.UNSET]", variable has type "Sequence[str] | Literal[Sentinel.FLAG_NEEDS_VALUE]")  [assignment]
                value = UNSET
                        ^~~~~
src/click/parser.py:412: error: Incompatible types in assignment (expression has type "Literal[Sentinel.UNSET]", variable has type "Sequence[str] | Literal[Sentinel.FLAG_NEEDS_VALUE]")  [assignment]
                    value = UNSET
```

These are addressed in this commit.

### DIFF
--- a/src/click/parser.py
+++ b/src/click/parser.py
@@ -83,7 +83,7 @@ def _unpack_args(
         if nargs == 1:
             rv.append(_fetch(args))
         elif nargs > 1:
-            x = [_fetch(args) for _ in range(nargs)]
+            x: list[str | T_UNSET] = [_fetch(args) for _ in range(nargs)]
 
             # If we're reversed, we're pulling in the arguments in reverse,
             # so we need to turn them around.
--- a/src/click/parser.py
+++ b/src/click/parser.py
@@ -50,7 +50,7 @@ V = t.TypeVar("V")
 
 def _unpack_args(
     args: cabc.Sequence[str], nargs_spec: cabc.Sequence[int]
-) -> tuple[cabc.Sequence[str | cabc.Sequence[str | None] | None], list[str]]:
+) -> tuple[cabc.Sequence[str | cabc.Sequence[str | T_UNSET] | T_UNSET], list[str]]:
     """Given an iterable of arguments and an iterable of nargs specifications,
     it returns a tuple with all the unpacked arguments at the first index
     and all remaining arguments as the second.
@@ -65,7 +65,7 @@ def _unpack_args(
     rv: list[str | tuple[str | T_UNSET, ...] | T_UNSET] = []
     spos: int | None = None
 
-    def _fetch(c: deque[V]) -> V | T_UNSET:
+    def _fetch(c: deque[str]) -> str | T_UNSET:
         try:
             if spos is None:
                 return c.popleft()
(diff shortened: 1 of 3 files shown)


src/click/parser.py
```diff
@@ -75,13 +75,13 @@ def _fetch(c: deque[V]) -> V | T_UNSET:
...
 
     while nargs_spec:
-        nargs = _fetch(nargs_spec)
-
-        if nargs is None:
-            continue
+        if spos is None:
+            nargs = nargs_spec.popleft()
+        else:
+            nargs = nargs_spec.pop()
```
AndreasBackx, line 81: This is basically inlining `_fetch` as we're certain it's never empty.

kdeldycke: I'm good with that kind of changes if it makes the type checker happy.

davidism (approved):

davidism:
These changes seem fine.

Note that sentinel values https://peps.python.org/pep-0661/ was accepted for Python 3.15. We should check that behavior is similar to that, as it gets closer to release.

kdeldycke:
> These changes seem fine.

I don't have strong opinions on types so I'm OK with whatever make type linters happy! 😇

> Note that sentinel values https://peps.python.org/pep-0661/ was accepted for Python 3.15. We should check that behavior is similar to that, as it gets closer to release.

TIL. Thanks for the pointer.

AndreasBackx:
> Note that sentinel values [peps.python.org/pep-0661](https://peps.python.org/pep-0661/) was accepted for Python 3.15. We should check that behavior is similar to that, as it gets closer to release.

I'll merge this for now. I can put up a diff to use the typing_extensions.sentinel backport that exists. Might wants to wait for our CI to have a 3.15 release candidate to test against? Anyhow, I've created #3406.

PR #3402, opened 2026-05-03
Fix `HelpFormatter.write_usage` with empty args

Im-Siyoun:
Fixes #3360.

This fixes `HelpFormatter.write_usage()` when it is called without `args`.

Previously, `write_usage("Program")` passed an empty string to `wrap_text()` while using the usage prefix as `initial_indent`. Since there was no text to wrap, `wrap_text()` returned an empty string and only a blank line was written.

This change preserves the usage prefix for the empty-args case, so `write_usage("Program")` now writes:

```python
"Usage: Program \n"
```

Tests were added to cover this behavior.

Tests run:

```text
pytest tests/test_formatting.py -q
pytest
```

### DIFF
--- a/src/click/formatting.py
+++ b/src/click/formatting.py
@@ -63,7 +63,7 @@ def wrap_text(
         replace_whitespace=False,
     )
     if not preserve_paragraphs:
-        return wrapper.fill(text)
+        return wrapper.fill(text) if text else initial_indent
 
     p: list[tuple[int, bool, str]] = []
     buf: list[str] = []
--- a/tests/test_formatting.py
+++ b/tests/test_formatting.py
@@ -433,3 +433,10 @@ def test_help_formatter_write_text():
     actual = formatter.getvalue()
     expected = "  Lorem ipsum dolor sit amet,\n  consectetur adipiscing elit\n"
     assert actual == expected
+
+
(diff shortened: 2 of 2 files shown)


Im-Siyoun:
I noticed after opening this PR that #3382 and #3383 previously attempted to address the same issue and were closed.

This PR takes a slightly different approach by preserving the existing trailing-space behavior from the usage prefix. In the empty-args case, `write_usage("Program")` writes `"Usage: Program \n"` instead of only `"\n"`.

If this behavior is not desired, or if there is a preferred direction for this issue, I’m happy to adjust the PR or close it.

kdeldycke:
@Rowlando13 this is likely an AI-produced fix comparing it to the number of related PR attempts (10+!). And it is a bit naive and lack some deep unittests. I will close that one and try to produce a new PR for 8.4.0.

Rowlando13:
Noted.

Im-Siyoun:
Thanks for the feedback.

Just to clarify, I used AI assistance for issue discovery and drafting the PR text due to language barriers, but the code changes and tests were written and reviewed by me.

I understand the concern, especially given the number of related PR attempts. I should have checked the related PR history more carefully before opening this PR, and I’ll review the 8.4.0 PR to learn from its approach and test coverage.

PR #3404, opened 2026-05-04
Dual-option value arbitration: `first` vs. `last` vs. `most explicit` policy

### LINKED ISSUE #3403: "default" behaviour in Click 8.3.x changes with enable/disable boolean flag pair
In click 8.2.1, defining a boolean option with both a "enable" and "disable" flag option works with the following logic:
```
import click

@click.command("foo")
@click.option("--without-xyz", "enable_xyz", flag_value=False)
@click.option("--with-xyz", "enable_xyz", flag_value=True, default=True)
def foo(enable_xyz):
    print(f"enable_xyz = {enable_xyz}")

foo()
```

With this code, using click 8.2.1:
* `python foo.py --with-xyz` returns True
* `python foo.py --without-xyz` returns False
* `python foo.py --with-xyz --without-xyz` returns False
* `python foo.py --without-xyz --with-xyz` returns True
* `python foo.py` returns False

In click 8.3.0, the output is the same for every case except the "default"  case - `python foo.py` returns True.

This is an extension of #3111; this [pattern is used in MkDocs](https://github.com/mkdocs/mkdocs/blob/2862536793b3c67d9d83c33e0dd6d50a791928f8/mkdocs/__main__.py#L257) (as of MkDocs 1.6.1) (see mkdocs/mkdocs#4032).

However, the fix for #3111 (#3239) doesn't fix this edge case.

The Properdocs fork of MkDocs [has worked around the issue](https://github.com/ProperDocs/properdocs/pull/14/files).

Environment:

- Python version: 3.12, 3.13, 3.14
- Click version: 8.3.0, 8.3.3

kdeldycke:
Address the value arbitrage regressions from the 8.3.x series, as reported in https://github.com/pallets/click/issues/3403.

### Explanation

Since Click inception, and up to the 8.2.x series, we did [`ctx.params[self.name] = value`](https://github.com/pallets/click/blob/f7cdff5cfc614f34025d1fc238642be029dca1f3/src/click/core.py#L2424). Options sharing the same parameter ID were processed sequentially, and always overwriting each other. Which means the last write was always winning. This is an accident: a policy that was never enforced, explained or checked against.

In #3030, I changed `consume_value`/`get_default` so that the absence of a value became a real manageable `UNSET` state.

But then the naive dual-options tests started failing: https://github.com/pallets/click/pull/3030#discussion_r2290982800 . The less intrusive fix for these tests was implemented and made the first of a dual option win. Which changed the implicit behavior.

The underlying policy or principle governing the dual options arbitrage was never noticed, debated or documented. Because it's been a silent side-effect of the code for 12 years.

The need for arbitraging dual-options were more or less known since the unit tests that were flapping in #3030 existed. But they were never deep enough to highlight and enforce the implicit *last write wins* policy. The regression escaped us because the test were not covering boolean `flag_values`.

### Solution: enforce policy

Re-introduce the *last write win* policy from pre-8.3.x series.

On tie, the *most explicit* source wins. For competing defaults, an explicit default beats an auto-derived one.                

For the implementation, we can not rely on the natural processing order of dual-options. We need to actively manage and track their state so we have a place to explicitly implement the policies.

### Plan

1. [x] Add a shit-load of tests along several combinations of variations:
   - [x] boolean and non-boolean flag_values
   - [x] explicit and non-explicit default
   - [x] explicit and auto-derived defaults
   - [x] competing user-provided dual-options in different orders
   - [x] duplicate competing options provided
   - [x] 2+ competing options sharing the same parameter ID
2. [x] Expect the tests to conform to the policy
3. [x] Make all the tests happy
4. [x] Document, explain and illustrate the policy in the public Sphinx documentation

### Context

The lack of distinction between the absence of value and value of absence, that was fixed by #3030, also add semantic implication for defaults on boolean flags.

It was anticipated by @janluke in https://github.com/pallets/click/issues/1992#issuecomment-892586673 which proposed stricter validation.

Even @ldionne independently proposed to raise an error on the nonsensical `flag_value=False` + `default=True` combinations in its original #3111 report that started the chain.

In both cases I deferred these aggressive changes to a future 9.x release to not change Click internals too much.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -57,6 +57,14 @@ Unreleased
     ``hide_input=True`` fails validation, instead of always showing a
     generic message. Built-in type messages mask the input value.
     :issue:`2809` :pr:`3256`
+-   Fix feature switch groups (several ``flag_value`` options sharing one
+    parameter name) silently dropping an explicit ``default`` when a sibling
+    option without an explicit default was declared first. Arbitration is now
+    source-aware: a more explicit :class:`ParameterSource` always wins, and
+    within ``ParameterSource.DEFAULT``, an option that received an explicit
+    ``default=`` keyword wins over a sibling whose default was auto-derived.
+    The 8.3.x first-wins fallback for remaining ties was reverted to the
+    pre-8.3.x last-wins fallback. :issue:`3403` :pr:`3404`
 
 Version 8.3.3
 -------------
--- a/docs/options.md
+++ b/docs/options.md
@@ -469,15 +469,131 @@ literally.
 
 #### Feature switch groups (multiple flags sharing one variable)
 
-When multiple `flag_value` options target the same parameter
-name, `default=True` on one of them marks it as the default
-choice.
+Several `flag_value` options can target the same parameter name to form a
+feature switch group. The user picks one flag on the command line, and the
+function receives the corresponding `flag_value`. When the user picks none,
+Click falls back to whichever option claims the slot under the arbitration
+rules described below.
+
+##### Non-boolean groups
+
+For non-boolean `flag_value` (strings, enum members, classes, ...), place
+`default=True` on the option that should win when no flag is passed. The
+substitution rule above resolves it to that option's `flag_value`. Any other
+explicit `default` is passed through literally.
 
 | Definition                                             | Not passed | `--upper` | `--lower` |
 |--------------------------------------------------------|------------|-----------|-----------|
 | `--upper` with `flag_value='upper'`, `default=True`    | `"upper"`  | `"upper"` | `"lower"` |
 | `--upper` with `flag_value='upper'`, `default='upper'` | `"upper"`  | `"upper"` | `"lower"` |
-| Both without `default`                                 | `None`     | `"upper"` | `"lower"` |
+| `--upper` with `flag_value='upper'`, `default=None`    | `None`     | `"upper"` | `"lower"` |
+| Neither option carries a `default`                     | `None`     | `"upper"` | `"lower"` |
+
+The third row is the three-state pattern: the function receives `None` when no
+flag is passed, distinguishable from either explicit choice.
+
+##### Boolean groups
+
+When `flag_value` is `True` or `False`, the substitution rule does not apply:
+`default=True` is the literal Python `True`. To make one flag in an
+enable/disable pair the default, set its `default=True` explicitly:
+
+```python
+@click.option("--without-xyz", "enable_xyz", flag_value=False)
+@click.option("--with-xyz", "enable_xyz", flag_value=True, default=True)
+```
+
+| Definition                                              | Not passed | `--with-xyz` | `--without-xyz` |
+|---------------------------------------------------------|------------|--------------|-----------------|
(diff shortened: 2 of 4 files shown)


kdeldycke:
This PR is ready to be reviewed.

davidism:
All this seems very thorough, I'm ok with it if you're sure it represents what you want to implement/maintain with this behavior. Once we document it more strongly like this, we're going to have a harder time changing it (which is fine, just be sure).

kdeldycke:
> All this seems very thorough, I'm ok with it if you're sure it represents what you want to implement/maintain with this behavior. Once we document it more strongly like this, we're going to have a harder time changing it (which is fine, just be sure).

Honestly I don't like the Click's API and behavior around options, flags, default and boolean flags. It was full of surprises and inconsistencies. I was a bit naive a year ago thinking I could fix it in a couple of PRs. But a year in and I'm still here discovering edge-cases and fixing them! 😅 But that's OK, I take full responsibility and will not abandon the project.

During that journey I updated my target: now my goal is to keep collecting edge-cases and real work examples, lockdown the use-cases in unit-tests, and implement work-around and special cases in the code. For me the value of Click is in the unit-tests. Which materialize the contract. Changing the contract and expectations is what cost us a lot in term of maintainance. So I will continue this kind of work for the whole 8.x series.

Then, once there is no more surprises, I will start a discussion with the community to change that contract. And clean up the public API. But that's for 9.x or later (or never, depending on the motivation, feedbacks and possible solutions).

As for that particular PR, the dual flag case is not new to me. I already had some suspicion from last year that their behavior was not deterministic and needed some active state management. It's just that I could not find a proper reproducer or see clearly the problem. https://github.com/pallets/click/issues/3403 report was a welcome reminder and an opportunity to think about this with a fresh mind.

So, to play it safe, I prefer to not changes the expectations of users for the 8.x series. Even at the price of complexity. And document all the subtleties. Click is way too entrenched in the Python eco-system and we still have a lot of work to do.

Rowlando13:
I think last write wins is the expected behavior for cli flags. So I think this change is right.

Rowlando13:
I think this should be the last significant feature for 8.4.0. The release is getting large.

Rowlando13 (approved):

Rowlando13:
I approve. I like how explicit the docs are. Later, I may try to make the docs more concise.

Rowlando13:
Once you have done your squash commits magic, I will merge.

kdeldycke:
> I think this should be the last significant feature for 8.4.0. The release is getting large.

Good to me! It's great to release often and have a small feedback loop. :)

kdeldycke:
> I approve. I like how explicit the docs are. Later, I may try to make the docs more concise.

Ah yes. Because I'm deep into the understanding of how that work I might be too verbose. Maybe placing too much internal knowledge and edge-cases in the user docs. So I'm happy to have you make a second pass to consolidate and make things more digestible.

kdeldycke:
> Once you have done your squash commits magic, I will merge.

Just squashed the commits and rebased on top of `main`. This is ready to be merged.

kdeldycke:

PR #1572, opened 2020-06-02
Add file-like pager: `click.get_pager_file()`

craigds:
In koordinates/kart we're in need of an output pager, and are already using click.

Trouble is, passing a generator to `echo_via_pager` requires that you're producing output in _one_ place. That's not the case with a reasonably complex command - functions and modules delegating to each other all over the place.

To use `echo_via_pager`, I essentially had to use some kind of concurrency - async or threads - and have my code feed output to a worker thread which would yield the output from a generator function, which `echo_via_pager` would consume.

That _worked_, but then I realised click could just expose the file-like object it's using internally (which is usually a pipe to `less`).

This PR does that. `click.get_pager_file` exposes a writable file like object which feeds into a pager. The semantics of `click.echo_via_pager` haven't changed, but it now uses `click.get_pager_file` under the hood and merely feeds the generator contents into the file.

PR #3405, opened 2026-05-04
`click.get_pager_file`: add tests (#1572 followup)

AndreasBackx:
This is a follow-up to #1572 which adds tests to validate the behaviour of `click.get_pager_file`. There are still some open-standing questions / todos depending on requirements:

- No Windows tests (yet) because I'm visiting family right now and only have access to a Linux machine.
	- [ ] What pager like a simple `cat` could we use to create simple tests? We might be able to install `cat` on our CI, though I haven't looked into it yet.
- What are some other edge cases we could check for? So far I've only added some simple tests:
	- Simple unit tests to test some internal implementation details. I'm not the happiest with this, but it seems like the termui testcode favours testing some internals and it was the easiest of the bunch to add.
	- Simple `cat`-based integration tests. We basically want to make sure that it works how we expect with a real pager, but using `cat` seems like the easiest way to validate this as I don't know how we'd test interactive CLIs. :thinking: 
	- [ ] Decide whether the current list of tests are complete enough or if we're missing anything, particularly edge cases.

### DIFF
--- a/tests/test_termui.py
+++ b/tests/test_termui.py
@@ -626,10 +626,14 @@ def test_pager_shlex_split(pager_env, expected_parts):
 
 
 def _get_real_pager_command() -> str:
-    """Return a platform pager used to exercise the BinaryIO pager branch."""
-    pager_name = "more" if WIN else "cat"
-    pager_path = shutil.which(pager_name)
-    assert pager_path is not None, f"{pager_name} not available"
+    """Return a real pager binary path used to exercise the pipe pager branch.
+
+    ..warning::
+        Unix-only for now: ``more.com`` on Windows is interactive and goes
+        through ``_tempfilepager`` rather than ``_pipepager``.
+    """
+    pager_path = shutil.which("cat")
+    assert pager_path is not None, "cat not available"
     return pager_path
 
 
@@ -655,6 +659,10 @@ def _write_pager_from_multiple_sites(pager):
     pager.write("suffix\n")
 
 
+@pytest.mark.skipif(
+    WIN,
+    reason="Exercises the pipe pager path; Windows uses _tempfilepager.",
+)
 @pytest.mark.parametrize(
     ("writer", "color", "expected"),
     [
@@ -685,7 +693,7 @@ def _write_pager_from_multiple_sites(pager):
 def test_get_pager_file_with_real_pager_binary_stream(
     monkeypatch, capfd, writer, color, expected
 ):
-    """A real pager should exercise the BinaryIO branch on Unix and Windows."""
+    """A real pager should exercise the BinaryIO branch."""
     output = _run_get_pager_file_with_real_pager(
         monkeypatch, capfd, writer, color=color
     )
@@ -693,6 +701,10 @@ def test_get_pager_file_with_real_pager_binary_stream(
     assert output == expected
 
 
+@pytest.mark.skipif(
+    WIN,
+    reason="Exercises the pipe pager path; Windows uses _tempfilepager.",
+)
 @pytest.mark.parametrize(
     ("color", "expected"),
     [
--- a/tests/test_termui.py
+++ b/tests/test_termui.py
@@ -1,11 +1,17 @@
+import contextlib
+import io
 import platform
 import shlex
+import shutil
+import subprocess
 import tempfile
 import time
+from functools import partial
 from unittest.mock import patch
 
 import pytest
 
+import click
 import click._termui_impl
 from click._compat import WIN
 from click._termui_impl import Editor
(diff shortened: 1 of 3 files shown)


kdeldycke (changes requested):
Can you add reference to that PR in the changelog along side the one from #1572?

i.e.:
```rst
-   Add ``click.get_pager_file`` for file-like access to an output
    pager. :pr:`1572` :pr:`3405`
```

Also update the `.. versionadded:: 8.2` that was left-over from #1572.

kdeldycke (dismissed):
There are workaround and late changes in #1572 about windows so yes, it's better to try to test on that platform too. If you don't have access to a Windows machine, just remove your skip decorator and use the feedback from GitHub workflow runs. Slow, but doable.

tests/test_termui.py
```diff
@@ -620,6 +626,160 @@ def test_pager_shlex_split(pager_env, expected_parts):
...
+        pager.write("suffix\n")
+
+    output = _run_get_pager_file_with_real_pager(monkeypatch, tmp_path, writer)
+
+    assert output == "prefix\nmiddle\nsuffix\n"
+
+
+@pytest.mark.skipif(WIN, reason="Exercises the pipe pager path.")
+@pytest.mark.parametrize(
+    ("text", "color", "expected"),
```
kdeldycke: I like this matrix of ANSI test cases! 👍

tests/test_termui.py
```diff
@@ -620,6 +626,160 @@ def test_pager_shlex_split(pager_env, expected_parts):
...
+    """An already-wrapped pager stream should be yielded unchanged."""
+    buffer = io.BytesIO()
+    wrapped = click._termui_impl.MaybeStripAnsi(buffer, color=True, encoding="utf-8")
+
+    @contextlib.contextmanager
+    def pager_contextmanager(color=None):
+        yield wrapped, "utf-8", color
+
+    monkeypatch.setattr(
+        click._termui_impl, "_pager_contextmanager", pager_contextmanager
```
kdeldycke: Instead of monkeypatching `_pager_contextmanager` everytwhere, you can just feed `get_pager_file` with `PAGER=`. That way you can check with `PAGER` being set to `cat`, `PAGER` left unset, `PAGER` set to a missing binary etc to test even more edge-cases.

tests/test_termui.py
```diff
@@ -620,6 +626,160 @@ def test_pager_shlex_split(pager_env, expected_parts):
...
+    monkeypatch.setattr(click._termui_impl, "isatty", lambda _: True)
+    monkeypatch.setitem(click._termui_impl.os.environ, "PAGER", pager_path)
+
+    with pager_out.open("w") as f:
+        force_subprocess_stdout = patch.object(
+            subprocess,
+            "Popen",
+            partial(subprocess.Popen, stdout=f),
+        )
+        with force_subprocess_stdout:
```
kdeldycke: Just a style nitpicking: why do you need the temporary `force_subprocess_stdout` variblae? Just fold everything in the with statement.

tests/test_termui.py
kdeldycke: Yes, I think that's the way to go because this is the top level API that is being modified by #1572. So if there are regressions, they will surface from here. Maybe you can even repack your tests to only target that higher-level API and demonstrate more use-cases, and skip testing implementation details.

And you can do the same with the other `echo_via_pager` function that was impacted by #1572.

tests/test_termui.py
```diff
@@ -620,6 +626,160 @@ def test_pager_shlex_split(pager_env, expected_parts):
...
+            with click.get_pager_file(color=color) as pager:
+                writer(pager)
+
+    return pager_out.read_text()
+
+
+@pytest.mark.skipif(WIN, reason="Exercises the pipe pager path.")
```
kdeldycke: I just learned the existence of the `more` utility on Windows. So you can probably test with that as a `PAGER`.

tests/test_termui.py
AndreasBackx: Explains a bit what it does and makes the with statement a bit less unsightly, just a preference and I don't think too much about it in test code.

tests/test_termui.py
AndreasBackx: I've given more a try. Let's see if it passes.

tests/test_termui.py
AndreasBackx: I've redesigned it all a bit because I made the mistake of diving into the differences between BinaryIO, TextIO, io.StringIO, and other io shenanigans. My brain is a bit fried and I'm just going to make dinner as I'm starving.

Let me know what you think. I don't think it's perfect yet, but honestly am just gonna put this down for today and give it another look another day.

AndreasBackx (commented):
Will still update the change entries, but likely not today.

kdeldycke:
Ah. You're right. Windows runners are failing hard on `more`. That's probably because `more` is interactive and waiting user input. I guess it is safe to skip on Windows given the circumstances.

I also spotted some opportunities to extend the tests with `@parametrize` decorator. Let me add a commit on top of your changes to show you what I mean.

kdeldycke:
@AndreasBackx I added a couple of tests around the `PAGER` env var, updated some references and re-introduced the Windows skipping behavior you originally implemented.

What do you think of these changes?

Rowlando13:
I don't want to release 8.4.0 until these go in. @kdeldycke Feel free to keep adding small fixes and I will merge until this is ready. However, when this is ready for merge, I will accept it and just cut the release, but I expect 8.4.1 to come out shortly so I don't expect the prs to sit for long.

kdeldycke:
> I don't want to release 8.4.0 until these go in. @kdeldycke Feel free to keep adding small fixes and I will merge until this is ready. However, when this is ready for merge, I will accept it and just cut the release, but I expect 8.4.1 to come out shortly so I don't expect the prs to sit for long.

Yes, you are right to push for stale PRs and prioritize coverage now. We need a bit of effort to arrive at a final 8.4.0.

kdeldycke:
> I don't want to release 8.4.0 until these go in. @kdeldycke Feel free to keep adding small fixes and I will merge until this is ready. However, when this is ready for merge, I will accept it and just cut the release, but I expect 8.4.1 to come out shortly so I don't expect the prs to sit for long.

Just rebased on top of `main`. This is ready to be merged upstream.

PR #3407, opened 2026-05-04
`click.prompt` typing clarifications / improvements

AndreasBackx:
This improves the typing of `click.prompt`. However there is still an opening question / issue...

# Brief summary

The current code makes the following change to the public API of `click.prompt`:

```diff
 def prompt(
     text: str,
-    default: t.Any | None = None,
+    default: V | None = None,
     hide_input: bool = False,
     confirmation_prompt: bool | str = False,
-    type: ParamType[t.Any] | t.Any | None = None,
-    value_proc: t.Callable[[str], t.Any] | None = None,
+    type: ParamType[V] | V | None = None,
+    value_proc: t.Callable[[str], V] | None = None,
     prompt_suffix: str = ": ",
     show_default: bool | str = True,
     err: bool = False,
     show_choices: bool = True,
-) -> t.Any:
+) -> V:
````

This changes the behaviour in that you can no longer pass types that aren't the same type as the type you expect at the end. For example:

```python
click.prompt("Iteration amount?", default="100", type=int)
```

Would no longer work under the current implementation because `"100"` is not of the type `int`. In the old implementation this worked because default values still were passed as `int("100")` before being returned. The remaining question is then what behaviour we should go for.

You could say, "why" not keep the current implementation? Well we could, but it's not actually typed correctly and actually allows behaviour we don't account for. That behaviour is that `value_proc: t.Callable[[str], V] | None = None` accepts only `str` to convert to the target type. So technically you shouldn't be allowed to pass `Any` to it. However.... `ParamType.__call__`, which makes `ParamType` also a callable, is used to convert the `default` to the target `type` as well and that accepts `Any`, not `str`. So there's not really a strict contract anywhere.

To make the decision on this PR easier I propose the following options:

## 1. Restrict `default` to same type (current implementation)

- Accept only the same type as the type passed in `type` in `default`.
- DO NOT pass `default` to `value_proc` or the constructor of `type`.

```diff
 def prompt(
     text: str,
-    default: t.Any | None = None,
+    default: V | None = None,
     hide_input: bool = False,
     confirmation_prompt: bool | str = False,
-    type: ParamType[t.Any] | t.Any | None = None,
-    value_proc: t.Callable[[str], t.Any] | None = None,
+    type: ParamType[V] | V | None = None,
+    value_proc: t.Callable[[str], V] | None = None,
     prompt_suffix: str = ": ",
     show_default: bool | str = True,
     err: bool = False,
     show_choices: bool = True,
-) -> t.Any:
+) -> V:
````

## 2. Public API typing change (or correction) with unlikely users

- Accept the same type as the type passed in `type` in `default` and also `str`.
- If `default` type is different from the type of `type`, pass it to `value_proc` or the constructor of `type` to convert it.
	- Make `ParamType.__call__` also onyl accept `str`, not `Any`.

```diff
 def prompt(
     text: str,
-    default: t.Any | None = None,
+    default: V | str | None = None,
     hide_input: bool = False,
     confirmation_prompt: bool | str = False,
-    type: ParamType[t.Any] | t.Any | None = None,
-    value_proc: t.Callable[[str], t.Any] | None = None,
+    type: ParamType[V] | V | None = None,
+    value_proc: t.Callable[[str], V] | None = None,
     prompt_suffix: str = ": ",
     show_default: bool | str = True,
     err: bool = False,
     show_choices: bool = True,
-) -> t.Any:
+) -> V:
````

```diff
    @t.overload
    def __call__(
        self,
        value: None,
        param: Parameter | None = None,
        ctx: Context | None = None,
    ) -> None: ...

    @t.overload
    def __call__(
        self,
-       value: t.Any,
+       value: str,
        param: Parameter | None = None,
        ctx: Context | None = None,
    ) -> ParamTypeValue: ...

    def __call__(
        self,
-       value: t.Any,
+       value: str,
        param: Parameter | None = None,
        ctx: Context | None = None,
    ) -> ParamTypeValue | None:
        if value is not None:
            return self.convert(value, param, ctx)
        return None

    def convert(
        self,
-       value: t.Any,
+       value: str,
		param: Parameter | None,
		ctx: Context | None,
    ) -> ParamTypeValue:

```

## 3. Change `value_proc` to accept `Any`

- Accept `Any` in `default`, keeping the same accepted type as before.
- If `default` type is different from the type of `type`, pass it to `value_proc` or the constructor of `type` to convert it.

```diff
 def prompt(
     text: str,
-    default: t.Any | None = None,
+    default: V | t.Any | None = None,
     hide_input: bool = False,
     confirmation_prompt: bool | str = False,
-    type: ParamType[t.Any] | t.Any | None = None,
-    value_proc: t.Callable[[str], t.Any] | None = None,
+    type: ParamType[V] | V | None = None,
+    value_proc: t.Callable[[t.Any], V] | None = None,
     prompt_suffix: str = ": ",
     show_default: bool | str = True,
     err: bool = False,
     show_choices: bool = True,
-) -> t.Any:
+) -> V:
````

# My preference?

Even though I've implemented the first option, I think the second option makes the most sense. I don't think there's a usecase for accepting beyond `str` for parsing.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -51,13 +51,11 @@ Unreleased
   parallel, use process-based isolation (such as `pytest-xdist`) rather
   than threads, since {meth}`CliRunner.invoke` also redirects the
   process-global standard streams. {issue}`3501` {issue}`3700` {pr}`3704`
-- `prompt()` is now generically typed and returns the type produced by `type`
-  or `value_proc` instead of `Any`. {class}`ParamType` takes a second optional
-  type parameter describing the input value it accepts (`ParamType[int, str]`
-  for a type converting strings to integers), defaulting to `Any`.
-  Additionally, a `prompt()` default that is already of the expected type is
-  returned as-is instead of doing a round trip through `value_proc` or the
-  type conversion. {pr}`3407`
+- `prompt()` is now generically typed and returns the type produced by
+  `type`, `value_proc`, or a matching `default` instead of `Any`.
+  {class}`ParamType` takes a second optional type parameter describing the
+  input value it accepts (`ParamType[int, str]` for a type converting
+  strings to integers), defaulting to `Any`. {pr}`3407`
 
 ## Version 8.4.2
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -51,6 +51,13 @@ Unreleased
   parallel, use process-based isolation (such as `pytest-xdist`) rather
   than threads, since {meth}`CliRunner.invoke` also redirects the
   process-global standard streams. {issue}`3501` {issue}`3700` {pr}`3704`
+- `prompt()` is now generically typed and returns the type produced by `type`
+  or `value_proc` instead of `Any`. {class}`ParamType` takes a second optional
+  type parameter describing the input value it accepts (`ParamType[int, str]`
+  for a type converting strings to integers), defaulting to `Any`.
+  Additionally, a `prompt()` default that is already of the expected type is
+  returned as-is instead of doing a round trip through `value_proc` or the
+  type conversion. {pr}`3407`
 
 ## Version 8.4.2
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,9 +14,6 @@ classifiers = [
     "Typing :: Typed",
 ]
 requires-python = ">=3.10"
-dependencies = [
-    "typing_extensions; python_version < '3.13'",
-]
 
 [project.urls]
 Donate = "https://palletsprojects.com/donate"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,9 @@ classifiers = [
     "Typing :: Typed",
 ]
 requires-python = ">=3.10"
+dependencies = [
+    "typing_extensions; python_version < '3.13'",
+]
 
 [project.urls]
 Donate = "https://palletsprojects.com/donate"
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import builtins
 import collections.abc as cabc
 import inspect
 import io
@@ -27,13 +26,7 @@ from .utils import echo
 if t.TYPE_CHECKING:
     from ._termui_impl import ProgressBar
 
-    if sys.version_info >= (3, 13):
-        from typing import TypeIs
-    else:
-        from typing_extensions import TypeIs
-
 V = t.TypeVar("V")
-C = t.TypeVar("C")
 
 # The prompt functions to use.  The doc tools currently override these
 # functions to customize how they work.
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

**Comment by kdeldycke** [comment]
The question isn't about the ideal type, it's what the other developers already rely on. We're in Hyrum's Law territory here: Click is too entranched in the ecosystem that we have users depending on behaviors we never strictly specified. Let's pin the current runtime behavior with tests before tightening anything.

Does option 3 breaks the actual unittests? Can you come up with tests that encode the actually behaviour we allow but we don't account for?

**Comment by AndreasBackx** [comment]
@kdeldycke I've made changes to the typing, though it has lead to the introduction of a new generic type being added to `ParamType`, one that identifies that input value type. This should be fine and should make typing work throughout.

However, to make it easier for people and not require custom `ParamType` instances (and the rest of our codebase atm) to use it yet, I've defaulted it to `t.Any`. That default however requires Python 3.13, or for us to start importing `typing_extensions` for `3.12<=`, which is currently not being done as it's only used for typing. Is it okay to add `typing_extensions` here as an import?

**Comment by kdeldycke** [comment]
> @kdeldycke I've made changes to the typing, though it has lead to the introduction of a new generic type being added to `ParamType`, one that identifies that input value type. This should be fine and should make typing work throughout.

Yeah that looks good! And indeed: it passes on Python 3.13 without issues.

> Is it okay to add `typing_extensions` here as an import?

If it would be my project I would say yes. But not sure what's our policy for Click. So thanks for asking! @davidism, @Rowlando13, @ThiefMaster, any opinion on adding `typing_extensions` dependency?

In the meantime @AndreasBackx add that dependency to this PR so you can demonstrate that it make your changes pass on all Pythons supported by Click. Could also help other maintainers form an opinion by reading the actual impact of your changes.

**Comment by AndreasBackx** [comment]
@kdeldycke all done. Thanks for the review.

**Comment by kdeldycke** [comment]
> @kdeldycke all done. Thanks for the review.

Let's now wait for others to chime in, especially about the added `typing-extensions` dependency. Other than that, I'm OK with these changes.

**Comment by davidism** [comment]
I've historically tried to keep static typing in its own space. Using string annotations, then `from __future__ import typing`, and not adding a runtime dependency on typing-extensions.

Admittedly, typing-extensions is basically the number one downloaded package, so I'm more confident about it now. But Pallets in general avoids external required dependencies. I'd prefer leaving it guarded by `TYPE_CHECKING`, but perhaps we can revisit it in a separate discussion than this.

**Comment by kdeldycke** [comment]
> I'd prefer leaving it guarded by `TYPE_CHECKING`, but perhaps we can revisit it in a separate discussion than this.

Yes, that's also [the idiom I adopted in my projects](https://github.com/kdeldycke/click-extra/blob/main/click_extra/config.py#L88-L93) as `TYPE_CHECKING` works well with `mypy`: 

```python
TYPE_CHECKING = False
if TYPE_CHECKING:
    from collections.abc import Callable, Sequence
    from typing import Any, Literal

    import click
```

kdeldycke:
Hey @AndreasBackx , I took the liberty to re-review this PR and rebased your proposal on top of `main`. I squashed all your original commits into one and then added my own to try to make it work. My intention is to clear the backlog of stale PRs.

Mainly: I tried to get rid of `typing-extensions` dependency so we keep Click without any, now that we have removed Colorama for the upcoming 8.5.0. For this I had to tweak some behavior that are now default in 3.13. Please review my `__class_getitem__` hack as I'm not quite sure this is the way to go. But at least this allow us to not depends on `typing-extensions`.

Also, by adding some tests I had to change your original `prompt()` signature, and remove `_is_expected_type` as it became unneccessary.

Rowlando13:
> Hey @AndreasBackx , I took the liberty to re-review this PR and rebased your proposal on top of `main`. I squashed all your original commits into one and then added my own to try to make it work. My intention is to clear the backlog of stale PRs.
> 
> Mainly: I tried to get rid of `typing-extensions` dependency so we keep Click without any, now that we have removed Colorama for the upcoming 8.5.0. For this I had to tweak some behavior that are now default in 3.13. Please review my `__class_getitem__` hack as I'm not quite sure this is the way to go. But at least this allow us to not depends on `typing-extensions`.
> 
> Also, by adding some tests I had to change your original `prompt()` signature, and remove `_is_expected_type` as it became unneccessary.

I think zero dependencies is the right move.

Rowlando13:
@kdeldycke looks good. Did you want to want for @AndreasBackx 's feedback before merge?

Rowlando13 (approved):

kdeldycke:
> @kdeldycke looks good. Did you want to want for @AndreasBackx 's feedback before merge?

I guess we can proceed with the merge as initially it was @AndreasBackx who requested my/our review on this PR (about the `typing-extensions` dependency)! 😅 And if I did something really wrong from @AndreasBackx point of view I'll be happy to correct it in a subsequent PR.

PR #3408, opened 2026-05-05
Organize tests utils

Rowlando13:


Rowlando13:
For test_termui.py look at changes after `feffea5` on Apr 2, 2026.

PR #3409, opened 2026-05-05
fix: handle empty args in HelpFormatter.write_usage

zons-zhaozhy:
When `write_usage(prog, args)` is called with an empty `args`,
the usage line is not printed (just a blank line). This happens because
`wrap_text("")` returns `""`, so only `"\n"` gets written.

Fix: add an early return when `args` is empty to write the usage
prefix and program name directly.

Fixes #3360

### DIFF
--- a/src/click/formatting.py
+++ b/src/click/formatting.py
@@ -155,6 +155,10 @@ class HelpFormatter:
         if prefix is None:
             prefix = f"{_('Usage:')} "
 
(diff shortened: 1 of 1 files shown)


zons-zhaozhy:
Closing this PR to revisit the fix with additional testing and validation. I'll submit a new PR after thorough verification.

PR #3412, opened 2026-05-08
Add pyrefly type checking.

AndreasBackx:
This adds pyrefly to the type checking. I'm working on a lot of typing improvements and it's helpful to have a third one because it's sometimes not really clear why pyright is saying typing is wrong. It already is saying some things are incorrect which I'm planning to look into.

I hope it's not going to lead to a conflict of typing issues, and if so, we can review this more closely in the future.

As of writing I've not fixed the typing issues, but I'll publish the PR once done.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -59,3 +59,18 @@ jobs:
           path: ./.mypy_cache
           key: mypy|${{ hashFiles('pyproject.toml') }}
       - run: uv run --locked --no-default-groups --group dev tox run -e typing
+  pyrefly:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57 # v8.0.0
+        with:
+          enable-cache: true
+          prune-cache: false
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version-file: pyproject.toml
+      - run: uv run --locked --no-default-groups --group dev tox run -e pyrefly -- --output-format=github
--- a/src/click/_compat.py
+++ b/src/click/_compat.py
@@ -16,6 +16,7 @@ if t.TYPE_CHECKING:
 else:
     OpenBinaryMode = OpenTextMode = str
 
+AnyStr = t.TypeVar("AnyStr", str, bytes)
 OpenFileMode = OpenTextMode | OpenBinaryMode
 
 CYGWIN = sys.platform.startswith("cygwin")
@@ -386,12 +387,12 @@ def _wrap_io_open(
     mode: OpenFileMode = "r",
     encoding: str | None = None,
     errors: str | None = "strict",
-) -> t.IO[t.Any]:
+) -> t.BinaryIO | t.TextIO:
     """Handles not passing ``encoding`` and ``errors`` in binary mode."""
     if "b" in mode:
-        return open(file, mode)
+        return t.cast(t.BinaryIO, open(file, mode))
 
-    return open(file, mode, encoding=encoding, errors=errors)
+    return t.cast(t.TextIO, open(file, mode, encoding=encoding, errors=errors))
 
 
 @t.overload
@@ -401,7 +402,7 @@ def open_stream(
     encoding: str | None = None,
     errors: str | None = "strict",
     atomic: bool = False,
-) -> tuple[t.BinaryIO, bool]: ...
+) -> tuple[t.IO[bytes], bool]: ...
 
 
 @t.overload
@@ -411,7 +412,7 @@ def open_stream(
     encoding: str | None = None,
     errors: str | None = "strict",
     atomic: bool = False,
-) -> tuple[t.TextIO, bool]: ...
+) -> tuple[t.IO[str], bool]: ...
 
 
 def open_stream(
@@ -420,7 +421,7 @@ def open_stream(
     encoding: str | None = None,
     errors: str | None = "strict",
     atomic: bool = False,
-) -> tuple[t.IO[t.Any], bool]:
+) -> tuple[t.IO[bytes] | t.IO[str], bool]:
     binary = "b" in mode
     filename = os.fspath(filename)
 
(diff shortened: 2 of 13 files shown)


kdeldycke:
Typing is not my forte. In my projects I just fire up a `mypy` and try to make it happy. Waiting for `ty` to mature a bit and I'll switch to it in the future.

So I'll let other give feedback on that topic and I'll just follow the movement! 😁

davidism:
I'd prefer to stick to mypy for now. If we're still testing with pyright, we can remove that; it was added when I thought we needed to test with all of them. But really, we just need to test with `mypy` for our own code, and `pyright --verifytypes` to know that we export complete types for others to use in _their_ code.

I know pyrefly is working on an export verifcation as well, but I tried last month and both its type checking and verifcation weren't ready yet. Same with ty. I'm really interested in those two, but prefer we wait for now. mypy, for all its quirks, is very mature and actively maintained.

davidism:
Regarding static typing in general, remember: annotations are there to serve us and our users, not the other way around. Click (and all of Pallets) predates static typing, and we would design it differently today to be better typed. So if we're doing something weird and typing can't express it (or it's too complicated to figure out), you're fine adding `# type: ignore[error code]` comments to insist that a return type is what we say it is, that a use of an argument is correct, etc.

AndreasBackx:
Wasn't expecting the comments just yet. :sweat_smile: I was primarily using this to experiment a bit on where we could improve typing. It's very neat to see where we could make the typing experience Click offers better, and also make it less of a pain to use inside of the codebase. So far it's teaching and frustrating me, as I'm expecting certain typing to be possible (as someone who uses Rust quite a bit) and then it seems to not be possible.

My current struggle is trying to plug through the binary/text file modes without having to repeat `if "b" in mode` everywhere and casting. There's fairly little that I can find on this and how others solve it.

I'll see where my experimentation gets me. But I'm not expecting all that much just yet.

So far Pyrefly is easier to use than `pyright`:
1. No waiting multiple seconds to type check the codebase.
2. Actually readable output from the CLI.

And `mypy` seems to sometimes not be as strict as `pyright` nor `pyrefly`, I've found it to be useful to see what the different ones say.

Regardless of where the typing experimentation goes, I do think it'd be useful to have multiple type checkers available. It sometimes helps to see how each one of them views a problem, even when we only care about a single one (mypy). Would we be able to have pyrefly and ty added and available via `uvx pyrefly/ty` which would make this easier (but not enabled in CI/tox)?

davidism:
Yeah it's fine to add config for it and use it to figure things out, just don't want to require it in CI yet.

Rowlando13:
I am interested to see where this goes. I switched to pyrefly mostly because the Language server works better in vscode.

.github/workflows/tests.yaml
```diff
@@ -59,3 +59,18 @@ jobs:
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+      - uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57 # v8.0.0
+        with:
+          enable-cache: true
+          prune-cache: false
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version-file: pyproject.toml
```
jorenham, line 75: There's no need for `setup-python` if you're using `setup-uv`

src/click/_compat.py
```diff
@@ -355,26 +364,64 @@ def get_text_stderr(encoding: str | None = None, errors: str | None = None) -> t
...
+def open_stream(
+    filename: str | os.PathLike[str],
+    mode: OpenTextMode = "r",
+    encoding: str | None = None,
+    errors: str | None = "strict",
+    atomic: bool = False,
+) -> tuple[t.IO[str], bool]: ...
+
+
+def open_stream(
```
jorenham, line 418: nit: those newlines here aren't necessary, and makes things easier to read IMO.


```suggestion
) -> tuple[t.IO[bytes], bool]: ...
@t.overload
def open_stream(
    filename: str | os.PathLike[str],
    mode: OpenTextMode = "r",
    encoding: str | None = None,
    errors: str | None = "strict",
    atomic: bool = False,
) -> tuple[t.IO[str], bool]: ...
def open_stream(
```

(btw, I like these overloads)

src/click/_compat.py
```diff
@@ -465,12 +520,11 @@ def close(self, delete: bool = False) -> None:
             return
         self._f.close()
         os.replace(self._tmp_filename, self._real_filename)
-        self.closed = True
 
     def __getattr__(self, name: str) -> t.Any:
         return getattr(self._f, name)
 
-    def __enter__(self) -> _AtomicFile:
+    def __enter__(self) -> _AtomicFile[AnyStr]:
```
jorenham, line 527:
```suggestion
    def __enter__(self) -> t.Self:
```

src/click/_compat.py
```diff
@@ -444,17 +491,25 @@ def open_stream(
...
+    text_af = _AtomicFile(text_f, tmp_filename, os.path.realpath(filename))
+    return t.cast(t.IO[t.Any], text_af), True
 
 
-class _AtomicFile:
-    def __init__(self, f: t.IO[t.Any], tmp_filename: str, real_filename: str) -> None:
-        self._f = f
+class _AtomicFile(t.IO[AnyStr], t.Generic[AnyStr]):
+    def __init__(self, f: t.IO[AnyStr], tmp_filename: str, real_filename: str) -> None:
+        self._f: t.IO[AnyStr] = f
```
jorenham, line 506: It's almost always better to avoid using `AnyStr` if you can (there are differences between how type-checkers treat it because it's very underspecified in the typing spec). And here, you can parametrize the entire `f` type, i.e. with 

```py
_FileT_co = t.TypeVar("_FileT_co", bound=t.IO[Any], covariant=True)
```

This way you capture more typing information, avoid issues with invariance, and avoid having to deal with the awkward `AnyStr`.

src/click/_winconsole.py
```diff
@@ -70,19 +70,18 @@
...
-    try:
-        # Using `typing_extensions.Buffer` instead of `collections.abc`
-        # on Windows for some reason does not have `Sized` implemented.
-        from collections.abc import Buffer  # type: ignore
-    except ImportError:
-        from typing_extensions import Buffer
+    from typing_extensions import Buffer
+
+    class _MSVCRTModule(t.Protocol):
+        def get_osfhandle(self, fd: int) -> int: ...
```
jorenham, line 76: nit:


```suggestion
        def get_osfhandle(self, fd: int, /) -> int: ...
```

src/click/shell_completion.py
```diff
@@ -436,7 +436,7 @@ def format_completion(self, item: CompletionItem) -> str:
         return f"{item.type}\n{value}\n{help_escaped}"
 
 
-ShellCompleteType = t.TypeVar("ShellCompleteType", bound="type[ShellComplete]")
+ShellCompleteType = t.TypeVar("ShellCompleteType", bound="ShellComplete")
```
jorenham, line 439: nit: Now that it's no longer bound to a `type`, the name isn't accurate anymore. The type variable naming convention is to use private (underscore-prefixed) naming and a `T` as suffix.

src/click/termui.py
```diff
@@ -696,7 +696,7 @@ def unstyle(text: str) -> str:
 
 def secho(
     message: t.Any | None = None,
-    file: t.IO[t.AnyStr] | None = None,
+    file: t.TextIO | t.BinaryIO | None = None,
```
jorenham, line 699: t.TextIO | t.BinaryIO` is a bit stricter (narrower) than `t.IO`, which might lead to false positives. 

```suggestion
    file: t.IO[Any] | None = None,
```

src/click/types.py
```diff
@@ -787,7 +794,7 @@ class FileInfoDict(ParamTypeInfoDict):
     encoding: str | None
 
 
-class File(ParamType[t.IO[t.Any]]):
+class File(ParamType[t.IO[AnyStr]], t.Generic[AnyStr, OpenFileMode]):
```
jorenham, line 797: Same as above; it's better to parametrize this using a covariant type variable with `bound=t.IO[Any]`

jorenham, line 797: I would strongly recommend against using `OpenFileMode` here: Pyright, for example, tends to upcasts `Literal` string values to `str` when used in generics, so this can lead to unexpected and different outcomes between type-checkers.

If I were in you should, I'd instead use a `TypeVar(..., bound=bytes | str)` here, and update the `__init__` overloads accordingly.

jorenham (commented):
Yay Pyrefly :tada:.

I took a quick look and left some comments; hope you don't mind.

PR #3422, opened 2026-05-12
Add type annotations for instance attributes in `utils`

jorenham:
I noticed that the type coverage of `click` was pretty high, but not quite at 100% yet (according to typestats: https://jorenham.github.io/typestats/dashboard/report/#click). So towards that end, this adds type annotations for the instance attributes of the classes in `click.utils`, increasing the type-coverage of `click` by almost 0.5% (90.97% -> 91.47%).

If you're open to it, I'd also like to open PRs for the other modules, so that we can reach that sweet 100% type coverage :)

### DIFF
--- a/src/click/utils.py
+++ b/src/click/utils.py
@@ -205,8 +205,10 @@ class LazyFile:
 
 
 class KeepOpenFile:
+    _file: t.IO[t.Any]
+
     def __init__(self, file: t.IO[t.Any]) -> None:
-        self._file: t.IO[t.Any] = file
+        self._file = file
 
     def __getattr__(self, name: str) -> t.Any:
         return getattr(self._file, name)
--- a/src/click/utils.py
+++ b/src/click/utils.py
@@ -117,6 +117,14 @@ class LazyFile:
     files for writing.
     """
 
+    name: str
+    mode: str
+    encoding: str | None
+    errors: str | None
+    atomic: bool
+    _f: t.IO[t.Any] | None
+    should_close: bool
+
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


kdeldycke:
I'm OK with these changes! :)

src/click/utils.py
```diff
@@ -117,21 +117,27 @@ class LazyFile:
...
 
+    name: str
+    mode: str
+    encoding: str | None
+    errors: str | None
+    atomic: bool
+    _f: t.IO[t.Any] | None
+    should_close: bool
+
     def __init__(
```
Rowlando13, line 128: I don't know enough about type hints to review this. Can you break down why list them out this way instead of the way that they are done in line 133 and 134?

kdeldycke: Probably for visual consistency or to please the type checker. Had to do these kind of things in the past. Sometimes introducing subtle bugs because of global class variables promotion. Hard to say in details without @jorenham feedback.

kdeldycke: Yeah, I'll classify that change as idiomatic. There is no danger as the class-level variable are not assigned any value by default.

kdeldycke:
One nitpick: if the goal is moving annotations to class level, you missed one in `KeepOpenFile`. So I just made that change on top of @jorenham . This is ready to be merged.

jorenham:
> One nitpick: if the goal is moving annotations to class level, I missed one in `KeepOpenFile`. So I just made that change on top of @jorenham . This is ready to be merged.

Ah very nice :ok_hand:

PR #3423, opened 2026-05-13
Add missing space between option help text and deprecation label

sjoubert:
Without the space, a deprecated documented option currently appears as:
`--deprecated-opt   Old option(DEPRECATED[: too old])`
After:
`--deprecated-opt   Old option (DEPRECATED[: too old])`

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2786,7 +2786,11 @@ class Option(Parameter):
                 if isinstance(deprecated, str)
                 else "(DEPRECATED)"
             )
-            help = help + deprecated_message if help is not None else deprecated_message
+            help = (
+                f"{help} {deprecated_message}"
+                if help is not None
+                else deprecated_message
+            )
 
         self.prompt = prompt_text
         self.confirmation_prompt = confirmation_prompt


kdeldycke:
Good catch. The other deprecated messages in `Command.get_short_help_str` and `Command.format_help_text` do introduce a space:
- https://github.com/pallets/click/blob/c943271a269e6941fcc51e3506ead074b9dda6be/src/click/core.py#L1138-L1145
- https://github.com/pallets/click/blob/c943271a269e6941fcc51e3506ead074b9dda6be/src/click/core.py#L1174-L1181

So this one is a legitimate inconsistency.

Now looking at that code maybe there is an opportunity for deduplicating code...

kdeldycke:
> So this one is a legitimate inconsistency.
> 
> Now looking at that code maybe there is an opportunity for deduplicating code...

I will merge this simple bug fix, and think about the code deduplication later.

kdeldycke:
> Now looking at that code maybe there is an opportunity for deduplicating code...

Good news, there is a real oportunity to deduplicate these messages. I proposed a PR at: https://github.com/pallets/click/pull/3430

Issue #3362, opened 2026-04-21
`HelpFormatter.write_usage` breaks options at a hyphen


Melebius:
If an option containing a hyphen in its name is printed at the line break limit, it’s broken at the hyphen.

### Reproduction

```python
import click

options = [
    "--enable-verbose-logging",
    "--output-file-path",
    "--max-retry-count",
    "--disable-cache-mode",
    "--config-file-location",
    "--user-auth-token",
    "--auto-update-interval",
    "--force-overwrite-existing",
    "--network-timeout-seconds",
    "--debug-trace-enabled",
]

f = click.HelpFormatter(width=65)
f.write_usage("program", " ".join(options))
print(f.getvalue())
```

### Expected output
```
Usage: program --enable-verbose-logging --output-file-path
               --max-retry-count --disable-cache-mode
               --config-file-location --user-auth-token
               --auto-update-interval --force-overwrite-existing
               --network-timeout-seconds --debug-trace-enabled
```

### Actual output
```
Usage: program --enable-verbose-logging --output-file-path --max-
               retry-count --disable-cache-mode --config-file-
               location --user-auth-token --auto-update-interval
               --force-overwrite-existing --network-timeout-
               seconds --debug-trace-enabled
```

### Environment

- Python version: 3.12.3
- Click version: 8.3.1

#### `width` option

This also happens without specifying `width`. The `width` option is altered in the example to make the issue more frequent.

#### Relevance to other use cases

This may likely only affect:

- direct users of `click.HelpFormatter`,
- users of `click.argument` with custom `metavar` containing hyphens. 

On the other hand, `click` prints `[OPTIONS]` instead of listing individual options on the usage line and uses underscores instead of hyphens in argument names (unless specifying `metavar`).

#### Cause

I looked into the code and found that `textwrap.TextWrapper` is used under the hood, which has the [`break_on_hyphens`](https://docs.python.org/3/library/textwrap.html#textwrap.TextWrapper.break_on_hyphens) option. However, there is no option for the `click` user to influence this, apart from reimplementing `HelpFormatter.write_usage` (in a subclass) and `wrap_text`.

Pitchfork-and-Torch:
I reproduced this on current `main` and pushed a fix branch:

https://github.com/pallets/click/compare/main...Pitchfork-and-Torch:click:fix/usage-hyphen-wrap?expand=1

[...]

PR #3425, opened 2026-05-14
 Fix usage wrapping for hyphenated options

RajvardhanPatil07:
## Summary
- add a break_on_hyphens option to Click's internal wrap_text helper
- disable hyphen breaks when wrapping HelpFormatter.write_usage argument strings
- add a regression test for hyphenated long option names in usage output

## Why
HelpFormatter.write_usage currently delegates to textwrap.TextWrapper with its default break_on_hyphens=True, so usage tokens like --max-retry-count can be split into --max- and retry-count. Usage output is easier to read when option names and metavars stay intact across wraps.

Fixes #3362.

## Validation
- Ran the issue repro with PYTHONPATH=src using Python 3.12.13
- .venv/bin/python -m pytest tests/test_formatting.py -q
- .venv/bin/python -m pytest -q

### DIFF
--- a/src/click/formatting.py
+++ b/src/click/formatting.py
@@ -34,6 +34,7 @@ def wrap_text(
     initial_indent: str = "",
     subsequent_indent: str = "",
     preserve_paragraphs: bool = False,
+    break_on_hyphens: bool = True,
 ) -> str:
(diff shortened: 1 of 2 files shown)


ThiefMaster:
LLM-generated PRs are not welcome in our projects.

PR #3430, opened 2026-05-15
Deduplicate deprecated label translation

kdeldycke:
A small bugfix PR at https://github.com/pallets/click/pull/3423 hinted at inconsistencies in the way deprecated messages were produced and translated.

We discussed these message translation on PR https://github.com/pallets/click/pull/2902 , which reminded me of duplicate messages.

Turns out we can deduplicate these messages with some simple helper.

CC @sjoubert @ic @v--

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -93,6 +93,23 @@ def _check_nested_chain(
     raise RuntimeError(message)
 
 
+def _format_deprecated_label(deprecated: bool | str) -> str:
+    """Return the parenthesized deprecation label shown in help text."""
+    label = _("deprecated").upper()
+    if isinstance(deprecated, str):
+        return f"({label}: {deprecated})"
+    return f"({label})"
+
+
+def _format_deprecated_suffix(deprecated: bool | str) -> str:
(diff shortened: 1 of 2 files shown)


src/click/core.py
```diff
@@ -93,6 +93,23 @@ def _check_nested_chain(
     raise RuntimeError(message)
 
 
+def _format_deprecated_label(deprecated: bool | str) -> str:
```
ic, line 96: For my understanding, why the `bool` argument type option in the two new helpers? Not used in the body.

ic (commented):
Thank you for sharing. Nice for consistent formatting. One question, perhaps leading to simpler type signature.

src/click/core.py
kdeldycke: > For my understanding, why the `bool` argument type option in the two new helpers? Not used in the body.

Users can have an option declared as `@my_option(deprecated=True)`, in which case the boolean will be passed to that function and produce `(DEPRECATED)` as a label. Without it we will produce the ugly `(DEPRECATED: )` string.

Issue #1550, opened 2020-05-09
Bug when setting options_metavar as empty in the Command class


yo-main:
Hi,

I have encountered an issue when using a Command object without any options.
As I wanted to prevent the '[OPTION]' from appearing in the Usage line, I tried to set the options_metavar parameter to None but got an exception.

I also tried to set it as an empty string rather than None, there's no exception then but the behavior isn't perfect neither as it prints a double space in the Usage row (the double space is between the filename and the argument)

Here's a code snippet to reproduce the issue with None

```python
@click.command("foo", options_metavar=None)
@click.argument("bar")
def cli():
    pass
```

### Expected Behavior
Running the above, I expected to see
```bash
$ python foobar.py --help
Usage: foobar.py BAR

Options:
  --help  Show this message and exit.
```

### Actual Behavior
An error is raised as the formatter was expecting a string instead of a None.

```pytb
Traceback (most recent call last):
  File "foobar.py", line 8, in <module>
    cli()
  File "/home/romain/projects/apps/myprss/.venv/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/romain/projects/apps/myprss/.venv/lib/python3.7/site-packages/click/core.py", line 781, in main
    with self.make_context(prog_name, args, **extra) as ctx:
  File "/home/romain/projects/apps/myprss/.venv/lib/python3.7/site-packages/click/core.py", line 700, in make_context
    self.parse_args(ctx, args)
  File "/home/romain/projects/apps/myprss/.venv/lib/python3.7/site-packages/click/core.py", line 1048, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "/home/romain/projects/apps/myprss/.venv/lib/python3.7/site-packages/click/core.py", line 1630, in handle_parse_result
    value = invoke_param_callback(self.callback, ctx, self, value)
  File "/home/romain/projects/apps/myprss/.venv/lib/python3.7/site-packages/click/core.py", line 123, in invoke_param_callback
    return callback(ctx, param, value)
  File "/home/romain/projects/apps/myprss/.venv/lib/python3.7/site-packages/click/core.py", line 950, in show_help
    echo(ctx.get_help(), color=ctx.color)
  File "/home/romain/projects/apps/myprss/.venv/lib/python3.7/site-packages/click/core.py", line 570, in get_help
    return self.command.get_help(self)
[...]

PR #3434, opened 2026-05-16
Fix `HelpFormatter.write_usage` producing spurious characters

kdeldycke:
This cleans-up the way the usage line is produce in the help by stripping all spurious trailing spaces produced by various parameters and indention setup. Also normalizes the second whitespace-only line that is used as separator for that section of the help screen.

Closes #3360.

The other related bug https://github.com/pallets/click/issues/1550 I mentioned in https://github.com/pallets/click/issues/3360#issuecomment-4287259287 has been fixed in an unrelated commit a while ago. But use the opportunity to cover that case in the new test suite.


kdeldycke:
This PR is ready to be reviewed and merged upstream.

PR #3450, opened 2026-05-18
`types`: fix variance issues and improve type coverage

jorenham:
The generic types in `click.types` were all parametrized using an invariant type parameter. This can easily lead to unexpected type errors for downstream users. For example, type-checkers would not allow you to assign `ParamType[bool]` to an `ParamType[int]`, even though `bool` is a subclass of `int`. This addresses these issues by using appropriate variance for these generic type parameters.

This also fixes several other typing issues, removing several `# type: ignore`s and `t,cast`s. 

I also took the parametrize some of the typed dicts, avoid several (intrinsically type-unsafe) `Any` annotations.

And just like https://github.com/pallets/click/pull/3422, this also fills in the missing annotations of attributes and constants. This consequently increases the type coverage by 2.18% (91.5% -> 93.68%).

I realize there are quite a lot of changes here. But I was afraid that spreading these out over multiple PRs would result in a merge conflict mess. But if you want me to split this up anyway, then that's also fine as far as I'm concerned.

### DIFF
--- a/src/click/types.py
+++ b/src/click/types.py
@@ -189,8 +189,18 @@ class CompositeParamType(ParamType[_ValueT_co]):
     def arity(self) -> int: ...  # type: ignore[override]
 
 
-class FuncParamTypeInfoDict(ParamTypeInfoDict, t.Generic[_ValueT_contra, _ValueT_co]):
-    func: t.Callable[[_ValueT_contra], _ValueT_co]
+if t.TYPE_CHECKING:
+    # on Python 3.10 this will raise a TypeError
+
+    class FuncParamTypeInfoDict(
+        ParamTypeInfoDict,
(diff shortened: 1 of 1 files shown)


Rowlando13:
I don't have enough knowledge of typing to review these changes. @kdeldycke

kdeldycke:
> I don't have enough knowledge of typing to review these changes. @kdeldycke

Same thing here, my brain breaks a bit with these kind of high-level typing. So no strong technical opinion on my side: the tests pass, the type checker is happy, the code does not introduce any malware, is purely type checking, and @jorenham already demonstrated reactivity and seriousness. So based on these circumstantial evidences, I would merge it and go ahead! Let's clear the backlog! 🤠

PR #3451, opened 2026-05-18
Add type annotations for instance attributes in `testing`

jorenham:
This is basically the story as #3422, and increases type coverage by 1.18& (91.5% -> 92.68%). 

I also took the libery of narrowing the contextmanager return types from `Iterator` to `Generator`, because (based)pyright was complaining that using `@contextmanager` with `-> Iterator` is deprecated:

image

### DIFF
--- a/src/click/testing.py
+++ b/src/click/testing.py
@@ -22,10 +22,15 @@ if t.TYPE_CHECKING:
 
     from .core import Command
 
-CaptureMode = t.Literal["sys", "fd"]
+CaptureMode: t.TypeAlias = t.Literal["sys", "fd"]
+ExceptionInfo: t.TypeAlias = tuple[type[BaseException], BaseException, TracebackType]
 
 
 class EchoingStdin:
+    _input: t.BinaryIO
+    _output: t.BinaryIO
+    _paused: bool
+
     def __init__(self, input: t.BinaryIO, output: t.BinaryIO) -> None:
         self._input = input
         self._output = output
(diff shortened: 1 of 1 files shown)


src/click/testing.py
```diff
@@ -22,10 +22,15 @@
 
     from .core import Command
 
-CaptureMode = t.Literal["sys", "fd"]
+CaptureMode: t.TypeAlias = t.Literal["sys", "fd"]
+_ExceptionInfo: t.TypeAlias = tuple[type[BaseException], BaseException, TracebackType]
```
AndreasBackx: Like other PR, I think no _ prefix is the standard.

AndreasBackx (approved):
Ctrl+c, ctrl+v: Just remove the version change and like other PR rebase on main instead perhaps?

jorenham:
> rebase on main instead perhaps?

@kdeldycke changed the target branch from main to stable, so I rebased onto that now. Should I switch back to main...?

src/click/testing.py
jorenham: I changed it, but keep in mind that without the `_` prefix, downstream users might start using this, thinking that it's public API. And while that's necessarily a problem, it might become one if you ever decide to remove it for some reason, as there's no mechanism (at all) to deprecate type aliases. So any change to it will then become a backwards incompatible breaking change. Precisely this is why, at NumPy, we're so incredibly careful with adding new type aliases to `numpy.typing`. Because once we do, we'd likely never be able to get rid of it, or at least, not without pissing off a whole bunch of users.

kdeldycke:
> > rebase on main instead perhaps?
> 
> @kdeldycke changed the target branch from main to stable, so I rebased onto that now. Should I switch back to main...?

No no: type changes are OK to target `stable` as long as they do not change the behavior, as per: https://github.com/pallets/click/pull/3463#issuecomment-4518579642

PR #3452, opened 2026-05-18
platform-dependent `testing.CaptureMode` type alias

jorenham:
`CliRunner.__init__` will raise a `ValueError` if `capture="fd"` on windows. By using the fact that type-checkers are able to understand basic `sys.platform` checks (https://typing.python.org/en/latest/spec/directives.html#version-and-platform-checking), we can use Python's type system to prevent this from happening, by having a definition for the `CaptureMode` type alias on windows, improving type-safety.

### DIFF
--- a/src/click/testing.py
+++ b/src/click/testing.py
@@ -22,7 +22,10 @@ if t.TYPE_CHECKING:
 
     from .core import Command
 
-CaptureMode: t.TypeAlias = t.Literal["sys", "fd"]
+if sys.platform == "win32":
+    CaptureMode: t.TypeAlias = t.Literal["sys"]  # pyright: ignore[reportRedeclaration]
+else:
+    CaptureMode: t.TypeAlias = t.Literal["sys", "fd"]  # pyright: ignore[reportRedeclaration]
 ExceptionInfo: t.TypeAlias = tuple[type[BaseException], BaseException, TracebackType]
 
 


src/click/testing.py
```diff
@@ -22,7 +22,11 @@
 
     from .core import Command
 
-CaptureMode = t.Literal["sys", "fd"]
+
+if sys.platform == "win32":
+    CaptureMode: t.TypeAlias = t.Literal["sys"]  # pyright: ignore[reportRedeclaration]
+else:
+    CaptureMode: t.TypeAlias = t.Literal["sys", "fd"]  # pyright: ignore[reportRedeclaration]
```
AndreasBackx, line 28: I wonder how you would make pyright happy here... 🤔️ Anyhow, this should still be fine I believe.

AndreasBackx (approved):
As other PRs: main instead of stable and no version change?

src/click/testing.py
jorenham: There's no redeclaration because these branches are mutually exclusive. So I'd say that this is false positive error, i.e. a bug in pyright, justifying this `# pyright: ignore`

src/click/testing.py
```diff
@@ -22,7 +22,11 @@
 
     from .core import Command
 
-CaptureMode = t.Literal["sys", "fd"]
+
+if sys.platform == "win32":
```
kdeldycke, line 25: Is this form work?
```python
from ._compat import WIN

if WIN:
    ...
```

jorenham: No unfortunately not.

kdeldycke: Thanks for the feedback! :)

PR #3453, opened 2026-05-18
Add type annotations for instance attributes in `exceptions`

jorenham:
Just like #3422 and #3451, this fills in the missing attribute type annotations for the error classes in `click.exceptions`. This increases type coverage by 0.89% (91.5% -> 92.39%).

### DIFF
--- a/src/click/exceptions.py
+++ b/src/click/exceptions.py
@@ -36,13 +36,16 @@ class ClickException(Exception):
     """An exception that Click can handle and show to the user."""
 
     #: The exit code for this exception.
-    exit_code = 1
+    exit_code: t.ClassVar[int] = 1
+
+    show_color: t.Final[bool | None]
+    message: t.Final[str]
 
     def __init__(self, message: str) -> None:
         super().__init__(message)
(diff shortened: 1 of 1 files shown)


src/click/exceptions.py
```diff
@@ -236,11 +250,15 @@ def __init__(
 
         super().__init__(message, ctx)
         self.option_name = option_name
-        self.possibilities: list[str] | None = None
+
+        # this temp variable avoids mypy complaining self.possibilities "reassignment"
```
AndreasBackx: Fine to remove this.

src/click/exceptions.py
```diff
@@ -266,11 +284,15 @@ def __init__(
 
         super().__init__(message, ctx)
         self.command_name = command_name
-        self.possibilities: list[str] | None = None
+
+        # this temp variable avoids mypy complaining self.possibilities "reassignment"
```
AndreasBackx: Same as other comment.

AndreasBackx (approved):
Same as other PRs, good to have. Just remove the version change and like other PR rebase on main instead perhaps?

PR #3454, opened 2026-05-18
Add type annotations for instance attributes in `formatting`

jorenham:
Here's another type-coverage improvement (+0.19%) like #3422, #3451, and #3453. 

I also noticed that some `Sequence` input annotations were only used in a for loop, so I took the liberty of widening those to `Iterable` (backwards compatible and type-safe); I hope you don't mind.

And as I explained in #3451, I also fixed the pyright deprecation warning for the `@contextmanager`s by changing their return types from `Iterator` to `Generator` (also backwards compatible and type-safe).

### DIFF
--- a/src/click/formatting.py
+++ b/src/click/formatting.py
@@ -119,6 +119,11 @@ class HelpFormatter:
                   width clamped to a maximum of 78.
     """
 
+    indent_increment: int
+    width: int
+    current_indent: int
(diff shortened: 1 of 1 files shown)


AndreasBackx (approved):
Just remove the version change and like other PR rebase on main instead perhaps?

PR #3455, opened 2026-05-18
Add type annotations for instance attributes in `shell_completion`

jorenham:
Here's another one like  #3422, #3451, #3453, and #3454 (+0.43% type coverage). 

This time, I somehow managed to keep myself from going out of scope, even though I saw a bunch of opportunities for improvement here in `click.shell_completion`. I'll address those in a follow-up (sorry @ reviewers, but it's for a good cause).

### DIFF
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -222,6 +222,11 @@ class ShellComplete:
     be provided by subclasses.
     """
 
+    cli: Command
+    ctx_args: cabc.MutableMapping[str, t.Any]
+    prog_name: str
+    complete_var: str
+
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


src/click/shell_completion.py
```diff
@@ -226,6 +225,11 @@ class ShellComplete:
     be provided by subclasses.
     """
 
+    cli: Command
+    ctx_args: cabc.MutableMapping[str, t.Any]
```
AndreasBackx, line 226: We should perhaps improve this later, but out of scope for this PR.

AndreasBackx (approved):
Just remove the version change and like other PR rebase on main instead perhaps?

PR #3460, opened 2026-05-18
Static typing improvements in `click.shell_completion`

jorenham:
As promised in #3455, this improves several (static) types in `click.shell_completion`:

- `shell_complete` narrowed return type
- `CompletionItem` now has an optional (defaulting to `Any` for backcompat) generic type parameter for its `.value`. I did not update the other modules to use this type parameter to avoid merge conflicts, but because of the `default=Any` there's no downside to this.
- `ShellComplete.source_vars` now returns a typed dict
- `add_completion_class` is now parametrized on the `cls` instance type instead of the type type, so that it's easier to see by looking at the annotations that `cls` is supposed to be a type rather than an instance. If I remember correctly, now all type checkers support using it as class decorator otherwise, but don't quote me on this.
- `get_completion_class` now has overloads for the builtin shell completion strings, so that `get_completion_class("zsh")` will now be inferred as `type[ZshComplete]` instead of `type[ShellComplete] | None` (the `| None` can be pretty annoying to deal with)

### DIFF
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -22,7 +22,7 @@ def shell_complete(
     prog_name: str,
     complete_var: str,
     instruction: str,
-) -> int:
+) -> t.Literal[0, 1]:
     """Perform shell completion for the given CLI program.
 
     :param cli: Command being called.
@@ -55,7 +55,16 @@ def shell_complete(
     return 1
 
 
-class CompletionItem:
+if t.TYPE_CHECKING:
+    from typing_extensions import TypeVar
+
+    # `Any` is used as default for backwards compatibility (instead of e.g. `str`)
+    _ValueT_co = TypeVar("_ValueT_co", covariant=True, default=t.Any)
+else:
+    _ValueT_co = t.TypeVar("_ValueT_co", covariant=True)
+
+
+class CompletionItem(t.Generic[_ValueT_co]):
     """Represents a completion value and metadata about the value. The
     default metadata is ``type`` to indicate special shell handling,
     and ``help`` if a shell supports showing a help string next to the
@@ -78,12 +87,12 @@ class CompletionItem:
 
     def __init__(
         self,
-        value: t.Any,
+        value: _ValueT_co,
         type: str = "plain",
         help: str | None = None,
         **kwargs: t.Any,
     ) -> None:
-        self.value: t.Any = value
+        self.value: _ValueT_co = value
         self.type: str = type
         self.help: str | None = help
         self._info = kwargs
@@ -198,6 +207,12 @@ complete --no-files --command %(prog_name)s --arguments \
 """
 
 
+class _SourceVarsDict(t.TypedDict):
+    complete_func: str
+    complete_var: str
+    prog_name: str
+
+
 class ShellComplete:
     """Base class for providing shell completion support. A subclass for
     a given shell will override attributes and methods to implement the
@@ -242,7 +257,7 @@ class ShellComplete:
         safe_name = re.sub(r"\W*", "", self.prog_name.replace("-", "_"), flags=re.ASCII)
         return f"_{safe_name}_completion"
 
-    def source_vars(self) -> dict[str, t.Any]:
+    def source_vars(self) -> _SourceVarsDict:
         """Vars for formatting :attr:`source_template`.
 
         By default this provides ``complete_func``, ``complete_var``,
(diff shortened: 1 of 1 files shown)


src/click/shell_completion.py
```diff
@@ -22,7 +22,7 @@ def shell_complete(
     prog_name: str,
     complete_var: str,
     instruction: str,
-) -> int:
+) -> t.Literal[0, 1]:
```
AndreasBackx, line 25: We should perhaps create a type for ExitCode, but at the same time, I'm not sure there's that much value. Kind of good to know the specific exit codes as well. 🤷️

src/click/shell_completion.py
```diff
@@ -43,19 +43,27 @@ def shell_complete(
...
 
     return 1
 
 
-class CompletionItem:
+if t.TYPE_CHECKING:
+    from typing_extensions import TypeVar
+
+    # `Any` is used as default for backwards compatibility (instead of e.g. `str`)
+    _ValueT_co = TypeVar("_ValueT_co", covariant=True, default=t.Any)
```
AndreasBackx, line 62: I don't think we need the underscore prefix, or at least I don't think we should as we don't do it elsewhere.

src/click/shell_completion.py
```diff
@@ -361,7 +377,7 @@ def get_completion_args(self) -> tuple[list[str], str]:
 
         return args, incomplete
 
-    def format_completion(self, item: CompletionItem) -> str:
+    def format_completion(self, item: CompletionItem[t.Any]) -> str:
```
AndreasBackx, line 377: Any reason it's Any here instead of str?

CHANGES.rst
```diff
@@ -1,23 +1,10 @@
 .. currentmodule:: click
 
-Version 8.4.2
--------------
-
-Unreleased
-
-
 Version 8.4.1
```
AndreasBackx: Let's not include version changes I think?

AndreasBackx (changes requested):
Just the underscore, the reason for the Any, and I think this should be rebased on main, not stable? Otherwise looks like fine typing improvements. Someone else can confirm the stable vs main thing.

Rowlando13:
If the types tighten at all, then I think it needs to be put on main, but I don't know types too well. If I were running runtime type checking, I would not expect types to change on a bug fix release. @davidism @kdeldycke can you both weigh in so we have a clear answer.

Rowlando13:
@jorenham sorry for the back and forth on branches. We have not accepted many types prs from outside contributors, so still figuring it out.

kdeldycke:
> If the types tighten at all, then I think it needs to be put on main, but I don't know types too well. If I were running runtime type checking, I would not expect types to change on a bug fix release. @davidism @kdeldycke can you both weigh in so we have a clear answer.

As I said earlier at https://github.com/pallets/click/pull/3463#issuecomment-4518579642:

> If typing improvements do not change the API or behavior, then I would consider them as I consider improved tests and bug fixes: OK for bug fix releases.

So to me it is OK to target `stable` and so the next 8.4.2.

davidism:
I'm fine with typing changes in fix releases.

Rowlando13:
Okay fix releases it is.

CHANGES.rst
jorenham: I didn't touch this; this diff is only here because @kdeldycke switched branches

src/click/shell_completion.py
jorenham: `str` would indeed also work, but in principle, this would work for any type of `item.value`, because it's only used in an f-strings here, where even the broadest type would work: `f"{object()}"`

src/click/shell_completion.py
jorenham: The convention is to use private naming (`_` prefix) for `TypeVar`, to highlight that they're not part of the public API. One reason for this is that, after Python <3.12 support is dropped, these `TypeVar`s can be removed in favor of the [PEP 695](https://peps.python.org/pep-0695/) generics syntax.

src/click/shell_completion.py
jorenham: Are there any other functions that return a return code like this, or would it be limited to this module? I had a quick look just now but couldn't find any others.

Rowlando13:
@AndreasBackx did @jorenham address your comments?

AndreasBackx:
> @AndreasBackx did @jorenham address your comments?

Should be good to go, satisfied with the answers as it was primarily nitpicks. Apologies for the delay.

kdeldycke:
> > @AndreasBackx did @jorenham address your comments?
> 
> Should be good to go, satisfied with the answers as it was primarily nitpicks. Apologies for the delay.

Thanks for the feedback! Let's merge then! :)

PR #3468, opened 2026-05-18
Add all dev groups to default groups

pamelafox:
Contributors currently have to run "uv sync --all-groups" to get all the groups required for all the dev tools. Instead of requiring that, let's add all those groups to uv default groups.
The drawback is that the install could take much longer to bring down large packages, but it doesn't seem like any of the extra groups are too heavy.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,7 +81,7 @@ exclude = [
 ]
 
 [tool.uv]
-default-groups = ["dev", "pre-commit", "tests", "typing"]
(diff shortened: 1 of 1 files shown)


pamelafox:
Nevermind, I chatted with @davidism about this. These packages are all used via `tox` commands, and those commands explicitly bring in their required dependency groups. We only put typing/tests in the default because people commonly run "pytest" and "mypy" on their own.

Closing the PR.

PR #3470, opened 2026-05-18
fix: Skip flaky pager test on macOS with free-threaded Python 3.14t

### LINKED ISSUE #2899: running tests in parallel with `tox p` fails

Jazzcort:
Temporarily fixes flaky test failure for `test_echo_via_pager` on macOS with 3.14t (free-threading) by skipping them for macOS + Python 3.14t combination until we figure out how to make the test result consistently pass when `test_echo_via_pager` and `test_stream_lifecycle` are running in parallel. 

fixes #2899

### DIFF
--- a/src/click/_compat.py
+++ b/src/click/_compat.py
@@ -12,6 +12,7 @@ from weakref import WeakKeyDictionary
 
 CYGWIN = sys.platform.startswith("cygwin")
 WIN = sys.platform.startswith("win")
+MAC = sys.platform == "darwin"
 auto_wrap_for_ansi: t.Callable[[t.TextIO], t.TextIO] | None = None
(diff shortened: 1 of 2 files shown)


Jazzcort:
@davidism @Rowlando13 
Thank you so much for the guidance! 😁

davidism:
Oops, got this backwards, should be `not sys._is_gil_enabled()`. I'll fix in a second.

davidism:
6cb3477

Jazzcort:
@davidism My bad! Should have caught it earlier. 😅

Issue #3481, opened 2026-05-19
Remove Python 2 utilities and little used utilities


Rowlando13:
Python2-centric utilities: 

- `get_binary_stream`
- `get_text_stream`

Little used Utilities: 

- `wrap_text`
- `getchar`

In the interest of making more time to focus on core click details, these utilities are planned to be deprecated in the next feature release, and removed from the public api in one after. If any of these are critical to your uses, please explain below, and we may consider other options.

Rowlando13:
@kdeldycke @davidism Any others to consider?

kdeldycke:
`get_binary_stream` and `get_text_stream` are genuinely Python 2/3-era shims for the `sys.stdin` vs `sys.stdin.buffer` split, so fair game to deprecate them. Can help identified other Python 2 left-over as suspected by @davidism at https://github.com/pallets/click/pull/3482#discussion_r3284863342

`wrap_text` is used internally a lot for `HelpFormatter`. Also used by @janluke in https://github.com/janluke/cloup/blob/d53d04f51846e3dbfe6a1b97ad677bfc5cdda47f/cloup/formatting/_formatter.py#L239-L240. What do you want to replace it with?

I have doubt about `getchar`. Is there a multi-platform equivalent in the `stdlib`?

davidism:
Not everything in utilities is compat stuff for Python 2. Pretty sure we use `getchar` because there's no built-in way to get a single character input (without pressing <kbd>Enter</kbd>) cross-platform.

Rowlando13:
Updated the issue. Plan to do a GitHub search for how much things are used. I got these from talking with people at Pycon, so just soliciting input. I am mainly concerned with reducing the public api. People are free to vendor the code, I just want to be sure it is really used if we keep it around. A fair amount of utilities have little to no test coverage, so there is maintenance work to be done if we want to keep them around.

davidism:
[...]

PR #3482, opened 2026-05-19
fix I/O operation on closed file error with CliRunner and echo_via_pager

### LINKED ISSUE #3449: "I/O operation on closed file" error with CliRunner and echo_via_pager under 8.4

Reproducer:
```python

@command()
def cli():

runner = CliRunner()
runner.invoke(cli)
```

Traceback:
```
    runner.invoke(cli)
    ~~~~~~~~~~~~~^^^^^
    sys.stdout.flush()
    ~~~~~~~~~~~~~~~~^^
```

Environment:

- Python version: 3.13.5
- Click version: 8.4

ChrisPappalardo:
Fixes #3449 by adding a thinner version of `MaybeStripAnsi` that is not subclassed from `io.TextIOWrapper` when handling `sys.stdout` and `sys.stderr` streams.

### DIFF
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -606,15 +606,35 @@ def _tempfilepager(
         os.unlink(f.name)
 
 
+class _SkipClose:
+    def __init__(self, stream: t.IO[t.Any]) -> None:
+        self.stream = stream
+
+    def __getattr__(self, name: str) -> t.Any:
+        return getattr(self.stream, name)
+
+    @property
+    def buffer(self) -> t.BinaryIO:
+        return _SkipClose(self.stream.buffer)  # type: ignore[attr-defined, return-value]
+
+    def close(self) -> None:
+        pass
+
+
 @contextlib.contextmanager
 def _nullpager(
     stream: t.TextIO, color: bool | None = None
 ) -> t.Iterator[tuple[t.TextIO, str, bool]]:
-    """Simply print unformatted text.  This is the ultimate fallback."""
+    """Simply print unformatted text. This is the ultimate fallback. Don't close the
+    output stream in this case, since it's coming from elsewhere rather than our
+    internal helpers.
+    """
     encoding = get_best_encoding(stream)
+
     if color is None:
         color = False
-    yield stream, encoding, color
+
+    yield _SkipClose(stream), encoding, color  # type: ignore[misc]
 
 
 class Editor:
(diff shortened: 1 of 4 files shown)


davidism:
Seems like we should be able to keep the single wrapper and override the close method to do nothing in the problematic case.

src/click/_termui_impl.py
```diff
@@ -436,10 +454,14 @@ def get_pager_file(color: bool | None = None) -> t.Generator[t.TextIO, None, Non
                   default is autodetection.
     """
     with _pager_contextmanager(color=color) as (stream, encoding, color):
-        # Split streams by capabilities rather than the abstract TextIO /
-        # BinaryIO annotations: buffered text streams can be unwrapped to bytes,
-        # while text-only streams are yielded as-is.
-        if _has_binary_buffer(stream):
+        # route stdout and stderr through a thinner wrapper
+        is_std = stream is sys.stdout or stream is sys.stderr
```
davidism: @Rowlando13 this is a good example of why I think the whole text/binary stream handling in Click is not needed. It's completely bypassed for stdin/stdout here, and all tests still pass. Fairly sure these checks were for Python 2/3 compat, but I didn't have enough time to check it all when I removed most of the compat years ago.

davidism:
After investigating from my initial comment, it does look like we need to wrap the stream returned by `nullpager` instead of this. The wrapper should pass everything through, except override close to do nothing. Will work on this tonight or tomorrow.

davidism:
Also needed to make sure `stream.buffer` wouldn't close either, since `get_pager_file` would wrap `stream.buffer` instead of `stream`. This all seems a bit messy, probably something that can be cleaned up eventually. @Rowlando13 @kdeldycke.

kdeldycke:
> Also needed to make sure `stream.buffer` wouldn't close either, since `get_pager_file` would wrap `stream.buffer` instead of `stream`. This all seems a bit messy, probably something that can be cleaned up eventually. @Rowlando13 @kdeldycke.

Also CC @AndreasBackx for the missing tests in https://github.com/pallets/click/pull/3405 covering https://github.com/pallets/click/pull/1572

src/click/_termui_impl.py
kdeldycke: This is being discussed at https://github.com/pallets/click/issues/3481

kdeldycke:
This fix was not complete. I proposed one at https://github.com/pallets/click/pull/3533 . @ChrisPappalardo can you review and test it?

PR #3484, opened 2026-05-20
Fix `get_parameter_source()` during type conversion and eager callbacks

### LINKED ISSUE #3458: `get_parameter_source()` returns `None` in 8.4.0
Starting with 8.4, get_parameter_source() returns None.

Repro:

```python
import click

class Source(click.ParamType):
    name = "source"
        return {'value': value, 'source': ctx.get_parameter_source(param.name)}

@click.command
@click.option('--default', type=Source(), default='/tmp/file')
@click.option('--nodefault', type=Source())
def main(default, nodefault):
    print("default:", default)
    print("nodefault:", nodefault)

if __name__ == '__main__':
    main()
```

Output:

```console
$ pip install click==8.4.0 -q
$ python source.py                  
default: {'value': '/tmp/file', 'source': None}
nodefault: None
$ python source.py --default cli --nodefault cli
default: {'value': 'cli', 'source': None}
nodefault: {'value': 'cli', 'source': None}

$ pip install click==8.3.3 -q                   
$ python source.py                              
default: {'value': '/tmp/file', 'source': <ParameterSource.DEFAULT: 5>}
nodefault: None
$ python source.py --default cli --nodefault cli
default: {'value': 'cli', 'source': <ParameterSource.COMMANDLINE: 2>}
nodefault: {'value': 'cli', 'source': <ParameterSource.COMMANDLINE: 2>}
```

Environment:

- Python version: 3.14.4 (seems to happen on all versions, on both macOS and Linux)
- Click version: 8.4.0

yurishevtsov:
### Summary 

Click 8.4.0 moved `set_parameter_source()` to after `process_value()` when fixing flag-group arbitration ([#3403](https://github.com/pallets/click/issues/3403), [#3404](https://github.com/pallets/click/pull/3404)). That broke a case wasn't meant to change: anything that calls `get_parameter_source()` during type conversion or in an eager callback saw `None`, even though `consume_value()` had already figured out where the value came from ([#3458](https://github.com/pallets/click/issues/3458)).

This is affecting Flask: with Click 8.4.0, `FLASK_DEBUG=1` in the environment was ignored because `_set_debug` runs as an eager callback and could no longer see that the `--debug` flag was only at its default ([pallets/flask#6025](https://github.com/pallets/flask/issues/6025)).

The fix is to record the source right after `consume_value()` and before `process_value()`, matching 8.3.3 timing for callbacks and custom types. Flag-group arbitration is the same as today. 

There is a second, narrower issue on the same code path: recording the source early means a *losing* option in a feature-switch group briefly overwrites `_parameter_source` before arbitration runs.  I am solving it by restoring the previous source when an option loses the slot. That path is easy to miss in tests (shared `opts` may give the loser the same source as the winner), so there is a dedicated case in `test_bool_flag_group_parameter_source`. More detail is in the comments around `handle_parse_result`.

fixes [#3458](https://github.com/pallets/click/issues/3458)

### Tests added / updated

- `tests/test_defaults.py` — issue repro (`ParamType.convert`), eager callback, Flask-style `_set_debug` guard
- `tests/test_options.py` — `test_bool_flag_group_parameter_source` (includes a loser-restore case with `envvar`, not only CLI + `default_map`)

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2609,6 +2609,11 @@ class Parameter(ABC):
         with augment_usage_errors(ctx, param=self):
             value, source = self.consume_value(ctx, opts)
 
+            # Record the source before processing so eager callbacks and type
+            # conversion can inspect it. Restored after arbitration if this
+            # option loses a feature-switch group.
+            ctx.set_parameter_source(self.name, source)
+
             # Display a deprecation warning if necessary.
             if (
                 self.deprecated
@@ -2654,14 +2659,13 @@ class Parameter(ABC):
         )
 
         if is_winner:
-            ctx.set_parameter_source(self.name, source)
             if self.expose_value:
                 ctx.params[self.name] = value
                 ctx._param_default_explicit[self.name] = self._default_explicit
-        elif existing_source is None:
-            # Nothing has claimed the slot yet. Record at least our source so downstream
-            # lookups don't return ``None``.
-            ctx.set_parameter_source(self.name, source)
+        elif existing_source is not None:
+            # Lost arbitration; restore the winning option's source.
+            ctx.set_parameter_source(self.name, existing_source)
+        # else: keep the provisional source recorded before process_value.
 
         return value, args
 
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2665,7 +2665,10 @@ class Parameter(ABC):
         elif existing_source is not None:
             # Lost arbitration; restore the winning option's source.
             ctx.set_parameter_source(self.name, existing_source)
-        # else: keep the provisional source recorded before process_value.
+        # else: ctx.params[self.name] was populated by code that bypassed
+        # handle_parse_result (from another option's callback for example). Keep
+        # the provisional source recorded before process_value so downstream
+        # lookups don't return ``None``.
 
         return value, args
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Version 8.4.1
 
 Unreleased
 
+-   ``get_parameter_source()`` is available during eager callbacks and type
+    conversion again. :issue:`3458`
 -   Zsh completion scripts parse correctly on Windows. :issue:`3277`
 -   Shell completion of `Choice` `Enum` values produces a valid completion
     result. :issue:`3015`
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,11 +6,10 @@ Version 8.4.1
 Unreleased
 
 -   ``get_parameter_source()`` is available during eager callbacks and type
-    conversion again. :issue:`3458`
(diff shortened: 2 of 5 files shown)


src/click/core.py
```diff
@@ -2609,6 +2609,11 @@ def handle_parse_result(
         with augment_usage_errors(ctx, param=self):
             value, source = self.consume_value(ctx, opts)
 
+            # Record the source before processing so eager callbacks and type
+            # conversion can inspect it. Restored after arbitration if this
+            # option loses a feature-switch group.
+            ctx.set_parameter_source(self.name, source)
```
yurishevtsov, line 2615: Restores 8.3.3 timing so `get_parameter_source()` works inside convert / eager callbacks (#3458). 

#3404 moved this to the end of `handle_parse_result`, which caused the regression.

src/click/core.py
```diff
@@ -2654,14 +2659,13 @@ def handle_parse_result(
...
             if self.expose_value:
                 ctx.params[self.name] = value
                 ctx._param_default_explicit[self.name] = self._default_explicit
-        elif existing_source is None:
-            # Nothing has claimed the slot yet. Record at least our source so downstream
-            # lookups don't return ``None``.
-            ctx.set_parameter_source(self.name, source)
+        elif existing_source is not None:
+            # Lost arbitration; restore the winning option's source.
+            ctx.set_parameter_source(self.name, existing_source)
```
yurishevtsov, line 2667: This option already wrote its provisional source above; if it loses arbitration, put the winner’s source back so `_parameter_source` matches `ctx.param`.

src/click/core.py
```diff
@@ -2654,14 +2659,13 @@ def handle_parse_result(
         )
 
         if is_winner:
-            ctx.set_parameter_source(self.name, source)
```
yurishevtsov, line 2657: Winner already has the right source from the early write. Only params need updating here.

kdeldycke (approved):
This fix is more generic than my own fix in Click Extra at: https://github.com/kdeldycke/click-extra/commit/9fe4e8bc397104b27300ebf66711b3a29434735b

Tests are good to me and checks all variations. One thing I realized: in https://click.palletsprojects.com/en/stable/advanced/#parameter-modifications , we document the overriding of one parameter from the callback of another. I didn't know how the `get_parameter_source` would react in that case so I added a unittest to lockdown the current behavior. Also used the opportunity to add an admonition in the docs to highlight the danger of that kind of bypass.

Also added a reference to the PR in the changelog.

kdeldycke:
An alternative, long term solution would be to restructure the pipeline and the parameter loop that is happening above. But that is a bigger refactor for later. So this PR is a good local minimal patch for 8.4.1.

kdeldycke:
Thanks @davidism for the merge! :)

PR #3493, opened 2026-05-21
Handle empty bytes in echo

### LINKED ISSUE #3487: Echoing empty bytes or bytearray raises TypeError
Hopefully this isn't too frivolous but I think it's an easy fix: echoing an empty byte string will cause a crash if `nl=True`.

To replicate:

```shell
$ python -c "import io; import click; click.echo(b'', io.BytesIO(), nl=True)"

  File "/myproject/click/src/click/utils.py", line 337, in echo
    ^^^^^^^^^^^^^^^
TypeError: a bytes-like object is required, not 'str'
```

Expected behaviour:

We expect `b"\n"` to be written to the stream and for `TypeError` not to be thrown.

Environment:

- Python version: 3.11.2
- Click version: 8.5.0.dev0

Suggested fix:

Change https://github.com/pallets/click/blob/dbf4d14c712750c31e15916b659925683150533b/src/click/utils.py#L297

```diff
    if nl:
-      out = out or ""
+      out = "" if out is None else out 
       if isinstance(out, str):
            out += "\n"
        else:
            out += b"\n"
```

Iain-S:
Previously, an empty byte string could cause an exception if `nl=True`.

This PR narrows the type scope of `out`, to `bytes() | bytearray() | str` and handles all cases exhaustively.

We also change from `typing.Any`, which effectively disables typing for a variable, to `object`, which is more restrictive. See  https://mypy.readthedocs.io/en/latest/dynamic_typing.html#any-vs-object for more.

fixes #3487

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ Unreleased
 -   Zsh completion scripts parse correctly on Windows. :issue:`3277`
 -   Shell completion of `Choice` `Enum` values produces a valid completion
     result. :issue:`3015`
+-   Fix empty byte-string handling in echo. :issue:`3487`
+
 
 Version 8.4.0
 -------------
--- a/src/click/utils.py
+++ b/src/click/utils.py
@@ -232,7 +232,7 @@ class KeepOpenFile:
 
 
 def echo(
-    message: t.Any | None = None,
+    message: object = None,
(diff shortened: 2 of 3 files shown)


davidism:
Don't merge yet, need to fix my commit signing on this new laptop.

PR #3494, opened 2026-05-21
AI junk

### LINKED ISSUE #3449: "I/O operation on closed file" error with CliRunner and echo_via_pager under 8.4

jesustorres-code:
Fixes #3449

## Root cause

`MaybeStripAnsi` subclasses `io.TextIOWrapper` and wraps `sys.stdout.buffer`. `io.TextIOWrapper` takes ownership of its underlying buffer and closes it when the wrapper is closed or garbage collected.

After `echo_via_pager` returns and the `with get_pager_file()` context exits, the `MaybeStripAnsi` instance goes out of scope. When Python GC collects it, `TextIOWrapper.__del__` closes the buffer — which effectively closes `sys.stdout`. Then `CliRunner.invoke`'s `finally` block calls `sys.stdout.flush()` and gets `ValueError: I/O operation on closed file`.

## Fix

After `stream.flush()` in `get_pager_file`'s `finally` block, call `stream.detach()` when the stream is a `MaybeStripAnsi`. `detach()` dissociates the `TextIOWrapper` from the buffer without closing it, so `sys.stdout.buffer` (and `sys.stdout`) remain open.

```python
finally:
    stream.flush()
    if isinstance(stream, MaybeStripAnsi):
        stream.detach()
```

## Test

Added `test_echo_via_pager_in_cli_runner` to `tests/test_testing.py` — a minimal reproduction of the reported bug that previously crashed with `ValueError`.

### DIFF
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -449,6 +449,12 @@ def get_pager_file(color: bool | None = None) -> t.Generator[t.TextIO, None, Non
             yield stream
         finally:
             stream.flush()
+            # MaybeStripAnsi wraps an existing buffer (e.g. sys.stdout.buffer).
+            # io.TextIOWrapper takes ownership of the buffer and closes it on
+            # GC, which would silently close sys.stdout. Detach first so the
+            # buffer is not closed when this wrapper is collected.
+            if isinstance(stream, MaybeStripAnsi):
+                stream.detach()
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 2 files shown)


davidism:
AI junk, ignored existing open PR

PR #3499, opened 2026-05-22
Fix pager test flackiness and add stress tests for `echo_via_pager`

### LINKED ISSUE #2899: running tests in parallel with `tox p` fails

kdeldycke:
This fix the root cause which was worked around in #3470, and builds on top of #3139.

The thing is: `_test_gen_func_fails` yields `test` then raises. Click cannot guarantee that the bytes pushed in the pipe to the pager is delivered before the subprocess dies, as this is under the system authority. So that's why the issue was only uncovered in Python `3.14t`, where the scheduling management is not deterministic.

We loose the "yield-then-raise" case, but that case was untestable by file content anyway.

To prevent this issue to re-occur, and to catch it early next time, I made the stress tests a bit more generic to collect all `stress` Pytest markers. And added a new stress test to detect `echo_via_pager` leaks.

Together with the b1bc8c9c / #3238 that was already pushed, this PR fully resolves #2899.

### DIFF
--- a/src/click/_compat.py
+++ b/src/click/_compat.py
@@ -12,7 +12,6 @@ from weakref import WeakKeyDictionary
 
 CYGWIN = sys.platform.startswith("cygwin")
 WIN = sys.platform.startswith("win")
-MAC = sys.platform == "darwin"
 auto_wrap_for_ansi: t.Callable[[t.TextIO], t.TextIO] | None = None
 _ansi_re = re.compile(r"\033\[[;?0-9]*[a-zA-Z]")
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -185,12 +185,11 @@ commands = [[
 ]]
 
 [tool.tox.env.stress]
-description = "stress tests for stream lifecycle race conditions"
+description = "high-iteration stress tests for race conditions"
 commands = [[
     "pytest", "-v", "--tb=short", "-x", "-m", "stress",
     "--basetemp={env_tmp_dir}",
     "--override-ini=addopts=",
-    "tests/test_stream_lifecycle.py",
     {replace = "posargs", default = [], extend = true},
 ]]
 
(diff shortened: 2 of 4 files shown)


kdeldycke:
> We loose the "yield-then-raise" case, but that case was untestable by file content anyway.

Just found a way to recover that original test intention in: https://github.com/pallets/click/pull/3499/changes/d4ed88f875107317c05a3586c319d464d682753e

kdeldycke:
This PR is ready to be reviewed and following my comment at https://github.com/pallets/click/issues/2899#issuecomment-4518428740 , will close https://github.com/pallets/click/issues/2899

src/click/_compat.py
```diff
@@ -12,7 +12,6 @@
 
 CYGWIN = sys.platform.startswith("cygwin")
 WIN = sys.platform.startswith("win")
-MAC = sys.platform == "darwin"
```
Rowlando13, line 15: Why did you remove this? Could be right, just down know much about _compat.py

Rowlando13:
@kdeldycke No changelog for test suite changes.

src/click/_compat.py
kdeldycke: > Why did you remove this? Could be right, just down know much about _compat.py

Because this is a 1:1 reversal of #3470 and not used anywhere else. So better to remove this unused one before someone outside rely on it! 😅

kdeldycke:
> @kdeldycke No changelog for test suite changes.

Indeed! Thanks for the reminder! :)

kdeldycke:
Fixed. This PR is ready to be merged.

PR #3500, opened 2026-05-22
Activate nightly workflow for random and stress tests

kdeldycke:
In https://github.com/pallets/click/pull/3139 we added stress tests and in https://github.com/pallets/click/pull/3151 we added randomization and parallel tests.

We decided to not activate them by default in https://github.com/pallets/click/pull/3151#issuecomment-3980528692 and rely on contributors diligence to run them by hand to catch regressions. In practice nobody does. I even often forget to run them on my own during my sessions, only focused on the task at hand.

I'd like to re-assess this decision in light of regressions that could have been caught if these test were active. See the following PRs:
- #3482
  Regression introduced in 8.4.0, could have been caught by `tests/test_stream_lifecycle.py`.
- #3499
  New stress test covering the root cause of #3470.
- #3470
  The bug exists because CI doesn't run tests in parallel.
- #2991
  The reason why stress test exist: *I originally only hit the race about once in 10k tries, so this seems difficult to test systematically.*.
- #3139
  Fix the regression from #2991, where only with the introduction of stress test keeps from recurring.
- #3238
  The random test already caught a leak nobody noticed via the default suite.

Also why even support parallel tests if we don't exercice them regularly? (ref: https://github.com/pallets/click/issues/2899)

Currently these tests adds ~15 minutes run-time. If you find this too long and invasive, I propose a compromise and run them nightly. But my preference would be to run them as part of the standard CI to incentivize contributors. And reduce surprises and and follow-up fixes.

Given the reversal in policy this change imply, I'm aiming for the next 8.5.0 minor release. I'm opening this PR as a way to discuss the best strategy forward.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,0 +1,42 @@
+name: Nightly race tests
+on:
+  schedule:
+    - cron: '0 6 * * *'
+  workflow_dispatch:
+permissions: {}
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+jobs:
+  race-tests:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os || 'ubuntu-latest' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - {name: Random 3.14, python: '3.14', tox: random}
+          - {name: Random 3.14t, python: '3.14t', tox: random}
+          - {name: Random 3.13, python: '3.13', tox: random}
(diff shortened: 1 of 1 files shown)


Rowlando13:
@kdeldycke I agree with the nightly run.

kdeldycke:
I think we were overly cautious just running these tests nightly as in the end they only take 5 minutes, not 15: https://github.com/pallets/click/actions/runs/26496766042/usage . To me the cost-benefit is excellent to trigger on every PR: a 5 minute wait is nothing in comparison to the effort and synchronization time required to fix leaky issues from non-core maintainers. But I will let that decision for later.

PR #3507, opened 2026-05-23
Mark commands as optional in the synopsis

### LINKED ISSUE #3059: With invoke_without_command set commands are not marked as optional

kdeldycke:
This PR show optional subcommands in the help screen as `[COMMAND]` instead of `COMMAND`.

The [bracket syntax is a convention described in POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) for optional argument and operands. There is no subcommand concept in POSIX, so a subcommand in the context of Click can be considered as an operand as it is positioned after the CLI invokation.

Closes #3059.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,9 @@ Unreleased
     :issue:`3043` :pr:`3504` :pr:`3508`
 -   Deprecated commands and options with empty or missing help text no longer
     render a stray leading space before the ``(DEPRECATED)`` label. :pr:`3509`
+-   A :class:`Group` with ``invoke_without_command=True`` marks its subcommand as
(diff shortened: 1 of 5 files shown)


davidism (approved):

kdeldycke:
This PR is ready to be merged.

PR #3510, opened 2026-05-23
docs: fix flag-pair separator and small typos in options guide

hernanr2205:
Small documentation fixes in `docs/options.md`. In the environment-variables caution note, a flag pair was written with a backslash separator instead of the forward slash that Click actually uses; it now reads `--flag/--no-flag`, matching the rest of the page (for example `--shout/--no-shout`). The same sentence now reads "with an `envvar`" instead of "with a `envvar`". Earlier in the page, "forward slash(`/`)" gains a missing space ("forward slash (`/`)"). These are wording and typo fixes only, with no behavior or content changes. Per the PR template, an issue is not required for documentation typo fixes.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -352,7 +352,7 @@ Use cases for this more explicit pattern include:
 * Shell aliases can set a flag, then an invocation can add a negation of the
   flag
 
-If a forward slash(`/`) is contained in your option name already, you can split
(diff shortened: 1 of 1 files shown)


kdeldycke:
Good nitpicking PR, let's merge! :)

PR #3511, opened 2026-05-24
AI junk

### LINKED ISSUE #3501: `isolated_filesystem()` is not thread safe

Aniketiitk21:
## Summary

Fix `CliRunner.isolated_filesystem()` so concurrent threads cannot race while swapping the process working directory.

The root cause is that `isolated_filesystem()` changes the global process `cwd` with `os.chdir()` and restores it later, but it does not synchronize that global mutation. Two threads entering the context at the same time can therefore stomp each other's working directory.

This change serializes the temporary `cwd` swap with a re-entrant lock and adds a regression test that verifies a second runner cannot enter its isolated filesystem until the first one exits.

Fixes #3501.

## Test plan

- `python -m pytest tests/test_testing.py -k isolated -q`
- `python -m pytest tests/test_testing.py -q`
- `python -m pytest tests/test_arguments.py::test_file_args tests/test_arguments.py::test_file_atomics tests/test_basic.py::test_file_option tests/test_shell_completion.py::test_files_closed tests/test_utils.py::test_open_file tests/test_utils.py::test_open_file_respects_ignore -q`

### DIFF
--- a/src/click/testing.py
+++ b/src/click/testing.py
@@ -8,6 +8,7 @@ import pdb
 import shlex
 import sys
 import tempfile
+import threading
 import typing as t
 from types import TracebackType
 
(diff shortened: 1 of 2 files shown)


ThiefMaster:
Do not send us LLM-generated PRs.

PR #3525, opened 2026-05-28
AI spam

### LINKED ISSUE #2634: `tests/test_types.py::test_file_surrogates` test is missing the `skipif not _non_utf8_filenames_supported` marker
The title says it all. There are a bunch of tests in [`tests/test_types.py`](https://github.com/pallets/click/blob/ca5e1c3d75e95cbc70fa6ed51ef263592e9ac0d0/tests/test_types.py#L147) that do some checks with non-utf8 filenames. Some filesystems don't support non-utf8 filenames, so the main `test_path_surrogates` test is conditionally disabled with

```python
@pytest.mark.skipif(
    not _non_utf8_filenames_supported(),
    reason="The current OS or FS doesn't support non-UTF-8 filenames.",
)
```

However, the next two tests (`test_file_surrogates` and `test_file_error_surrogates`) are missing these marks.

The fix should be pretty simple - add the above-mentioned decorator to these tests as well.

MohammadYusif:
Added `@pytest.mark.skipif` guards to `test_file_surrogates` and
`test_file_error_surrogates` in `tests/test_types.py`. Both tests use
surrogate escape sequences (`\udcff`) in filenames, which require the
OS/filesystem to support non-UTF-8 filenames. The sibling test
`test_path_surrogates` already had this guard via
`_non_utf8_filenames_supported()`, but the other two were missing it,
causing spurious test failures on macOS HFS+, some FUSE mounts, and
other filesystems that reject non-UTF-8 filenames.

fixes #2634

- Tests pass on supporting filesystems and skip gracefully on
  filesystems that reject non-UTF-8 filenames — no logic change,
  only the missing skip decorators added.
- No public API changed; no `.. versionchanged::` entries needed.
- Added entry to `CHANGES.rst` under Version 8.5.0.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,9 @@ Unreleased
     Colorama is no longer a dependency and is not used. :issue:`2986` :pr:`3505`
 -   :class:`Argument` accepts a ``help`` parameter, and help output includes
     a ``Positional arguments`` section when argument help is available. :issue:`2983` :pr:`3473`
+-   ``test_file_surrogates`` and ``test_file_error_surrogates`` now skip on
+    filesystems that do not support non-UTF-8 filenames, matching the existing
+    guard on ``test_path_surrogates``. :issue:`2634`
 
 
 Version 8.4.2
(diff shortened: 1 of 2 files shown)


kdeldycke:
This is a naive fix. Probably AI slop.

MohammadYusif:
Utility fix = ai slop, fair point

PR #3526, opened 2026-05-28
Accept ZFS error-message for non-UTF-8 names

### LINKED ISSUE #2634: `tests/test_types.py::test_file_surrogates` test is missing the `skipif not _non_utf8_filenames_supported` marker

kdeldycke:
This PR update the list of file-system specific error messages for a test that is supposed to fail for non-UTF8 filesystem.

Closes: #2634.

Relates to: 
- #2945
- #2946

CC @ruro

### DIFF
--- a/tests/test_types.py
+++ b/tests/test_types.py
@@ -243,13 +243,22 @@ def test_path_surrogates(tmp_path, monkeypatch):
     ],
 )
 def test_file_surrogates(type, tmp_path):
-    path = tmp_path / "\udcff"
+    """Ensures that the error handling in ``click.File`` is robust.
 
-    # - common case: �': No such file or directory
-    # - special case: Illegal byte sequence
-    # The special case is seen with rootless Podman. The root cause is most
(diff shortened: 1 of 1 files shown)


ruro:
Seems reasonable to me :+1:

PR #3530, opened 2026-05-29
Fix thread safety in CliRunner.isolated_filesystem

### LINKED ISSUE #3501: `isolated_filesystem()` is not thread safe

tyler-talbert:
fixes #3501 

isolated_filesystem calls os.chdir on the process working directory. Since CWD is per-process, two threads entering the context manager at the same time can end up in each other's temp directory. Added a module-level lock around the chdir and teardown so only one thread is inside at a time.

Tradeoff: tests that use isolated_filesystem across threads now run serially through this lock. Picked it over passing the dir through explicitly, which would break the existing API.

Added a test that uses a Barrier to force both threads into the race window. Full suite still passes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,8 @@ Unreleased
 -   A :class:`Group` with ``invoke_without_command=True`` marks its subcommand as
     optional in the usage help, showing ``[COMMAND]`` instead of ``COMMAND``.
     :issue:`3059` :pr:`3507`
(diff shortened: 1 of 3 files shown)


tyler-talbert:
Hi @davidism 

no problem with the close; could you share what would make this acceptable, or whether the underlying bug isn't one you want to fix? Just trying to learn for future contributions.

tyler-talbert:
@Rowlando13 

Wanted to add your visibility here as well

davidism:
There's a bigger discussion going on regarding this. Adding locks isn't the solution.

PR #3533, opened 2026-05-29
Do not let the GC close a borrowed `stdout`

### LINKED ISSUE #3449: "I/O operation on closed file" error with CliRunner and echo_via_pager under 8.4

kdeldycke:
Complete the partial fix proposed in https://github.com/pallets/click/pull/3482 for https://github.com/pallets/click/issues/3449.

This PR reproduce the case highlighted in #3449 where a stream that is not managed by the pager machinery is closed by it unconditionally.

As  extra bonuses:
- replace `_SkipClose` with `utils.KeepOpenFile`
- document `KeepOpenFile`
- eliminate a dead `elif isinstance(stream, t.BinaryIO)` branch as non-buffered text streams can be yielded as-is

This is a follow up of #3482 which is a follow up of #3405 which is a follow up of #1572.

CC @ChrisPappalardo @AndreasBackx @craigds

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,10 @@ Unreleased
 -   ``echo_via_pager`` flushes after each write, so passing a generator streams
     output to the pager incrementally instead of staying hidden until the pipe
     buffer fills. :issue:`3242` :issue:`2542` :pr:`3534`
+-   ``echo_via_pager`` and ``get_pager_file`` no longer close a borrowed stdout
+    stream when no external pager runs, completing the partial
+    ``I/O operation on closed file`` fix from :pr:`3482`. :issue:`3449`
+    :pr:`3533`
 
 
 Version 8.4.1
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -29,6 +29,7 @@ from ._compat import term_len
 from ._compat import WIN
 from .exceptions import ClickException
 from .utils import echo
+from .utils import KeepOpenFile
 
 V = t.TypeVar("V")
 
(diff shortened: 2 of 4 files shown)


davidism:
I realized there's already `utils.KeepOpenFile` which does all the dunder stuff. Let's get rid of `_SkipClose`, it's redundant. (Again noticing just how many wrappers we have.)

davidism:
There also seems to be way too many context managers and levels of indirection for the pager machinery at this point. I'm not sure how we got here, but I'd sure love to refactor and simplify it somehow.

kdeldycke:
> I realized there's already `utils.KeepOpenFile` which does all the dunder stuff. Let's get rid of `_SkipClose`, it's redundant. (Again noticing just how many wrappers we have.)

TIL. Let me play a bit with `utils.KeepOpenFile` to see how it works...

kdeldycke:
> I realized there's already `utils.KeepOpenFile` which does all the dunder stuff. Let's get rid of `_SkipClose`, it's redundant. (Again noticing just how many wrappers we have.)

Seems to do the trick! Thanks for the hint! :)

kdeldycke:
> There also seems to be way too many context managers and levels of indirection for the pager machinery at this point. I'm not sure how we got here, but I'd sure love to refactor and simplify it somehow.

This part of Click is insane. The only thing that gives me confidence is the amount of collected cases from users that we transformed into unittests.

kdeldycke:
I squashed the commits, this is ready to be reviewed. I'll go down the pager rabbit hole in other PRs.

Rowlando13:
@ChrisPappalardo can you review this?

frenzymadness:
I can confirm that this fixes all the I/O problems for us. We still have 30 failed tests, but I haven't had time to investigate them yet.

ChrisPappalardo:
> @ChrisPappalardo can you review this?

LGTM - this has evolved pretty far from my initial PR, but I think this approach works

PR #3553, opened 2026-06-04
Run nightly tests on both `main` and `stable` branches

kdeldycke:
Since #3500, we now runs a workflow on the `main` branch every night to detect race conditions and leaks with randomized and stress tests.

This PR update the workflow to check both `main` and `stable` branches, instead of `main`-only, to ensure we catch these kind of regressions even earlier in the development cycle.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -9,12 +9,13 @@ concurrency:
   cancel-in-progress: false
 jobs:
   race-tests:
-    name: ${{ matrix.name }}
-    runs-on: ${{ matrix.os || 'ubuntu-latest' }}
+    name: ${{ matrix.config.name }} (${{ matrix.branch }})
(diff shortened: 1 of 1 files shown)


kdeldycke:
Thanks for the merge! Before any 8.4.2 release, I need to double-check how this workflow runs on both `main` and `stable`, as I suspect there is a small regression we introduced by piling up changes. Just a hunch I'll need to verify after the first successful run,

kdeldycke:
> Thanks for the merge! Before any 8.4.2 release, I need to double-check how this workflow runs on both `main` and `stable`, as I suspect there is a small regression we introduced by piling up changes. Just a hunch I'll need to verify after the first successful run,


Just triggered that workflow manually (was not sure I had the rights), and everything is green: https://github.com/pallets/click/actions/runs/27082285856 . I guess we can plan a 8.4.2 release.

PR #3560, opened 2026-06-06
Move changes to markdown

Rowlando13:


kdeldycke:
That's a big conversion! 😅 But we had it coming. Don't know if that was your intention but this change ended up in `stable` so will be part of the upcoming 8.4.2. I'm OK with that, just wanted to double-check it was your target branch.

Rowlando13:
Yes. Intentionally did the change on main and stable. It was easier to resolve the conflicts once both were formatted the same.

Issue #2786, opened 2024-10-09
Multi-value options override the value returned from `callback` when not all specified


mxndtaylor:
I discovered this when I wanted to add a `flag_value`-style option to shortcut a custom `str` parameter option.

I want to have 2 options:
1. one that accepts a parameter of any `str` value like the name of a file or something
2. another that is a flag that fetches the value from a remote server instead of me specifying it

Since these both represent the same thing, my initial thought was just to pass them under the same parameter name to the function, which seems to almost work.

the problem seems to be that if the "custom" value is not specified while the "flag" option is set, then click will use the `flag_value` from the "flag" as the default value for the "custom" value.

And `click` seems to process unspecified options last (which I can see why this was done), but since "custom" is not specified, it processes the default after the "flag" option has already set a value for the function parameter that they share.

```python
# bad.py
import click

SENTINEL = "$_fetch"

def fetch():
    # some logic dependent on an external resource
    # return requests.get("...").json()["value")
    print('resource fetched')
    return "foo"

def callback(ctx, param, value):
    if value is SENTINEL:
        return fetch()
    return value

@click.command()
@click.option("--custom", "custom")
@click.option("--fetch", "custom", flag_value=SENTINEL, callback=callback)
def cli(custom):
    print(f'custom={custom}')

if __name__ == '__main__':
    cli()
```

results in the `SENTINEL` internal value being passed to the function when `--fetch` is specified alone:
```sh
# call with just --custom <arg>, get <arg>, as expected
$ python bad.py --custom bar
custom=bar

# call with just --fetch, get SENTINEL, the internal value that should not ever appear in the values
$ python bad.py --fetch
resource fetched
custom=$_fetch

# call with --custom <arg> first and --fetch after, get the fetched resource, as expected
$ python bad.py --custom bar --fetch
resource fetched
custom=foo
```
It's that last use-case that really rubs the dirt in; paradoxically, I have to specify the longer option I was trying to avoid using in order to get my "shortcut" flag option to work.

However, we can use a workaround to fix it, sort of, by passing the same `callback` to both options:
```python
# bad.py
# ..
@click.option("--custom", "custom", callback=callback)
@click.option("--fetch", "custom", flag_value=SENTINEL, callback=callback)
# ..
```

results in fetch being called twice when `--fetch` is the last specified option:
```sh
# call with just --custom <arg>, get <arg>, as expected
$ python bad.py --custom bar
custom=bar

# call with just --fetch, get foo, but the external resource is pinged twice
[...]

PR #3562, opened 2026-06-06
AI spam

suzunn:
﻿## Problem

Shared options that target the same parameter can process the same parser value more than once. When one option has a callback and another sibling option does not, the callback result can be overwritten by the sibling's raw parser value. In #2786 this allows an internal `flag_value` sentinel to reach the command callback when only the shortcut flag is provided.

## Solution

I tightened the shared-parameter arbitration so a callback-less option with the same command-line source does not replace an existing processed value with a different raw value. This preserves the callback result while keeping the existing explicit-source ordering behavior for normal shared options.

I also added regression coverage for the flag-only case and both mixed ordering cases:

- `--fetch` keeps the callback result
- `--custom typed --fetch` lets the later flag win
- `--fetch --custom typed` lets the later explicit value win

## Testing

- `uv run --group tests pytest tests\test_options.py -k "shared_option_callback_value_is_preserved or default_dual_option_callback" -q`
- `uv run --group tests pytest tests\test_options.py -q`
- `uv run ruff check src\click\core.py tests\test_options.py`
- `uv run --group tests pytest -q`
- `git diff --check`

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2674,10 +2674,21 @@ class Parameter(ABC):
             and existing_default_explicit
             and not self._default_explicit
         )
+        unprocessed_would_downgrade_callback = (
+            same_source
+            and source == ParameterSource.COMMANDLINE
+            and self.callback is None
+            and existing_value is not UNSET
+            and existing_value != value
+        )
         is_winner = (
             slot_empty
             or more_explicit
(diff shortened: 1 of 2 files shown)


ThiefMaster:
@noezhiya-dot Please do not dump AI-generated comments in other people's repos who did not ask for them.

PR #3568, opened 2026-06-07
AI junk

### LINKED ISSUE #3022: Support FORCE_COLOR and NO_COLOR environment variables
It would be great if the `FORCE_COLOR` and `NO_COLOR` environment variables could be used during calls to `click.echo`/`click.secho` to determine the default preference for colouring of stdout/stderr when an explicit preference for outputting colour is not set.

It appears that currently, click simply inspects whether the output stream considers itself to be a physical TTY. This is problematic on build systems such as GitLab or Jenkins or GitHub Actions, as logs that could be output in colour will not be since the output stream is still technically a file.

Supporting this would allow simpler catering for outputs that are designed to help visually impaired users when viewing automated build outputs, but may have other applications as well.

These environment variables are a good candidate to be used, because [the Python interpreter already considers these specific variables](https://docs.python.org/3/using/cmdline.html#controlling-color), and other libraries such as `colorlog` also respect this.

madhavcodez:
Adds support for the `NO_COLOR` and `FORCE_COLOR` environment variables, as requested in #3022.

`resolve_color_default()` is the single place every color-aware output path goes through (`echo`, `secho`, `prompt`, and error display). Today it returns the explicit `color` argument if one is given, otherwise the active context's `color`, otherwise `None` (auto-detect from whether the stream is a TTY). This adds one more fallback step: when neither the caller nor the context expresses a preference, the environment is consulted.

- A non-empty `NO_COLOR` disables color (per https://no-color.org).
- A non-empty `FORCE_COLOR` forces it on.
- `NO_COLOR` takes precedence over `FORCE_COLOR`; empty values are ignored.

This helps in CI (GitHub Actions, GitLab, Jenkins), where stdout is not a TTY but colored logs are often wanted, and conversely for users who set `NO_COLOR` globally. An explicit `color=` argument or a context `color` setting still wins, so nothing changes for callers that already opt in or out, and behavior is unchanged when neither variable is set.

fixes #3022

### Changes

- `src/click/globals.py` — `resolve_color_default()` falls back to `NO_COLOR` / `FORCE_COLOR`, with a `.. versionchanged:: 8.5` note.
- `tests/test_utils.py` — tests for the env fallback, `NO_COLOR` precedence over `FORCE_COLOR`, empty-value handling, explicit-argument override, and a direct unit test of `resolve_color_default()`.
- `CHANGES.md` — changelog entry.

### Notes

I went with the no-color.org convention where a variable counts only when set to a non-empty string (so `NO_COLOR= mytool` does *not* disable color), rather than mere presence. Happy to switch to presence-based checks if you'd rather match the interpreter's exact semantics.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,11 @@
 
 Unreleased
 
+- `echo`, `secho`, and other color-aware output respect the `NO_COLOR` and
+  `FORCE_COLOR` environment variables when no explicit color preference is
+  set by the caller or context, matching the behavior of the Python
+  interpreter. `NO_COLOR` takes precedence and empty values are ignored.
+  {issue}`3022`
 - Supported versions of Windows enable ANSI terminal styles by default.
   Colorama is no longer a dependency and is not used. {issue}`2986` {pr}`3505`
 - {class}`Argument` accepts a `help` parameter, and help output includes
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@ Unreleased
   `FORCE_COLOR` environment variables when no explicit color preference is
   set by the caller or context, matching the behavior of the Python
   interpreter. `NO_COLOR` takes precedence and empty values are ignored.
-  {issue}`3022`
+  {issue}`3022` {pr}`3568`
 - Supported versions of Windows enable ANSI terminal styles by default.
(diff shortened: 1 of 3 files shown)


ThiefMaster:
This PR feels like it's LLM-generated...

Issue #3572, opened 2026-06-08
ANSI codes not stripped from click.confirm after upgrading to click 8.4


alasdairnicol:
I had failing unit tests after upgrading to click 8.4.1. I was able to address some 
by using `runner.invoke(cmd, color=False)`, but that didn't help for commands
that use `click.confirm()`

This test script `test_repro.py` replicates the issue.
```
import click
from click.testing import CliRunner


@click.command()
def cmd():
    click.echo(click.style('Hello World!', fg='green'))


@click.command()
def cmd_confirm():
    click.confirm(click.style('Hello World!', fg='green'), abort=True)


def test_echo_strips_ansi_with_color_false():
    runner = CliRunner()
    result = runner.invoke(cmd, color=False)
    assert result.output == "Hello World!\n"


def test_confirm_strips_ansi_with_color_false():
    runner = CliRunner()
    result = runner.invoke(cmd_confirm, input="Y", color=False)
    assert result.output == "Hello World! [y/N]: Y\n"  # FAILS: ANSI codes present
```

With click 8.3.3 both tests pass. With click 8.4.0 and 8.4.1, the `test_confirm_strips_ansi_with_color_false` test fails:

```
uvx --with click==8.4 pytest test_repro.py
...
================================================ FAILURES ================================================
_______________________________ test_confirm_strips_ansi_with_color_false ________________________________

    def test_confirm_strips_ansi_with_color_false():
        runner = CliRunner()
        result = runner.invoke(cmd_confirm, input="Y", color=False)
>       assert result.output == "Hello World! [y/N]: Y\n"  # FAILS: ANSI codes present
```

The ansi codes should be stripped when running click.confirm, as they are for click.echo.

Environment:

- Python version: 3.13
- Click version: 8.4.1
- MacOS

kdeldycke:
Confirmed: this is a regression introduced by https://github.com/pallets/click/pull/2969 . I just proposed a fix at https://github.com/pallets/click/pull/3653 . Can you test it and confirm it fixes the regression?

alasdairnicol:
Thanks @kdeldycke. I've tested #3653 and my test suite is passing again, without having to add `color=False` 🎉

PR #3573, opened 2026-06-08
AI spam

ShubhamPhapale:
Fixes #3572

`CliRunner` writes prompt text directly to stdout via `visible_input` /
`hidden_input` rather than through `echo`, so `color=False` did not strip ANSI
codes from prompts. This regressed in 8.4.0, when `confirm()` / `prompt()` output
moved off the `echo` path for readline support. The fix strips ANSI in those two
functions when `color=False`, matching what `echo` does.

- **Test:** added `test_prompt_strips_ansi_with_color_false`, asserting prompts are
  stripped with `color=False` and preserved with `color=True`. It fails on `main`
  without this change.
- **Changelog:** added an entry under the unreleased 8.4.2 section in `CHANGES.md`.

LinkedIn: redacted

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,9 @@ Unreleased
   stream when no external pager runs, completing the partial
   `I/O operation on closed file` fix from {pr}`3482`. {issue}`3449`
   {pr}`3533`
+- `CliRunner` strips ANSI codes from prompts when `color=False`, matching the
+  behavior of `echo`. This was broken in `8.4.0` when prompt text moved off the
+  `echo` path for readline support. {issue}`3572`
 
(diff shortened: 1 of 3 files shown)


ShubhamPhapale:
Thanks — both addressed in the latest push:

- **Hidden-input coverage:** extended the regression test with a `hide_input=True`
  prompt, asserting ANSI is stripped with `color=False` and preserved with
  `color=True` (so `hidden_input` is now exercised too).
- **Duplication:** factored the shared `prompt or "" → strip` logic into a local
  `_normalize_prompt` helper used by both `visible_input` and `hidden_input`.

PR #2969, opened 2025-06-07
Fix readline backspace/line-wrapping on linux

yorickvP:
termui contains a workaround for windows where it prints a readline prompt itself. On linux, this confuses `readline` so things like backspace/ctrl-backspace and line wrapping don't work correctly.

On platforms other than windows, just let readline handle the prompt.

fixes #2968

Note: while I used claude code to offload some manual work creating this PR, I've steered and reviewed all the changes and take personal responsibility for the contents.

PR #3575, opened 2026-06-10
AI spam

binggao1230:
Since 8.4.0, `_readline_prompt` passes the full prompt text to `input()` so readline handles line editing correctly (#2969). That bypasses `echo()` and its ANSI handling, so `click.prompt()` and `click.confirm()` leak styling codes when color is disabled — `click.confirm` as reported, and `click.prompt` equally (same code path).

The fix strips the codes in `_readline_prompt` using the same `should_strip_ansi(stream, resolve_color_default())` decision `echo()` uses, with the stream chosen by `err`, restoring the pre-8.4 behavior where the prompt went through `echo`. The lookup goes through the `_compat` module attribute because `CliRunner` replaces `should_strip_ansi` during isolation.

Tests cover `prompt` and `confirm` with color disabled and enabled (codes must survive `color=True`), and the `err=True` stderr path; the three stripping tests fail without the change.

fixes #3572

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,9 @@ Unreleased
   stream when no external pager runs, completing the partial
   `I/O operation on closed file` fix from {pr}`3482`. {issue}`3449`
   {pr}`3533`
+- `prompt` and `confirm` strip ANSI codes from the prompt text again when
+  color is disabled. The prompt is passed to `input()` directly since 8.4.0,
+  bypassing `echo` and its ANSI handling. {issue}`3572` {pr}`3575`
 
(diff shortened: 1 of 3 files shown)


binggao1230:
Thanks for the careful read! I'd lean toward leaving it as-is — `resolve_color_default()` is just a thread-local context lookup (it returns `get_current_context().color`), so for an interactive prompt the cost is negligible, and I'd rather keep this PR scoped to the 8.4 regression than add caching state to the hot path. Happy to revisit if it ever shows up in a profile.

PR #3579, opened 2026-06-10
AI spam

### LINKED ISSUE #3476: `test_flag_group_competition_duplicate_option_name` relies on `-Werror`
The following test fails when the test suite of click-8.4.0 is run with `-Wdefault`:

```pytb
$ pytest -Wdefault tests/test_options.py::test_flag_group_competition_duplicate_option_name
========================================================= test session starts =========================================================
platform linux -- Python 3.13.13, pytest-9.0.3, pluggy-1.6.0
rootdir: /tmp/click-8.4.0
configfile: pyproject.toml
collected 1 item                                                                                                                      

tests/test_options.py F                                                                                                         [100%]

============================================================== FAILURES ===============================================================
__________________________________________ test_flag_group_competition_duplicate_option_name __________________________________________

runner = <click.testing.CliRunner object at 0x7fa0db07e660>

    def test_flag_group_competition_duplicate_option_name(runner):
        """The same option name declared twice on the same command is a user
        error.
        """
    
        @click.command()
        @click.option("--xyz", default="first")
        @click.option("--xyz", default="second")
        def cli(xyz):
            click.echo(repr(xyz), nl=False)
    
        result = runner.invoke(cli, [])
>       assert result.exit_code == 1
E       assert 0 == 1
E        +  where 0 = <Result okay>.exit_code

tests/test_options.py:3248: AssertionError
========================================================== warnings summary ===========================================================
tests/test_options.py::test_flag_group_competition_duplicate_option_name
  /tmp/click-8.4.0/src/click/core.py:1262: UserWarning: The parameter --xyz is used more than once. Remove its duplicate as parameters should be unique.
    parser = self.make_parser(ctx)

tests/test_options.py::test_flag_group_competition_duplicate_option_name
  /tmp/click-8.4.0/src/click/core.py:1255: UserWarning: The parameter --xyz is used more than once. Remove its duplicate as parameters should be unique.
    self.parse_args(ctx, args)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
======================================================= short test summary info =======================================================
FAILED tests/test_options.py::test_flag_group_competition_duplicate_option_name - assert 0 == 1
==================================================== 1 failed, 2 warnings in 0.13s ====================================================
```

We are running tests with `-Wdefault`, since it is important that they don't regress over unrelated warnings. If they expect a warning, they should check for a warning rather than rely on the pytest invocation implicitly turning warnings into exceptions.

Environment:

- Python version: 3.13.13
- Click version: 8.4.0

kshitiz235:
Fixes #3476.

`test_flag_group_competition_duplicate_option_name` asserted on
`result.exception`, which only works because the test suite turns warnings
into errors (`filterwarnings = ["error"]`). When the test runs under the
default warning filter (`pytest -W default`), the `UserWarning` is emitted
but not raised, so the command exits 0 and the assertions fail. This breaks
downstream packagers who run the suite with different warning filters.

This uses `pytest.warns(UserWarning, match="used more than once")` to assert
the warning directly, so the test passes regardless of the active warning
filter (it still passes under the project's warnings-as-errors config).

- [x] Reproduced the failure with `pytest -W default`
- [x] Test passes under both `-W default` and the default warnings-as-errors config
- [x] Full `tests/test_options.py` suite passes (644 passed)
- [x] `ruff check` and `ruff format` clean
- [x] Changelog entry added

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,8 @@ Unreleased
   stream when no external pager runs, completing the partial
   `I/O operation on closed file` fix from {pr}`3482`. {issue}`3449`
   {pr}`3533`
+- The test suite no longer relies on warnings being treated as errors, so it
+  passes when run with the default warning filter. {issue}`3476`
 
 ## Version 8.4.1
 
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -3320,10 +3320,8 @@ def test_flag_group_competition_duplicate_option_name(runner):
     def cli(xyz):
         click.echo(repr(xyz), nl=False)
 
-    result = runner.invoke(cli, [])
-    assert result.exit_code == 1
-    assert isinstance(result.exception, UserWarning)
-    assert "used more than once" in str(result.exception)
+    with pytest.warns(UserWarning, match="used more than once"):
+        runner.invoke(cli, [])
 
 
 @pytest.mark.parametrize(


davidism:
@kshitiz235  we are not accepting work generated with AI. Do not submit AI output as your own again.

PR #3589, opened 2026-06-12
AI junk

### LINKED ISSUE #2740: Display exception notes when raising `ClickException`
Since 3.11 Python supports [adding notes on exceptions](https://docs.python.org/3/tutorial/errors.html#enriching-exceptions-with-notes). Basically this is completion of the exception message, and it can be useful to add context.

Notes are currently not displayed when `ClickException` are raised:

```python
import click

@click.command()
def foobar():
    exc = click.ClickException("foo")
    exc.add_note("bar")
    raise exc

if __name__ == '__main__':
    foobar()
```

```shell
$ python foobar.py
Error: foo
```

I suggest to display the notes in addition to the main message.

What do you think?

kinghenesey:
Fixes #2740

Since Python 3.11, exceptions support `add_note()` which appends to
`__notes__`. When a `ClickException` is raised, these notes were
silently ignored.

This change updates `ClickException.show()` and `UsageError.show()`
to display any `__notes__` after the main error message, matching
how Python's default exception display handles notes.

### DIFF
--- a/src/click/exceptions.py
+++ b/src/click/exceptions.py
@@ -64,6 +64,9 @@ class ClickException(Exception):
             color=self.show_color,
         )
 
+        for note in getattr(self, "__notes__", []):
+            echo(note, file=file, color=self.show_color)
+
 
 class UsageError(ClickException):
     """An internal exception that signals a usage error.  This typically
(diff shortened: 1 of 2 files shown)


kinghenesey:
Hi @davidism, thanks for taking a look. Could you share any feedback on why this was closed? I'd love to understand if there's a different approach you'd prefer, or if this feature is out of scope for the project. thank you once again.

PR #3597, opened 2026-06-15
tests/test_basic: Fix to work with newer versions of pytest

rpurdie:
Fix an error with newer versions of pytest (9.1.0) by converting the generator into a tuple to resolve:

E   pytest.PytestRemovedIn10Warning: Passing a non-Collection iterable to parametrize is deprecated.
E   Test: tests/test_basic.py::test_boolean_conversion, argvalues type: chain
E   Please convert to a list or tuple.
E   See https://docs.pytest.org/en/stable/deprecations.html#parametrize-iterators
ERROR: tests/test_basic.py:tests/test_basic.py

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -258,9 +258,11 @@ def test_boolean_flag(runner, default, args, expect):
 
 @pytest.mark.parametrize(
     ("value", "expect"),
-    chain(
-        ((x, "True") for x in ("1", "true", "t", "yes", "y", "on")),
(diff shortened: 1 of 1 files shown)


Rowlando13:
@kdeldycke Whats the deal with this one? There are already some changes in the exact spot.

kdeldycke:
This PR is a duplicate of https://github.com/pallets/click/pull/3656 . It has already been addressed.

PR #3618, opened 2026-06-20
Fix incorrect variable name in pipeline example

Rfannn:
Fixed input to in in the command pipeline example. The variable in is passed as an argument and should be used in the generator expression.


Rfannn:
yo

PR #3626, opened 2026-06-23
optimize core parser backend using slots

Vizonex:
Greetings, I am looking to help contribute with this project as I see a need to make my own competitor to the typer library in the future (since I wanted to add my own context manager systems, hooks, config files, msgspec tooling (I take inspiration from the argparse library for dataclass) and other magical features) that will allow for more creativity and productivity in the long run. It will utilize this library hence why I am here to begin with.

One of the things I was primarily looking at doing as for right now was optimizing the backend with `__slots__` since we don't expect users to be accessing these deprecated classes anytime soon, unless I'm mistaken by the upcoming deprecation of these objects I hope that this at least speeds up some of the more heavier sections for the time being.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Unreleased
   Colorama is no longer a dependency and is not used. {issue}`2986` {pr}`3505`
 - {class}`Argument` accepts a `help` parameter, and help output includes
   a `Positional arguments` section when argument help is available. {issue}`2983` {pr}`3473`
+- Optimized parser backend to enhance performance {pr}`3626`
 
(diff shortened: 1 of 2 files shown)


Vizonex:
Feel free to let me know if I need to do anything else since I am a little new around here.

davidism:
This doesn't seem necessary, these will be removed soon, and the number of parser objects is not large enough for slots to meaningfully affect memory.

Vizonex:
@davidism No problem I had no clue if this should be here or not, I'll figure something else out to look at optimizing then :)

PR #3627, opened 2026-06-23
Optimize `split_arg_string` with `extend(...)` instead of a for loop system.

Vizonex:
This PR is an optimization that extends the lexer's iterables inside of `split_arg_string` instead of calling append every time. This ultimately saves a bit of code but also enhances performance.

### DIFF
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -502,7 +502,7 @@ def split_arg_string(string: str) -> list[str]:
     lex = shlex.shlex(string, posix=True)
     lex.whitespace_split = True
     lex.commenters = ""
-    out = []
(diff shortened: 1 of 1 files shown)


kdeldycke:
Just a heads up: this will be part of Click 8.5.0 and not 8.4.2 as it was merged to `main` and not `stable`.

Vizonex:
> Just a heads up: this will be part of Click 8.5.0 and not 8.4.2 as it was merged to `main` and not `stable`.

@kdeldycke No problem I am actually used to the wait. I've contributed to many packages previously.

Rowlando13:
@Vizonex if you find any more optimizations of similar scope, we would be happy to take a look.

PR #3641, opened 2026-06-27
Streamline option flag management

kdeldycke:
This refactor is an attempt in streamlining the code around flag management in `Option`.

I hope this is not too big to digest. I did not went too far and just focused on `Option` internals and flag-management. My goal is to colocate decision code by domain. Like having all prompt-related code in the same block. Or all type-resolution code together. Sometimes in the same code block, sometimes under a private utility function. My heuristic is mostly driven by the number of lines and the content of the docstring.

Another idiom I introduced was using `@property` for `is_bool_flag`. I didn't see that much propoerties used in Click, but these are easier to reason about for me.

I also used the oportunity to introduce a type inference point in `_pick_type` which is a small step echoing what @sirosen [was proposing last year](https://github.com/pallets/click/pull/3079#issuecomment-3324790897) about having a separation between "data which is stored (where you might have UNSET) from the data which is viewed". That also resurface the infamous `UNSET` a bit, but hidden into a new `flag_activation_value` property. And `_hide_unset` is a way to delimit the internals state from the presetation state.

This also contains some small syntax cleanups that were flagged by my linters.

Proof that this is the right time to introduce this refactor: after accumulating a lot of unitetsts over the past few years, I did not break any. And I did not discovered any new blind spot in the test suite.

I hope this will make code easier to read and maintain. But I need other eyes to check that.

My initial goal was way too ambitious as I want to create a `FlagOption` subclass of `Option`. But this is way too big and not sure I can pull it yet. So this is like a first step in that general direction.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,11 @@ Unreleased
   {issue}`2819` {pr}`3678`
 - `unstyle` and the ANSI handling behind help-text wrapping now strip the full
   CSI escape-sequence grammar.
+- Streamline `Option` flag handling: the flag-kind, type, lazy-default and
+  validation steps in `Option.__init__` move into focused helpers, and
+  `flag_value` and `default` keep their unset sentinel at construction
+  (resolved lazily on read) so `is UNSET` reliably tells a user-supplied value
+  from an auto-derived one. Behavior is unchanged. {pr}`3641`
 
 ## Version 8.4.2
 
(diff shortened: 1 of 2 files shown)


Rowlando13:
@kdeldycke This looks good. @davidism Can you take a look?

kdeldycke:
Given https://github.com/pallets/click/pull/3676 , I guess the milestone change to 9.0.0 is just the blast radius of the recent `main` <-> `stable` dance. So I retarget this to 8.5.0 unless an explicit decision is made.

PR #3651, opened 2026-06-29
AI junk

kmg0308:
## Summary
- Compare Bash shell completion versions as numbers so 4.10 and 10.0 are not treated as older than 4.4.

## Validation
- shell completion tests, full pytest, and ruff passed.

## Notes
- Opened as a draft PR for maintainer review.


ThiefMaster:
As you may have seen, we do not accept LLM-generated PRs.

ThiefMaster:
And honestly, this change is so trivial, you could have done that without all the resources wasted by putting an LLM on it.

PR #3656, opened 2026-07-01
Don't parameterize tests using non-Collection iterables

cjwatson:
This is deprecated as of pytest 9.1; see
https://docs.pytest.org/en/stable/deprecations.html#non-collection-iterables-in-pytest-mark-parametrize.


kdeldycke:
Ah! Nice catch. Let's merge.

PR #3663, opened 2026-07-06
Fix Windows prompt err=True stdout leak in_readline_prompt

### LINKED ISSUE #3662: 8.4.x: prompt/confirm(err=True) leaks prompt tail and CliRunner reply echo to stdout on Windows
On the stable 8.4.x line, `prompt()` / `confirm()` with `err=True` still write part of the prompt interaction to **stdout** on Windows, because `_readline_prompt`'s `WIN` branch bypasses the stderr redirect that the POSIX branch applies:

```python
# src/click/termui.py (8.4.2)
def _readline_prompt(func, text, err):
    if WIN:
        echo(text[:-1], nl=False, err=err)   # prompt body honors err ...
        return func(text[-1:])               # ... but this call is NOT redirected
    if err:
        with redirect_stdout(sys.stderr):    # POSIX: fully redirected
            return func(text)
    return func(text)
```

Consequences on Windows only:

1. **Real console**: `input(text[-1:])` writes the prompt's last character (the trailing space of `"... [y/N]: "`) to stdout. For CLIs that promise "stdout is a single machine-readable document" under a `--json` flag (prompt chrome on stderr via `err=True`), that's a stray byte in the payload stream — benign for JSON parsers since it's whitespace, but the contract is broken.
2. **`CliRunner`** (the bigger practical hit): test isolation replaces `termui.visible_prompt_func` with a shim that echoes the prompt tail **and the typed reply** to `sys.stdout`. Because the `WIN` branch doesn't wrap the call in `redirect_stdout`, stdout becomes `" n\n" + <payload>`. Any test doing `json.loads(result.stdout)` on a prompt-then-JSON flow passes on POSIX and fails on Windows with `Expecting value: line 1 column 2 (char 1)`.

## Reproduction

Deterministic on a Windows runner; on POSIX you can force the branch to see the exact same behavior:

```python
import json
import click, click.termui

@click.command()
def cmd():
    if not click.confirm("Continue?", default=False, err=True):
        click.echo(json.dumps({"ok": False}))

# click.termui.WIN = True   # <- uncomment on POSIX to simulate
result = CliRunner().invoke(cmd, [], input="n\n")
print(repr(result.stdout))
# POSIX / main:  '{"ok": false}\n'
# Windows 8.4.2: ' n\n{"ok": false}\n'
```

Environment: click 8.4.2, Python 3.12, observed on GitHub Actions `windows-latest` (and reproduced via the `WIN = True` simulation above on macOS/Linux).

## Note on main

This is already fixed on `main` as a side effect of the colorama removal (74546476376b39d05186ec1085fd2faf9d86fa20) — `_readline_prompt` no longer has a `WIN` branch, so the `err=True` redirect applies on all platforms. So this report is effectively a request to consider backporting the redirect part to the 8.4.x stable line (the `if WIN:` branch could keep the separate `echo` for coloring but wrap the `func` call in `redirect_stdout(sys.stderr)` when `err=True`).

Downstream context, if useful: this surfaced as two Windows-only CI failures in memtomem (bisect + analysis in https://github.com/memtomem/memtomem/issues/1640); we've since worked around it by bypassing the prompt machinery for our `--json` paths.

asnamobin-hue:
Fixes issue where Windows _readline_prompt bypassed stderr redirection when err=True,
causing prompt output to leak into stdout in CliRunner-based tests.

This change ensures consistent behavior with POSIX by correctly applying
redirect_stdout(sys.stderr) for prompt execution when err=True.

Fixes #3662

### DIFF
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -94,11 +94,15 @@ def _readline_prompt(func: t.Callable[[str], str], text: str, err: bool) -> str:
         echo(text[:-1], nl=False, err=err)
         # Echo the last character to stdout to work around an issue
         # where readline causes backspace to clear the whole line.
-        
         if err:
             with redirect_stdout(sys.stderr):
-                return func(text)
-        return func(text)
+                return func(text[-1:])
+
+        return func(text[-1:])
+    if err:
+        with redirect_stdout(sys.stderr):
+            return func(text)
+    return func(text)
 
 
 def _build_prompt(
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-
+from ._compat import WIN
 import collections.abc as cabc
 import inspect
 import io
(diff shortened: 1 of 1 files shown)


davidism:
I don't want to deal with backporting this.

PR #3420, opened 2026-05-12
Make `TextWrapper` ANSI-aware

kdeldycke:
### Bug description

`click.formatting.wrap_text` and the underlying `click.formatting.TextWrapper` use the Python standard library implementation inherited from `textwrap.TextWrapper`. But it is not aware of ANSI codes. Which are invisible to the user, but artificially inflate the line limit accounting used to decide when a long line of terminal text is going to be cut and put on a new line.

### Solution

To fix this, I copied the `textwrap.TextWrapper._wrap_chunks` implementation from the standard library and replaced the used of raw `len()` by our own `click._compat.term_len()`, which strip ANSI before counting.

Patching it there allow to fix the shared behavior of `wrap_text` itself, but also `HelpFormatter.write_usage`, `write_text`, `write_dl`, `HelpFormatter.write_text` and `write_paragraph`.

[Python's own `_wrap_chunks`](https://github.com/python/cpython/blob/0534774a472424f6b9be2dc4ce9784384bc02401/Lib/textwrap.py#L238-L339) is monolithic and too rigid, and couldn't find a clean way of patching it. So my last resort was to brutally copy it and change its code locally.

This is I think the right thing to do in principle, as Click as its own subclass for `textwrap.TextWrapper`, and already ships with the perfect `click._compat.term_len()` utility for this task.

### Diff with Python's stdlib

For reference, here are the changes I applied compared to the inherited class:

```diff
--- cpython 3.14.3 Lib/textwrap.py TextWrapper._wrap_chunks
+++ click src/click/_textwrap.py TextWrapper._wrap_chunks
@@ -1,13 +1,13 @@
-def _wrap_chunks(self, chunks):
+def _wrap_chunks(self, chunks: list[str]) -> list[str]:
     lines = []
     if self.width <= 0:
-        raise ValueError("invalid width %r (must be > 0)" % self.width)
+        raise ValueError(f"invalid width {self.width!r} (must be > 0)")
     if self.max_lines is not None:
         if self.max_lines > 1:
             indent = self.subsequent_indent
         else:
             indent = self.initial_indent
-        if len(indent) + len(self.placeholder.lstrip()) > self.width:
+        if term_len(indent) + term_len(self.placeholder.lstrip()) > self.width:
             raise ValueError("placeholder too large for max width")

     chunks.reverse()
@@ -22,27 +22,27 @@
         else:
             indent = self.initial_indent

-        width = self.width - len(indent)
+        width = self.width - term_len(indent)

[...]

PR #3666, opened 2026-07-07
AI junk

sean-kim05:
### Summary

`click.style()` (and therefore `secho()`) silently drops the 256-color index `0` (black) when passed as `fg` or `bg`, producing uncolored text.

The color arguments are guarded with a truthiness check:

```python
if fg:
    bits.append(f"\033[{_interpret_color(fg)}m")
if bg:
    bits.append(f"\033[{_interpret_color(bg, 10)}m")
```

`0` is a valid 8-bit color index (documented range "[0, 255]") but is falsy, so the guard skips it and no escape is emitted. Every other input works — nonzero indices, named colors, and even RGB black `(0, 0, 0)` (a truthy tuple) — which makes index `0` an inconsistency, not a design choice.

### Reproduction

```pycon
>>> import click
>>> click.style("x", fg=0)      # 256-color black
'x\x1b[0m'                       # color dropped
>>> click.style("x", fg=1)      # works
'\x1b[38;5;1mx\x1b[0m'
>>> click.style("x", fg=(0, 0, 0))  # RGB black works
'\x1b[38;2;0;0;0mx\x1b[0m'
```

### Fix

Guard on `is not None` instead of truthiness, matching how the other style attributes (`bold`, `dim`, …) in the same function are already handled. `None` (the default) stays suppressed, unknown color names still raise `TypeError`, and index `0` is now emitted correctly.

The truthiness guards date to the original 2014 implementation when only named colors (all truthy) existed; integer/RGB support was later added inside the same blocks without updating them.

### Tests

Added `{"fg": 0}` and `{"bg": 0}` cases to the parametrized `test_styling` table in `tests/test_utils.py`. They fail on `main` and pass with this change; the full suite is unaffected (1682 passed).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,9 @@ Unreleased
   This stripping was lost in `8.4.0` when {pr}`2969` began writing the
   prompt with `input()` directly. {issue}`3572` {pr}`3653`
 - Fix test failures when using pytest >= 9.1. {pr}`3656`
+- `style()` and `secho()` no longer silently drop the 256-color index `0`
+  (black) passed as `fg` or `bg`. A truthiness check treated the valid index
+  `0` as unset. {pr}`3666`
 
 ## Version 8.4.2
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,8 +12,11 @@ Unreleased
   prompt with `input()` directly. {issue}`3572` {pr}`3653`
 - Fix test failures when using pytest >= 9.1. {pr}`3656`
 - `style()` and `secho()` no longer silently drop the 256-color index `0`
-  (black) passed as `fg` or `bg`. A truthiness check treated the valid index
-  `0` as unset. {pr}`3666`
+  (black) passed as `fg` or `bg`, and now validate color arguments: unknown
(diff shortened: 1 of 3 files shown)


kdeldycke:
This is right on time as I am myself digging into the ANSI-side of Click (see: https://github.com/pallets/click/pull/3653 and https://github.com/pallets/click/pull/3420). I confirm the issue, which also made me aware of a new class of edge-cases the `style()` method is not processing. This affects both my Click Extra project and Cloup (cc @janluke).

So I just updated your PR with an additional commit to illustrate all the edge-cases not covered in unittests, and also fix the behavior by validating all colors passed to `style()`.

Now I need a quick pass from other maintainers to check that I did not miss anything but to me this is good to merged in the upcoming 8.5.0 release.

kdeldycke:
@davidism do you want me to redo it?

kdeldycke:
> @davidism do you want me to redo it?

I guess this is another case of https://github.com/pallets/click/pull/3165#issuecomment-4207585879 , so I just reproduced my fix at https://github.com/pallets/click/pull/3677

PR #3673, opened 2026-07-08
Stable

Rowlando13:

PR #3672, opened 2026-07-08
Move test utils to a module and each function to its own file

Rowlando13:

PR #3505, opened 2026-05-23
Remove colorama

davidism:
Tested `echo`, `prompt`, and other output utilities in a script, and styles seem to work fine. Tested in both Terminal and cmd.exe. I'm not clear _when_ Windows 11 enabled ANSI by default, but realistically we can only support the latest version since that's the only version we and most people can get. An application that wants to support older Windows can still require colorama and call `colorama.init()` before `main()`.

closes #2986

PR #3674, opened 2026-07-08
Merge stable into main. 

Rowlando13:


kdeldycke:
Relates to:
- https://github.com/pallets/click/pull/3672
- https://github.com/pallets/click/pull/3673

kdeldycke:
This introduced the drop of a change in tests that was previously introduced in #3505. A fix is proposed at #3676.

PR #3508, opened 2026-05-23
Backport #3504 from `main` to `stable` branch

kdeldycke:
My intention was to merge https://github.com/pallets/click/pull/3504 to `stable` to include it for the upcoming 8.4.2 release, but was distracted by the topic of the PR itself and forgot to retarget the base branch of the PR from `main` to `stable`. This is the well known issue discussed at https://github.com/pallets/click/discussions/3337.

To fix this, this PR just cherry-pick the c3535905 commit from `main` and target `stable` inclusion. The next sync of `stable` to `main` will be clean and be reconciled automaticcaly by a forward-merge.

As a bonus, I added the missing changelog entry for #3504, updated the admonition and added a regression test.

PR #3676, opened 2026-07-08
Restore `test_echo_color_flag` from #3505

kdeldycke:
In #3674 to push split of `test_utils.py`, a test change in #3505 while dropping Colorama has been lost.

This restore it.

See recent @Rowlando13 `main` <-> `stable` dance in:
- https://github.com/pallets/click/pull/3672
- https://github.com/pallets/click/pull/3673
- https://github.com/pallets/click/pull/3674

### DIFF
--- a/tests/test_utils/test_echo.py
+++ b/tests/test_utils/test_echo.py
@@ -74,22 +74,18 @@ def test_echo_color_flag(monkeypatch, capfd):
     out, err = capfd.readouterr()
     assert out == f"{styled_text}\n"
 
-    isatty = True
     click.echo(styled_text)
     out, err = capfd.readouterr()
     assert out == f"{styled_text}\n"
 
     isatty = False
-    # Faking isatty() is not enough on Windows;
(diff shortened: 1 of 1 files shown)


Rowlando13:
@kdeldycke Thanks for the double check. How did you spot this?

kdeldycke:
> @kdeldycke Thanks for the double check. How did you spot this?

Because I just remembered the last time I had to do the `main` <-> `stable` dance I had to painfully reconcile the two. Remember https://github.com/pallets/click/pull/3508 ? 😁 So when I just saw you playing with the branches and other tickets being changed their milestone, I was pretty sure something was dropped along the way. Just my instincts. So I compared the code in both branches and found the discrepancy.

But yeah, something is probably wrong in the way we sync or manage our branches as this is not the first time this happens. In my projects I just have a `main` and if I need a hotfix, I just branch the last tagged release, create a new release from that and reconcile the history by hand.

PR #3677, opened 2026-07-08
Validate `style()` color arguments

kdeldycke:
Also fix an explicit black color argument (`0`) being ignored.

### DIFF
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -570,14 +570,28 @@ def clear() -> None:
 
 
 def _interpret_color(color: int | tuple[int, int, int] | str, offset: int = 0) -> str:
-    if isinstance(color, int):
-        return f"{38 + offset};5;{color:d}"
+    """Interprets a color value and returns the corresponding ANSI code."""
+    if isinstance(color, str) and color in _ansi_colors:
+        return str(_ansi_colors[color] + offset)
 
-    if isinstance(color, (tuple, list)):
+    # bool is an int subclass: without the exclusion, True and False would
+    # silently render as the palette indices 1 and 0.
+    elif isinstance(color, int) and not isinstance(color, bool):
+        if 0 <= color <= 255:
+            return f"{38 + offset};5;{color:d}"
+
+    elif (
+        isinstance(color, (tuple, list))
+        and len(color) == 3
+        and all(
+            isinstance(c, int) and not isinstance(c, bool) and 0 <= c <= 255
+            for c in color
+        )
+    ):
         r, g, b = color
         return f"{38 + offset};2;{r:d};{g:d};{b:d}"
 
-    return str(_ansi_colors[color] + offset)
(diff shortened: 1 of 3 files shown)


src/click/termui.py
```diff
@@ -570,14 +570,23 @@ def clear() -> None:
...
+        if len(color) == 3 and all(
+            isinstance(c, int) and not isinstance(c, bool) and 0 <= c <= 255
+            for c in color
+        ):
+            r, g, b = color
+            return f"{38 + offset};2;{r:d};{g:d};{b:d}"
+    elif isinstance(color, str) and color in _ansi_colors:
+        return str(_ansi_colors[color] + offset)
+
+    raise TypeError(_("Unknown color {colour!r}").format(colour=color))
```
davidism: `ValueError` makes more sense here.

kdeldycke: Indeed! Fixed in https://github.com/pallets/click/compare/d602e9f596b64f760bb3112c7498905bdee7ae34..4860fa974a079fef7ddccab1296afaf8762af73d

src/click/termui.py
```diff
@@ -570,14 +570,23 @@ def clear() -> None:
...
+        if 0 <= color <= 255:
+            return f"{38 + offset};5;{color:d}"
+    elif isinstance(color, (tuple, list)):
+        if len(color) == 3 and all(
+            isinstance(c, int) and not isinstance(c, bool) and 0 <= c <= 255
+            for c in color
+        ):
+            r, g, b = color
+            return f"{38 + offset};2;{r:d};{g:d};{b:d}"
+    elif isinstance(color, str) and color in _ansi_colors:
```
davidism: This should come first now, since it's probably the most common and cheapest check.

src/click/termui.py
```diff
@@ -655,6 +664,12 @@ def style(
                   string which means that styles do not carry over.  This
                   can be disabled to compose styles.
 
+    .. versionchanged:: 8.5
+        The 256-color index ``0`` is no longer ignored. All other invalid
+        ``fg`` and ``bg`` values now raise :exc:`ValueError`: previously,
```
davidism: Can omit the `previously`. A little more concise:

```
All invalid color values raise ``ValueError``. 256-color index 0 is not ignored.
```

src/click/termui.py
kdeldycke: That's sensible. Done in https://github.com/pallets/click/compare/b674be1194d9745852cbc86c7a6b3e60fef84e28..07c909f23f0f83b5ca137c167b9a134d66201f67

src/click/termui.py
kdeldycke: You're right, that is more concise. Done in https://github.com/pallets/click/compare/b674be1194d9745852cbc86c7a6b3e60fef84e28..07c909f23f0f83b5ca137c167b9a134d66201f67

PR #3678, opened 2026-07-08
Fix parsing when a parameter is named `help`

### LINKED ISSUE #2819: help not resolving automatically

kdeldycke:
This PR fixes #2819 by having ou pre-configured `--help` parameter automatically change its internal name used to store its value, in the case of a user providing its own argument or option conflicting with ours.

Note that I arbitrarily used a convention of adding underscores `_` to that internal name until no other parameter conflict with it. I'm ready to change that strategy if it is does not look good.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,14 @@ Unreleased
   [discussion #3527](https://github.com/pallets/click/discussions/3527). {pr}`3581`
 - `style()` and `secho()` no longer silently drop the 256-color index `0`
   (black) passed as `fg` or `bg`, and now validate color arguments. {pr}`3677`
+- The automatic help option stores its value under the reserved name
+  `_click_default_help` instead of `help`, so a parameter named `help` no
+  longer breaks parsing. The new name is visible in
+  {meth}`Command.to_info_dict` output. Parameters that overwrite each other's
+  value trigger a warning: an argument sharing its name with another
+  parameter, or any parameter claiming the reserved name. Options may still
+  share a name to compete for the same value (feature switches).
+  {issue}`2819` {pr}`3678`
 
 ## Version 8.4.2
 
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -322,7 +322,7 @@ The default placeholder variable ([meta variable](https://en.wikipedia.org/wiki/
 
 ## Help Parameter Customization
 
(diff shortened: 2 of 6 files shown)


davidism:
Perhaps we could just always force an internal name, like `_click_default_help`?

Rowlando13:
I think that is probably best.

kdeldycke:
> Perhaps we could just always force an internal name, like `_click_default_help`?

Yeah why not! Need to check if there is a nasty case in which a user can create its own argument or option that get assigned this internal ID. 😈

Wait. Hold on to that PR. I might have found that this issue not only affects `--help` but any parameter... 🤔 Let me verify that...

kdeldycke:
I just changed the strategy to always naming Click's own `--help` with the `_click_default_help` constant. And also make the name collision detection more generic and systematic: it now warns at the parameters fetching level. This greatly simplifies the tests and code.

The only compromise: the public `to_info_dict()` of `--help` now returns unconditionally the `_click_default_help` storage ID instead of `help`. I can restore part of my strategy of only adopting `_click_default_help` dynamically for `--help` if a colliding argument is detected.

PR #3684, opened 2026-07-12
AI junk

### LINKED ISSUE #2869: Support pathlib.Path in edit filename
Would you welcome a PR to avoid this?

```
Argument of type "Path" cannot be assigned to parameter "filename" of type "str | None" in function "edit"
  Type "Path" is not assignable to type "str | None"
    "Path" is not assignable to "str"
    "Path" is not assignable to "None"Pylance[reportArgumentType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportArgumentType)
```

Easy workaround with:
```py
click.edit(filename=str(export_common.path))
```

but would be nice if didn't need `str`.

arturict:
## What changed

- accept a single `os.PathLike[str]` or an iterable containing path-like filenames in `click.edit`
- normalize path-like values with `os.fspath` before invoking the editor
- cover both runtime behavior and static typing

## Why

`click.edit(filename=Path(...))` was rejected by type checkers and treated the path as an iterable at runtime. This makes the public API consistent with other Click file APIs that already accept path-like objects.

Fixes #2869

## Validation

- `uv run --frozen pytest -q` — 1832 passed, 95 skipped, 1 xfailed
- `uv run --frozen ruff check src tests`
- `uv run --frozen mypy tests/typing/typing_termui.py`
- `git diff --check`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 Unreleased
 
+- `edit()` accepts path-like objects for its `filename` parameter.
+  {issue}`2869`
 - Supported versions of Windows enable ANSI terminal styles by default.
   Colorama is no longer a dependency and is not used. {issue}`2986` {pr}`3505`
 - {class}`Argument` accepts a `help` parameter, and help output includes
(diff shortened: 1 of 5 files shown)


arturict:
Closing after finding the project's maintainer guidance that AI-generated contributions are not wanted. Sorry for missing that before opening this draft.

PR #3637, opened 2026-06-26
Add built-in PowerShell shell completion support

doctorlai-msrc:
## Summary

Adds `PowerShellComplete` as a fourth built-in shell completer, alongside the
existing `BashComplete`, `ZshComplete`, and `FishComplete`. Users of Click-based
CLIs on Windows PowerShell 5.1+ or PowerShell (pwsh) 7+ can now opt into TAB
completion the same way they do for the other shells:

```powershell
# In $PROFILE
$env:_FOO_BAR_COMPLETE = 'powershell_source'
foo-bar | Out-String | Invoke-Expression
Remove-Item Env:_FOO_BAR_COMPLETE
```

…or generate the script once and source it:

```powershell
$env:_FOO_BAR_COMPLETE = 'powershell_source'
foo-bar | Out-File -Encoding utf8 ~/.foo-bar-complete.ps1
Remove-Item Env:_FOO_BAR_COMPLETE
# then in $PROFILE:
. ~/.foo-bar-complete.ps1
```

PowerShell support has been a long-standing gap in Click; downstream projects
have repeatedly worked around it by shipping ad-hoc completer scripts. This PR
makes it a first-class built-in so every Click-based CLI gets it for free.

PR #3697, opened 2026-07-17
Pre-8.5.0 release changelog and doc fixes

kdeldycke:
> [!IMPORTANT]
> **Keep this PR in draft mode** until we are closing in on 8.5.0 release.

A small PR to accumulate all missing documentation details for the upcoming 8.5.0 release.

- Missing changelog entries and references from:
  - #3637
  - #3681
  - #3642
- Mention of new `custom_version_option` in the docs from #3581
- Rename of camel-cased test files from #3672
- Admonition styling
- `.. versionadded` styling
- Markdown format
- Example CLI consistency
- A rogue type lint

Before removing the `draft` status, review all changes in the pending release diff: https://github.com/pallets/click/compare/8.4.2...main

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@ Unreleased
 - Add built-in shell completion support for PowerShell (Windows PowerShell
   5.1+ and pwsh 7+) alongside the existing `bash`, `zsh`, and `fish`
   completers. Use `_FOO_BAR_COMPLETE=powershell_source foo-bar` to generate
-  the completion script.
+  the completion script. {issue}`2672` {pr}`3637`
 - Supported versions of Windows enable ANSI terminal styles by default.
   Colorama is no longer a dependency and is not used. {issue}`2986` {pr}`3505`
 - {class}`Argument` accepts a `help` parameter, and help output includes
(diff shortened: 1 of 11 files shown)


Rowlando13:
@kdeldycke I think we should leave test file names as is. The normal convention is test_module_name.py but since we are testing by function or class we should use test_ExactName.py. Then it is very clear what our structure is.

kdeldycke:
> @kdeldycke I think we should leave test file names as is. The normal convention is test_module_name.py but since we are testing by function or class we should use test_ExactName.py. Then it is very clear what our structure is.

`test_CamelCase.py` and `test__private_function.py` looks ugly to me, so let's ask @davidism to make a call on that naming convention, and I will apply it! 😅 My own rationale is it introduce less surprises when invoking pytest with wildcards like `pytest "tests/test_stuff_*"`.

kdeldycke:
@Rowlando13 @davidism , unless there is something big I uncover in the next few days, this is my last PR for the 8.5.0 release. After this if you think it is ready, we can think about releasing 8.5.0. What do you think?

Rowlando13:
Okay. I think 8.5.0 is about the right size. The only thing left I want to do is move all the tests into separate files so when we merge main to stable it will be done.

Rowlando13:
If we can leave the renaming of test files alone, I will merge this so we are all ready as splitting test files does not need a change log.

kdeldycke:
> If we can leave the renaming of test files alone, I will merge this so we are all ready as splitting test files does not need a change log.

I just reverted the test filenames to your original version in https://github.com/pallets/click/compare/4a735fa183dbc44b6d4c89e26b946a432ed7df58..0f4738df88e3ea47c40a4a442103596a61cfee79 . There is no point to delay the work on 8.5.0 for this kind of bikesheding.

Feel free to merge! :)

Rowlando13 (approved):

PR #3695, opened 2026-07-16
Mark clearly functions private status. Deprecated Python 2 utilities.

Rowlando13:
Closes #3099 and #3481

PR #3704, opened 2026-07-17
Deprecate `isolated_filesystem` and document its limits

### LINKED ISSUE #3501: `isolated_filesystem()` is not thread safe

### LINKED ISSUE #3700: Deprecate `isolated_filesystem()`
As @davidism pointed out in https://github.com/pallets/click/issues/3123#issuecomment-3433922728 :

> Nowadays you can also use `pytest` directly for this, or use with `tempfile.TemporaryDirectory()`. I don't think we'd add `isolated_filesystem` today, I think it was added before Python 3 and `pytest` were in common use.

Relates to: https://github.com/pallets/click/issues/3501

kdeldycke:
This is an attempt to deprecate `isolated_filesystem` from the `CliRunner`. It has been discussed in #3501.

In addition to the deprecation, I added some explicit documentation about thread-safety as discussed in #3501. I used our own usage of `pytest-xdist` as a good example to follow.

All occurrences of `isolated_filesystem` in our own test suite have been removed and we are now relying on Pytest's `tmp_path`.

I also inaugurated the `docs/upgrade-guides.md` section introduced a while ago by @Rowlando13 .

This PR closes #3700.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,6 +43,14 @@ Unreleased
   `DeprecationWarning` until Click 9.0: `LazyFile`, `KeepOpenFile`,
   `make_default_short_help`, `PacifyFlushWrapper`, and `safecall`.
   {issue}`3099` {pr}`3695`
+- Deprecate {meth}`CliRunner.isolated_filesystem`. It relies on
+  {func}`os.chdir`, which mutates process-global state and is not
+  thread-safe. The helper predates Python 3 and modern pytest: use a
+  temporary directory ({class}`tempfile.TemporaryDirectory` or pytest's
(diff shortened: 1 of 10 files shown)


Rowlando13:
Just a nit. Can you add a test of the deprecation warning?

Rowlando13 (commented):
Looks good except nit.

kdeldycke:
> Just a nit. Can you add a test of the deprecation warning?

Ah yes, you probably missed it because it was in `tests/test_testing.py`. So I just applied [your convention from your other PR](https://github.com/pallets/click/pull/3695#discussion_r3603319679) and moved it to `tests/test_deprecations.py`.

Issue #3571, opened 2026-06-08
`click.progressbar` doesn't show full completion when using `show_pos=True` combined with `update_min_steps`


maarten-ic:
Using a `click.progressbar` with:
- `update_min_steps` which isn't a divisor of `length`
- `show_pos=True`

won't show full completion at the end. This can be reproduced as follows:

```python
import time

import click

with click.progressbar(
    range(20),
    show_pos=True,
    update_min_steps=7,
) as bar:
    for i in bar:
        time.sleep(0.1)
```

with (final) output in the terminal

```text
  [####################################]  14/20          
```

**Expected behaviour**: I had expected the output to show `20/20` at the end.
This would be consistent with the default percentage formatting, as can be seen by commenting the line `show_pos=True` and re-running the reproduction:

```text
  [####################################]  100%          
```

Environment:

- Python version: tested with 3.12.3 and 3.14.3
- Click version: 8.4.1

davidism:
If anyone wants to work on it, they're welcome to. We are not accepting AI generated PRs though.

PR #3720, opened 2026-07-23
AI junk

cf3901646:
Fixes #3571.

### Root Cause
When `update_min_steps > 1` is used with `show_pos=True`, trailing steps that do not reach `update_min_steps` remain buffered in `_completed_intervals`. When `finish()` is called, these buffered steps were not applied to `self.pos`, causing `format_pos()` to display an incomplete position (e.g. `14/20`).

### Fix
Flushes remaining `_completed_intervals` in `ProgressBar.finish()` by calling `make_step()` before setting `self.finished = True`. Added a regression unit test `test_progressbar_update_min_steps_show_pos` in `tests/test_termui.py`.

### DIFF
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -348,6 +348,9 @@ class ProgressBar(t.Generic[V]):
             self._completed_intervals = 0
 
     def finish(self) -> None:
+        if self._completed_intervals:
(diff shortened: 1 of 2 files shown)


Samielakkad (commented):
Retracted.

PR #3728, opened 2026-07-25
Sort help args to make them reproducible

jspricke:
get_help_option_names() uses a set which is not ordered.
The return value is used by mkdocs-click making the
mkdocs-doc package in Debian not reproducible:

https://reproduce.debian.net/all/unstable.html#mkdocs-doc

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -63,6 +63,8 @@ Unreleased
   {class}`ParamType` takes a second optional type parameter describing the
   input value it accepts (`ParamType[int, str]` for a type converting
   strings to integers), defaulting to `Any`. {pr}`3407`
+- {meth}`Command.get_help_option_names` returns the help option names in the
+  order they were declared. {pr}`3728`
 
 ## Version 8.4.2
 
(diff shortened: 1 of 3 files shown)


davidism:
I don't think the set or this simple sort are correct. We probably want to apply some specific order, such as long options then short options, or preserve the order they were originally specified. @kdeldycke I think you added this (and maybe I suggested the set in review, can't remember)? What do you think?

kdeldycke:
> I don't think the set or this simple sort are correct. We probably want to apply some specific order, such as long options then short options, or preserve the order they were originally specified. @kdeldycke I think you added this (and maybe I suggested the set in review, can't remember)? What do you think?

Ah yes, the `sort` proposed by @jspricke is the right idea. We want stability and predictability. And the `set`-based arithmetics implemented here 12 years ago (see c2e187b7361a8d7bc2ddab27c159132b94722ef9) is sound, but does not preserved order.

So the fix is to replicate the arithmetics, but with order preserving lists/tuples. It sounds like something I solved 2 years ago in #2811 / 70c673d37eb91ba42a129be9037caf3ebed62f3e when I added a `test_iter_params_for_processing` parametrized test to demonstrate the mangling of invocation and declaration orders.

I will amend this PR with something in the same spirit to speed up the process and have it reach the upcoming 8.5.0 release.

kdeldycke:
Just made the changes, this PR is ready to be merged.

PR #3729, opened 2026-07-25
AI junk

genrichez:
## Fix: progressbar not showing full completion with `update_min_steps`

Fixes #3571.

### Problem

When using `click.progressbar` with `show_pos=True` and an `update_min_steps` that isn't a divisor of the iterable length, the final display shows an incomplete position (e.g., `14/20` instead of `20/20` ).

### Root Cause

`finish()` marks the bar as complete but does not flush the remaining `_completed_intervals` accumulated since the last render. The `pos` attribute stays at the last rendered value rather than the actual total.

### Fix

Flush any remaining `_completed_intervals` via `make_step()` at the start of `finish()`, before setting `finished = True`. This is a no-op when `_completed_intervals` is already 0 (i.e., when `update_min_steps` evenly divides the length).

### Testing

- Full existing test suite: 1940 passed, 0 failures
- Added 2 regression tests covering the flush case and the no-remainder case

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 Unreleased
 
+- Fix `progressbar` not showing full completion when `update_min_steps`
+  is not a divisor of the iterable length. `finish()` now flushes
+  remaining accumulated steps so that `show_pos` displays the correct
+  final count. {issue}`3571`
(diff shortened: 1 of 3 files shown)


davidism:

PR #3730, opened 2026-07-27
AI junk

n0liu:
Fixes #3571

### The bug

When `update_min_steps` does not divide the length of the iterable and `show_pos=True`, the progress bar ends showing a stale count such as `18/20` even though the bar and percentage are at 100%.

`ProgressBar.update()` buffers steps in `self._completed_intervals` and only advances `self.pos` (via `make_step`) once the buffer reaches `update_min_steps`. The trailing steps that never reach the threshold stay in the buffer. When the iterator is exhausted, `generator()` calls `finish()`, which sets `self.finished = True` but never flushes that buffer. Since `pct` returns `1.0` once `finished`, the bar and percentage read 100% while `show_pos` renders the stale `self.pos` — a self-contradictory line.

Reproduced against `main` (`update_min_steps=3`, length `20`): after the loop `pos == 18` with `2` steps still buffered, and `finish()` leaves `pos == 18`.

### The fix

Flush the buffered `_completed_intervals` into `self.pos` in `finish()`, so a fully consumed iterator reports its final position. `finish()` is only called when the iterator is exhausted (end of `generator()`); early exit via the context manager goes through `render_finish()`, so this does not falsely complete a bar that was abandoned.

### Tests

Added `test_progressbar_finish_reports_full_position` in `tests/test_termui.py`: iterating a bar with `update_min_steps=3` over length `20` and `show_pos=True` now ends with `progress.pos == progress.length`. It fails on `main` (`assert 18 == 20`) and passes with the change. `tests/test_termui.py` is green (222 passed), and `ruff check` / `ruff format --check` are clean.

I'll add a `CHANGES.md` entry once this PR has a number.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,9 @@ Unreleased
   This stripping was lost in `8.4.0` when {pr}`2969` began writing the
   prompt with `input()` directly. {issue}`3572` {pr}`3653`
 - Fix test failures when using pytest >= 9.1. {pr}`3656`
+- `progressbar()` with `show_pos=True` shows the final position when
+  `update_min_steps` does not divide the length, instead of a stale count
+  below the total. {issue}`3571` {pr}`3730`
 - {class}`Path` with `allow_dash=True` no longer triggers a `BytesWarning`,
   an error under `python -bb`, when checking a value against the `-`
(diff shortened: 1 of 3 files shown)


davidism:

PR #3733, opened 2026-07-29
AI junk

Sasireddy001:
## Description

The Click documentation uses a custom Sphinx directive .. click:example:: that pre-imports click into the execution namespace when building the docs. However, the rendered source code blocks do not include import click.

This means users who copy the examples from the documentation will get NameError: name click is not defined unless they add the import themselves.

This change adds import click to each .. click:example:: and python example in docs/prompts.md so the snippets are self-contained.

## Verification

- Confirmed the click:example directive does not render import click automatically.
- Verified examples use @click decorators and click.echo/click.option without an explicit import.
- No open or merged PR was found addressing this.

### DIFF
--- a/docs/prompts.md
+++ b/docs/prompts.md
@@ -31,7 +31,7 @@ Example:
 ```{eval-rst}
 .. click:example::
 
-    @click.command()
+    import click\n    \n    @click.command()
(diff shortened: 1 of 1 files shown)


davidism:

PR #3736, opened 2026-07-30
AI junk

Zuhef:
The new PowerShell completer (#3637, unreleased in 8.5.0) forwards the command text verbatim and splits it Python-side with `split_arg_string`, which runs `shlex` in POSIX mode. In POSIX mode a backslash is an escape character, but in PowerShell it is an ordinary character — PowerShell escapes with a backtick. So every unquoted Windows path loses its separators before the completion logic ever sees it:

```pycon
>>> from click.shell_completion import split_arg_string
>>> split_arg_string(r"cli --config C:\Users\me\app.cfg run")
['cli', '--config', 'C:Usersmeapp.cfg', 'run']
```

That corrupts both values `PowerShellComplete.get_completion_args` returns. The `incomplete` value is the visible one — completing a path offers nothing at all, because the prefix being matched is not the prefix that was typed:

```pycon
>>> # user typed:  cli --config C:\Users\m<TAB>
>>> os.environ["COMP_WORDS"] = r"cli --config C:\Users\m"
>>> os.environ["COMP_CWORD"] = "2"
>>> PowerShellComplete(cli, {}, "cli", "_CLI_COMPLETE").get_completion_args()
(['--config'], 'C:Usersm')
```

A `shell_complete` callback filtering on `incomplete.startswith(...)` returns an empty list, so the user sees no completions. The preceding `args` are mangled the same way, so any completion that depends on an earlier path value is computed from the wrong value.

The quoted form happens to survive, since `shlex` keeps a backslash inside double quotes, which is probably why this was not noticed — but PowerShell does not quote paths for you as you type.

## Change

`split_arg_string` gains a keyword-only `escape` parameter, and `PowerShellComplete` passes `escape=False`. bash, zsh and fish keep POSIX splitting unchanged, which they need. I also corrected the comment above `_SOURCE_POWERSHELL`, which cited reuse of "the same shlex-based splitting used by the bash/zsh completers" as the reason for forwarding the text verbatim.

## Testing

Seven cases added to `tests/test_shell_completion.py`: the `escape=False` splitting (including quoted and multi-token forms), a guard that the default is still POSIX, that `get_completion_args` keeps backslashes in `args`, and an end-to-end completion of a Windows path value through the runner.

With `src/click/shell_completion.py` reverted, 6 of the 7 fail. The end-to-end one shows the user-visible symptom directly — it produces no completions at all:

```
FAILED tests/test_shell_completion.py::test_powershell_completes_windows_path_value
  assert '\n' == 'plain\nC:\\Users\\me\\app.cfg\n_\n'
```

Full suite on Windows with Python 3.13, plus the pre-commit linters:

```console
$ python -m pytest
1876 passed, 95 skipped, 1 xfailed

$ python -m ruff check src/click/shell_completion.py tests/test_shell_completion.py   # All checks passed!
$ python -m ruff format --check src/click/shell_completion.py tests/test_shell_completion.py   # 2 files already formatted
$ python -m mypy src/click/shell_completion.py   # Success: no issues found
```

I also confirmed the generated source script still parses in real PowerShell, via `[System.Management.Automation.Language.Parser]::ParseFile` on the output of `_CLI_COMPLETE=powershell_source`.

I did not add a `CHANGES.md` entry, since the completer itself is unreleased and its existing 8.5.0 bullet covers this. Happy to add one if you would rather have it recorded.

### DIFF
--- a/src/click/shell_completion.py
+++ b/src/click/shell_completion.py
@@ -528,11 +528,11 @@ class PowerShellComplete(ShellComplete):
     source_template: t.ClassVar[str] = _SOURCE_POWERSHELL
 
     def get_completion_args(self) -> tuple[list[str], str]:
-        # PowerShell does not use a backslash as an escape character, so
-        # splitting must leave it alone. Otherwise an unquoted Windows path
-        # such as C:\\Users\\me loses its separators and neither the
-        # incomplete value nor the preceding args match what was typed.
-        cwords = split_arg_string(os.environ["COMP_WORDS"], escape=False)
+        # PowerShell escapes with a backtick, not a backslash. Splitting with
+        # the default would eat the separators of an unquoted Windows path such
+        # as C:\\Users\\me, so neither the incomplete value nor the preceding
+        # args would match what was typed.
+        cwords = split_arg_string(os.environ["COMP_WORDS"], escape="`")
         cword = int(os.environ["COMP_CWORD"])
         args = cwords[1:cword]
 
@@ -604,7 +604,7 @@ def get_completion_class(shell: str) -> type[ShellComplete] | None:
     return _available_shells.get(shell)
 
 
-def split_arg_string(string: str, *, escape: bool = True) -> list[str]:
+def split_arg_string(string: str, *, escape: str = "\\") -> list[str]:
     """Split an argument string as with :func:`shlex.split`, but don't
     fail if the string is incomplete. Ignores a missing closing quote or
     incomplete escape sequence and uses the partial token as-is.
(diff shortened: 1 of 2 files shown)


Zuhef:
﻿Pushed a correction. My first attempt just turned escaping off, which fixed Windows paths but broke the other half: PowerShell escapes with a backtick, so `C:\my` file` was still being split in two and `a`"b` lost its quote.

The parameter now takes the escape character rather than a flag, and PowerShell passes a backtick. I checked the expectations against PowerShell's own parser (`[Parser]::ParseInput`) rather than guessing, and each case in the new test is a single command element there.

davidism:

Zuhef:
﻿Closing this myself: I missed that the Pallets contributing guide does not accept AI-generated contributions. That is on me for not checking the org-level guide before opening it, and I am not asking for a review here.

Leaving the finding in plain text in case it is useful to someone: `split_arg_string` splits `COMP_WORDS` with `shlex` in POSIX mode, so for the PowerShell completer a backslash is treated as an escape and an unquoted Windows path such as `C:\Users\me` loses its separators before completion runs. PowerShell escapes with a backtick, not a backslash.

PR #3737, opened 2026-07-30
AI junk

ghost:
### Summary of Changes
- Enhanced click.DateTime parameter type to support datetime.fromisoformat parsing fallback when custom format strings aren't matched.
- Enables seamless parsing of standard ISO 8601 strings (including timezone offsets like +00:00 or Z and subsecond resolution).
- Added unit test 	est_datetime_iso8601 in 	ests/test_types.py.

### Testing
- pytest tests/test_types.py (44 passed)
- mypy src/click/types.py (0 type errors)

### DIFF
--- a/src/click/types.py
+++ b/src/click/types.py
@@ -564,6 +564,12 @@ class DateTime(ParamType[datetime]):
             if converted is not None:
                 return converted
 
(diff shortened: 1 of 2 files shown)


davidism:

PR #3738, opened 2026-07-31
AI junk

moongoin:
Flushes pending throttled progress updates when a progress bar finishes.

When update_min_steps holds increments below the next refresh boundary, the final render can retain the previous position even after iteration is exhausted. finish() now flushes those pending increments before the bar is marked finished.

The regression test exercises a real command with show_pos=True and verifies that the final output reaches 20/20.

Fixes #3571


davidism:

PR #3739, opened 2026-07-31
Changed `tempfile.NamedTemporaryFile` mode to `w`

### LINKED ISSUE #3731: echo_via_pager on Windows fails with WinError 32, masking a TypeError from writing str to a binary temp file (8.4.0 regression)
On Windows, `click.echo_via_pager()` fails with a `PermissionError: [WinError 32]` naming a temp file. The reported error is cleanup noise masking the real one: a `TypeError` from writing `str` into a binary-mode temp file.

Regression introduced in 8.4.0 by the `get_pager_file` / `_pager_contextmanager` rewrite. Reproduced on 8.4.0, 8.4.1, 8.4.2, and the code shape is unchanged on `main`.

Windows-only. `_pipepager`'s `Popen(..., text=True).stdin` is a `TextIOWrapper` with a real `.buffer`, so POSIX takes a different path and is unaffected.

### Cause

`_tempfilepager` yields a `NamedTemporaryFile(mode="wb")` to the caller:

```python
f = tempfile.NamedTemporaryFile(mode="wb", delete=False)
try:
    yield t.cast(t.BinaryIO, f), encoding, color
```

`get_pager_file` then decides whether to wrap it for text writes:

```python
if _has_binary_buffer(stream):
    wrapper = MaybeStripAnsi(stream.buffer, color=color, encoding=encoding)
    stream = wrapper
```

`_has_binary_buffer` is `getattr(stream, "buffer", None) is not None`. A `NamedTemporaryFile` returns a `_TemporaryFileWrapper`, which has **no** `.buffer` attribute, so the check is False and the text wrapper is never installed. `echo_via_pager` then writes `str` directly to a binary handle:

```
TypeError: a bytes-like object is required, not 'str'
```

That exception unwinds into `_tempfilepager`'s cleanup:

```python
finally:
    os.unlink(f.name)
```

The handle is still open in the same process, and Windows will not unlink an open file, so `PermissionError` replaces the `TypeError` and becomes the only thing the user sees. Anyone hitting this will reasonably suspect antivirus or file locking rather than a type error.

Side effect of the same masking: when the paging path *does* succeed the file is closed first, so the `finally` is only reachable in a broken state.

### Reproduction

Needs a tty on stdin/stdout to reach `_tempfilepager`, so this forces the branch to stay runnable under captured stdio:

```python
import click
from click import _termui_impl as ti

ti.isatty = lambda stream: True
ti.WIN = True

click.echo_via_pager(f"line {i}\n" for i in range(50))
```

```
  File "click/termui.py", line 346, in echo_via_pager
    with get_pager_file(color=color) as pager:
  File "click/_termui_impl.py", line 633, in _tempfilepager
    os.unlink(f.name)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\...\Temp\tmpsswp_chu'
```

On a real Windows terminal, no monkeypatching needed — any output taller than the window reproduces it. Surfacing the masked error:

```python
from click import _termui_impl as ti
gen = ti._tempfilepager(["cat"], color=False)
f, enc, color = gen.__enter__()
f.write("line\n")   # TypeError: a bytes-like object is required, not 'str'
```

### Downstream impact

Found via [mycli](https://github.com/dbcli/mycli) 2.7.0, which pins `click~=8.4.2` and calls `echo_via_pager` whenever query output exceeds the terminal. Any `SHOW TABLES` on a database with more tables than terminal rows dies with WinError 32. Pinning `click==8.3.3` is a clean workaround, since the pre-8.4 `_tempfilepager` buffered and encoded internally rather than yielding the handle.

### Suggested fix

Wrap the temp file so it both accepts `str` and satisfies the `.buffer` check, and close before unlinking:

```diff
-    f = tempfile.NamedTemporaryFile(mode="wb", delete=False)
+    raw = tempfile.NamedTemporaryFile(mode="wb", delete=False)
+    filename = raw.name
+    f = io.TextIOWrapper(raw, encoding=encoding)
     try:
-        yield t.cast(t.BinaryIO, f), encoding, color
-        f.flush()
-        f.close()
-        subprocess.call([str(cmd_path), f.name])
+        yield f, encoding, color
+        f.flush()
+        f.close()
+        subprocess.call([str(cmd_path), filename])
     finally:
-        os.unlink(f.name)
+        if not f.closed:
+            f.close()
+        os.unlink(filename)
```

Capturing `filename` up front matters because `f.name` is not reachable through the wrapper the same way. Guarding `f.close()` in the `finally` is what stops a mid-write exception from turning into WinError 32 and hiding itself.

Verified locally: `str` writes are accepted, the pager subprocess receives the content, and the unlink succeeds.

Two alternatives if you'd rather not change the yielded type:

- Have `_has_binary_buffer` also accept objects exposing a `file` attribute (`_TemporaryFileWrapper.file`), keeping the yield binary.
- Use `tempfile.mkstemp()` + `open(filename, "w", encoding=...)`, which sidesteps the wrapper class entirely.

### Environment

- Windows 11, Python 3.11
- click 8.4.0 / 8.4.1 / 8.4.2 all affected; 8.3.3 not affected
- `PAGER` unset and set to `less` both reproduce; `less` present on PATH

### LINKED ISSUE #3732: echo_via_pager broken on Windows since 8.4.0
Since the 8.4.0 pager rewrite, echo_via_pager on Windows dies with PermissionError WinError 32. Real cause: the temp file _tempfilepager yields has no .buffer, so str gets written to a binary handle (TypeError), then the finally tries to unlink the still-open file and Windows refuses. 8.3.3 works.

### LINKED ISSUE #3740: [Bug] On Windows, get_pager_file returns a typing.BinaryIO object NOT typing.TextIO
Inside `src/click/_termui_impl.py`, the function `get_pager_file` delegates the creation of a tempfile to `_tempfilepager`, which returns a `BinaryIO` object of the opened file:

```python
f = tempfile.NamedTemporaryFile(mode="wb", delete=False)
try:
    yield t.cast(t.BinaryIO, f), encoding, color
...
```
Which is called by _pager_contextmanager in
```python
...
if pager_cmd_parts:
        if WIN:
            return _tempfilepager(pager_cmd_parts, color)
        return _pipepager(pager_cmd_parts, color)
...
    if WIN or sys.platform.startswith("os2"):
        return _tempfilepager(["more"], color)
    return _pipepager(["less"], color)
```
In the end `get_pager_file`
```python
def get_pager_file(color: bool | None = None) -> t.Generator[t.TextIO, None, None]:
    ...
    with _pager_contextmanager(color=color) as (stream, encoding, color):
    # Split streams by capabilities rather than the abstract TextIO /
    # BinaryIO annotations: buffered text streams can be unwrapped to bytes,
    # while other streams are yielded as-is.
    wrapper: MaybeStripAnsi | None = None
        if _has_binary_buffer(stream):
            # Text stream backed by a binary buffer.
            wrapper = MaybeStripAnsi(stream.buffer, color=color, encoding=encoding)
            stream = wrapper
        try:
            # Narrow the BinaryIO | TextIO union that _pager_contextmanager
            # yields; the caller writes text to the pager.
            yield t.cast(t.TextIO, stream)
        ...
```
In the above `yield t.cast(t.TextIO, stream)` stream object will be `BinaryIO`, contradicting the `TextIO` annotation of the return value, all because it was created with the mode `"wb"` earlier.

Then if a caller tries: `stream.write("some text")` it will raise an exception:
`TypeError: a bytes-like object is required, not 'str'`

FIX:
In `_tempfilepager`, create the temp file with `mode='w'`
```python
f = tempfile.NamedTemporaryFile(mode="w", delete=False)
```
I went ahead and created a pull request:
https://github.com/pallets/click/pull/3739

H-Sorkatti:
Inside `src/click/_termui_impl.py`, the function `get_pager_file` delegates the creation of a tempfile to `_tempfilepager`, which returns a `BinaryIO` object of the opened file:

```python
f = tempfile.NamedTemporaryFile(mode="wb", delete=False)
try:
    yield t.cast(t.BinaryIO, f), encoding, color
...
```
Which is called by _pager_contextmanager in
```python
...
if pager_cmd_parts:
        if WIN:
            return _tempfilepager(pager_cmd_parts, color)
        return _pipepager(pager_cmd_parts, color)
...
    if WIN or sys.platform.startswith("os2"):
        return _tempfilepager(["more"], color)
    return _pipepager(["less"], color)
```
In the end `get_pager_file`
```python
def get_pager_file(color: bool | None = None) -> t.Generator[t.TextIO, None, None]:
    ...
    with _pager_contextmanager(color=color) as (stream, encoding, color):
    # Split streams by capabilities rather than the abstract TextIO /
    # BinaryIO annotations: buffered text streams can be unwrapped to bytes,
    # while other streams are yielded as-is.
    wrapper: MaybeStripAnsi | None = None
        if _has_binary_buffer(stream):
            # Text stream backed by a binary buffer.
            wrapper = MaybeStripAnsi(stream.buffer, color=color, encoding=encoding)
            stream = wrapper
        try:
            # Narrow the BinaryIO | TextIO union that _pager_contextmanager
            # yields; the caller writes text to the pager.
            yield t.cast(t.TextIO, stream)
        ...
```
In the above `yield t.cast(t.TextIO, stream)` stream object will be `BinaryIO`, contradicting the `TextIO` annotation of the return value, all because it was created with the mode `"wb"` earlier.

Then if a caller tries: `stream.write("some text")` it will raise an exception:
`TypeError: a bytes-like object is required, not 'str'`

FIX:
In `_tempfilepager`, create the temp file with `mode='w'`
```python
f = tempfile.NamedTemporaryFile(mode="w", delete=False)
```

Closes #3740

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -65,6 +65,11 @@ Unreleased
   strings to integers), defaulting to `Any`. {pr}`3407`
 - {meth}`Command.get_help_option_names` returns the help option names in the
   order they were declared. {pr}`3728`
+- {func}`get_pager_file` yields a text stream on Windows again. The temporary
+  file backend opened its file in binary mode, so writing a `str` to the pager
+  raised `TypeError: a bytes-like object is required, not 'str'`, and the
+  `color` argument was ignored on that path. Regression introduced in `8.4.0`
+  by {pr}`1572`. {issue}`3731` {issue}`3732` {issue}`3740` {pr}`3739`
 
 ## Version 8.4.2
 
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -623,7 +623,7 @@ def _tempfilepager(
     # On Windows, NamedTemporaryFile cannot be opened by another process
     # while Python still has it open, so we use delete=False and clean up manually
     # rather than using a contextmanager here.
-    f = tempfile.NamedTemporaryFile(mode="wb", delete=False)
+    f = tempfile.NamedTemporaryFile(mode="w", delete=False)
     try:
         yield t.cast(t.BinaryIO, f), encoding, color
         f.flush()
--- a/tests/test_termui.py
+++ b/tests/test_termui.py
@@ -936,6 +936,77 @@ def test_get_pager_file_flushes_stream_on_exception(monkeypatch):
     assert stream.flush_calls == 1
 
 
+def _force_tempfile_pager(monkeypatch, pager_cmd="cat"):
+    """Route ``get_pager_file`` through the ``_tempfilepager`` backend.
+
+    That backend is only reachable on Windows, so the platform flag and the tty
+    probes are faked to exercise it from any runner.
+    """
+    cmd_path = shutil.which(pager_cmd)
+    assert cmd_path is not None, f"{pager_cmd} not available"
+    monkeypatch.setattr(click._termui_impl, "isatty", lambda _: True)
+    monkeypatch.setattr(click._termui_impl, "WIN", True)
+    monkeypatch.setitem(click._termui_impl.os.environ, "PAGER", cmd_path)
+
+
+def _page_with_get_pager_file():
+    with click.get_pager_file() as pager:
+        pager.write("hello\n")
+
+
+def _page_with_echo_via_pager():
+    click.echo_via_pager("hello")
+
+
+@pytest.mark.skipif(shutil.which("cat") is None, reason="cat not available")
+@pytest.mark.parametrize(
+    "page",
+    [
+        pytest.param(_page_with_get_pager_file, id="get_pager_file"),
+        pytest.param(_page_with_echo_via_pager, id="echo_via_pager"),
+    ],
+)
+def test_tempfile_pager_accepts_text(monkeypatch, capfd, page):
+    """The temp file backend must yield a text stream (issues #3731, #3740).
+
+    ``_tempfilepager`` opens its temp file in binary mode and yields the
+    ``_TemporaryFileWrapper`` straight to ``get_pager_file``. That wrapper
+    exposes no ``.buffer``, so the ``_has_binary_buffer`` probe is False and no
+    ``MaybeStripAnsi`` wrapper is installed. The caller then writes ``str`` to a
+    binary handle and gets ``TypeError: a bytes-like object is required``.
+    """
+    _force_tempfile_pager(monkeypatch)
+
+    page()
+
+    out, err = capfd.readouterr()
+    assert err == ""
+    assert out.replace("\r\n", "\n") == "hello\n"
+
+
+@pytest.mark.skipif(shutil.which("cat") is None, reason="cat not available")
+@pytest.mark.parametrize(
+    ("color", "expected"),
+    [
+        pytest.param(False, "hello\n", id="strip ansi"),
+        pytest.param(True, click.style("hello", fg="red") + "\n", id="preserve ansi"),
+    ],
+)
+def test_tempfile_pager_handles_ansi(monkeypatch, capfd, color, expected):
+    """The temp file backend honors ``color`` like the pipe backend does.
+
+    ANSI stripping lives in ``MaybeStripAnsi``, which the binary temp file
+    never gets wrapped in, so ``color`` is silently ignored on this path.
+    """
+    _force_tempfile_pager(monkeypatch)
+
+    click.echo_via_pager(click.style("hello", fg="red"), color=color)
+
+    out, err = capfd.readouterr()
+    assert err == ""
+    assert out.replace("\r\n", "\n") == expected
+
+
 def test_editor_unclosed_quote():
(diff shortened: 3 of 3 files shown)


H-Sorkatti:
@kdeldycke 
I'm not AI by the way. This a genuine human fix. But of course, it's your library.

kdeldycke:
> @kdeldycke I'm not AI by the way. This a genuine human fix. But of course, it's your library.

Ah. Sorry for the confusion. Maintainers are overloaded with noise these past few months so I wrongly flagged that. Let me review your proposal then.

kdeldycke:
OK this is confirmed as a regression from https://github.com/pallets/click/pull/1572 . And it was not caught because `_tempfilepager` has 0 test coverage.

kdeldycke:
Just added tests to cover the regression and added a changelog entry. Also added test for color handling. This PR is ready to be merged upstream.

Rowlando13 (approved):

PR #3741, opened 2026-08-01
AI junk

AJ-ing:
Flush leftover _completed_intervals in finish() so the final position matches a completed bar. Fixes #3571.


davidism:

PR #3743, opened 2026-08-02
AI junk

rudrakumar07:
Fixes #2847.

When a long option uses = to separate option from value (e.g. --color=value), shells replace the entire current word with the completion. Without tracking the option prefix, the completion replaces --color= with just auto, losing the option name.

This detects the = separator in get_completions(), saves the prefix (--color=), and prepends it to each completion item so the shell replaces --color= with --color=auto as expected.

Tested with bash and zsh completion scenarios. All existing tests pass.


davidism:

PR #3745, opened 2026-08-02
AI junk

fyrsta7:
## Summary

This caches whether an internal parser `_Option` takes a value when the option object is created, instead of recomputing the same `nargs` comparison through a property on every access.

## Why this can improve performance

The parser consults `takes_value` repeatedly while processing command-line tokens. The value is derived only from `_Option.nargs`, which is already fixed during `_Option` construction, so storing the boolean directly removes repeated property calls on the parsing path.

## Validation

- `git diff --check`
- `python3 -m compileall -q src/click/parser.py`
- Behavior check for `_Option.takes_value` with flag-like and value-taking options

## Benchmark

Current machine, CPython microbenchmark, 9 samples, median:

| Benchmark | Base | This change | Delta |
|---|---:|---:|---:|
| Repeated `_Option.takes_value` checks | 21,147.820 ns/op | 7,140.603 ns/op | 66.23% faster |

### DIFF
--- a/src/click/parser.py
+++ b/src/click/parser.py
@@ -158,13 +158,10 @@ class _Option:
 
         self.dest = dest
         self.action = action
+        self.takes_value = action in ("store", "append")
         self.nargs = nargs
         self.const = const
         self.obj = obj
-
-    @property
-    def takes_value(self) -> bool:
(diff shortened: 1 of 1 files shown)


davidism:

PR #3748, opened 2026-08-05
AI junk

bunlongheng:
## Summary

Fixes #3571.

When `update_min_steps > 1` and `length % update_min_steps != 0`, the final
batch of items may not meet the threshold, so `_completed_intervals` is left
non-zero when the iteration ends. `finish()` set `self.finished = True` but
never flushed those pending intervals into `self.pos`. The progress bar then
rendered with `pos < length` (e.g. `14/20` instead of `20/20`) even though
the bar itself appeared full.

## Reproduction

```python
import click

with click.progressbar(range(20), show_pos=True, update_min_steps=7) as bar:
    for i in bar:
        pass
# Before fix: [####################################]  14/20
# After fix:  [####################################]  20/20
```

## Fix

Flush any remaining `_completed_intervals` via `make_step()` at the start of
`finish()`, before marking the bar complete - mirroring what `render_progress()`
already does on each intermediate step.

```diff
 def finish(self) -> None:
+    if self._completed_intervals > 0:
+        self.make_step(self._completed_intervals)
+        self._completed_intervals = 0
+
     self.eta_known = False
     self.current_item = None
     self.finished = True
```

All 1939 existing tests pass with this change.

### DIFF
--- a/src/click/_termui_impl.py
+++ b/src/click/_termui_impl.py
@@ -348,6 +348,10 @@ class ProgressBar(t.Generic[V]):
             self._completed_intervals = 0
 
     def finish(self) -> None:
+        if self._completed_intervals > 0:
+            self.make_step(self._completed_intervals)
+            self._completed_intervals = 0
+
         self.eta_known = False
         self.current_item = None
         self.finished = True


davidism:

PR #3751, opened 2026-08-08
AI junk

### LINKED ISSUE #2740: Display exception notes when raising `ClickException`

TheRodzz:
Fixes #2740

### DIFF
--- a/src/click/exceptions.py
+++ b/src/click/exceptions.py
@@ -63,6 +63,9 @@ class ClickException(Exception):
             file=file,
             color=self.show_color,
         )
+        if hasattr(self, "__notes__"):
+            for note in self.__notes__:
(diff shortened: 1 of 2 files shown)


davidism:

PR #3757, opened 2026-08-11
AI junk

pcruz1905:
`textwrap` breaks on hyphens by default, so an option name that happens to
reach the wrap column gets split in the middle of itself:

```
               --force-overwrite-existing --network-timeout-
               seconds --debug-trace-enabled
```

A usage line is a list of option and metavar tokens rather than prose, so
there is never a useful place to break inside one. Wrapping it with
`break_on_hyphens=False` moves the whole token to the next line instead:

```
```

Both branches of `write_usage` are covered: arguments placed beside the
prefix, and arguments wrapped onto their own line when the prefix is too
long to share.

The ticket also notes there is currently no way to influence this short of
reimplementing `wrap_text` and `write_usage` in a subclass, so `wrap_text`
takes a `break_on_hyphens` argument. It defaults to `True`, leaving prose
wrapping in help text and epilogs unchanged; there is a test covering that
so the default does not drift.

fixes #3362

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -65,6 +65,11 @@ Unreleased
   strings to integers), defaulting to `Any`. {pr}`3407`
 - {meth}`Command.get_help_option_names` returns the help option names in the
   order they were declared. {pr}`3728`
+- {meth}`HelpFormatter.write_usage` no longer splits an option name at one of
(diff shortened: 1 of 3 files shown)


davidism:

Issue #2676, opened 2024-02-13
Support for Automatically Parsing Negative Number Arguments


thomask77:
I would like to re-open the discussion from issue #555 and #1624.

My program makes heavy use of numeric arguments, including negative numbers, special float values (NaN, Infinity..), negative hex numbers, etc.

At the moment I'm using the following monkey patch as a work-around:

```python
from click.parser import OptionParser, ParsingState


def _is_number(arg: str) -> bool:
    try:
        _ = int(arg, 0)
    except ValueError:
        try:
            _ = float(arg)
        except ValueError:
            return False
    return True


# Allow negative numbers as arguments
#
# Based on click/parser.py v8.1.7
#
def _process_args_for_options(self, state: ParsingState) -> None:
    while state.rargs:
        arg = state.rargs.pop(0)
        if arg == "--":
            return
        elif arg[:1] in self._opt_prefixes and len(arg) > 1 and not _is_number(arg):
            self._process_opts(arg, state)
        elif self.allow_interspersed_args:
            state.largs.append(arg)
        else:
            state.rargs.insert(0, arg)
            return


# Monkey-patch the OptionParser class
#
OptionParser._process_args_for_options = _process_args_for_options
```

It would be nice, if click would offer an official mode that accepts negative numbers.
I can also submit a PR, but would need further input on how to properly integrate the feature.

Rowlando13:
Based on looking at the previous issues, the issue results from an argument(-1 or --1)  looking like options, and so get parsed that way. In the documentation there are clear work arounds such as pass ``--``  or passing an ignore_unknown_options to the context settings. I am not sure that its worth the parsing complexity. Possibly for integers and floats. @AndreasBackx Thoughts? 

https://click.palletsprojects.com/en/stable/arguments/#option-like-arguments

thomask77:
We are using click in a power-electronics control application, and it's absolutely critical.

There's just no way, we can explain to our customers that they have to use "--" when they want to output a negative voltage.

The additional complexity is minimal, as shown by the monkey patch above.

AndreasBackx:
I think it's something that should be possible indeed. Though I'm unsure if the proposed solution is ideal. Clap has solved this in a different way which also allows for non-number values that start with dashes: https://github.com/clap-rs/clap/issues/1245. We could possibly explore a similar solution.

I think for now it's good there's a workaround for those who need it. @davidism, would the parser rewrite of yours possible be able to accommodate this more easily?

[...]

PR #3758, opened 2026-08-12
AI junk

dallascrilley:
Addresses the docs discussion in #2676.

Adds a "Negative Number Arguments" section to the arguments docs, directly after the existing `ignore_unknown_options` example it builds on. It covers:

- why `-6` is rejected before the parameter's type is consulted, shown with a runnable example
- the `--` separator, connected back to the section above it
- the `ignore_unknown_options` recipe for commands whose arguments are routinely negative numbers
- the caveat that comes with it: unknown options become arguments, so mistyped or misplaced flags arrive as values, and string-typed arguments accept them silently. The section closes with the practical guidance to validate values when using this pattern.

I hit this exact problem building a mixer-control CLI (fader values like `-6`) and shipped the `ignore_unknown_options`-plus-validation pattern there, including learning the caveat the hard way: a misplaced `--json` was silently accepted as a string value until validation was added. This section is the write-up I wish the docs had had.

No parser changes, per the maintainer discussion in #2676 (the feature itself is deferred to the parser rewrite). The examples are `click:run` blocks, so their output is generated by executing the code at docs build time; `sphinx-build -E -W -b dirhtml docs docs/_build/dirhtml` passes locally with no warnings. Lines are wrapped at 80 characters. No CHANGES entry since this is docs-only; happy to add one if wanted.

### DIFF
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -138,6 +138,59 @@ And from the command line:
     invoke(touch, ['-foo.txt', 'bar.txt'])
 ```
 
+## Negative Number Arguments
+
+Negative numbers hit the same rule: the parser decides what is an
+option before a parameter's type is consulted, so a value like `-6`
+is rejected as an unknown option even on a `FLOAT` argument:
+
+```{eval-rst}
+.. click:example::
+
+    @click.command()
+    @click.argument('volume', type=float)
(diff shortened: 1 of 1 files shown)


davidism:

PR #3763, opened 2026-08-13
AI junk

rawsun007:
fixes #3362

`write_usage` wraps through `wrap_text`, which builds a `TextWrapper` with textwrap's default `break_on_hyphens=True`. On a narrow line that splits an option after any internal hyphen:

```
```

Neither half of `--max-` / `retry-count` is copyable, and at a glance it reads as two separate options. After:

```
               --auto-update-interval
```

## Approach

Hyphen breaking is right for prose and wrong for a line made of tokens that must stay whole. Rather than hardcode that inside `write_usage`, `wrap_text()` gains a `break_on_hyphens` parameter defaulting to `True` — help text and every other caller are unchanged — and `write_usage` passes `False`.

Exposing it also answers the issue's second point: someone hitting this had no way to influence the wrapper short of reimplementing `write_usage` in a subclass.

## Both call sites

`write_usage` has two `wrap_text` calls and only one runs per invocation. The second is reached when the prefix is too long for the args to sit beside it:

```python
if text_width >= (term_len(usage_prefix) + 20):
    ...   # args to the right of the prefix
else:
    ...   # args on the next line
```

Fixing only the first leaves the bug live behind a long program name, so both pass `break_on_hyphens=False`.

## Tests

Three, each failing on `main`:

| Test | Pins |
| --- | --- |
| `test_write_usage_does_not_break_options_at_hyphens` | the reported case, asserting the exact three output lines |
| `test_write_usage_does_not_break_hyphens_on_the_wrapped_prefix_path` | the long-prefix branch |
| `test_wrap_text_still_breaks_hyphens_by_default` | prose is unaffected; only usage lines opt out |

The second one is worth a note. My first version passed on `main`, because a single short option fits on one line and never reaches the wrapping code in that branch at all — it looked like coverage while testing nothing. It now passes three options so the args genuinely have to wrap inside that branch, and it fails on `main` like the others.

The third exists because the obvious wrong fix is to set `break_on_hyphens=False` inside `wrap_text` itself, which would silently change every help paragraph in every Click app.

```
$ pytest -q
1955 passed, 25 skipped, 1 xfailed

$ ruff check src tests    All checks passed!
$ ruff format --check     63 files already formatted
```

`CHANGES.md` entry under 8.5.0, plus `.. versionchanged:: 8.4.0` notes on `wrap_text` and `write_usage`. Say the word if those should read 8.5.0 instead — the file's existing `versionchanged` markers point at 8.4.0 while the changelog is on 8.5.0, and I was not sure which is current.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,11 @@
 
 Unreleased
 
+- `HelpFormatter.write_usage` no longer splits an argument after a hyphen
+  when the usage line wraps, so `--enable-verbose-logging` stays whole
+  instead of becoming `--enable-verbose-` / `logging`. `wrap_text()` takes a
+  new `break_on_hyphens` parameter, defaulting to `True`, for callers that
+  need the same control. {issue}`3362`
 - Add built-in shell completion support for PowerShell (Windows PowerShell
   5.1+ and pwsh 7+) alongside the existing `bash`, `zsh`, and `fish`
   completers. Use `_FOO_BAR_COMPLETE=powershell_source foo-bar` to generate
--- a/src/click/formatting.py
+++ b/src/click/formatting.py
@@ -34,6 +34,7 @@ def wrap_text(
     initial_indent: str = "",
     subsequent_indent: str = "",
     preserve_paragraphs: bool = False,
+    break_on_hyphens: bool = True,
 ) -> str:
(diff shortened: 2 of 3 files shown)


davidism:

PR #3765, opened 2026-08-14
AI junk

### LINKED ISSUE #2184: Click option completion callback doesn't have access to context.params for parsed values of already typed arguments.
Click option completion callback doesn't have access to context.params for parsed values of already typed arguments during shell completion:

```python
import click

def test_completer(ctx, param, incomplete):
    import sys
    print(f"1: {param= }", file=sys.stderr)
    print(f"2: {incomplete= }", file=sys.stderr)
    print(f"3: {ctx.params = }", file=sys.stderr)
    return []

@click.command()
@click.argument('type', shell_complete=test_completer)
@click.option('--test', shell_complete=test_completer)
def app(type, count):
    pass
```

Output:
```
app foo --test [TAB] [TAB]
1: param= <Option test>
2: incomplete= ''
3: ctx.params = {'type': None, 'test': None}
```
Thus the context inside option callback does not have access to already typed argument.

But vice versa works:
```
app --test bar [TAB] [TAB] 
1: param= <Argument type>
2: incomplete= ''
3: ctx.params = {'test': 'bar', 'type': None}
```
The context inside argument callback have access to already typed option.

I think it should be the same in both cases.

Environment:

- Python version: 3.8
- Click version: 8.0.3

fff122:
Fixes #2184

## Summary

In resilient parsing mode, `_OptionParser.parse_args` stopped after `_process_args_for_options` raised a `UsageError`. As a result, positional arguments that had already been identified were not processed before shell completion callbacks inspected `ctx.params`.

This change continues with positional-argument processing when resilient parsing is enabled, while preserving the existing error behavior for normal parsing. The regression test verifies that a shell-completion callback sees the parsed positional argument.

## Validation

- `PYTHONPATH=src python3 -m pytest -q` — 1954 passed, 24 skipped, 1 xfailed
- Ruff check passed
- `git diff --check` passed

### DIFF
--- a/src/click/parser.py
+++ b/src/click/parser.py
@@ -307,10 +307,16 @@ class _OptionParser:
         state = _ParsingState(args)
         try:
             self._process_args_for_options(state)
+        except UsageError:
+            if self.ctx is None or not self.ctx.resilient_parsing:
+                raise
+
+        try:
             self._process_args_for_args(state)
         except UsageError:
             if self.ctx is None or not self.ctx.resilient_parsing:
                 raise
+
         return state.opts, state.largs, state.order
 
     def _process_args_for_args(self, state: _ParsingState) -> None:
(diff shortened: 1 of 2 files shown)


davidism:

PR #3771, opened 2026-08-16
AI junk

Hashim1999164:
write_usage wraps args with wrap_text. The wrapper inherited textwrap hyphen breaks, so tokens such as option names were split across lines.

Disable hyphen breaks for usage args so each whitespace separated token stays together.

Fixes #3362


davidism:

Hashim1999164:
not generated by llm, just rephrased my pr message because of weak english and searched gpt for possible fixes, but okay

PR #3772, opened 2026-08-16
AI junk

maharshi-coding:
Fixes #3362.

## The problem

`HelpFormatter.write_usage` wraps the argument list with `textwrap`, whose default `break_on_hyphens=True` splits words at hyphens. When an option name falls at the wrap boundary it is broken mid-name:

```
```

## The change

- Add a `break_on_hyphens` parameter to `wrap_text` (default `True`, so existing behavior — including help-text wrapping — is unchanged).
- Pass `break_on_hyphens=False` from `HelpFormatter.write_usage`, so option names stay intact on the usage line.

The reproduction from the issue now produces the expected output, with option names kept whole:

```
```

## Tests

- Added a regression test in `tests/test_formatting.py`.
- Full local test suite passes (`1899 passed, 95 skipped`).

---

*Disclosure: this contribution was prepared with AI assistance; I reviewed the change and verified the full test run before submitting.*

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -74,6 +74,9 @@ Unreleased
   does not divide the total. Steps below that threshold are applied when the
   bar finishes, so `show_pos` renders `20/20` rather than the last multiple
   it reached. {issue}`3571` {pr}`3769`
+- `HelpFormatter.write_usage()` no longer splits option names containing
+  hyphens across lines; e.g. `--max-retry-count` stays intact. `wrap_text()`
+  gained a `break_on_hyphens` parameter. {issue}`3362` {pr}`3772`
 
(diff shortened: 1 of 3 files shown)


ThiefMaster:

> If you are a new contributor, do not use LLM or AI tools to contribute at all; you do not have the trust or experience to take responsibility for what is generated. If you are someone who is familiar to the maintainers, and know that they already trust you to make high quality contributions, your contribution must still genuinely be understood and created personally by you, regardless of what tools you may use to assist you.

maharshi-coding:
Withdrawing this — it was an unsolicited AI-assisted PR that I should not have opened. Apologies for the noise.

PR #3774, opened 2026-08-17
AI junk

### LINKED ISSUE #2869: Support pathlib.Path in edit filename

JiuYue0820:
``click.edit(filename=...)`` only treated a ``str`` as a single file.
``pathlib.Path`` is not a ``str``. On Python 3.12+ it is also not
iterable, so the call raises ``TypeError: 'WindowsPath' object is not
iterable``. On earlier versions it iterated path parts and opened each
segment as a separate file.

This matches ``open_file`` and the existing ``click.Path`` param type,
which already accept ``os.PathLike``.

fixes #2869

- Tests cover a single ``Path`` and an iterable of ``Path`` values.
- ``CHANGES.md`` (the changelog file in this repo) has an 8.5.0 entry.
- Docstring notes the 8.5.0 PathLike change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 Unreleased
 
+- {func}`edit` accepts :class:`os.PathLike` values, including
+  :class:`pathlib.Path`, for ``filename``. Passing a path object used to
+  raise ``TypeError`` because ``Path`` is neither a ``str`` nor an
+  iterable of strings. {issue}`2869`
 - Add built-in shell completion support for PowerShell (Windows PowerShell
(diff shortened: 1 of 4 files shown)


davidism:

PR #3775, opened 2026-08-17
AI junk

### LINKED ISSUE #3652: Automatically append ellipsis (`...`) to metavars when `multiple=True` in options
When building a command-line interface with `click.option(..., multiple=True)`, the auto-generated usage string does not visually signal to the user that the option can be repeated. For instance, an option --foo=FOO looks identical whether it accepts single or multiple arguments. This is different from the common convention for multi-value items in shell interfaces often suggests adding an ellipsis like `--foo=FOO...`

Expected behavior if option "foo" has `multiple=True`:
```
Usage: script.py [OPTIONS]

Options:
  --foo TEXT...  A list of foo strings.
```

karandhaodiyal28-hash:
Fixes #3652

## Problem
When an option has `multiple=True`, the help text shows a single metavar like
`--file FILENAME` instead of the conventional `...` indication that the option
accepts multiple values. The current output is misleading for users.

Example (before):
```
--file FILENAME
--tag TAG
```

Example (after):
```
--file FILENAME...
--tag TAG...
```

## Fix
Append `...` to the rendered metavar inside `_write_opts` when the option
has `multiple=True`. The change is purely cosmetic — it only affects help
text rendering, not option parsing. nargs>1 already implies multiple in click,
so options like `nargs=2` get the suffix too.

## Tests
Added 5 tests in `tests/test_options.py`:
- `test_multiple_metavar_has_ellipsis` — basic multiple=True option
- `test_metavar_default_no_ellipsis` — default (no multiple) is unchanged
- `test_metavar_explicit_no_ellipsis` — explicit metavar without multiple
- `test_metavar_nargs_multiple_ellipsis` — nargs=2 (auto-multiple) shows ellipsis
- `test_metavar_with_short_option_ellipsis` — short and long both show ellipsis

All 5 pass. Full test suite: **1899 passed, 0 failed, 99 skipped, 1 xfailed** in
8.13s.

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -3350,7 +3350,10 @@ class Option(Parameter):
                 any_prefix_is_slash = True
 
             if not self.is_flag and not self.count:
-                rv += f" {self.make_metavar(ctx=ctx)}"
+                metavar = self.make_metavar(ctx=ctx)
+                if self.multiple:
+                    metavar += "..."
+                rv += f" {metavar}"
 
             return rv
 
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -72,6 +72,68 @@ def test_deprecated_empty_help_no_leading_space(help_text, deprecated, expected)
     assert opt.get_help_record(ctx)[1] == expected
 
 
+# See #3652
+@pytest.mark.parametrize("metavar", [None, "FOO"])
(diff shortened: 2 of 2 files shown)


davidism:

PR #3778, opened 2026-08-19
Add FAQ entry about `UnicodeEncodeError` on Windows

### LINKED ISSUE #2121: `UnicodeEncodeError` on Windows when there are Unicode chars in the help message
I have come across an error when I try to print the help message for my app (`--help`) on Windows (using bash, cmd, and powershell). My help message has unicode characters in it (the project name) which is what seems to be causing the problem:

* https://github.com/leouieda/nene/blob/main/nene/cli.py#L72

This PR tests running the app with `--help` and it fails on Windows and Python 3.6 and 3.10: https://github.com/leouieda/nene/pull/12

Here is a minimum example that fails:

```python
# example.py
import click

@click.command(context_settings={"help_option_names": ["-h", "--help"]})
def main():
    """
    App description with Unicode ‣
    """
    pass

if __name__ == '__main__':
    main()
```

```
$ python example.py -h
  File "example.py", line 11, in <module>
    main()
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\click\core.py", line 1128, in __call__
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\click\core.py", line 1052, in main
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\click\core.py", line 914, in make_context
    self.parse_args(ctx, args)
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\click\core.py", line 1370, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\click\core.py", line 2347, in handle_parse_result
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\click\core.py", line 2309, in process_value
    value = self.callback(ctx, self, value)
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\click\core.py", line 1270, in show_help
    echo(ctx.get_help(), color=ctx.color)
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\click\utils.py", line 298, in echo
  File "C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u2023' in position 62: character maps to <undefined>
```
I can confirm that it's the Unicode characters in the docstring of the function wrapped with the main `@click.command` that causes the issue. Removing them fixes the problem (the second CI run on https://github.com/leouieda/nene/pull/12). This issue does not happen on Linux and Mac.

For now, I'll remove the unicode characters so I'm not pushing a broken package but it would be great to be able to include the proper spelling of the package name in the future.

Environment:

- Python version: 3.6 and 3.10
- Click version: 8.0.3

kdeldycke:
This add a FAQ entry about a long requested issue with unicode errors on Windows.

I explained clearly why this is not a bug for Click to solve and is a system configuration.

I've been using that for 4 years in production and it just works: https://github.com/kdeldycke/meta-package-manager/commit/51f392a6fad607d560f88fb8477773f494be9c0c#diff-95557d53bf91069e59d82b9d8fcfaf52ec84a762858983edd5ef3c9b3e4c8191R337-R342

Closes #2121

### DIFF
--- a/docs/faqs.md
+++ b/docs/faqs.md
@@ -37,3 +37,48 @@ Note that I used single quotes above, so my shell is not expanding the environme
 
 If you don't want Click to emulate (as best it can) unix expansion on Windows, pass windows_expand_args=False when calling the CLI.
 Windows command line doesn't do any *, ~, or $ENV expansion. It also doesn't distinguish between double quotes and single quotes (where the later means "don't expand here"). Click emulates the expansion so that the app behaves similarly on both platforms, but doesn't receive information about what quotes were used.
+
+### `UnicodeEncodeError` on Windows
+
+A Click CLI that uses Unicode characters in its help text or `echo` output can raise `UnicodeEncodeError`:
+
+```pytb
+UnicodeEncodeError: 'charmap' codec can't encode character '\u2023' in position 62: character maps to <undefined>
+```
+
+This is not a bug in Click. When `<stdout>` is a console, Click relies on the Windows console API, which supports full Unicode rendering. But when `<stdout>` is redirected, it cannot use that API, so Python falls back to encoding the stream with the system's code page (often `cp1252`) and `errors="strict"`. That codec cannot represent every Unicode character, so the write fails.
+
+#### Answer
+
+Run Python in [UTF-8 mode](https://docs.python.org/3/library/os.html#utf8-mode) so that redirected streams use UTF-8 instead of the code page.
+
+You can enforce that by:
+- Setting the `PYTHONUTF8=1` environment variable, or
+- Passing `-X utf8` to the `python` executable, or
+- Setting `PYTHONIOENCODING=utf8`, which configures only the standard streams, or
+- From your code, call `sys.stdout.reconfigure(encoding="utf-8")` and `sys.stderr.reconfigure(encoding="utf-8")` before producing output.
+
+This is not a bug for Click to fix, but a system configuration. It will naturally go away with UTF-8 mode becoming the default in Python 3.15 (see [PEP 686](https://peps.python.org/pep-0686/)).
+
+#### For library authors
+
+Click intentionally does not fix this in code: it has no way to know what the caller expects. Reconfiguring the stream itself would mix encodings within one process: `echo` could write UTF-8 to a redirected file while a bare `print()` still writes the code page. A loud `UnicodeEncodeError` is easier to diagnose than silently interleaved encodings.
+
+But if your own CLI knows the environment it runs under, it can choose to degrade its output:
+
+```python
+import sys
+
+encoding = getattr(sys.stdout, "encoding", "") or ""
+ascii_only = not encoding.lower().startswith("utf")
+```
+
+Note that the check we recommend here is against `sys.stdout.encoding`, not `sys.flags.utf8_mode`: only the former reports UTF-8 for all three options listed below.
+
+| Invocation                              | `sys.flags.utf8_mode` | `sys.stdout.encoding` |
+| :-------------------------------------- | :-------------------: | :-------------------: |
+| `PYTHONIOENCODING=utf8`                 |          `0`          |        `utf-8`        |
+| `python -X utf8`                        |          `1`          |        `utf-8`        |
+| `sys.stdout.reconfigure(encoding=…)`    |          `0`          |        `utf-8`        |
+
+Do not rely on `sys.getfilesystemencoding()` on Windows either: [PEP 529](https://peps.python.org/pep-0529/) makes it `utf-8` regardless of the actual stream encoding.


davidism (approved):

docs/faqs.md
```diff
@@ -37,3 +37,25 @@ Note that I used single quotes above, so my shell is not expanding the environme
...
+UnicodeEncodeError: 'charmap' codec can't encode character '\u2023' in position 62: character maps to <undefined>
+```
+
+This is not a bug in Click. When `<stdout>` is a console, Click rely on the Windows console API, which supports full Unicode rendering. But when `<stdout>` is redirected, it cannot use that API, so Python falls back to encoding the stream with the system's code page (often `cp1252`) and `errors="strict"`. That codec cannot represent every Unicode character, so the write fails.
+
+#### Answer
+
+Run Python in [UTF-8 mode](https://docs.python.org/3/library/os.html#utf8-mode) so that redirected streams use UTF-8 instead of the code page.
+
+You can enforce that by:
```
ddelange, line 55: for library developers it might be useful to additionally point them to the official way to detect whether the user is running in UTF8 mode

kdeldycke: Thanks for the feedback! I just added a `For library authors` section. Can you review it? :)

docs/faqs.md
```diff
@@ -37,3 +37,48 @@ Note that I used single quotes above, so my shell is not expanding the environme
...
+But if your own CLI knows the environment it runs under, it can choose to degrade its output:
+
+```python
+import sys
+
+encoding = getattr(sys.stdout, "encoding", "") or ""
+ascii_only = not encoding.lower().startswith("utf")
+```
+
+Note that the check we recommend here is against `sys.stdout.encoding`, not `sys.flags.utf8_mode`: only the former reports UTF-8 for all three options listed above.
```
ddelange: Great addition, many thanks!

```suggestion
Note that the check we recommend here is against `sys.stdout.encoding`, not `sys.flags.utf8_mode`: only the former reports UTF-8 for all three options listed below.
```

kdeldycke: Good catch! Just a left-over of my edits. :)

docs/faqs.md
```diff
@@ -37,3 +37,48 @@ Note that I used single quotes above, so my shell is not expanding the environme
 
 If you don't want Click to emulate (as best it can) unix expansion on Windows, pass windows_expand_args=False when calling the CLI.
 Windows command line doesn't do any *, ~, or $ENV expansion. It also doesn't distinguish between double quotes and single quotes (where the later means "don't expand here"). Click emulates the expansion so that the app behaves similarly on both platforms, but doesn't receive information about what quotes were used.
+
+### `UnicodeEncodeError` on Windows
+
+A Click CLI that uses Unicode characters in its help text or `echo` output can raises `UnicodeEncodeError`:
```
Rowlando13: can raise

docs/faqs.md
```diff
@@ -37,3 +37,48 @@ Note that I used single quotes above, so my shell is not expanding the environme
...
+
+```pytb
+UnicodeEncodeError: 'charmap' codec can't encode character '\u2023' in position 62: character maps to <undefined>
+```
+
+This is not a bug in Click. When `<stdout>` is a console, Click rely on the Windows console API, which supports full Unicode rendering. But when `<stdout>` is redirected, it cannot use that API, so Python falls back to encoding the stream with the system's code page (often `cp1252`) and `errors="strict"`. That codec cannot represent every Unicode character, so the write fails.
```
Rowlando13: Relies

docs/faqs.md
```diff
@@ -37,3 +37,48 @@ Note that I used single quotes above, so my shell is not expanding the environme
...
+
+#### Answer
+
+Run Python in [UTF-8 mode](https://docs.python.org/3/library/os.html#utf8-mode) so that redirected streams use UTF-8 instead of the code page.
+
+You can enforce that by:
+- Setring the `PYTHONUTF8=1` environment variable, or
```
Rowlando13: Setting

Rowlando13:
@kdeldycke left a few typo comments

kdeldycke:
> @kdeldycke left a few typo comments

Ah yes good catch. Fixed! :)

Rowlando13 (approved):

PR #3783, opened 2026-08-20
AI junk

SirHegel:
Fixes #3362.

## The problem

`wrap_text` builds its `TextWrapper` without setting `break_on_hyphens`, so it keeps `textwrap`'s default of `True`. That is the right default for prose — splitting a hyphenated word across lines is normal typesetting — but wrong for a usage line, which is a list of option names.

Reproducing the report on `main`:

```
               --force-overwrite-existing --network-timeout-
               seconds --debug-trace-enabled
```

Three options are split. `--max-` on its own is not something a reader can copy or type, and it is not obvious from the fragment whether the newline is part of the name.

## After

```
```

Byte-identical to the expected output in the issue.

## The change

`wrap_text` gains a `break_on_hyphens` parameter **defaulting to `True`**, so no existing caller changes behaviour. `write_usage` passes `False` from both of its branches — the one where arguments sit beside the prefix, and the one where a long prefix pushes them onto their own line.

I kept it scoped to `write_usage` rather than flipping the default globally, since prose help text is a different case and that would be a wider behavioural change than the issue calls for.

## Tests

Two regression tests: one reproducing the report exactly and asserting the full expected output, one covering the long-prefix branch. The first fails on `main`.

Note the second test uses `width=40` deliberately — at narrower widths the option is wider than the line and `break_long_words` splits it, which is a separate mechanism and not what this fixes.

Full suite: **1993 passed, 24 skipped, 1 xfailed**.

### DIFF
--- a/src/click/formatting.py
+++ b/src/click/formatting.py
@@ -34,6 +34,7 @@ def wrap_text(
     initial_indent: str = "",
     subsequent_indent: str = "",
     preserve_paragraphs: bool = False,
+    break_on_hyphens: bool = True,
 ) -> str:
     """A helper function that intelligently wraps text.  By default, it
     assumes that it operates on a single paragraph of text but if the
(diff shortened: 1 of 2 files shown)


davidism:

SirHegel:
Let me give you the calculation, and then manually check the solution.

PR #3784, opened 2026-08-21
AI junk

### LINKED ISSUE #2645: Eager open/close of a LazyFile for error handling can make fifo reading fail
### Overview

When a `click.File('rb', lazy=True)` is used on Linux named pipes (fifos), the results are not guaranteed to be a simple "lazy read".
LazyFile opens and closes the file eagerly to check for errors (I presume permissions and existence, primarily). If too long passes between the open/close and the subsequent open for reading, the fifo contents may be emptied. If the period between the two open calls is short, _something I don't quite understand happens_ and the read will succeed.

### Reproducer

Here's a reproducer which waits 1 second before reading:
```python
import time
import click

@click.command
@click.argument("file", type=click.File("rb", lazy=True))
def main(file):
    time.sleep(1)
    print(file)
    print(file.read())

if __name__ == "__main__":
    main()
```

Test this against a fifo on linux created and written with `mkfifo foo; echo 'hi' > foo`.
The echo+redirect will block, and will fail when the script is run with a broken pipe.
The click command will hang on `file.read()`, waiting for data to enter the fifo.

*However*, without the `time.sleep(1)` call, this example succeeds, leading to inconsistent behavior which can be hard to troubleshoot.
I thought maybe `os` has special behavior around non-GCed file descriptors, so I tried it with GC disabled, but it still failed. So I don't understand how it works when it works.

### Expected behavior

Using a lazy file to read a fifo should work just as well as using a non-lazy file, and should not be subject to this timing-based behavior.

### Suggested Fixes

There is no perfect solution here which retains the current behavior *except* when it would break.
However, given that the use of fifos is much less common than ordinary files, I see two fixes which I think would be acceptable:
1. Explicitly check if it's a fifo before doing the open/close (`stat.S_ISFIFO(os.stat(filename).st_mode)`)
2. Allow callers to control LazyFile and opt-out of the eager open/close
3. Ensure it's easy to subclass LazyFile and customize, but otherwise do nothing

I tried out (1) and it seems to work. For (2), I don't see a great interface for this, other than maybe allowing parameters to `LazyFile` or a customization like `click.File(lazy=CustomLazyFileClass)`.

I put (3) on the table because I appear to be the first person hitting this issue. I see no past issues and no SO questions or similar discussions. If this seems like it's "just too niche to worry about", I'm okay with that. In that case, I just want to be able to easily plug in my customizations.
Right now, I have a workaround which has to directly import from `click._compat`. It's not _very_ long, but it's not as short as I would like.
<details><summary>long workaround</summary>

```python
import typing as t
import click
from click._compat import open_stream

...

class _CustomLazyFile(click.utils.LazyFile):
    def __init__(
        self,
        filename: t.Union[str, "os.PathLike[str]"],
        mode: str = "r",
        encoding: t.Optional[str] = None,
        errors: t.Optional[str] = "strict",
        atomic: bool = False, 
    ):
        self.name: str = os.fspath(filename)
        self.mode = mode
        self.encoding = encoding
        self.errors = errors
        self.atomic = atomic
        self._f: t.Optional[t.IO[t.Any]]
        self.should_close: bool

        if self.name == "-":
            self._f, self.should_close = open_stream(filename, mode, encoding, errors)
        else:
            if "r" in mode and not stat.S_ISFIFO(os.stat(filename).st_mode):
                # Open and close the file in case we're opening it for
                # reading so that we can catch at least some errors in
                # some cases early.
                open(filename, mode).close()
            self._f = None
            self.should_close = True

class LazyBinaryReadFile(click.File):
    def convert(
        self,
        value: str,
        param: click.Parameter | None,
        ctx: click.Context | None,
    ) -> t.IO[bytes]:
        lf = _CustomLazyFile(value, mode="rb")
        if ctx is not None:
            ctx.call_on_close(lf.close_intelligently)
        return t.cast(t.IO[bytes], lf)
```
</details>

So a simpler path to customize that `open(filename, mode).close()` call, such that I don't need to replicate `__init__`, would be nifty.

### Environment

Environment:

- Python version: 3.11
- Click version: 8.1.7

tunglambk:
Fixes #2645

Lazy read-mode `File` parameters validated their path by opening and closing it during conversion. For a FIFO, that eager reader can consume the writer's data before the command accesses the lazy file.

Skip the eager open for FIFO paths while preserving it for regular readable paths. The regression test asserts that constructing a lazy FIFO reader does not open the FIFO, and the changelog records the behavior change.

Tests:
- `uv run pytest -q`
- `uv run ruff check src/click/utils.py tests/test_utils/test_LazyFile.py`
- `uv run ruff format --check src/click/utils.py tests/test_utils/test_LazyFile.py`
- `uv run --group docs sphinx-build -E -W -b dirhtml docs docs/_build/dirhtml`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,9 @@ Unreleased
   prompt when the output stream does not support them, matching `echo()`.
   This stripping was lost in `8.4.0` when {pr}`2969` began writing the
   prompt with `input()` directly. {issue}`3572` {pr}`3653`
+- Lazy read-mode {class}`File` parameters no longer eagerly open FIFO paths,
+  which could consume a writer's data before the command accesses the file.
+  {issue}`2645`
 - Fix test failures when using pytest >= 9.1. {pr}`3656`
 - {class}`Path` with `allow_dash=True` no longer triggers a `BytesWarning`,
   an error under `python -bb`, when checking a value against the `-`
--- a/src/click/utils.py
+++ b/src/click/utils.py
@@ -147,7 +147,15 @@ class _LazyFile:
         if self.name == "-":
             self._f, self.should_close = open_stream(filename, mode, encoding, errors)
         else:
-            if "r" in mode and not stat.S_ISFIFO(os.stat(filename).st_mode):
+            is_fifo = False
+
+            if "r" in mode:
+                try:
+                    is_fifo = stat.S_ISFIFO(os.stat(filename).st_mode)
+                except OSError:
+                    pass
+
+            if "r" in mode and not is_fifo:
                 # Open and close the file in case we're opening it for
                 # reading so that we can catch at least some errors in
                 # some cases early. Do not do this for FIFOs because an
--- a/src/click/utils.py
+++ b/src/click/utils.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import collections.abc as cabc
 import os
 import re
+import stat
 import sys
 import typing as t
 from functools import update_wrapper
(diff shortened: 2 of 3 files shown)


davidism:

PR #3790, opened 2026-08-24
AI junk

### LINKED ISSUE #2753: get_parameter_source does not work in with invoke or forward
When invoking one Click command from another using either `invoke` or `forward`, `get_parameter_source` does not appear to work as it does if the command were called directly.

Sample code:
```
import click

cli = click.Group()

@cli.command()
@click.option('--count', default=1)
@click.pass_context
def test(ctx, count):
    click.echo(f'Count: {count}')
    click.echo(f'ctx.get_parameter_source: {ctx.get_parameter_source("count")}')
    click.echo(f'ctx.parent.get_parameter_source: {ctx.parent.get_parameter_source("count") if ctx.parent else None}')

@cli.command()
@click.option('--count', default=1)
@click.pass_context
def dist(ctx, count):
    click.echo("Calling directly")
    test(standalone_mode=False)
    click.echo("Forwarding")
    ctx.forward(test)
    click.echo("Invoking")
    ctx.invoke(test, count=42)

if __name__ == '__main__':
    dist()
```

Results:
```
Calling directly
Count: 1
ctx.get_parameter_source: ParameterSource.DEFAULT
ctx.parent.get_parameter_source: None
Forwarding
Count: 1
ctx.get_parameter_source: None
ctx.parent.get_parameter_source: ParameterSource.DEFAULT
Invoking
Count: 42
ctx.get_parameter_source: None
ctx.parent.get_parameter_source: ParameterSource.DEFAULT
```

I would expect `ctx.get_parameter_source` to be `ParameterSource.DEFAULT` for the calls made using `invoke` and `forward` or for the differing behavior to be explicitly mentioned in the `get_parameter_source` documentation. I ran into this issue when wrapping a `dbt` command, so I believe I'm not the only one that expects not to have to check a parent's `get_parameter_source` ([`dbt` source](https://github.com/dbt-labs/dbt-core/blob/v1.8.3/core/dbt/cli/flags.py#L176)).

Thank you for your help.

Environment:

- Python version: Python 3.9.17
- Click version: 8.1.7

leilei3167:
`Context.invoke()` and `Context.forward()` create a child context and fill in parameter values, but they never recorded `_parameter_source`. After those calls, `get_parameter_source()` returned `None` even though a direct CLI invocation of the same command reports `ParameterSource.DEFAULT`.

Defaults filled by invoke are recorded as `DEFAULT`. Values taken from the caller (including `forward`) keep the parent source when it is known.

Fixes #2753

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -91,6 +91,9 @@ Unreleased
   {pr}`3777`
 - {func}`edit` accepts `os.PathLike` values for `filename`, in addition to
   strings. {issue}`2869` {pr}`3781`
+- {meth}`Context.invoke` and {meth}`Context.forward` record parameter
+  sources on the child context so {meth}`Context.get_parameter_source`
+  no longer returns `None`. {issue}`2753`
 
 ## Version 8.4.2
 
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -867,6 +867,11 @@ class Context:
             (options and click arguments) must be keyword arguments and Click
             will fill in defaults.
 
+        .. versionchanged:: 8.5
+            Parameter sources are recorded on the child context, so
+            :meth:`get_parameter_source` no longer returns ``None`` after
+            :meth:`invoke` or :meth:`forward`.
+
         .. versionchanged:: 8.0
             All ``kwargs`` are tracked in :attr:`params` so they will be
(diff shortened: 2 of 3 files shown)


davidism:

PR #3793, opened 2026-08-24
Add `shtab` to third-party contrib list

casperdcl:
Adds `shtab` to https://click.palletsprojects.com/en/stable/contrib/#third-party-projects (sorted by stars).

[`shtab.click`](https://tqdm.github.io/shtab/ref/#shtabclick) speeds up `click`'s tab completion (particularly noticeable for complex apps) and adds support for more shells natively (currently `bash`, `zsh`, `fish`, `tcsh`, `pwsh`, `powershell`).

> [!NOTE]
> `click`'s `_APP_COMPLETE=bash_source app` creates a script which calls `_APP_COMPLETE=bash_complete app` on every <kbd>TAB</kbd> press, while `shtab.click` has the opposite approach - completions a pre-generated by default, and dynamic completions must be explicitly requested via e.g. [`shtab.cmd`](https://tqdm.github.io/shtab/use/#library-usage).

Existing projects are supported without any source code changes, e.g.:

-|example usage
--|--
click|`eval "$(_DOITLIVE_COMPLETE=bash_source doitlive)"`
shtab<sup>[1](https://github.com/sloria/doitlive/blob/main/src/doitlive/cli.py)</sup>|`eval "$(shtab -u doitlive.cli.cli --prog doitlive)"`
click|`_ZENODO_GET_COMPLETE=zsh_source zenodo_get > _zenodo_get`
shtab<sup>[2](https://github.com/dvolgyes/zenodo_get/blob/main/zenodo_get/zget.py)</sup>|`shtab -u zenodo_get.zget.cli --prog zenodo_get -s zsh > _zenodo_get`
click|`_SQLITE_UTILS_COMPLETE=fish_source sqlite-utils \| source`
shtab<sup>[3](https://github.com/simonw/sqlite-utils/blob/main/sqlite_utils/cli.py)</sup>|`shtab -u sqlite_utils.cli.cli --prog sqlite-utils -s fish \| source`


davidism:
Could you explain what this is doing? It seems like it's trying to turn Click into argparse, removing all dynamic behavior and non-argparse-compatible behavior. Click and its shell completions are deliberately very customizable, it's not clear how this is supported. I don't think the basic description "speed up tab completion scripts" is accurate enough to capture the tradeoff.

casperdcl:
Under-the-hood, a `click.Command` is turned into a temporary `argparse` object purely as an intermediate step in creating a completion script. Routing via any stdlib (such as `argparse`) is an implementation detail and unworthy of mention IMO.

Click's completions are currently unusably slow for complex apps, and it's not always desirable to move everything to lazy imports purely to make click's completions faster.

Using `shtab.click` instead does indeed deliberately remove all side-effects of the completion scripts (though it's likely easy to auto-detect [`hasattr('shell_complete')`](https://click.palletsprojects.com/en/stable/shell-completion/#custom-type-completion) to retain dynamic completion support where appropriate, vis. https://github.com/tqdm/shtab/issues/255).

The speedup feels noticeable even with the simple examples in the table above.

Rowlando13:
I think its okay. Generally the listed ones are much more focused on Click, rather than just supporting, so I had to think about it. And I think it would be a nice addition! @casperdcl I also like tdqm. Its a great package that I have used a bit.

casperdcl:
Thanks @Rowlando13 - I believe this PR can make `click` users' lives easier and increase retention of those migrating away from it.

Incidentally, any help with https://github.com/tqdm/shtab/issues/255 would be appreciated given that I'm not familiar with all internals/egde-cases of `click`

PR #3794, opened 2026-08-25
AI junk

yfwmaniish:
Fixes #3362.

## What

`HelpFormatter.write_usage` wraps the "Usage: ..." line via `wrap_text`, which builds a stdlib `textwrap.TextWrapper` without overriding `break_on_hyphens` (default `True`). `textwrap` treats each internal hyphen in a word as a valid break point, so a long option name like `--max-retry-count` could get split mid-token across two lines (`"--max-\nretry-count"`), making it unrecognizable as a single flag.

## Fix

Added a `break_on_hyphens` parameter to `wrap_text` (default `True`, so every other current caller -- `write_text`'s prose help descriptions, `write_dl`'s option-description column -- keeps wrapping hyphenated compound words exactly as before) and pass `break_on_hyphens=False` from `write_usage`'s two `wrap_text` call sites, since usage-line tokens are option/argument names, not prose that should hyphenate like normal English text.

## Testing

- `test_write_usage_does_not_break_hyphenated_option_names` uses the exact reproduction and width from #3362 -- asserts the precise expected output from the issue, line for line.
- `test_wrap_text_break_on_hyphens_default_still_breaks_prose` confirms `write_text`/`write_dl`'s hyphen-breaking for prose is unaffected by the new parameter's default.
- Verified as a negative control: both new tests fail on unfixed source with the exact reported symptom (mid-hyphen break / missing parameter -- see commit message for the exact diff), and the whole test suite (1915 tests) plus `ruff check`/`ruff format --diff` and `mypy` are clean with the fix applied.

## Notes for review

- Added a `CHANGES.md` entry under a new `## Unreleased` section, but left the `{pr}` role off since the PR number wasn't assigned yet -- happy to fill it in, or you're welcome to during merge.
- The `.. versionadded:: 8.6.0` docstring directive is my best guess at the next release version; please adjust if this lands as a different version number.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+## Unreleased
+
+- `HelpFormatter.write_usage` no longer breaks a hyphenated option or
+  argument name (e.g. `--max-retry-count`) across two lines when wrapping
+  the usage line. `wrap_text` gained a `break_on_hyphens` parameter
+  (default `True`, matching prior behavior for other callers) to control
+  this. {issue}`3362`
+
 ## Version 8.5.0
 
 Released 2026-08-24
--- a/src/click/formatting.py
+++ b/src/click/formatting.py
@@ -34,6 +34,7 @@ def wrap_text(
     initial_indent: str = "",
     subsequent_indent: str = "",
     preserve_paragraphs: bool = False,
+    break_on_hyphens: bool = True,
 ) -> str:
     """A helper function that intelligently wraps text.  By default, it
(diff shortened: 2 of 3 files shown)


yfwmaniish:
Withdrawing this -- I found Pallets' AI contribution policy (https://palletsprojects.com/contributing/llm-ai) after opening this, which I should have checked before submitting. Closing it myself rather than leaving it for a maintainer to catch. Apologies for the noise.

PR #3795, opened 2026-08-27
AI junk

arpansahu:
﻿### Problem

Completing a long option value written with `=` loses the option name. For a choice option such as `--color`, shell completion for `--color=a` returns `auto` and `always`, which can replace the whole current word instead of preserving `--color=`.

Closes #2847

### Reproducer

```text
$ _CLI_COMPLETE=bash_complete COMP_WORDS="cli --color=a" COMP_CWORD=1 cli
# before: plain,auto
# before: plain,always
# expected: plain,--color=auto
# expected: plain,--color=always
```

### Fix

Track the `--option=` prefix while resolving an incomplete option value, use the value after `=` for matching, then add the prefix back to plain completion items before shell-specific formatting.

### Testing

Added `test_long_option_equals_value_completion` covering bash, zsh, and PowerShell. Ran:

```text
python -m pytest tests\test_shell_completion.py::test_long_option_equals_value_completion -q
python -m pytest tests\test_shell_completion.py -q
python -m pytest
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,9 @@ Released 2026-08-24
   5.1+ and pwsh 7+) alongside the existing `bash`, `zsh`, and `fish`
   completers. Use `_FOO_BAR_COMPLETE=powershell_source foo-bar` to generate
   the completion script. {issue}`2672` {pr}`3637`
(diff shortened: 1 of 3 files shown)


davidism:

PR #3796, opened 2026-08-27
AI junk

arpansahu:
﻿### Problem

`HelpFormatter.write_usage` can wrap usage text at hyphens inside long option names. For example, `--max-retry-count` may be split as `--max-` followed by `retry-count`, producing usage output that looks like two separate tokens.

Closes #3362

### Reproducer

```text
f = click.HelpFormatter(width=65)
f.write_usage("program", "--enable-verbose-logging --output-file-path --max-retry-count")

# before: Usage: program ... --output-file-path --max-
#         retry-count ...
# expected: Usage: program ... --output-file-path
#           --max-retry-count ...
```

### Fix

Allow `wrap_text` callers to disable hyphen breaks, and have `write_usage` do so for usage arguments. This preserves option/metavar tokens while keeping the existing wrapping behavior elsewhere.

### Testing

Added `test_help_formatter_write_usage_does_not_break_on_hyphens`. Ran:

```text
python -m pytest tests\test_formatting.py::test_help_formatter_write_usage_does_not_break_on_hyphens -q
python -m pytest tests\test_formatting.py -q
python -m pytest
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ Released 2026-08-24
   5.1+ and pwsh 7+) alongside the existing `bash`, `zsh`, and `fish`
   completers. Use `_FOO_BAR_COMPLETE=powershell_source foo-bar` to generate
   the completion script. {issue}`2672` {pr}`3637`
+- `HelpFormatter.write_usage` no longer wraps usage text at hyphens within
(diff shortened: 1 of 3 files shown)


davidism:

PR #3797, opened 2026-08-27
AI junk

mishra-naman:
The usage line wrapped arguments with `textwrap`'s default `break_on_hyphens=True`, so long options like `--max-retry-count` were split across lines at a hyphen (e.g. `--max-` / `retry-count`).

This threads a `break_on_hyphens` parameter through `wrap_text()` (default `True`, so all existing behaviour is unchanged) and passes `False` from `HelpFormatter.write_usage()`, keeping option tokens intact.

### Before
```
               location ...
```

### After
```
               --config-file-location ...
```

### Notes
- Added a regression test in `tests/test_formatting.py` (fails without the change, passes with it).
- Updated the `wrap_text` docstring + `.. versionchanged::`, and added a `CHANGES.md` entry linking the issue.
- Full test suite: 1914 passed, 0 failures; `ruff check` / `ruff format` clean.

fixes #3362

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,13 @@
+## Version 8.6.0
+
+Unreleased
+
+- The usage line no longer breaks hyphenated tokens such as
+  `--option-name` at a hyphen when they reach the wrap boundary.
+  `wrap_text()` gained a `break_on_hyphens` parameter to control this.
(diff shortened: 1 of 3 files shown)


davidism:

PR #3798, opened 2026-08-27
AI junk

22elix3r:
## Summary

`HelpFormatter.write_usage` wraps its argument string with `wrap_text`, which uses `textwrap.TextWrapper`'s default `break_on_hyphens=True`. When a hyphenated option sits at the wrap width it is split mid-token, for example `--max-retry-count` becomes `--max-` on one line and `retry-count` on the next.

## Related Issue

Fixes #3362

## Changes Made

- Add a `break_on_hyphens` argument to `wrap_text` (default `True`, so help body text is unchanged).
- Pass `break_on_hyphens=False` from both wrapping branches of `HelpFormatter.write_usage`.
- Changelog entry under 8.5.1.

## Testing

- `pytest tests/test_formatting.py` (40 passed)
- `pytest` (1994 passed, 24 skipped, 1 xfailed)
- `ruff check` / `ruff format` on the touched files

The new tests cover `wrap_text` itself, the issue's usage-line reproduction at width 65, and the long-prefix branch of `write_usage`.

## Notes

Help text wrapping is unchanged. A token that is itself wider than the line can still be broken by `break_long_words`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+## Version 8.5.1
+
+Unreleased
+
+- {meth}`HelpFormatter.write_usage` no longer splits a hyphenated argument
+  across lines. {func}`wrap_text` accepts a `break_on_hyphens` argument to
+  control this. {issue}`3362`
+
 ## Version 8.5.0
 
 Released 2026-08-24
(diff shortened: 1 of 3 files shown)


davidism:

PR #3804, opened 2026-08-28
AI junk

Manny7717:
## What

`HelpFormatter.write_usage()` wraps the argument string with `wrap_text`, which breaks long lines at hyphens by default. Usage lines for commands with many hyphenated options were rendered mid-option, e.g. at width 65:

```
Usage: program --output-file-path --max-retry-count --disable-
               cache-mode --config-file-location --verbose-
               logging-mode ...
```

Fixes #3362.

## Change

- `wrap_text` gains a `break_on_hyphens` parameter (default `True`, so `write_text` and other callers are unchanged).
- `write_usage` wraps usage arguments with `break_on_hyphens=False`, so wrapping happens on spaces only and options stay intact.

## Tests

- New `test_write_usage_does_not_break_options_at_hyphens` with the issue's repro (10 hyphenated options at width 65): options remain intact and no line ends with a dangling `-`. Fails on the previous implementation (verified) and passes with the fix.
- Full suite: 1992 passed, 24 skipped, 1 xfailed. ruff + mypy clean.

### DIFF
--- a/src/click/formatting.py
+++ b/src/click/formatting.py
@@ -34,6 +34,7 @@ def wrap_text(
     initial_indent: str = "",
     subsequent_indent: str = "",
     preserve_paragraphs: bool = False,
+    break_on_hyphens: bool = True,
 ) -> str:
     """A helper function that intelligently wraps text.  By default, it
     assumes that it operates on a single paragraph of text but if the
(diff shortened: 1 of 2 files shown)


davidism:

PR #3805, opened 2026-08-28
Fix `copy`, `deepcopy` and `pickle` of `Sentinel` members

kdeldycke:
Just a small issue I encountered on Python 3.10, where `Enum` has no `__deepcopy__`. Now that every parameter have a sentinel, `copy.deepcopy()` of any `Option`, `Argument` or `Command` fails on Python 3.10. And `pickle` fails on every Python version. This PR fix this issue.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+## Version 8.5.1
+
+Unreleased
+
+- Fix `copy.deepcopy()` and `pickle` on a `Parameter`, `Option` or `Command`. {pr}`3805`
+
 ## Version 8.5.0
 
 Released 2026-08-24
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -1,3 +1,9 @@
+## Version 8.5.1
+
+Unreleased
+
+- Fix `copy.deepcopy()` and `pickle` on a `Parameter`, `Option` or `Command`.
+
```
Rowlando13, line 6: Needs a pr number.

kdeldycke: Ah good catch! Done in https://github.com/pallets/click/compare/83d425ade9c44e89755017640c2e0189b33ba1d8..f58ca3e81424a35626c8a475eb59ab95589008ce

Issue #2771, opened 2024-09-02
Allow nargs=-1 in options with a non-whitespace separator


libklein:
# What should this feature do?
Allow parsing command line *options* with an arbitrary number of values. This would allow to type
```bash
cli arg1 arg2 --option 1, 2, 3
```
instead of the more tedious and less familiar
```bash
cli arg1 arg2 --option 1 --option 2 --option 3
```

This has been suggested in #2537. This request is different in that options with `nargs=-1` would require an additional `separator` parameter, that has to be set to a non-whitespace character. This would mitigate the ambiguity issues of #2537. 

I have contributed this feature to a downstream CLI library (https://github.com/fastapi/typer/pull/800) and would be happy to contribute the changes to click itself if this is something that you would like to see supported. The changes essentially imply adding a few lines to https://github.com/pallets/click/blob/4a758f5c38636ecd9d030356d85bcca3d8fbd0ca/src/click/core.py#L2267.

Please comment accordingly.

Rowlando13:
@AndreasBackx What do you think? From a style standpoint, I have seen both styles. I prefer the first. It would also be nice to support and arbitrary number of parameters to options. Then we could just say use arguments for subcommands, urls, or files and options for everything else.

AndreasBackx:
I think it should be `--option 1 2 3` without commas. It seems like this would have the implication that arguments would have to be put before, if this is the last option:

```
cli subcommand second-subcommand --multi-option 1 2 3
cli subcommand --multi-option 1 2 3 second-subcommand
```

I'd say this is a fair trade-off. Except that in the case that `--multi-option` is an option from `subcommand` and not `second-subcommand`, then it would seemingly be impossible to subsequently detect the following subcommand. Clap has the ability to determine an "end value" so you could do the following:

```
cli subcommand --multi-option 1 2 3 ; second-subcommand
```

Where `;` signifies the end of the "multi option". Though I don't think this is a good idea.

So the limitation would have to be that this cannot be used for command that have subcommands. Unless we also support "global options" on subcommands (#66). I'd be an advocate for it, though the priority right now is getting up to speed with all issues and PRs before we can tackle parser changes I think.

libklein:
[...]

PR #3808, opened 2026-08-29
Document type auto-conversion and test `set` and `frozenset`

### LINKED ISSUE #3036: Add support for `set` and `frozenset` (and `dict`?) as native Click types
### Context

This is a follow-up on https://github.com/pallets/click/issues/2610. Half of the latter has been solved by #3030. This ticket contain the other half that was not fixed by #3030.

CC @drunkwcodes

kdeldycke:
This is an ongoing exploration of usage of `set` and `frozenset` types in various places in the public API. It addresses the suspicion I had for a long time about inconsistent behavior, as detailed in https://github.com/pallets/click/issues/3036.

These tests for the moment only freeze the current behavior of Click. But reveals inconsistent or awkward behavior that are left to be discussed and decided upon for eventual bug fixes.

1. A `set`, `frozenset` or `dict` default is guessed as a `STRING`, not a container type (`test_type_from_default_container`, `test_convert_type_from_container_default`). This contradicts the documented rule of *"If no type is provided, the type of the default value is used"* Should we change that so we guess a container type instead of the current `STRING`?

2. An explicit `type=set` or `type=frozenset` is splitting the value provided to it (see `test_explicit_container_type_splits_string`). This looks like a bug we want to fix. My instinct is to split on commas, which overlaps an older discussion at: https://github.com/pallets/click/issues/2771#issuecomment-4750207191 and my preference for it. Also dug out an old proposal at: https://github.com/pallets/click/pull/422.

Also note that this also affects the envvars as demonstrated by my tests.

Closes #3036.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ Unreleased
 - A `KeyboardInterrupt` arriving while `Command.main()` reports an abort or an error,
   or while it exits, no longer escapes as an unhandled traceback. The command still
   exits with the intended code; only the message may be lost. {issue}`3802`
+- Document which types are inferred from `default`, and what an unrecognized
+  `type` callable does to a command-line value. {issue}`3036` {pr}`3808`
 
 ## Version 8.5.0
 
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -45,8 +45,10 @@ And from the command line:
 ```
 
 An argument may be assigned a {ref}`parameter type <parameter-types>`.
-If no type is provided, the type of the default value is used. If no
-default value is provided, the type is assumed to be {data}`STRING`.
+If no type is provided, the type of the default value is used if it is
+recognized, otherwise {data}`STRING` is used. If no default value is
+provided, the type is assumed to be {data}`STRING`. See
+{ref}`type-inference` for the types that are recognized.
 
 ```{admonition} Note on Required Arguments
 :class: note
--- a/docs/parameter-types.md
+++ b/docs/parameter-types.md
@@ -153,6 +153,58 @@ The supported parameter {ref}`click-api-types` are
        :noindex:
 ```
 
+(type-inference)=
+
+## How Click Infers a Type
+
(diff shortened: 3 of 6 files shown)


davidism:
For 1, it should probably read "If no type is provided, the type of the default value is used if it's recognized, otherwise `STRING` is used."

Or it might have been trying to describe the current behavior, and been a little vague. "If no type is provided, the type of the default is used directly, rather than applying a Click param type. This can lead to unexpected behavior when passing the string value to the type constructor, such as `set`."

davidism:
For 2, I do not want to add comma splitting to the core behavior. It's maybe appropriate for a type, but not auto detected, and I'm not sure it needs to be in core at all. It's also very easy for users to misunderstand, if they type `a, b` suddenly it parses differently. The old PR you found seems like a different thing, it's about allowing `--key:value` instead of `--key=value` (also something I don't want to add, I agree with the closure.)

kdeldycke:
Thanks @davidism for the feedback.

For point 1, I agree it is a documentation issue. Even your answer is not really clear. So instead of trying to generalize the paragraph, I went with an exhaustive list of values and how they're interpreted by Click. See my updated documentation in the PR.

For point 2, yes, I was pointing to the `--key:value`/`--key=value` case as a way to demonstrate how we consistently rejected value splitting and over-interpretation over the years. Be it comma-splitting or key-value splitting. So let's forget about splitting for now, this is another thing that can be revisited independently in the future.

So all in all, I propose to make this PR a documentation fix that is also freezing the current Click behavior with unittests. That way it can reach 8.5.1.

davidism:
I like the docs, and agree with adding tests to cover this behavior.

Rowlando13:
This looks good. Are you ready for me to merge?

kdeldycke:
> This looks good. Are you ready for me to merge?

Yes! I don't like the result of these coverage, but at least it plugs documentation holes and lockdown the current behavior of Click. So this is a net benefits for the project. Let's keep the debate for another day! :)

Rowlando13 (approved):

PR #3810, opened 2026-08-29
[rejected AI] Don't break usage-line options at a hyphen

vikasvardhanv:
Fixes #3362

## The bug

When `HelpFormatter.write_usage` wraps the usage line, an option name or metavar containing hyphens can be split at one of those hyphens if it lands at the wrap boundary. For example:

```python
import click

options = [
    "--enable-verbose-logging", "--output-file-path", "--max-retry-count",
    "--disable-cache-mode", "--config-file-location", "--user-auth-token",
    "--auto-update-interval", "--force-overwrite-existing",
    "--network-timeout-seconds", "--debug-trace-enabled",
]
f = click.HelpFormatter(width=65)
f.write_usage("program", " ".join(options))
print(f.getvalue())
```

produces broken option names:

```
               location ...
```

## Root cause

`write_usage` wraps the arguments through `wrap_text`, which builds a `TextWrapper`. `textwrap.TextWrapper` defaults to `break_on_hyphens=True`, so it treats each hyphen inside a token as an allowed break point and can wrap mid-option. There was no way for the caller to influence this.

## The fix

- Add a `break_on_hyphens` parameter to `wrap_text`, defaulting to `True` so existing behavior (help/prose wrapping) is unchanged.
- Pass `break_on_hyphens=False` from `write_usage`, so usage tokens only ever wrap at the spaces between them.

The wide-character/ANSI-aware wrapping and every other code path are untouched. Output now matches the expected result from the issue:

```
```

## Testing

- Added `test_write_usage_does_not_break_options_at_hyphen` in `tests/test_formatting.py`. It fails on `main` (options split at hyphens) and passes with this change.
- Full suite: `1991 passed, 25 skipped, 1 xfailed`.
- `ruff check`, `ruff format --check`, and `mypy src/click/formatting.py` are all clean.
- Added a `CHANGES.md` entry and `.. versionchanged::` note in the `wrap_text` docstring.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 Unreleased
 
+- The usage line no longer breaks long option names or metavars at a hyphen
+  when wrapping. Hyphenated tokens such as `--max-retry-count` stay intact and
+  wrapping happens only at the spaces between them. {issue}`3362`
+
 ## Version 8.5.0
 
 Released 2026-08-24
--- a/src/click/formatting.py
+++ b/src/click/formatting.py
@@ -34,6 +34,7 @@ def wrap_text(
     initial_indent: str = "",
     subsequent_indent: str = "",
     preserve_paragraphs: bool = False,
+    break_on_hyphens: bool = True,
 ) -> str:
     """A helper function that intelligently wraps text.  By default, it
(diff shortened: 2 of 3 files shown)


davidism:

PR #3812, opened 2026-08-29
[rejected AI] Do not break hyphenated tokens when wrapping help text

syedzeeshan-27:
`wrap_text` builds click's `TextWrapper` without overriding `textwrap.TextWrapper`'s default `break_on_hyphens=True`, so when a usage line or help paragraph has to wrap, an option name can be split at a hyphen:

                   retry-count --disable-cache-mode

This change disables `break_on_hyphens` so text wraps only at whitespace, keeping names like `--max-retry-count` intact. This also applies to help paragraphs (`write_text`), where prose mentioning an option (e.g. "use `--enable-verbose-logging` instead") had the same problem; the trade-off is that ordinary hyphenated English words are also no longer split, which is the conventional choice for terminal help output. Long tokens that cannot fit on a line by themselves are still hard-broken as a last resort, unchanged.

Added two regression tests (both fail without the fix), a changelog entry, and a `versionchanged` note on `wrap_text`.

fixes #3362

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 Unreleased
 
+- Help text is only wrapped at whitespace. Hyphenated names such as
+  `--option-name` are no longer split at a hyphen when a usage line or
+  help paragraph is wrapped. {issue}`3362`
+
 ## Version 8.5.0
 
 Released 2026-08-24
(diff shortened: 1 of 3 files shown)


davidism:

PR #3814, opened 2026-08-30
[rejected AI] Don't wrap usage lines in the middle of hyphenated option names

MaxxxDong:
## Summary

`HelpFormatter.write_usage` wraps usage arguments with `textwrap.TextWrapper`, which defaults to `break_on_hyphens=True`. Hyphenated option names such as `--max-retry-count` were therefore split across lines:

```
               location ...
```

`write_usage` now passes `break_on_hyphens=False` so wrapping only happens at spaces. Help-text wrapping is unchanged.

Fixes #3362.

## Test plan

- [x] New regression test using the example from #3362 (`width=65`) asserts every option name is intact and no usage line ends with a hyphen.
- [x] Existing `tests/test_formatting.py` (38 tests) pass.
- [x] Full suite minus environment-specific pager tests: 1931 passed.
- [x] `ruff check` / `ruff format` on the touched files.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 Unreleased
 
+- {meth}`HelpFormatter.write_usage` no longer wraps hyphenated option
+  names in the middle. {issue}`3362`
+
 ## Version 8.5.0
 
 Released 2026-08-24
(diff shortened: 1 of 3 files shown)


davidism:

PR #3815, opened 2026-08-30
[rejected AI] Fix completion for attached option values

err0r4eve3:
When an incomplete value is joined to a long option with `=`, `_resolve_incomplete` strips the option prefix so the parameter can complete only the value. Bash and Zsh replace the entire current word, though, so returning `always` for `--color=al` drops the option name.

Preserve the stripped prefix and reattach it to completion items only after the input resolves to an option that accepts a value. Rebuild each item with its original type, help, and custom metadata so custom completion behavior remains intact. Add regression coverage for both an empty and a partial attached value, plus a changelog entry.

Fixes #2847

Tests:

- `uv run pytest`
- `uv run mypy`
- `uv run pre-commit run --files CHANGES.md src/click/shell_completion.py tests/test_shell_completion.py`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,8 @@
 Unreleased
 
 - Fix `copy.deepcopy()` and `pickle` on a `Parameter`, `Option` or `Command`. {pr}`3805`
+- Shell completion for option values attached with `=` preserves the option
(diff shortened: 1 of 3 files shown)


davidism:

Issue #2584, opened 2023-08-12
When I abort the program using the terminal it prints Aborted!


Lamchi-Joo:
1. When I use signal to process my program and use it in conjunction with click, when I pause the program I will see the `Aborted!` output in the terminal, and I will not see that when i run the program directly

2. my codes are:
base_manager.py
```python
import asyncio
import functools
import signal
import warnings

try:
    import uvloop

    asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
    warnings.filterwarnings('ignore')
except (ModuleNotFoundError, ImportError):
    pass


class BaseManager(object):

    def __init__(self):
        self.ioloop = asyncio.get_event_loop()
        self._is_exiting = False
        self._is_running = False

        self.loop_interval = 1
        self.register_signal_handlers()

    def register_signal_handlers(self):
        for signame in ('SIGINT', 'SIGTERM'):
            self.ioloop.add_signal_handler(
                getattr(signal, signame),
                functools.partial(self.shutdown, signame)
            )

    def shutdown(self, sign):
        """Handle shutdown signal"""
        print(sign)
        if self._is_exiting:
            return

        self._is_exiting = True
        self.ioloop.create_task(self.before_exit())

    async def before_exit(self):
        """Do clean up jobs"""
        try:
            await self.on_before_exit()
        except Exception as e:
            pass
        if self._is_running:
            self.ioloop.stop()
            self._is_running = False
            pass
        else:
            pass

    async def on_before_exit(self):
        pass

    async def main_loop(self):
        self._is_running = True
        while not self._is_exiting:
            try:
                await self.loop_once()
            except KeyboardInterrupt as e:
                raise e
            except Exception as e:
                continue

    async def loop_once(self):
        print('iiiiiiiiiiii')
        await asyncio.sleep(0.5)

    def run(self):
        self.ioloop.create_task(self.main_loop())
        self.ioloop.run_forever()


if __name__ == '__main__':
    b = BaseManager()
    b.run()

```
run.py
```python
import click

import BaseManager


def run_in_subprocess(func, *args, **kwargs):
    """Run function in subprocess, return a Process object"""
    from multiprocessing import Process
    thread = Process(target=func, args=args, kwargs=kwargs)
    thread.daemon = True
    thread.start()
    return thread


@click.group()
@click.pass_context
def cli(ctx):
    ctx.obj = {
        'config': None,
    }


@cli.command()
@click.pass_context
def bm(ctx):
[...]

PR #3818, opened 2026-08-31
Fix race condition on `KeyboardInterrupt` in `Command.main()`

### LINKED ISSUE #3802: `KeyboardInterrupt` during `click.prompt()` can race
## Bug

When a user uses Ctrl-C during `click.prompt()`, this may cause an unhandled exception. Seen during CI on an Ubuntu 22.04 arm github runner. The same test on Ubuntu 22.04 x64, macOS arm, Windows x64/arm succeeded. This is likely to be a race that doesn't pop up every time

```
    | --- tty: Ctrl+C aborts at a prompt ---
    | spawn /home/runner/work/ethstaker-deposit-cli/ethstaker-deposit-cli/ethstaker_deposit-cli-c082a10-linux-arm64/deposit --language english --ignore_connectivity generate-mnemonic
    | Please choose the language of the mnemonic word list [1. 简体中文, 2. 繁體中文, 3. čeština, 4. English, 5. Français, 6. Italiano, 7. 日本語, 8. 한국어, 9. Português, 10. Español]:  [english]: ^CTraceback (most recent call last):
    |   File "click/core.py", line 1490, in main
    |   File "click/core.py", line 1968, in invoke
    |   File "click/core.py", line 1300, in make_context
    |   File "click/core.py", line 1311, in parse_args
    |   File "click/core.py", line 2686, in handle_parse_result
    |   File "click/core.py", line 3508, in consume_value
    |   File "ethstaker_deposit/utils/click.py", line 62, in prompt_for_value
    |   File "click/core.py", line 3372, in prompt_for_value
    |   File "click/termui.py", line 218, in prompt
    |   File "click/termui.py", line 201, in prompt_func
    | click.exceptions.Abort
    | 
    | During handling of the above exception, another exception occurred:
    | 
    | Traceback (most recent call last):
    |   File "deposit.py", line 128, in <module>
    |   File "deposit.py", line 121, in run
    |   File "click/core.py", line 1569, in __call__
    |   File "click/core.py", line 1532, in main
    |   File "click/utils.py", line 337, in echo
    |   File "click/_compat.py", line 509, in should_strip_ansi
    |   File "click/_compat.py", line 578, in isatty
    | KeyboardInterrupt
    | [PYI-3853:ERROR] Failed to execute script 'deposit' due to unhandled exception!
    | 
    | [TEST FAIL] Unexpected EOF while waiting for: Aborted
```

## Replication

Probably a `click.prompt()` with Ctrl-C on a slower machine, run often enough to eventually trigger this race condition. In our case it was the free github `ubuntu-22.04-arm` runner.

## Expected

click handles the KeyboardInterrupt exception gracefully, regardless of timing or architecture.

Environment:

- Python version: 3.14
- Click version: 8.4.2

kdeldycke:
This PR address https://github.com/pallets/click/issues/3802 by unifying the policy that was already implemented elsewhere in the code.

This applies the policy the pager has had since #351, where a late interrupt must not replace the intended outcome:
https://github.com/pallets/click/blob/8b19813f2bfca99f1018a587a8cf54fc959f2e5d/src/click/_termui_impl.py#L618-L632

So now in `main()`, every handler takes the same two steps:
- propagate when standalone mode is off, or
- collect the message and the exit code.

A final step then prints the message and exits, so a late interrupt can only lose the message, not the code.

This save one nesting level in the `try...except` cascade of `main()`.

This is in accordance with older policy choices from:
- #2380: where we decided that the first interrupt still reaches non-standalone callers as `Abort`, and keeps the interrupt as its `__cause__`
- #2584: where we decided not to change the `Aborted!` message

With these changes, we do not crash anymore, and a late interrupt after a successful run now exits 0, instead of printing a traceback.

This PR also adds stress tests that introduce random delays in the execution so we can catch regression at different points in time with real SIGINTs.

Off course we can discuss the policy we rely on and I'm ready to change the code accordingly.

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -99,6 +99,11 @@ def _check_nested_chain(
     raise RuntimeError(message)
 
 
+def _echo_aborted() -> None:
+    """Write the final abort message to standard error."""
+    echo(_("Aborted!"), file=sys.stderr)
+
+
 def _format_deprecated_label(deprecated: bool | str) -> str:
     """Return the parenthesized deprecation label shown in help text."""
     label = _("deprecated").upper()
@@ -1546,6 +1551,15 @@ class Command:
         # Process shell completion requests and exit early.
         self._main_shell_completion(extra, prog_name, complete_var)
 
+        # Every handler below takes the same two steps: propagate when
+        # standalone mode is disabled, otherwise collect the message and the
+        # exit code, and leave both to the teardown after the ``try``. The
+        # teardown writes the message and exits, and the outermost handler
+        # holds one policy for every interrupt arriving that late: the
+        # message may be lost, the intended exit code still wins.
+        report: cabc.Callable[[], None] | None = None
+        exit_code = 1
+
         try:
             try:
                 with self.make_context(prog_name, args, **extra) as ctx:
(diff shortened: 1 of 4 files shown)


src/click/core.py
```diff
@@ -1546,6 +1551,15 @@ def main(
         # Process shell completion requests and exit early.
         self._main_shell_completion(extra, prog_name, complete_var)
 
+        # Every handler below takes the same two steps: propagate when
+        # standalone mode is disabled, otherwise collect the message and the
+        # exit code, and leave both to the teardown after the ``try``. The
+        # teardown writes the message and exits, and the outermost handler
+        # holds one policy for every interrupt arriving that late: the
+        # message may be lost, the intended exit code still wins.
+        report: cabc.Callable[[], None] | None = None
```
davidism, line 1560: I'm not sure if we'd ever expect to do anything else with `report` besides show a message on stderr. Maybe we can do `exit_message: str | None = None` and `if exit_message is not None: echo(exit_message)`

tests/test_abort_interrupt.py
```diff
@@ -0,0 +1,305 @@
...
+
+        real_exit(code)
+
+    monkeypatch.setattr(sys, "exit", interrupted_exit)
+
+    assert _exit_code(cli, RecordingStderr(), monkeypatch) == 0
+    assert codes == [0, 0]
+
+
+# Same scenario, driven with real signals through a real pty.
```
davidism: I think this is a bit too much and can be left out. It also relies on timeouts, meaning it could be flaky in CI.

davidism:
> This save one nesting level in the try...except cascade of main().

Nice! It's always hard for me to follow this code at first. It would probably be good to document every way to "interrupt" a script, and what happens for each.

tests/test_abort_interrupt.py
kdeldycke: This is in fact the test I first started with to reproduce the issue in real conditions and introduce some randomness to reproduce the race condition and better understand what's going on. Only after that I created the set of more deterministic tests above in this suite. So yeah technically we can remove it as it did its job guiding this small refactor.

kdeldycke:
> Nice! It's always hard for me to follow this code at first. It would probably be good to document every way to "interrupt" a script, and what happens for each.

I can improve that part of the code by adding more comments if you find them not too superfluous. I can reuse some of the comments I have in the test suite and spread them at the right place in that `try...except` tree.

src/click/core.py
kdeldycke: I choose `report` as a variable name to steer users complaining about the `Aborted!` message in https://github.com/pallets/click/issues/2584 to use that to customize their own CLI. But the main reason is this is not always a message: it could be a `e.show` method.

kdeldycke:
> I can improve that part of the code by adding more comments if you find them not too superfluous. I can reuse some of the comments I have in the test suite and spread them at the right place in that `try...except` tree.

Better idea: I just updated the existing documentation in `docs/exceptions.md` file.

tests/test_abort_interrupt.py
kdeldycke: Removed in https://github.com/pallets/click/compare/de7ad0c7f7263352ba0b1c0a4e0084c991eabb7f..038450ff754217023a3f3cc4b03ac5fee8515e15

src/click/core.py
kdeldycke: I documented the `e.show` case in the table in the `docs/exceptions.md` documentation.

davidism:
Perfect, that's what I meant.

PR #3820, opened 2026-08-31
Document calling a command as a regular function

### LINKED ISSUE #1054: Use function both as API and click function?
I am trying to reduce boilerplate code and it was my idea to just decorate an existing function with click decorators to make it available via click. It's unclear to me from the documentation if this use case is supported. I also search the issues for the below error message and couldn't find anything related.

When I decorate a function with click, I am not able to use it as a function within Python:

Here's the function code:

```python
@click.command()
@click.argument('datestr')
def nasa_date_to_iso(datestr):
    """Convert the day-number based NASA format to ISO.

    Parameters
    ----------
    datestr : str
        Date string in the form Y-j

    Returns
    -------
    Datestring in ISO standard yyyy-mm-ddTHH:MM:SS.MMMMMM
    """
    date = dt.datetime.strptime(datestr, nasa_date_format)
    click.echo(date.isoformat())
    return date.isoformat()
```

Simple enough, right?
I install it via setup.py like so:
```
    entry_points={
        "console_scripts": [
            'nasa_date_to_iso = planetpy.utils:nasa_date_to_iso',
        ]
    },
```
but when I call it within ipython I get this:
```python
In [1]: from planetpy.utils import nasa_date_to_iso

In [2]: nasa_date_to_iso('2017-090')
Usage: ipython [OPTIONS] DATESTR

Error: Got unexpected extra arguments (0 1 7 - 0 9 0)
An exception has occurred, use %tb to see the full traceback.

SystemExit: 2

/Users/klay6683/miniconda3/envs/stable/lib/python3.6/site-packages/IPython/core/interactiveshell.py:2971: UserWarning: To exit: use 'exit', 'quit', or Ctrl-D.
  warn("To exit: use 'exit', 'quit', or Ctrl-D.", stacklevel=1)
```

So, is it not possible to have the same function both as a click-cli function and an python-internal API?

System: Py3.6 via conda on macOS 10.12.6, click version 6.7

kdeldycke:
Document an old open issue on how to call a command as a regular function.

Closes #1054.

### DIFF
--- a/docs/faqs.md
+++ b/docs/faqs.md
@@ -7,6 +7,49 @@
 
 ## General
 
+### Calling a command as a regular function
+
+I decorated a function with `@click.command()` to make it a CLI. When I import that name and call it from Python, it does not run my function. It parses arguments, prints a usage error and exits the process:
+
+```pytb
+>>> nasa_date_to_iso("2017-090")
+Usage: cli [OPTIONS] DATESTR
+Try 'cli --help' for help.
+
+Error: Got unexpected extra arguments (0 1 7 - 0 9 0)
+SystemExit: 2
+```
+
+#### Answer
+
+This is not a bug in Click. The decorator replaced the function with a {class}`Command` object. Calling that object runs the Click pipeline, and its argument is the list of command line arguments to parse, not the function's parameters.
+
+The recommended pattern is to keep a plain function and wrap it with a thin command:
+
+```python
+def nasa_date_to_iso(datestr): ...
+
+
+@click.command("nasa_date_to_iso")
+@click.argument("datestr")
+def nasa_date_to_iso_command(datestr):
+    click.echo(nasa_date_to_iso(datestr))
+```
+
+The function stays available as a Python API, and the command adds the CLI on top. Extracting behavior into a shared function used by several entry points is a common refactor.
+
+If you still want to run the command itself from Python, call it with a list of arguments and `standalone_mode=False` so it returns the result instead of exiting. This runs the Click pipeline, including parsing, validation and callbacks:
+
+```python
+nasa_date_to_iso_command(["2017-090"], standalone_mode=False)
+```
+
+An alternative form is:
+
+```python
+nasa_date_to_iso_command.main(["2017-090"], standalone_mode=False)
+```
+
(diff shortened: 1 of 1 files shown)


docs/faqs.md
```diff
@@ -7,6 +7,55 @@
...
+```pytb
+>>> nasa_date_to_iso("2017-090")
+Usage: cli [OPTIONS] DATESTR
+Try 'cli --help' for help.
+
+Error: Got unexpected extra arguments (0 1 7 - 0 9 0)
+SystemExit: 2
+```
+
+This is not a bug in Click. The decorator replaced the function with a {class}`Command` object. Calling that object runs the Click pipeline, and its argument is the list of command line arguments to parse, not the function's parameters.
```
Rowlando13: Move this line to answer

davidism:
I'm not sure I want to document calling callback directly. It feels like an internal implementation detail. I think the other strategies are sufficient.

Rowlando13:
If it's an implementation detail we should document as private and deprecate public api in the next feature release.

docs/faqs.md
kdeldycke: Done in https://github.com/pallets/click/compare/18924fda3f3369c060ca12b53ba60aba1300def5..bb86cdea6e83bc8e2711290eeffb4c8654baa7a3

kdeldycke:
> I'm not sure I want to document calling callback directly. It feels like an internal implementation detail. I think the other strategies are sufficient.

Removed in https://github.com/pallets/click/compare/18924fda3f3369c060ca12b53ba60aba1300def5..bb86cdea6e83bc8e2711290eeffb4c8654baa7a3 .

Now what do you think of @Rowlando13 proposition to make it a private method and deprecate it?

davidism:
We don't need to go that far, it's fine that it's accessible as "the thing the command calls", but we don't need to promote it as a thing to call directly.

kdeldycke:
> We don't need to go that far, it's fine that it's accessible as "the thing the command calls", but we don't need to promote it as a thing to call directly.

OK good, so this PR is ready to be merged upstream! :)

alpha33000 (commented):
Reggie

PR #3821, opened 2026-08-31
Add `Option.get_help_spec` to extract the help's left part of an hidden option

kdeldycke:
This PR splits the production of an option's left help column into its own dedicated column. This always returns the invocation specification of an option, even if that option is hidden.

`get_help_record()` on the other hand is unmodified and still keep returning the whole option's help that is to be displayed.

That distinction help separate the rendering layer:
- `get_help_record()` is returning what should be displayed to the user 
- `get_help_spec()` is returning what the option specification is, regardless of the decision to display it to the user

This allows finer introspection of the Click machinery and its rendering pipeline.

This is the other half of a work I initiated 2 years ago with:
- #2516
- #2517

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,9 @@ Unreleased
 - A `KeyboardInterrupt` arriving while `Command.main()` reports an abort or an error,
   or while it exits, no longer escapes as an unhandled traceback. The command still
   exits with the intended code; only the message may be lost. {issue}`3802`
+- Add {meth}`Option.get_help_spec`, which returns the option's left help
+  column even when the option is hidden. {meth}`Option.get_help_record` still
+  returns `None` for hidden options, so help screens are unchanged. {pr}`3821`
 - Document which types are inferred from `default`, and what an unrecognized
   `type` callable does to a command-line value. {issue}`3036` {pr}`3808`
 
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -3363,10 +3363,15 @@ class Option(Parameter):
                 nargs=self.nargs,
             )
 
-    def get_help_record(self, ctx: Context) -> tuple[str, str] | None:
-        if self.hidden:
-            return None
+    def get_help_spec(self, ctx: Context) -> str:
+        """Returns the left column of the option's help record: its spellings
+        and metavar, like ``-v, --verbose`` or ``-c, --config TEXT``.
 
+        Unlike :meth:`get_help_record`, the spec is produced even when the
+        option is :attr:`hidden`.
+
+        .. versionadded:: 8.5.1
+        """
         any_prefix_is_slash = False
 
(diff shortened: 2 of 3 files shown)


janluke:
> This will allow better integration for help-enhancing tools wrapping Click like Cloup (cc @janluke) 

Can you explain how this new method help us or how do you plan to use it? We already get `(left_col, right_col)` from `get_help_record`.

kdeldycke:
> > This will allow better integration for help-enhancing tools wrapping Click like Cloup (cc @janluke)
> 
> Can you explain how this new method help us or how do you plan to use it? We already get `(left_col, right_col)` from `get_help_record`.

Ah. I apologize. I was a bit quick to invoke you.😅 I just had an old impression that this was something you would rely on but I conflated my use-case and yours. I should have tripled-checked it. 🙇

I need to be more precise: this PR allows me to extract the `left_col` even when the option is hidden (`self.hidden is True`). This is the exact application of a workaround I had to introspect all parameters of a CLI, including those who are hidden: https://github.com/kdeldycke/click-extra/blob/68868d4489b87dca8247b2d73fc731167265febe/click_extra/parameters.py#L593-L613

This allows me to print metadata about an option (in this particular case its invocation specifications), outside its help screen context, and so outside the decision of either printing it to the user or not. Because an option hidden from the help screen is still an option recognize by the CLI and still has an effect on it. This is useful to implement deprecated options or preview options you don't want to advertise to the user. Still, you might need to print its specification in deprecation warnings, logs or error messages. That's where this new method can be useful.

In Click Extra I need the option specs to produce [machine-readable specs in JSON, Markdown and Carapace](https://kdeldycke.github.io/click-extra/machine-readable.html), and [live introspection](https://kdeldycke.github.io/click-extra/wrap.html#restricting-columns) rendering of third-party Click CLI.

Rowlando13 (approved):

Rowlando13:
I am okay with this.

janluke (approved):
LGTM

davidism:
Maybe there's a better name for this. Why is this the "spec"? Perhaps, `format_help_names` or something?

kdeldycke:
> Maybe there's a better name for this. Why is this the "spec"? Perhaps, `format_help_names` or something?

I guess `name` is even more confusing, given the whole PR I am trying to land around the real `name` property of `Argument` and `Option` at https://github.com/pallets/click/pull/3827 . I choose `spec` to differentiate it with `name` for that reason. And differentiate it from `decls`.

So the transformation pipeline become:
- `decls`: what the developer provide to its argument or option
- `spec`: what the user see in `--help` and what it provide to the CLI
- `name`: what is used internally in Click to identify that argument or option

If I had full reigns over the Click codebase, I would rename that cursed `name` into `id` or `uid`.

PR #3825, opened 2026-09-01
[rejected AI] atomic open_file: set mode via fchmod and close fd on failure (#3824)

### LINKED ISSUE #3824: [rejected AI] Atomic `open_file`: path-based `chmod` can follow a symlink, and a failed `chmod` leaks the fd
`click.open_file(..., atomic=True)` (and `click.File` with `atomic=True`) creates a sibling temp file with `os.open(..., O_CREAT|O_EXCL)`, then:

```python
fd = os.open(tmp_filename, flags, 0o666 if perm is None else perm)
if perm is not None:
    os.chmod(tmp_filename, perm)  # restore bits umask stripped
```

The atomic branch has two related problems:

1. **Leaked fd.** If `os.chmod` raises, `fd` is never closed and never wrapped in a file object. The exception propagates from `open_stream`.
2. **chmod by name.** `os.open` already has `fd`. Linux `os.chmod(path)` follows symlinks. Between `open` and `chmod`, a writer who can create names in the destination directory can replace `tmp_filename` with a symlink; `chmod` then changes the **target** of that symlink. `os.fchmod(fd, perm)` would only touch the inode that was just created.

# Reproduction

## 1) Leaked fd when `chmod` fails

```python
import errno
import os
import tempfile

import click

workdir = tempfile.mkdtemp()
dest = os.path.join(workdir, "target.txt")
with open(dest, "w") as f:
    f.write("old\n")
os.chmod(dest, 0o644)

real_chmod = os.chmod

def boom_chmod(path, mode, *args, **kwargs):
    if os.path.basename(str(path)).startswith(".__atomic-write"):
        raise OSError(errno.EPERM, "chmod denied", path)
    return real_chmod(path, mode, *args, **kwargs)

before = len(os.listdir("/proc/self/fd"))
os.chmod = boom_chmod
try:
    click.open_file(dest, "w", atomic=True)
except OSError:
    pass
after = len(os.listdir("/proc/self/fd"))
os.chmod = real_chmod
print(f"fd before={before} after={after} leaked={after - before}")
```

Observed:

```
fd before=9 after=10 leaked=1
```

Expected:
```
fd before=9 after=9 leaked=0 
```

## 2) `chmod` follows a replacement symlink

The swap below is what a concurrent directory writer can do in the window between `os.open` and `os.chmod`. This requires a concurrent writer with access to the destination directory; it is not a remote exploit.

```python
import errno
import os
import stat
import tempfile

import click

workdir = tempfile.mkdtemp()
dest = os.path.join(workdir, "atomic-out.txt")
victim = os.path.join(workdir, "victim.txt")
with open(dest, "w") as f:
    f.write("dest\n")
with open(victim, "w") as f:
    f.write("secret\n")
os.chmod(dest, 0o644)
os.chmod(victim, 0o600)
print("victim before", oct(stat.S_IMODE(os.stat(victim).st_mode)))

real_chmod = os.chmod

def attack_chmod(path, mode, *args, **kwargs):
    path = os.fspath(path)
    if os.path.basename(path).startswith(".__atomic-write"):
        os.unlink(path)
        os.symlink(victim, path)
        real_chmod(path, mode, *args, **kwargs)
        print("victim after chmod", oct(stat.S_IMODE(os.stat(victim).st_mode)))
        raise OSError(errno.EPERM, "stop after chmod", path)
    return real_chmod(path, mode, *args, **kwargs)

os.chmod = attack_chmod
try:
    click.open_file(dest, "w", atomic=True)
except OSError:
    pass
os.chmod = real_chmod
print("victim final", oct(stat.S_IMODE(os.stat(victim).st_mode)))
```
Observed:

```
victim before 0o600
victim after chmod 0o644
victim final 0o644
```

Expected:
```
victim before 0o600
victim final 0o600
```

The victim file was never passed to `open_file`. Its mode changed because `os.chmod` followed the symlink.

## Environment:

- Python version: 3.12.3
- Click version: 8.5.1.dev

linhongyu510:
## Summary

Closes #3824.

The atomic-write branch of `open_file(..., atomic=True)` set the temp file's mode with `os.chmod(tmp_filename, perm)`. This has two problems:

1. **Leaked descriptor.** If `os.chmod` raised, the fd returned by `os.open(O_CREAT|O_EXCL)` was never closed and never wrapped in a file object; the exception propagated and the fd leaked.
2. **`chmod` follows symlinks.** The mode was set by path rather than by fd. Between `open()` and `chmod()`, a concurrent writer with create access to the destination directory could replace the temp name with a symlink, so the mode change landed on the symlink's target (an unrelated file). `os.fchmod(fd, ...)` acts on the inode that was just created.

## Fix

- Set the mode with `os.fchmod(fd, perm)` where available, falling back to the path-based `os.chmod` on platforms without `fchmod` (Windows, where chmod only toggles the read-only bit).
- If setting the mode fails, close the fd and remove the temp file before re-raising, so no descriptor leaks.

The normal atomic-write behavior (mode preservation for existing files, umask for new files) is unchanged.

## Verification

Reproduced both issues from the issue report on macOS (fd count via `os.fstat` over the fd table; the symlink swap via a monkeypatched `os.chmod`):

- **Before:** mode failure leaks 1 fd; the symlink-swap changes the unrelated victim's mode `0o600 -> 0o644`.
- **After:** no fd leak on failure; the mode is applied through the fd, so no path-based `chmod` targets the temp file and the victim stays `0o600`; successful atomic writes still preserve the destination mode/content.

Added two regression tests in `tests/test_utils/test_open_file.py` (both fail on the current code — one with a `ResourceWarning` from the leaked fd — and pass with the fix). Full `test_open_file.py`, `test_basic.py`, `test_compat.py`: 290 passed.

### DIFF
--- a/src/click/_compat.py
+++ b/src/click/_compat.py
@@ -445,7 +445,27 @@ def open_stream(
             raise
 
     if perm is not None:
-        os.chmod(tmp_filename, perm)  # in case perm includes bits in umask
+        # Set the mode on the open fd rather than via the path: os.chmod()
+        # follows symlinks, so a concurrent writer who replaces the temp
+        # name between open() and chmod() could redirect the mode change
+        # to an unrelated file. fchmod() operates on the inode that was
+        # just created (it is unavailable on Windows, where chmod only
+        # toggles the read-only bit and we fall back to the path-based
+        # call). If setting the mode fails, close the fd and remove the
+        # temp file so no descriptor leaks.
+        fchmod = getattr(os, "fchmod", None)
+        try:
+            if fchmod is not None:
+                fchmod(fd, perm)  # in case perm includes bits in umask
+            else:
+                os.chmod(tmp_filename, perm)
+        except OSError:
+            os.close(fd)
+            try:
+                os.unlink(tmp_filename)
+            except OSError:
+                pass
+            raise
 
     f = _wrap_io_open(fd, mode, encoding, errors)
     af = _AtomicFile(f, tmp_filename, os.path.realpath(filename))
--- a/tests/test_utils/test_open_file.py
+++ b/tests/test_utils/test_open_file.py
@@ -1,11 +1,25 @@
 import os
 import pathlib
+import resource
 import stat
 
 import pytest
 
 import click
 from click._compat import WIN
+
+
+def _fd_is_open(fd):
+    try:
+        os.fstat(fd)
+        return True
+    except OSError:
+        return False
+
+
+def _count_open_fds():
+    soft, _ = resource.getrlimit(resource.RLIMIT_NOFILE)
+    return sum(1 for fd in range(soft) if _fd_is_open(fd))
 
 
 def test_open_file(runner, tmp_path):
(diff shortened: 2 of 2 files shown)


linhongyu510:
Fixed the Windows collection failure in 5920263. The production path was not failing: the new test module imported the Unix-only `resource` module at collection time, before its Windows-skipped descriptor test could be skipped. The import is now local to `_count_open_fds()`, which is only called by that Unix-only test. Verification: all 13 tests in `tests/test_utils/test_open_file.py` pass in an isolated test environment; loading the test module with imports of `resource` forced to fail now prints `WINDOWS_COLLECTION_OK`; targeted Ruff lint/format and `git diff --check` pass. Fresh CI is running on the new head.

davidism:

PR #3827, opened 2026-09-01
WIP: Require an automatic parameter name to be a Python identifier

### LINKED ISSUE #2483: Missing auto-generated environment variables in help screen & case-sensitivity

kdeldycke:
> [!WARNING]
> This PR is **waiting for #3866** to be settled/merged into the 8.x release cycle to be rebased on it, as both shares extensive unit tests.

This PR changes the behavior of automatic naming:
- An argument must produce a valid Python identifier. `0-file`, `foo.bar` and `foo bar` are now refused.
- `expose_value=False` no longer bypasses that check, on both arguments and options.
- An argument or option requires at least one declaration that can't be empty. `click.argument(expose_value=False)` and `click.option(expose_value=False)` used to produce a `""` parameter.
- An awkward explicit option name is normalized: `click.option("--in-file", "Input_File")` names its parameter `input_file`.

What does not change:

- The transformation itself: every `-` becomes `_` and the result is lower cased.
- An exposed option still raise if not producing a Python identifier, like `click.option("--0-file")`.

How to migrate:

- Give an option an explicit name: `click.option("--0-file", "zero_file")`.
- Rename an argument and pass `metavar` to keep the old display: `click.argument("zero_file", metavar="0-FILE")`.

This PR shares the majority of the tests from #3866 which lockdown all other edge-cases and canonical naming behavior.

### Context

It is in the same vein as my previous PR at https://github.com/pallets/click/pull/3808 , and is also based on a collection of edge-cases I accumulated over the years. It also covers platform differences between Windows and Unix-like (see: https://github.com/pallets/click/issues/2483).

Note: this PR started as simple test coverage expansion but ended up implementing sanitizing fixes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,18 @@
+## Version 9.0.0
+
+Unreleased
+
+- Breaking: an automatically derived parameter name must be a valid Python
+  identifier, or `TypeError` is raised. {class}`Argument` and {class}`Option`
+  derive their name the same way and apply the same check. {pr}`3827`
+- Breaking: `expose_value=False` no longer bypass the name check for a parameter.
+  {pr}`3827`
+- Breaking: an {class}`Option` now normalizes its declaration written as a Python
+  identifier to determine the parameter name. {pr}`3827`
+- Breaking: neither kind builds a parameter without a declaration.
+  `click.argument(expose_value=False)` and `click.option(expose_value=False)`
+  used to name a parameter `""`. {pr}`3827`
+
 ## Version 8.5.1
 
 Unreleased
--- a/docs/arguments.md
+++ b/docs/arguments.md
@@ -50,6 +50,32 @@ recognized, otherwise {data}`STRING` is used. If no default value is
 provided, the type is assumed to be {data}`STRING`. See
 {ref}`type-inference` for the types that are recognized.
 
+(argument-names)=
+
+## Argument Names
+
+The single declaration is not used as the name verbatim. Every `-` is replaced
+with `_` and the result is lower cased, so `click.argument("input-file")` names
+its parameter `input_file`. That is the same transform options apply, and it is
+likewise not reversible.
+
+The name must satisfy {meth}`str.isidentifier`, so that the callback can
+receive it as a keyword argument. {ref}`Options <option-names>` derive their
+name the same way and apply the same check. The
+{ref}`caution about reserved keywords <keyword-names>` applies here too.
+
+An argument takes exactly one declaration, and passing more raises
+{exc}`TypeError`. That declaration becomes the name through the
+{ref}`transform every parameter shares <name-transform>`, where the examples
+live.
+
+`expose_value=False` is no exception, because the name is also the key the
+parser stores the value under.
+
+An argument takes exactly one declaration, where an option takes several and
+{ref}`picks one of them <option-names>`. Past that choice both kinds derive the
+name the same way.
+
 ```{admonition} Note on Required Arguments
 :class: note
 
--- a/docs/options.md
+++ b/docs/options.md
@@ -60,40 +60,66 @@ and converting underscores to dashes.
     invoke(echo, args=['--string-to-echo', 'Hi!'])
 ```
 
-More formally, Click will try to infer the decorated function argument name as
-follows:
+(option-names)=
+
+## Option Names
+
+An option carries several declarations, so Click picks one of them to name the
+parameter:
 
 1. If a positional argument is a valid [Python identifier](https://docs.python.org/3/reference/lexical_analysis.html#identifiers) (and thus does not have dashes), it is chosen.
 2. If multiple positional arguments are prefixed with `--`, the first one
   declared is chosen.
 3. Otherwise, the first positional argument prefixed with `-` is chosen.
 
-To get the argument name, the chosen positional argument is converted to lower
-case, a leading `-` or `--` is removed if found, and any remaining `-`
-characters are replaced with `_`.
+The chosen declaration is not used as the name verbatim. Its `-` or `--` prefix
+is dropped, every remaining `-` is replaced with `_` and the result is lower
+cased, so `click.option("--input-file")` names its parameter `input_file`. That
+holds for rule 1 too: an identifier declaration says which declaration names the
+parameter, not what the name is spelled like.
+
+The name must satisfy {meth}`str.isidentifier`, so that the callback can
+receive it as a keyword argument. {ref}`Arguments <argument-names>` derive their
+name the same way and apply the same check, including the
+{ref}`caution about reserved keywords <keyword-names>`.
 
 ```{eval-rst}
 .. list-table:: Examples
-    :widths: 15 15
+    :widths: 25 15
     :header-rows: 1
 
     * - Decorator Arguments
-      - Inferred Argument Name
+      - Declaration Chosen
     * - ``"-f", "--foo-bar"``
-      - foo_bar
-    * - ``"-x"``
-      - x
-    * - ``"-f", "--filename", "dest"``
-      - dest
-    * - ``"--CamelCase"``
-      - camelcase
-    * - ``"-f", "-fb"``
-      - f
+      - ``--foo-bar``
     * - ``"--f", "--foo-bar"``
-      - f
+      - ``--f``
+    * - ``"-f", "-fb"``
+      - ``-f``
+    * - ``"-f", "--filename", "dest"``
+      - ``dest``
+    * - ``"-f", "--filename", "Dest"``
+      - ``Dest``
     * - ``"---f"``
-      - _f
+      - ``---f``
 ```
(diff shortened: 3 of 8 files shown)


davidism:
Both of these seem like bugs. I feel like the automatic names should always be normalized and pass `str.isidentifier`.

kdeldycke:
> Both of these seem like bugs. I feel like the automatic names should always be normalized and pass `str.isidentifier`.

OK cool, that was also my feeling. We don't want special treatments of the `name` of parameters, whatever their kind (options or arguments). I cannot find any reasons with the transformation should be different between the two kinds. So in the end yes, `str.isidentifier` should be the minimal check common to the two.

kdeldycke:
> Both of these seem like bugs. I feel like the automatic names should always be normalized and pass `str.isidentifier`.

@davidism Done. I did not pushed the normalization too much.

So there is one bothering difference between `Argument` and `Option`: the latter makes `click.option("--in-file", "Input_File")` names its parameter `Input_File`, while Argument forces it to `input_file`. Should I align both? I would tend to yes.

kdeldycke:
I just pushed the alignment and normalization of naming a bit more and updated the differences in the body of that PR.

kdeldycke:
OK these were my final edits. You can review this PR.

docs/arguments.md
```diff
@@ -50,6 +50,80 @@ recognized, otherwise {data}`STRING` is used. If no default value is
...
+```{caution}
+Only the ASCII hyphen is replaced, so a separator that merely looks like one is
+refused, as is any character that renders nothing.
+
+The identifier set itself moves with the Unicode table Python ships: the
+zero-width joiner (`U+200D`) entered it in Python 3.13, so a declaration holding
+one is refused up to Python 3.12 and names a parameter from 3.13 on. Prefer a
+declaration that is already a lower-case identifier with `-` for `_`.
+```
+
```
davidism, line 121: This is probably too much detail. I doubt anyone's writing these, but linking to `isidentifier` should already be enough.

docs/options.md
```diff
@@ -87,12 +83,60 @@ characters are replaced with `_`.
...
+read the other's value. Pass an explicit name instead:
+`click.option("--0-file", "zero_file", expose_value=False)`.
+
+```{caution}
+Transformation from option name to argument name is not reversible. And is many-to-one: several option names can map to the same argument name.
+
+For example, `--foo-bar`, `--Foo-Bar` and `--FOO-BAR` all map to `foo_bar`.
+
+This is allowed so that options can deliberately form a [feature switch group](#feature-switch-group).
 ```
```
davidism, line 140: This also seems like too much detail.

The example at the end about capitalization can be added to the table at the top if it's not there.

CHANGES.md
```diff
@@ -2,6 +2,14 @@
 
 Unreleased
 
+- {class}`Argument` derives its name the same way {class}`Option` does, and must be
+  a valid Python identifier, else raises `TypeError`. {pr}`3827`
+- `expose_value=False` no longer excuses that check on either kind. Pass an
+  explicit name to {class}`Option`, or rename an {class}`Argument` and pass
+  `metavar` to keep its display. {pr}`3827`
+- Neither kind builds a parameter without a declaration. `click.argument()` and
```
davidism, line 10: Is this saying that `Parameter.name` can no longer be `None`? Do we need to update the annotation? Would be great if so.

davidism:
This needs to be targeted at 9.0 and listed high up in the changelog, the name normalization is a breaking change. Along with the keyword check I mention below, this could be significantly disruptive. We should probably issue a deprecation warning and accept names that are currently accepted, before applying the new rules.

It might be good for the `TypeError` to give explicit instructions: `Argument {decls} tried to use 'name' but it is not a valid Python identifier. Add a valid name to the parameter declaration.`

Overall I think the docs are a little _too_ verbose about this. Giving a few examples and linking to `isidentifier` should be enough.

You bring up a good point about keywords. I think the intention of this code was clear: to only accept names that can be used in signatures rather than `kwargs`. We can also check that the name isn't a keyword with https://docs.python.org/3/library/keyword.html. I bet `from` and `in` are pretty common option names though, perhaps this is not worth it?

src/click/core.py
```diff
@@ -2436,9 +2434,47 @@ def __repr__(self) -> str:
...
+
+        :raises TypeError: when no name was derived, or the one derived is not
+            an identifier.
+        """
+        if name is not None and name.isidentifier():
+            return name
+
+        raise TypeError(
+            _(
+                "Could not determine name for {param_type} with declarations {decls!r}"
```
davidism, line 2474: "Could not determine a valid Python identifier for param_type decls."

Rowlando13 (commented):
I agree with @davidism. The docs are too verbose and these changes should go through normal deprecation process. Probably a best course of action is to make a pr against stable that has whatever tests you think we should have to lock in existing behavior. Then can merge stable into main. And you can make the new pr against main with the deprecations.

kdeldycke:
Ahah agreed 100% about the verbosity of the docs. They accumulates all my changes and discovery through the process of producing this PR. It was a way for me to highlight how clunky the ID derivation is at the moment. My thinking being that if I can write the docs easily and make them flow naturally, it probably means the underlying concept and implementation is clean.

Ok for a 9.0.0 target then.

What I will do:
- keep targeting this for a clean break for 9.0.0
- propose another PR for 8.5.1/8.6.0 with proper detection of the edge-cases, deprecation warnings and migration documentation so the future break has time to diffuse in the ecosystem

docs/arguments.md
kdeldycke: > This is probably too much detail. I doubt anyone's writing these, but linking to `isidentifier` should already be enough.

Agreed. These details are materialized in the test suite anyway so no need for this level of verbosity. Removed.

docs/options.md
kdeldycke: I was able to simplify these explanation with examples added to the table, but more importantly, but aligning the Option and Argument processing strictly, and moving the whole explanation to a dedicated section to the common `parameters.md` docs.

CHANGES.md
kdeldycke: Ah good catch! And yes, `Parameter.name` is declared `name: str` at the top level. And your comment made me realized I could simplify some annotation and branch checks. Done.

src/click/core.py
kdeldycke: Fixed.

kdeldycke:
> You bring up a good point about keywords. I think the intention of this code was clear: to only accept names that can be used in signatures rather than `kwargs`. We can also check that the name isn't a keyword with https://docs.python.org/3/library/keyword.html. I bet `from` and `in` are pretty common option names though, perhaps this is not worth it?

Problem is the [list of Python keywords](https://docs.python.org/3/reference/lexical_analysis.html#keywords) would prevent some options like `--import` or `--from`. Which is a hard constraint to impose a developer of a CLI. I prefer to not enforce checks too much here.

kdeldycke:
OK so I think I am done with this PR and it now implements the ideal target behavior for 9.0.0. I will now create in the next few days a sibling PR for 8.5.1/8.6.0 that detects the changes and emits deprecation warnings. If I discover new stuff while producing this other PR I might bring some changes here.

PR #3830, opened 2026-09-02
[rejected AI] remove indentation from epilog text

### LINKED ISSUE #2881: Allow epilog without indentation.
When adding an epilog to a group, the text appears to be indented under the commands list.

```python
@click.group(
    context_settings={"max_content_width": 100, "show_default": True},
    epilog="""\b
    Example Usage:
        # Run this stuff first
        $ some-command
        
        # Then run this command
        my-click-script go
    """,
)
def regression() -> None:
    """This is a tool with lots of commands."""
    pass
```

```txt
$ regression --help

Usage: regression [OPTIONS] COMMAND [ARGS]...

  This is a tool with lots of commands.

Options:

Commands:
  bundle  Wrap it all up.
  guard   Run lots of verifications.
  report  Generate a report from a bundle.
  run     Do stuff.
  store   Don't use this; it's for CI/CD only.
  verify  Compare two reports and verify they are identical.

  Example Usage:
      # Run this stuff first
      $ some-command
      
      # Then run this command
      regression guard
```

I want the entire epilog to not be indented. Perhaps `indent-epilog=False`?

raghibhaque:
## Description
Fixes #2881

When an epilog is added to a Click group or command, it was being indented along with the commands list. This change removes the indentation context so the epilog appears at the same level as other top-level help sections.

## Changes
- Removed `with formatter.indentation():` context from `format_epilog()` method in `src/click/core.py`
- Added two test cases to verify epilog is not indented
- Updated CHANGES.md with the fix

## Testing
All 39 existing tests pass, plus 2 new tests specifically for this fix.

The epilog will now be printed without indentation, allowing users to control its formatting via the epilog text itself (e.g., using the `\b` prefix if needed).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 Unreleased
 
+- The epilog text in help output is no longer indented. This allows for
+  better control over epilog formatting. {issue}`2881`
+
 ## Version 8.5.0
 
 Released 2026-08-24
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -1321,9 +1321,7 @@ class Command:
         if self.epilog:
             epilog = inspect.cleandoc(self.epilog)
             formatter.write_paragraph()
-
-            with formatter.indentation():
-                formatter.write_text(epilog)
+            formatter.write_text(epilog)
 
     def make_context(
         self,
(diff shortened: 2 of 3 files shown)


davidism:

PR #3839, opened 2026-09-05
[rejected AI] Show repeated option metavars in help

### LINKED ISSUE #3652: Automatically append ellipsis (`...`) to metavars when `multiple=True` in options

Nikhi00718:
Fixes #3652.

Automatically append `...` to generated option metavars when `multiple=True`, so help output communicates that an option may be repeated (for example, `--foo TEXT...`). Added a help-output regression test.

Tests: `python -m pytest tests/test_options.py -q` (647 passed). The full suite also exposed two unrelated sandbox subprocess-handle failures.

### DIFF
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2427,7 +2427,7 @@ class Parameter(ABC):
         if metavar is None:
             metavar = self.type.name.upper()
 
-        if self.nargs != 1:
+        if self.nargs != 1 or self.multiple:
             metavar += "..."
 
         return metavar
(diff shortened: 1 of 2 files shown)


davidism:

PR #3842, opened 2026-09-05
[rejected AI] Fix count=True ignoring non-zero default

### LINKED ISSUE #3841: [rejected AI] count=True ignores non-zero default when counting flags
When an option uses `count=True` with a non-zero `default`, each occurrence of the flag counts from **0** instead of from the configured default. Passing fewer flags than the default can therefore produce a value **lower than the default** (or leave `default=1` unchanged when the flag is passed once).

This differs from `argparse.ArgumentParser` with `action="count"`, which increments relative to the default (for example `default=3` plus one flag occurrence yields `4`).

```python
import argparse
import click

# argparse (expected-aligned behavior)
p = argparse.ArgumentParser()
p.add_argument("-v", action="count", default=3)
print("argparse no flags", p.parse_args([]).v)        # 3
print("argparse one flag", p.parse_args(["-v"]).v)    # 4
print("argparse two flags", p.parse_args(["-vv"]).v)  # 5

# click
@click.command()
@click.option("-v", count=True, default=3)
def cli(v):
    click.echo(v)

runner = CliRunner()
print("click no flags", runner.invoke(cli, []).output.strip())       # 3
print("click one flag", runner.invoke(cli, ["-v"]).output.strip())   # 1  ← bug
print("click two flags", runner.invoke(cli, ["-vv"]).output.strip()) # 2  ← bug
```

Root cause appears to be in `click/parser.py`, where the count action always starts from `0`:

```python
state.opts[self.dest] = state.opts.get(self.dest, 0) + 1
```

Once a command-line count is present, `consume_value()` takes that value and never applies the option default as a baseline.

Expected (argparse-compatible): command-line counts are **additive** on top of the option default.

| Invocation | Expected | Actual (Click) |
|---|---|---|
| `(none)` | `3` | `3` |
| one `-v` | `4` | `1` |
| `-vv` | `5` | `2` |
| `-vvv` | `6` | `3` |

With `default=1`, a single flag occurrence should yield `2`, but currently yields `1` (same as omitting the flag).

Environment:

- Python version: 3.12.10
- Click version: 8.5.1.dev0 (main)

ShamikOfficial:
When `count=True` options have a non-zero `default`, Click previously counted from `0` whenever the flag appeared on the command line. That could produce a value **lower than the default** (for example `default=3` with one `-v` yielded `1`).

This change makes command-line counts **additive relative to the option default** (and `Context.default_map` when present), matching `argparse` `action="count"`.

fixes #3841

- Added parametrized tests for non-zero defaults and a `default_map` baseline test.
- Documented behavior in the `Option` docstring (`.. versionchanged:: 8.5.1`).
- Added a `CHANGES.md` entry under 8.5.1.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,13 @@
 
 Unreleased
 
+-   `count=True` options with a non-zero `default` (or a `default_map`
+    baseline) now increment relative to that baseline when the flag is
+    passed on the command line, matching `argparse` `action="count"`.
+    Previously the parser always counted from `0`, so fewer flag
+    occurrences than the default could produce a value lower than the
+    default. {issue}`3841`
+
 ## Version 8.5.0
 
 Released 2026-08-24
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -2893,6 +2893,9 @@ class Option(Parameter):
                      in how it works but supports arbitrary number of
                      arguments.
     :param count: this flag makes an option increment an integer.
+                  Occurrences on the command line are counted relative to
+                  the option's default (including :attr:`Context.default_map`),
+                  matching :mod:`argparse`'s ``action="count"`` behavior.
     :param allow_from_autoenv: if this is enabled then the value of this
                                parameter will be pulled from an environment
                                variable in case a prefix is defined on the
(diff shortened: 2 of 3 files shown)


roli-lpci:
`count=True` also accepts string defaults, which Click normally converts through
the option's integer type. On this head, the new addition happens before that
conversion, so an explicit count crashes for either default source:

```python
@click.command()
@click.option("-v", count=True, default="3")
def cli(v):
    click.echo(v)

runner.invoke(cli, []).output       # "3\n" (an int in the callback)
runner.invoke(cli, ["-v"]).exception
# TypeError: can only concatenate str (not "int") to str
```

The same exception occurs with `default_map={"v": "3"}`. Both baselines are
still raw strings in `consume_value`, while the parser count is an `int`; normal
type conversion only runs later in `handle_parse_result`. Could the baseline be
converted through the option type before adding it, with coverage for string
values from both `default` and `default_map`?

davidism:

PR #3844, opened 2026-09-05
[rejected AI] Reject incomplete envvar batches for multiple+nargs options

### LINKED ISSUE #3843: [rejected AI] Envvar for multiple=True + nargs>1 silently drops incomplete trailing tokens
﻿When an option uses `multiple=True` and `nargs > 1`, an environment-variable value whose whitespace-split token count is **not a multiple of `nargs`** is accepted with exit code 0, and the trailing incomplete group is **silently discarded**.

The same values on the CLI are rejected (e.g. `Got unexpected extra argument`). Non-multiple `nargs > 1` options already reject incomplete envvar values with `BadParameter` (`Takes N values but M were given.`). The `multiple=True` path should do the same instead of truncating.

Root cause: `Option.value_from_envvar()` calls `batch()`, which uses `zip(..., strict=False)` and drops remainder tokens.

## Reproduction

```python
import click

runner = CliRunner()

@click.command()
@click.option("--arg", nargs=2, multiple=True)
def cli(arg):
    click.echo(repr(arg))

# Envvar: silently drops "c"
r = runner.invoke(cli, [], auto_envvar_prefix="TEST", env={"TEST_ARG": "a b c"})
print(r.exit_code, r.output.strip())
# Actual: 0  "(('a', 'b'),)"

# CLI: rejects the extra token
r2 = runner.invoke(cli, ["--arg", "a", "b", "c"])
print(r2.exit_code)
print(r2.output)
# Actual: exit 2, "Got unexpected extra argument (c)"
```

Also reproduces with `envvar="ARG"` and `env={"ARG": "a b c"}`.

With five tokens (`"a b c d e"`), the result is `(('a', 'b'), ('c', 'd'))` — `e` is silently discarded.

## Expected behavior

Envvar parsing should raise `BadParameter` (same family of arity error used for non-multiple `nargs` envvars), instead of silently dropping tokens and exiting 0.

## Environment

- Python version: 3.12.10
- Click version: main @ 36baa15 (8.5.1 unreleased)

**AI assistance disclosure:** This bug report was prepared with assistance from Cursor (AI coding agent). I reproduced and validated the behavior locally.

ShamikOfficial:
For options with `multiple=True` and `nargs > 1`, environment-variable values whose token count is not a multiple of `nargs` were silently truncated by `batch()` (`zip(..., strict=False)`), so incomplete values exited 0 and dropped trailing tokens. CLI parsing rejects the same incomplete input.

This change validates that the split envvar token count is a multiple of `nargs` in `Option.value_from_envvar()` and raises `BadParameter` (same message family as non-multiple `nargs` envvar checks) before batching.

**AI assistance disclosure:** Developed with assistance from Cursor (AI coding agent). I reproduced the bug, reviewed the fix/tests, and validated locally.

fixes #3843

- [x] Tests added (`test_multiple_nargs_envvar_rejects_incomplete_batch`)
- [x] `CHANGES.md` entry added for 8.5.1
- [x] No docs/`versionchanged` needed (bugfix aligning envvar path with existing arity errors)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,12 @@
 
 Unreleased
 
+-   Options with `multiple=True` and `nargs > 1` no longer silently drop
+    trailing tokens when an environment variable value's token count is
+    not a multiple of `nargs`. Incomplete envvar values now raise
+    `BadParameter`, matching arity checks for non-multiple `nargs`
+    options. {issue}`3843`
+
 ## Version 8.5.0
 
 Released 2026-08-24
--- a/src/click/core.py
+++ b/src/click/core.py
@@ -3580,6 +3580,19 @@ class Option(Parameter):
         if value_depth > 0:
             multi_rv = self.type.split_envvar_value(rv)
             if self.multiple and self.nargs != 1:
+                # batch() uses zip(strict=False) and would silently drop a
+                # trailing incomplete group. Reject incomplete envvar values
+                # the same way type_cast_value rejects wrong-arity groups.
+                if len(multi_rv) % self.nargs != 0:
+                    raise BadParameter(
+                        ngettext(
+                            "Takes {nargs} values but 1 was given.",
+                            "Takes {nargs} values but {len} were given.",
+                            len(multi_rv),
+                        ).format(nargs=self.nargs, len=len(multi_rv)),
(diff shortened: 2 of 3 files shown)


davidism:

PR #3852, opened 2026-09-06
[rejected AI] Deprecate public names for internal type helpers

### LINKED ISSUE #3847: More private utilities that need to be marked and public api deprecated
In types.py
- [ ]  `convert_type` - rename to _convert_type - currently has no top level import
- [ ] `FuncParamType` - maybe typing thing? - currently has no top level import

ryux1:
Rename the internal implementations to `_FuncParamType` and `_convert_type`, while keeping the old module attributes available through `__getattr__` with a deprecation warning until Click 9.0. Internal call sites and tests now use the private names.

The existing `to_info_dict()` output remains compatible (`param_type` is still `Func`).

Validation:
- `uv run pytest` (2062 passed, 24 skipped, 1 xfailed)
- `uv run tox run -e style,typing`

Developed with assistance from OpenAI Codex.

Fixes #3847

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,9 @@ Unreleased
   returns `None` for hidden options, so help screens are unchanged. {pr}`3821`
 - Document which types are inferred from `default`, and what an unrecognized
   `type` callable does to a command-line value. {issue}`3036` {pr}`3808`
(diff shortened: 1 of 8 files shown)


davidism:

PR #3854, opened 2026-09-07
[rejected AI] Mark `convert_type` and `FuncParamType` as private

### LINKED ISSUE #3847: More private utilities that need to be marked and public api deprecated

ump45nose:
Mark `convert_type` and `FuncParamType` as private in `click.types`. Neither
name is exported at the top level, so this only affects code that reaches into
`click.types` directly.

- Rename `convert_type` to `_convert_type` and update its internal callers
  (`Parameter`, `Option`, `Choice.get_metavar`, `Tuple.__init__`, `prompt`).
- Rename `FuncParamType` to `_FuncParamType` and `FuncParamTypeInfoDict` to
  `_FuncParamTypeInfoDict`.
- Ignore a leading underscore in `ParamType.to_info_dict` so a private
  `ParamType` subclass still serializes its public name (`"Func"`), keeping the
  existing info-dict output unchanged.

Tests: `ruff check` and `ruff format --check` pass, `mypy` (strict) and
`pyright --ignoreexternal --verifytypes click` pass, and the full test suite
passes (`2058 passed`).

fixes #3847

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 Unreleased
 
+- Mark `convert_type` and `FuncParamType` as private API in `click.types`.
+  {issue}`3847`
 - Fix `copy.deepcopy()` and `pickle` on a `Parameter`, `Option` or `Command`. {pr}`3805`
 - A `KeyboardInterrupt` arriving while `Command.main()` reports an abort or an error,
   or while it exits, no longer escapes as an unhandled traceback. The command still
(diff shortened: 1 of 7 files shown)


davidism:

PR #3859, opened 2026-09-08
Add pyrefly type checking

MarcoGorelli:
I got curious about taking https://github.com/pallets/click/pull/3412 forwards, and this is the result. It's a fairly minimal diff, so I've opened a PR

### DIFF
--- a/src/click/termui.py
+++ b/src/click/termui.py
@@ -779,7 +779,7 @@ def unstyle(text: str) -> str:
 
 def secho(
     message: t.Any | None = None,
-    file: t.IO[t.AnyStr] | None = None,
+    file: t.IO[t.Any] | None = None,
     nl: bool = True,
     err: bool = False,
     color: bool | None = None,
--- a/src/click/testing.py
+++ b/src/click/testing.py
@@ -22,6 +22,11 @@ if t.TYPE_CHECKING:
 
     from .core import Command
 
+    if sys.version_info >= (3, 13):
+        from typing import TypeIs
+    else:
+        from typing_extensions import TypeIs
+
 if sys.platform == "win32":
     CaptureMode: t.TypeAlias = t.Literal["sys"]  # pyright: ignore[reportRedeclaration]
 else:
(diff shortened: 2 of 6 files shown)


src/click/termui.py
```diff
@@ -779,7 +779,7 @@ def unstyle(text: str) -> str:
 
 def secho(
     message: t.Any | None = None,
-    file: t.IO[t.AnyStr] | None = None,
+    file: t.IO[t.Any] | None = None,
```
MarcoGorelli, line 782: `AnyStr` [is a typevar](https://docs.python.org/3/library/typing.html#typing.AnyStr), so it's not meant to be used in only a single place in a function

I've updated `t.IO[t.AnyStr]` to `t.IO[t.Any]` then

src/click/testing.py
```diff
@@ -22,6 +22,11 @@
 
     from .core import Command
 
+    if sys.version_info >= (3, 13):
+        from typing import TypeIs
+    else:
+        from typing_extensions import TypeIs
```
MarcoGorelli, line 28: this `typing_extensions` import only happens within a `if TYPE_CHECKING` block, so no runtime dependency is introduced

src/click/testing.py
```diff
@@ -208,12 +213,16 @@ def mode(self) -> str:
...
+
+
 def make_input_stream(
     input: str | bytes | t.IO[t.Any] | None, charset: str
 ) -> t.BinaryIO:
     # Is already an input stream.
-    if hasattr(input, "read"):
-        rv = _find_binary_reader(t.cast("t.IO[t.Any]", input))
+    if _has_read(input):
+        rv = _find_binary_reader(input)
```
MarcoGorelli, line 225: using `TypeIs` instead of just `hasattr` (which doesn't give type-checkers any info) we avoid the need for `cast`s

src/click/types.py
```diff
@@ -730,6 +730,7 @@ def __repr__(self) -> str:
         return "INT"
 
 
+# pyrefly: ignore [invalid-inheritance]  # https://github.com/facebook/pyrefly/issues/4842
 class IntRange(_NumberRangeBase[int, int], IntParamType):
```
MarcoGorelli, line 734: it's under discussion whether pyrefly should flag this (it probably should) https://github.com/facebook/pyrefly/issues/4842

for now I've just suppressed it

Rowlando13:
@MarcoGorelli Thanks for taking the time to add this, but I am not sure that we can accept it. All the pallets projects try to run the same type checker on the implementation which is mypy, until @davidism decides for us to make a switch. But we are interested is adding other type checkers to check the types we export as the type checkers become mature. That is also another call by @davidism.

PR #3473, opened 2026-05-19
Add help parameter to arguments

yurishevtsov:
Adds a `help` parameter to `click.argument()` and an Arguments section on `--help` when argument help is set. Commands without argument help are unchanged.

Implementation follows options: `Argument.get_help_record()` returns the metavar and help text for the help formatter (the base `Parameter` class returns `None`, so arguments never appeared in help tables before).

`Argument.to_info_dict()` now includes `help`, matching `Option`, so parameter metadata stays consistent for doc tools; this is not required for CLI help output.

Deprecated arguments without explicit `help` show a deprecation-only row in Arguments, matching deprecated options.

fixes #2983

PR #3860, opened 2026-09-09
Always list all positional arguments

kdeldycke:
This is inconsistent with the way options are displayed, as every options are displayed including those without help.

The reason to display all arguments in the `Positional arguments` section is to not introduce subtle confusion, as an argument without help would only be noticeable in the usage line.

Cloup already renders the section this way (see https://github.com/janluke/cloup/issues/210) and is the source discussion with @janluke for that bug fix.

In this implementation, I refrained to always display the `Positional arguments` section. So I made the decision to not print the `Positional arguments` section in full if and only if at least one of its argument has an `help`. This decision is open to discussion.

This is a follow up on https://github.com/pallets/click/pull/3473

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,9 @@ Unreleased
 - Add {meth}`Option.get_help_spec`, which returns the option's left help
   column even when the option is hidden. {meth}`Option.get_help_record` still
   returns `None` for hidden options, so help screens are unchanged. {pr}`3821`
+- The `Positional arguments` help section lists every argument of the command,
+  not only the documented ones. An argument with no `help` gets an empty
+  description, the same way an option with no `help` does. {pr}`3860`
 - Document which types are inferred from `default`, and what an unrecognized
   `type` callable does to a command-line value. {issue}`3036` {pr}`3808`
 - `path_type` narrows the converted value's type for `convert` and `prompt`.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@ Unreleased
 - The `Positional arguments` help section lists every argument of the command,
   not only the documented ones. An argument with no `help` gets an empty
   description, the same way an option with no `help` does. {pr}`3860`
+- {meth}`Argument.get_help_record` always returns a tuple. It used to return
+  `None` for an argument with no `help`. {pr}`3860`
 - Document which types are inferred from `default`, and what an unrecognized
   `type` callable does to a command-line value. {issue}`3036` {pr}`3808`
 - `path_type` narrows the converted value's type for `convert` and `prompt`.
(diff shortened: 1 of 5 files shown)


janluke (commented):
Of course, I approve the change in behavior, since that's what I decided for Cloup as well.

The code looks correct, but I think we should consider changing `get_help_record` to always return a tuple (never `None`; unless this turns useful elsewhere):
```python
    def get_help_record(self, ctx: Context) -> tuple[str, str]:
        return self.make_metavar(ctx), self.help or ""
```
Returning `None` when `help is None` is only hurting us now.

Also, this might read a bit more Pythonic:
```python
    def format_arguments(self, ctx: Context, formatter: HelpFormatter) -> 
        args = [
            param
            for param in self.get_params(ctx)
            if isinstance(param, Argument)
        ]
        if any(arg.help for arg in args):
            # assuming arg.get_help_record() always returns a help re
            help_records = [arg.get_help_record(ctx) for arg in args]
            ...
```

kdeldycke:
> Of course, I approve the change in behavior, since that's what I decided for Cloup as well.
> 
> The code looks correct, but I think we should consider changing `get_help_record` to always return a tuple (never `None`; unless this turns useful elsewhere):
> 
> ```python
>     def get_help_record(self, ctx: Context) -> tuple[str, str]:
>         return self.make_metavar(ctx), self.help or ""
> ```
> 
> Returning `None` when `help is None` is only hurting us now.

The reason for keeping `get_help_record` returning `tuple[str, str] | None` is to keep the `Parameter`, `Option` and `Argument` classes aligned.

I propose to evaluate this change in another PR as this opened a related issue I just stumbled upon. I will open a new one today and ping you there for feedback. Let's keep this PR focused on `Positional arguments` behavior for now.

> Also, this might read a bit more Pythonic:
> 
> ```python
>     def format_arguments(self, ctx: Context, formatter: HelpFormatter) -> 
>         args = [
>             param
>             for param in self.get_params(ctx)
>             if isinstance(param, Argument)
>         ]
>         if any(arg.help for arg in args):
>             # assuming arg.get_help_record() always returns a help re
>             help_records = [arg.get_help_record(ctx) for arg in args]
>             ...
> ```

Good point. Just applied your changes.

kdeldycke:
> > Returning `None` when `help is None` is only hurting us now.
> 
> The reason for keeping `get_help_record` returning `tuple[str, str] | None` is to keep the `Parameter`, `Option` and `Argument` classes aligned.
> 
> I propose to evaluate this change in another PR as this opened a related issue I just stumbled upon. I will open a new one today and ping you there for feedback. Let's keep this PR focused on `Positional arguments` behavior for now.

OK I get it now. You are right, this change should be bundled in this PR, as it highlight the policy of refusing to allow an argument to be `hidden`. Just made the changes.

This PR is ready to be reviewed.

janluke (approved):

Rowlando13 (approved):

PR #3863, opened 2026-09-10
[rejected AI] Fix shell completion for equals-separated option values

LouisDeconinck:
Fixes #2847

## Summary
- preserve the option prefix when completing values entered as `--option=value`
- add regression coverage for Bash and Zsh completion output
- document the fix in the unreleased changes

## Tests
- `uv run pytest -q`
- `uv run ruff check src/click/shell_completion.py tests/test_shell_completion.py


davidism:

PR #3701, opened 2026-07-17
Bad attempt: Deprecate `isolated_filesystem` and document its limits

kdeldycke:
This is an attempt to deprecate `isolated_filesystem` from the `CliRunner`. It has been discussed in #3501.

In addition to the deprecation, I added some explicit documentation about thread-safety as discussed in #3501. I used our own usage of `pytest-xdist` as a good example to follow.

All occurrences of `isolated_filesystem` in our own test suite have been removed and we are now relying on Pytest's `tmp_path`.

I also inaugurated the `docs/upgrade-guides.md` section introduced a while ago by @Rowlando13 .

This PR closes #3700.

PR #3866, opened 2026-09-11
Deprecate parameter names that are: not a Python identifier, or a Python Keyword, or not lower-cased

kdeldycke:
This PR prepares the ground for #3827, which is going to refuse parameter names that are not valid Python identifiers.

This add `DeprecationWarning` that detect when `Parameter`, `Argument` or `Option` gets a `name` that will be not allowed in the future 9.0.0 release:
- a parameter that is not a Python identifier (check against `str.identifier`)
- a parameter that is a Python keyword (check against `keyword.iskeyword`)
- an option name that is nor lower-cased

This PR also copy the majority of tests that were originally proposed in #3827, so the behavior of the `name` is locked down and frozen. This will prevent tweaking or introducing regressions during the full 8.x cycle.

This PR also inaugurate the 8.6.0 release on `main`.

### Historical details

Of course, I fucked-up the same way I fucked-up 2 months ago with https://github.com/pallets/click/pull/3701 , when juggling between `main` and `stable` branches made me push to `stable`. 😫

So:
- I pushed by mistake https://github.com/pallets/click/commit/e2bddbc801dabc4efc3c3c1e32b53b3f938d3efc
- I reverted my mistake on `stable` with https://github.com/pallets/click/commit/d036881798e34289a49b18a5c550c7cf687e5a7a
- And this PR is what it should have been all along

Anyway, this was resolved by resyncing both branches with https://github.com/pallets/click/pull/3876 .


davidism:
What do you think, should we also check `keyword.iskeyword`? Might be good to warn about it and see if anything comes up.

kdeldycke:
> What do you think, should we also check `keyword.iskeyword`? Might be good to warn about it and see if anything comes up.

Ah. That might be a good idea: test drive that check during a long 8.x cycle and see if it show up... If I implement this, what kind of severity should it triggers? A `DeprecationWarning` like non-Python-IDs ?

davidism:
Deprecation seems fine. Like I said in the other thread, it could be more disruptive, but it also seems to match the intention of the `str.isidentifier` check.

kdeldycke:
Just implemented the double `isidentifier()` / `iskeyword()` deprecation check in a common `_check_name_is_usable()` internal function. The later is maybe a bit too verbose as I distinguished the 2 cases and reused the deprecation explanation from `docs/upgrade-guides.md`. Can make it way smaller if you want to.

This is ready to be reviewed for eventual merge.

Rowlando13:
@kdeldycke this should merge into main. I don't we should issue new deprecations on bug fix versions.

kdeldycke:
> @kdeldycke this should merge into main. I don't we should issue new deprecations on bug fix versions.

OK make sense. So I retargeted the `main` branch, added a new commit in this PR to inaugurate the 8.6.0 release and updated the doc to target that release. And this time I hope I did not made any mistake juggling with the two branches... 🫣

Rowlando13 (approved):

PR #3870, opened 2026-09-15
[rejected AI] Test envvar list picks the first non-empty value; make chain xfail strict

ArnauFerma:
Two small test-only changes, no behaviour change.

- `envvar=[...]` is documented (`core.py`, "only the first non-empty value is returned") but no test passes a list or tuple to `envvar`. Changing `if rv:` to `if rv is not None:` in `resolve_envvar_value` survives the whole suite. The new `test_envvar_list_uses_first_non_empty` fails under that change with `assert 'name: \n' == 'name: second\n'` and passes on `main`.
- `test_group_chaining` is `@pytest.mark.xfail` without `strict`, and `xfail_strict` isn't set in `pyproject.toml`, so when the behaviour is fixed it will report XPASS and nothing will flag the marker for removal. `strict=True` makes that a failure. It still xfails today.

Found while running a mutation-based check over the suite; the rest of what was tried (boolean/choice/range parsing, path existence, exit codes, stderr routing, help output, `confirm`) was caught, most of it by many tests. Write-up: https://github.com/ArnauFerma/falsifiable-tests/blob/main/case-studies/click.md

- [x] Tests pass (`2060 passed, 24 skipped, 1 xfailed`)
- [ ] No changelog entry needed — test-only


davidism:

